PR #3, opened 2017-02-28
added yml/json/toml support, --help flag, and example directory

tscholl2:


cmd/task/task.go
```diff
@@ -1,9 +1,34 @@
...
+      "echo \"I am going to write a file named 'output.txt' now.\"",
+      "echo \"hello\" > output.txt"
+    ],
+    "generates": [
+      "output.txt"
+    ]
+  }
+}
+'''
+`)
```
andreynering: I think this example should be in YAML.

YAML will be considered the recommended default, in examples and documentation.

example/Taskfile.json
```diff
@@ -0,0 +1,11 @@
...
+  "hello": {
+    "cmds": [
+      "echo \"I am going to write a file named 'output.txt' now.\"",
+      "echo \"hello\" > output.txt"
+    ],
+    "generates": [
+      "output.txt"
+    ]
+  }
+}
```
andreynering: Missing line break.

andreynering: Also missing TOML example.

tscholl2: Done. I don't know if these examples really highlight the parts of task that you want them to, so you might consider changing them. But I hope this helps some.

andreynering:
Thank you again.

Merged as https://github.com/go-task/task/commit/61b1aa8559dda4469dae5579b3bbc17f1c4683c7 and https://github.com/go-task/task/commit/a53fcf8bbac78616906db9444826587297b34220

PR #4, opened 2017-03-01
Added chdir to task

sascha-andres:
I wanted to run tools in sub directories of the directory Taskfile.yml lives in. For me it is more convenient to provide a directory to run the command(s) in then prefixing each command with `cd <path> && `

### DIFF
--- a/task.go
+++ b/task.go
@@ -40,7 +40,7 @@ type Task struct {
 	Deps      []string
 	Sources   []string
 	Generates []string
-	Chdir     string
+	Dir       string
 }
 
 type taskNotFoundError struct {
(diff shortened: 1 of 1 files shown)


sascha-andres:
A sample file looks like this:

    build:
      cmds:
        - ./run.sh
      chdir: ./build/

task.go
```diff
@@ -40,6 +40,7 @@ type Task struct {
 	Deps      []string
 	Sources   []string
 	Generates []string
+	Chdir     string
```
andreynering: I think this can be called just "dir".

task.go
```diff
@@ -125,13 +126,16 @@ func isTaskUpToDate(t *Task) bool {
...
 
-func runCommand(c string) error {
+func runCommand(c, path string) error {
 	var cmd *exec.Cmd
 	if ShExists {
 		cmd = exec.Command(ShPath, "-c", c)
 	} else {
 		cmd = exec.Command("cmd", "/C", c)
 	}
+	if "" != path {
```
```go
path != ""
```

sascha-andres:
Updated code

andreynering:
Thank you very much

PR #5, opened 2017-03-02
Simple cyclic dependency detection

sascha-andres:
A very simple detection for cyclic dependencies like this:

```
build:
  deps:
  - build2
  cmds:
  - echo a

build2:
  deps: 
  - build
  cmds:
  - echo b
```


andreynering:
Thank you again.

PR #6, opened 2017-03-03
Variable handling

sascha-andres:
Use of variables ( task local, variable file and environment ) in Taskfiles


variableHandling.go
```diff
@@ -0,0 +1,82 @@
...
+		localVariables[key] = value
+	}
+	return localVariables, nil
+}
+
+// ReplaceVariables writes variables into initial string
+func ReplaceVariables(initial string, variables map[string]string) string {
+	replaced := initial
+	for name, val := range variables {
+		replaced = strings.Replace(replaced, fmt.Sprintf("{{%s}}", name), val, -1)
```
andreynering: I think I'll replace this with [Go's template engine](https://golang.org/pkg/text/template/) after merge.

So we can have some variables assigned by default (like OS, Architecture, etc). And will allow users to program if's and range's. Example:

```yml
acmd:
  cmds:
    - {{if eq .GOOS "windows"}}windows-command{{else}}unix-command{{end}}
```

But don't change that in this PR, otherwise it'll become too big.

variableHandling.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	// VariableFilePath file containing additional variables
+	VariableFilePath = "Variables"
```
andreynering: Maybe `Taskvars`?

README.md
```diff
@@ -42,6 +42,55 @@ task assets build
...
+
+They are evaluated in the following order:
+
+Task local variables are overwritten by variables found in `Variables`. Variables found in `Variables` are overwritten with variables from the environment. The output of the last command is stored in the environment. So you can do something like this:
+
+```yml
+build:
+  deps: [setvar]
+  cmds:
+  - echo "{{prefix}} '{{THEVAR}}'"
```
andreynering: I think `prefix` should be `PREFIX`. Uppercase should be a convention for variables, just like environment variables.

README.md
```diff
@@ -42,6 +42,55 @@ task assets build
...
+  deps: [setvar]
+  cmds:
+  - echo "{{prefix}} {{THEVAR}}"
+  variables:
+    prefix: "Path:"
+
+setvar:
+  cmds:
+  - echo "{{PATH}}"
+  set: THEVAR
```
andreynering: `setvar` instead of `set`?

task.go
```diff
@@ -81,21 +83,38 @@ func RunTask(name string) error {
...
 			return &taskRunError{name, err}
 		}
+		var (
+			output string
+			err    error
+		)
+		if output, err = runCommand(ReplaceVariables(c, vars), ReplaceVariables(t.Dir, vars)); err != nil {
+			return &taskRunError{name, err}
+		}
+		fmt.Println(output)
```
sascha-andres: Thought about a quiet switch, but that would imply more, I will change code accoring to your suggestion

sascha-andres: You have already...

variableHandling.go
```diff
@@ -0,0 +1,82 @@
+package task
```
sascha-andres: Changed

README.md
sascha-andres: Shouldn't this be a choice of the user instead of the tool?

README.md
sascha-andres: `set` is the property with the variable name. I searched for a valid name of the most simplistic task to set a variable.

task.go
```diff
@@ -81,21 +83,38 @@ func RunTask(name string) error {
 		return &taskNotFoundError{name}
 	}
 
-	for _, d := range t.Deps {
-		if err := RunTask(d); err != nil {
-			return err
-		}
-	}
-
 	if !Force && isTaskUpToDate(t) {
```
sascha-andres: @andreynering  Why running dependencies for a task when the task itself would not run? That was at least my assumption, as you would create side-effects that possible could change what the task would generate itself. I create an issue for discussion and commit a change running deps first.

variableHandling.go
sascha-andres: Changed

PR #9, opened 2017-03-06
Allow specifying additional environment settings

sascha-andres:
Some tools need to have environment settings. This feature adds the ability to provide settings on a per task basis.

### DIFF
--- a/task.go
+++ b/task.go
@@ -2,6 +2,7 @@ package task
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
@@ -47,6 +48,7 @@ type Task struct {
 	Dir       string
 	Vars      map[string]string
 	Set       string
+	Env       map[string]string
 }
 
 // Run runs Task
@@ -104,7 +106,7 @@ func RunTask(name string) error {
 	}
 
 	for i := range t.Cmds {
-		if err = t.runCommand(i); err != nil {
+		if err = t.runCommand(i, t.Env); err != nil {
 			return &taskRunError{name, err}
 		}
 	}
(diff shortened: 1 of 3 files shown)


andreynering:
Hmm... we could just set `vars` as env variables instead? Is there any benefit having 2 separated options?

task.go
```diff
@@ -104,7 +106,7 @@ func RunTask(name string) error {
 	}
 
 	for i := range t.Cmds {
-		if err = t.runCommand(i); err != nil {
+		if err = t.runCommand(i, t.Env); err != nil {
```
andreynering: Don't need to pass `Env` as parameter. Task is available as a receiver of the func.

task.go
```diff
@@ -151,6 +153,21 @@ func (t *Task) runCommand(i int) error {
 	if dir != "" {
 		cmd.Dir = dir
 	}
+	if nil != envVariables {
```
```go
t.Env != nil
```

sascha-andres:
@andreynering  We use to have services running that take decide a mandator based on an environment. Devs set their mandator globally and when they wanted to test with another they had to source the profile. The ability to set the env of the child process to a variable would make them get rid of that process.

andreynering:
@sascha-andres I'm OK with that. 👍 

What's you opinion on change proposed on https://github.com/go-task/task/issues/11 ?

E.g. change `vars` attribute and `Taskvars` file so it will be actually a command, and the result will be assigned to the var?

It's a backward incompatible change, but we'd be doing that early. If we decide to follow that, for consistency maybe that should be done on `env` props, too?

sascha-andres:
@andreynering I would keep env properties as they are - you can fill in variables there. I think that would complicate stuff.

Regarding vars we could go for a fallback scenario like trying to execute and when it fails assign the text. One thing that comes into my mind is, that even for a short Taskfile the number of processes spawned might be very high.

PR #10, opened 2017-03-06
add help task

zbindenren:
For large projects sometimes it is nice to have a help task, that lists the most important tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,37 @@ printos:
     - echo 'Is SH? {{IsSH}}'
 ```
 
+### Help
+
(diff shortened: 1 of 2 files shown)


sascha-andres:
I like the idea. Maybe we could amend this to print out dependencies?

zbindenren:
I would do the depgraph with an option.

task -d target shows depgraph for target, task -d shows depgraph for all targets?

andreynering:
Thank you very much.

Merged manually as https://github.com/go-task/task/commit/e4c1cc3e7765d25c4f7cedd0544405872ba37241

zbindenren:
Thx

PR #12, opened 2017-03-07
fix trailing white space for set instruction

zbindenren:
Otherwise you have trailing white space in your vars.


sascha-andres:
Agree that there is probably no use for whitespace in front or behind

Issue #11, opened 2017-03-06
Feature: Would be nice to have a more dynamic Taskvar file


zbindenren:
I think the following would be really nice:

```yaml
GIT_USER: git config user.name
GIT_EMAIL: git config user.email
STATIC_VAR: echo static
```

Also if #7 will be implemented, OS specific Taskvar would be nice.

What do you think?

andreynering:
You can have that currently with:

```yml
setgituser:
  set: GIT_USER
  cmds:
    - git config user.name
```

But I also like your idea. Would be also welcome `vars` attribute of a task. 🤔

zbindenren:
Yes, I am already doing it :-). But it is more verbose and I have to set the dependencies. If you have a lot of vars that means a lot of dependencies.

sascha-andres:
@andreynering I would keep env properties as they are - you can fill in variables there. I think that would complicate stuff.

Regarding vars we could go for a fallback scenario like trying to execute and when it fails assign the text. One thing that comes into my mind is, that even for a short Taskfile the number of processes spawned might be very high.

sascha-andres:
After thinking a whole on this, I had an idea of a different approach. Let's assume a Taskvar like this:

```
SIMPLE_VAR: content
EXEC_VAR:
  cmd: echo a
```

The former is just an assignment, the latter one an executed one. This would make reading the Taskvar harder but the syntax cleaner

andreynering:
Both ideas are valid.

I just had an idea for having it in a clear and backward compatible way. We could have a special syntax like prefixing commands with `@`.

```yml
# Taskvars.yml
STATIC: foobar
DYNAMIC: @git config user.name
```

```yml
# Taskfile.yml
build:
  cmds:
    - go build main.go
  vars:
    STATIC: foobar
    DYNAMIC: @git config user.name
```

sascha-andres:
I like the idea of the `@` in front. I think I have some spare time this evening and will give this a try

zbindenren:
The idea us not bad, but could confuse make users. @command means suppress output.

Why not using the make approach:

DYNAMIC: $(shell command)

or similar:

DYNAMIC: $(command)

zbindenren:
When we are at variables. In a Makefile it is possible to define a default value for an env var. Would this also be possible with task?

sascha-andres:
@zbindenren You can define env variables at task level. Or you set a variable to a constant which essentially writes the variable to the processe's environment

[...]

PR #13, opened 2017-03-07
Feature/os specific

sascha-andres:
Implemented a first throw to load OS specific handling. Later we might use commandline switches to load any number of Taskfiles in order of appearence


sascha-andres:
@andreynering If there is nothing from your side I would merge this before working on #11

sascha-andres:
Test with sample from README successful

PR #14, opened 2017-03-08
Feature/#11 dynamic taskvar

sascha-andres:
Easy implementation of dynamic variables


andreynering:
Merged manually on master.

I just changed `@` to `$` because YAML don't likes `@` at beginning of strings unless you quote it. For `$`, a quote is not necessary.

PR #16, opened 2017-03-09
Support for a default task

sascha-andres:
If no argument is given try default. @andreynering Just a proposal


andreynering:
I like the idea. Only misses documentation on README.

sascha-andres:
Adding README and merging

PR #25, opened 2017-04-22
Migrate from os/exec.Cmd to a native Go sh interpreter

### LINKED ISSUE #23: Use native Go sh interpreter? github.com/mvdan/sh
With this project, maybe it's possible to get rid of `cmd` to run commands on Windows.

https://github.com/mvdan/sh

andreynering:
Closes #23


execext/exec.go
```diff
@@ -2,22 +2,49 @@ package execext
...
-	ShPath, err = exec.LookPath("sh")
-	ShExists = err == nil
-}
+// RunCommand runs a shell command
+func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+
+	p, err := syntax.Parse(strings.NewReader(opts.Command), "", syntax.ParseMode(0))
```
mvdan: You can just use `0` as the last argument

execext/exec.go
```diff
@@ -2,22 +2,49 @@ package execext
...
+		Dir:    opts.Dir,
+		Env:    opts.Env,
+		Stdin:  opts.Stdin,
+		Stdout: opts.Stdout,
+		Stderr: opts.Stderr,
+	}
+	if err = r.Run(); err != nil {
+		return err
+	}
+	return nil
```
mvdan: `return r.Run()`

execext/exec.go
andreynering: Thanks. What's the intention of this param? Accept different shells in the future? (`sh`, `bash`, `zsh`, etc?).

mvdan: It also does other things, like tell the parser whether or not to keep comments: https://godoc.org/github.com/mvdan/sh/syntax#ParseMode

In the future (2.0) I think I'll change this, because mixing in such different things in a bitfield is a bit confusing.

PR #35, opened 2017-07-03
Print on set

smyrman:
Enables printing of the run command, even when "set" is specified. I don't know if this is what we want, but it felt more consistent. I would have raised an issue first, but since the change was so small, I thought I would skip directly to the PR instead.

This PR also updates the help text to follow standard conventions. I renamed "target" in the help text to "task", since "target" often refers to the "target output file name".

Feel free to reject or request changes.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -14,14 +14,14 @@ var (
 	version = "master"
 )
 
-func main() {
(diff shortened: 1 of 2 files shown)


andreynering:
@smyrman Thanks!

Issue #42, opened 2017-07-09
Variable templating not handled when calling task via deps


smyrman:
Given the following Taskfile:
```yaml
taskA:
  vars:
    A: null
    B: "42"
  cmds:
    - echo "{{.A}} {{.B}} {{.C}}"

taskB:
  vars:
    A: "taskB"
    B: "ignored"
    aha: "weee"
  deps:
    - task: taskA
      vars:
        A: "dep:{{.A}}"
        C: "{{.aha}}"
  cmds:
    - task: taskA
      vars:
        A: "cmd:{{.A}}"
        C: "{{.aha}}"

default:
  deps:
    - task: taskB
      vars: {A: "default"}
```

I would expect `task` to output:
```sh
task: No argument given, trying default task
echo "dep:default 42 weee"
dep:default 42 weee
echo "cmd:default 42 weee"
cmd:default 42 weee
```
Instead, we get:
```sh
task: No argument given, trying default task
echo "dep:{{.A}} 42 {{.aha}}"
dep:{{.A}} 42 {{.aha}}
echo "cmd:default 42 weee"
cmd:default 42 weee
```

smyrman:
UPDATE: I was on a branch. actually, on master, even the templating in cmd no longer works:

```go
task: No argument given, trying default task
echo "dep:{{.A}} 42 {{.aha}}"
dep:{{.A}} 42 {{.aha}}
echo "cmd:{{.A}} 42 {{.aha}}"
cmd:{{.A}} 42 {{.aha}}
```

PR #43, opened 2017-07-09
Bugfix: allow templating when calling deps

smyrman:
Fixes issue #42 by allowing for template evaluatation on task override
variables when the task is launched as dependency.

### DIFF
--- a/task.go
+++ b/task.go
@@ -165,8 +165,16 @@ func (e *Executor) runDeps(ctx context.Context, call Call) error {
 			if err != nil {
 				return err
 			}
+			depVars := make(Vars, len(d.Vars))
+			for k, v := range d.Vars {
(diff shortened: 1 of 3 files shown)


task.go
```diff
@@ -165,8 +165,16 @@ func (e *Executor) runDeps(ctx context.Context, call Call) error {
 			if err != nil {
 				return err
 			}
+			depVars := make(Vars, len(d.Vars))
+			for k, v := range d.Vars {
+				v, err := e.ReplaceVariables(v, call)
```
smyrman: As an FYI, it would be good with some restructuring to avoid code duplication in command and deps handling.

This commit only makes things work.

andreynering: Sure, we should deduplicate it. =)

andreynering:
Thank you very much! =)

Issue #47, opened 2017-07-12
"generates" does not handle absolute paths.


smyrman:
It may seem like that the files listed in "generates" (and maybe the ones in "sources"; haven't checked) are prefixed with the task dir, even when an absolute path is given.

Below is a principal example for reproducing the problem:

Taskvars.yml:
```yaml
ROOT: $pwd
```

Taskfile.yml:
```yaml
default:
  dir: sub
  cmds:
    - cat file.txt > '{{.ROOT}}/file2.txt'
  sources:
    - file.txt
  generates:
    - "{{.ROOT}}/file2.txt"
```

sub/file.txt
```txt
hello world
```

If i run this task twice, it would not be found up to date:

``` sh
$ go get -u github.com/go-task/task/cmd/task   # Ensure task is up-to-date
$ task -v
task: No argument given, trying default task
task: dynamic variable: "$pwd", result: "/Users/smyrman/.local/src/github.com/searis/play"
cat file.txt > '/Users/smyrman/.local/src/github.com/searis/play/file2.txt'
$ task -v
task: No argument given, trying default task
task: dynamic variable: "$pwd", result: "/Users/smyrman/.local/src/github.com/searis/play"
cat file.txt > '/Users/smyrman/.local/src/github.com/searis/play/file2.txt'
```

If I _change_ the generates section in the Taskfile.yml to be a relative path, it works:
```yaml
  generates:
    - "../file2.txt"
```
Relaunch task:
```sh
$ task -v        
task: No argument given, trying default task
task: dynamic variable: "$pwd", result: "/Users/smyrman/.local/src/github.com/searis/play"
task: Task "default" is up to date
```

As for the use-case of absolute paths in generates, it matters when you want to use the same (configurable) output directory for several build targets that are built from various sub-directories.

smyrman:
I will do a PR

andreynering:
This fix is as simple as changing:

```go
p = filepath.Join(dir, p)
```

to

```go
if !filepath.IsAbs(p) {
        p = filepath.Join(dir, p)
}
```

[Here][here1] and [here][here2].

[here1]: https://github.com/go-task/task/blob/master/file.go#L26
[here2]: https://github.com/go-task/task/blob/master/file.go#L37

smyrman:
> This fix is as simple as changing (...)

I found that pretty quick, but I used for ever to figure out why the new tests didn't work:-)

smyrman:
All fixed now in PR #48.

PR #48, opened 2017-07-12
Allow absolute path in generates section.

smyrman:
Fixes issue #47 and adds a test for it. Absolute paths should also be
allowed in sources after this change, but this has not been specifically
tested.


testdata/generates/Taskfile.yml
```diff
@@ -0,0 +1,34 @@
...
+    - "../rel.txt"
+
+sub/src.txt:
+  desc: generate source file
+  cmds:
+    - mkdir -p sub
+    - echo "hello world" > sub/src.txt
+    # Sleep to avoid time-stamp issue. Some filesystems, E.g. HFS+ for Mac OS,
+    # only support one second precision on file timestamps.
+    - sleep 1
```
smyrman: I will remove this sleep as part of a PR for #49.

smyrman:
Updated to add .gitignore file.

andreynering:
Hi @smyrman 

Both PR's are fine. Thank you for them.

I made a simple comment on #50. You can merge both when you're confident with the code.

smyrman:
Thanks @andreynering. I saw the comment:-)

PR #50, opened 2017-07-12
Equal time (depends on #48)

### LINKED ISSUE #49: Allow equal time-stamps to be considered up-to-date
# Proposal
A task should be considered up-to-date if the newest timestamp for `sources` is exactly equal to the oldest timestamp from `generates`.

# Background
On Mac, the default filesystem, HFS, only supports second precession for file modification timestamps. Windows NTFS (in theory) supports 100-nanosecond interval precession, while ext4 on Linux rely on cached timestamps from the kernel which in most cases would at least be about 100 ms apart.

Especially on HFS,  this can create some hard-to-debug issues where tasks some times are considered up-to-date, and sometimes not.

While writing the tests for #47, it took me a while to figure out _why_ my test worked (sometimes) when I run `task` manually in "testdata/generates", but never when I run it from task_test.go. For real-life use-cases, this would help with getting a more consistent behavior when we have task-output from one task used as input in the next (short-running) task.

smyrman:
Let's wait for #48 to be merged, then we can rebase this PR. This commit fixes #49.

### DIFF
--- a/task.go
+++ b/task.go
@@ -250,8 +250,7 @@ func (e *Executor) isUpToDateTimestamp(ctx context.Context, call Call) (bool, er
 	if err != nil || generatesMinTime.IsZero() {
 		return false, nil
 	}
-
-	return generatesMinTime.After(sourcesMaxTime), nil
+	return !generatesMinTime.Before(sourcesMaxTime), nil
 }
 
 func (e *Executor) runCommand(ctx context.Context, call Call, i int) error {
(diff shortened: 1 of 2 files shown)


task.go
```diff
@@ -250,8 +250,8 @@ func (e *Executor) isUpToDateTimestamp(ctx context.Context, call Call) (bool, er
 	if err != nil || generatesMinTime.IsZero() {
 		return false, nil
 	}
-
-	return generatesMinTime.After(sourcesMaxTime), nil
+	upToDate := generatesMinTime.After(sourcesMaxTime) || generatesMinTime.Equal(sourcesMaxTime)
+	return upToDate, nil
```
```go
return !generatesMinTime.Before(sourcesMaxTime), nil
```

smyrman: better:-)

Issue #40, opened 2017-07-06
Allow template evaluation in task variables


smyrman:
Issue #38 is about allowing a variable to be passed onto a sub-task or dependency. The chosen way to handle it, at least initially, was to allow for template variables to be evaluated when passing on the vars. See #39.

However, it might make sense to allow _global_ variables and task _parameters_ to be evaluated in a task's local vars and env section as well. The example below is maybe not the best use-case, but it should illustrate what I mean:

```yaml
# Taskvars.yml
GOOS: $go env GOOS
GOARCH: $go env GOARCH
```

```yaml
# Taskfile.yml
default:
  vars:
    TARGET_DIR: "build/{{.GOOS}}_{{.GOARCH}}"
  env:
    GOOS: "{{.GOOS}}"
    GOARCH: "{{.GOARCH}}"
  cmds:
   - mkdir -p "{{.TARGET_DIR}}"
   - go build -o '{{.TARGET_DIR}}/myapp' .

cross-compile:
  deps:
    - task: default
      vars: {GOOS: linux}
    - task: default
      vars: {GOOS: windows, GOARCH: 386}
    - task: default
      vars: {GOOS: windows, GOARCH: amd64}
    - task: default
      vars: {GOOS: darwin}
```

Another reason for adding this feature is to make the behaviour more consistent, and to allow for easier documentation of the detailed behaviour. E.g.:
- task scoped variables and env have access to global variables and environment variables, as well as passed in task override variables.
- variables set when calling a task, has access to all of the above, pluss any variables or environemnt variables set on the task.

andreynering:
Working on master branch. I'll keep this open until it can be tested.

smyrman:
Cool, that's fast!

andreynering:
I'm closing this by now. If any issue rises, it can be reopened.

smyrman:
Dosn't seam to work...

```yaml
# component-go-bindata is a reusable task to buils go binary assets from web
# assets.
component-go-bindata:
  vars:
    COMPONENT: null       # component to build bindata for (configurator, web)
    WEB_TARGET_SUFFIX: "" # suffix for webpack target folder name
    WEB_BUILD_OPTIONS: "" # prefix for npm build command
    TARGET: "{{.BUILD_DIR_ANY}}/web/{{.COMPONENT}}{{.WEB_TARGET_SUFFIX}}"
  deps:
    - go-bindata-assetfs
  dir: "cmd/{{.COMPONENT}}"
  cmds:
[...]

PR #56, opened 2017-07-20
Resolve issues with template- and dynamic variable resolution order and improve testing

smyrman:
UPDATE: This PR now ensures:
- Template evaluation finds placed _before_ copying the task for execution
- Template evaluation finds place _before_ dynamic variables are resolved
- Variable _priority order_ is in line with the documentation (also for default/named task)
- Template evaluation gets access to the following variables:
   - Taskvars.yml: uncompiled static Taskvars.yml overwritten by environment variables.
   - Task variables: resolved Taskvars.yml overwritten by uncompiled static task variables, environment variables and fully _resolved_ call parameters.
   - Everything _inside_ a task: fully resolved variables for the task.

PS! This is now possible:
```yaml
task:
   vars:
       A: "no templating or sh:"
       B: "{{.A}}"
```

These changes should finally resolve issue #40, and ensure it keeps working.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -14,6 +14,121 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// fileContentTest provides a basic reusable test-case for running a Taskfile
+// and inspect generated files.
+type fileContentTest struct {
+	Dir       string
+	Target    string
+	TrimSpace bool
+	Files     map[string]string
+}
+
+func (fct fileContentTest) name(file string) string {
+	return fmt.Sprintf("target=%q,file=%q", fct.Target, file)
+}
+
+func (fct fileContentTest) Run(t *testing.T) {
+	for f := range fct.Files {
+		_ = os.Remove(filepath.Join(fct.Dir, f))
+	}
+
+	e := &task.Executor{
+		Dir:    fct.Dir,
+		Stdout: ioutil.Discard,
+		Stderr: ioutil.Discard,
+	}
+	assert.NoError(t, e.ReadTaskfile(), "e.ReadTaskfile()")
+	assert.NoError(t, e.Run(fct.Target), "e.Run(target)")
+
+	for name, expectContent := range fct.Files {
+		t.Run(fct.name(name), func(t *testing.T) {
+			b, err := ioutil.ReadFile(filepath.Join(fct.Dir, name))
+			assert.NoError(t, err, "Error reading file")
+			s := string(b)
+			if fct.TrimSpace {
+				s = strings.TrimSpace(s)
+			}
+			assert.Equal(t, expectContent, s, "unexpected file content")
+		})
+	}
+
+}
+
+func TestVars(t *testing.T) {
+	tt := fileContentTest{
+		Dir:       "testdata/vars",
+		Target:    "default",
+		TrimSpace: true,
+		Files: map[string]string{
+			// hello task:
+			"foo.txt":              "foo",
+			"bar.txt":              "bar",
+			"baz.txt":              "baz",
+			"tmpl_foo.txt":         "foo",
+			"tmpl_bar.txt":         "<no value>",
+			"tmpl_foo2.txt":        "foo2",
+			"tmpl_bar2.txt":        "bar2",
+			"shtmpl_foo.txt":       "foo",
+			"shtmpl_foo2.txt":      "foo2",
+			"nestedtmpl_foo.txt":   "{{.FOO}}",
+			"nestedtmpl_foo2.txt":  "foo2",
+			"foo2.txt":             "foo2",
+			"bar2.txt":             "bar2",
+			"baz2.txt":             "baz2",
+			"tmpl2_foo.txt":        "<no value>",
+			"tmpl2_foo2.txt":       "foo2",
+			"tmpl2_bar.txt":        "<no value>",
+			"tmpl2_bar2.txt":       "<no value>",
+			"shtmpl2_foo.txt":      "<no value>",
+			"shtmpl2_foo2.txt":     "foo2",
+			"nestedtmpl2_foo2.txt": "{{.FOO2}}",
+			"equal.txt":            "foo=bar",
+			"override.txt":         "bar",
+		},
+	}
+	tt.Run(t)
+	// Ensure identical results when running hello task directly.
+	tt.Target = "hello"
+	tt.Run(t)
+}
+
+func TestVarsInvalidTmpl(t *testing.T) {
+	const (
+		dir         = "testdata/vars"
+		target      = "invalid-var-tmpl"
+		expectError = "template: :1: unexpected EOF"
+	)
+
+	e := &task.Executor{
+		Dir:    dir,
+		Stdout: ioutil.Discard,
+		Stderr: ioutil.Discard,
+	}
+	assert.NoError(t, e.ReadTaskfile(), "e.ReadTaskfile()")
+	assert.EqualError(t, e.Run(target), expectError, "e.Run(target)")
+}
+
+func TestParams(t *testing.T) {
+	tt := fileContentTest{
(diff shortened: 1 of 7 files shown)


smyrman:
Added a _potential_ fix to the tests. I am still not sure if this solves all cases, so we might need some additional tests for this, but at least it improves on the situation.

variables.go
```diff
@@ -176,13 +178,14 @@ func (e *Executor) handleDynamicVariableContent(v Var) (string, error) {
...
 	r := varReplacer{vars: vars}
+
 	new := Task{
 		Desc:      r.replace(t.Desc),
 		Sources:   r.replaceSlice(t.Sources),
 		Generates: r.replaceSlice(t.Generates),
 		Status:    r.replaceSlice(t.Status),
 		Dir:       r.replace(t.Dir),
-		Vars:      r.replaceVars(t.Vars),
+		Vars:      nil, // no longer needed.
```
smyrman: I remove this, because I am afraid it's just a source of confusion when it's no longer used. If we where to set it to _anything_, it should probably be the full, passed in `vars`.

variables.go
```diff
@@ -121,6 +121,8 @@ func (e *Executor) getVariables(t *Task, call Call) (Vars, error) {
 	if err := merge(call.Vars); err != nil {
 		return nil, err
 	}
+	r := varReplacer{vars: result}
+	result = r.replaceVars(result)
```
smyrman: The way I have implemented things with this proposal, I am afraid you might get some slightly _unexpected_ behavior, although I am not sure if it's really a problem:

Taskvars.yml:
```
A = "{{.B}}!" // B not defined here.
```

Taskfile.yml:
```
default:
  vars:
    B: "aha"
  cmds:
    - echo "{{.A}}}"
```
`task default` would echo "aha!".

This basically means that no templates are resolved before _all_ variables are resolved. I don't really know if this is the desired behavior, but it's better than no variables getting compiled. I think personally, something along the lines of the following _might_ be more in line with what _I_ would expect:

1. At startup, Taskvars and environment variables are merged/resolved and compiled. These get's stored on the executor (when/if support for `set` is dropped). As long as `set` is needed, it _might_ be better to store a map of _environment variables_ in the Context. See *(1).
2. For each task, fetch task variables, override with _compiled_ call variables and executor level variables, and compile the remaining variables.
3. Retrieve new task with "ReplaceVariables" (although I think this method might need a better name to be honest; it does way more than replacing variables).

@andreynering, what's your opinion on how this should work?

*(1) To ensure better concurrent behavior when relying on the `set` keyword, It might be better to store a new _copy_ of the overwritten environment variables (as map) in the Context for each call to `set`, and use that context for any task running _after_ `set` in the dependency graph. I.e., ensure this always gives the same results, no matter wich order task-a and task-b ends up being called in:
```yaml
default:
  deps:
    - task-a
    - task-b
task-a:
  deps:
    - task: set-evar
      vars: {SET: "A"}
  cmds:
    - echo "$EVAR" // use context to ensure this always echo "A"
task-b:
  deps:
    - task: set-evar
      vars: {SET: "B"}
  cmds:
    - echo "$EVAR"  // use context to ensure this always echo "B"
set-evar:
  cmds:
    - echo {{.SET}}
  set: EVAR
```
is the only reason you might want to store this in
- For each task,

andreynering:
Hi @smyrman 

I haven't personally tried it yet. I plan to do with time.

> The way I have implemented things with this proposal, I am afraid you might get some slightly unexpected behavior, although I am not sure if it's really a problem:
>
> Taskvars.yml:
>
> ```yml
> A = "{{.B}}!" // B not defined here.
> Taskfile.yml:
> ```
>
> ```yml
> default:
>   vars:
>     B: "aha"
>   cmds:
>     - echo "{{.A}}}"
> ```

I think the above behavior *is* unexpected. It seems confusing to me that Taskvars have access to any variable defined in a task. But I'm open to discuss this, if you think this is desirable.

> 1. At startup, Taskvars and environment variables are merged/resolved and compiled. These get's stored on the executor (when/if support for set is dropped). As long as set is needed, it might be better to store a map of environment variables in the Context. See *(1).
> 2. For each task, fetch task variables, override with compiled call variables and executor level variables, and compile the remaining variables.
> 3. Retrieve new task with "ReplaceVariables" (although I think this method might need a better name to be honest; it does way more than replacing variables).

Agree completely with (3). (1) and (2) makes sense to me, I hope it don't make the code much more complicated.

I still haven't tried the last issue you posted on #40, but I think it may be simple to fix. Maybe we should focus on fixing that in this PR, and pushing refactorings to future commits?

Thank you for adding a test showing the issue. It will help to find the fix.

smyrman:
> I think the above behavior is unexpected. It seems confusing to me that Taskvars have access to any variable defined in a task.

I agree.

> Maybe we should focus on fixing that in this PR, and pushing refactorings to future commits?

We can add some tests for the unexpected behavior as well maybe, and then see how much we need to do to get rid of that. I won't look more on it today though. Maybe tomorrow.

smyrman:
Ok, noticed now (on my bygger real-world Taskfile), that `handleDynmaicVariables` is also called _before_ variables are compiled in the Variable `sh` section. Thus it will try to run commands with `{{.SOME_VAR}}`. We should have a test for that also.

smyrman:
Pushed a new version that should resolve all issues. I will push some more tests for it before I merge - given you think it looks OK.

andreynering:
The code looks fine. Thank you.

You can merge this when done.

smyrman:
After implementing this, I saw that the _documented_ (not actual) priority order of variables places task  variable priority _higher_ than `Taskvars` variables. I will change this PR to match what's documented, and have suggested some improvements for later in #57.

smyrman:
@andreynering, I had do quite a bit of changes to resolve some final last-minute issues, do you want to take another look to make sure it looks fine?

task_test.go
```diff
@@ -14,6 +14,106 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// fileContentTest provides a basic reusable test-case for running a Taskfile
+// and inspect generated files.
+type fileContentTest struct {
```
andreynering: 👍

task_test.go
```diff
@@ -14,6 +14,106 @@ import (
...
+	}
+	tt.Run(t)
+	// Ensure identical results when running hello task directly.
+	tt.Target = "hello"
+	tt.Run(t)
+}
+
+func TestParams(t *testing.T) {
+	tt := fileContentTest{
+
```
andreynering: Remove blank line

variables.go
```diff
@@ -95,33 +100,79 @@ func init() {
...
+			for k, v := range src {
+				dest[k] = v
+			}
+		}
+	}
+	varsKeys := func(srcs ...Vars) []string {
+		m := make(map[string]bool)
+		for _, src := range srcs {
+			for k := range src {
+				m[k] = true
```
```go
m := make(map[string]struct{})
for _, src := range srcs {
	for k := range src {
		m[k] = struct{}{}
```

variables.go
```diff
@@ -95,33 +100,79 @@ func init() {
...
+	}
+	replaceVars := func(dest Vars, keys []string) {
+		r := varReplacer{vars: dest}
+		for _, k := range keys {
+			v := dest[k]
+			dest[k] = Var{
+				Static: r.replace(v.Static),
+				Sh:     r.replace(v.Sh),
+			}
+		}
```
andreynering: `return r.err`?

smyrman: Good catch!

variables.go
smyrman: Yes, that's better.

variables.go
smyrman: fixed

smyrman:
Ok, I will merge this now.

PR #60, opened 2017-08-04
Cancel and restart long running process on file change by using context.WithCancel(..)

JackMordaunt:
Tasks are now called in a goroutine meaning the order of completion is *not* guaranteed. 
I'm not sure if task order is a requirement.

The reason they have to be called in a `goroutine` is because long running tasks would block `watchTasks(..)`, such that it doesn't read-in file change events.

The `goroutines` don't leak because the `context.CancelFunc` cancels the currently executing Task, causing `Executor.RunTask(..)` to return, thus exiting the `goroutine`.

I tested it locally and it appears to work correctly: long running processes are cancelled and re-started when a file change event occurs.

Resolves #60

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -13,18 +13,11 @@ import (
 func (e *Executor) watchTasks(args ...string) error {
 	e.printfln("task: Started watching for tasks: %s", strings.Join(args, ", "))
 
-	taskCancellers := map[string]context.CancelFunc{}
-
+	ctx, cancel := context.WithCancel(context.Background())
 	for _, a := range args {
-		// Use of goroutines means task order is not guaranteed.
-		// Not sure if task order is a requirement. (jackmordaunt)
-		go func(a string) {
-			ctx, cancel := context.WithCancel(context.Background())
-			taskCancellers[a] = cancel
-			if err := e.RunTask(ctx, Call{Task: a}); err != nil {
-				e.println(err)
-			}
-		}(a)
+		if err := e.RunTask(ctx, Call{Task: a}); err != nil {
+			e.println(err)
+		}
 	}
 
 	watcher, err := fsnotify.NewWatcher()
@@ -46,17 +39,12 @@ loop:
 	for {
 		select {
 		case <-watcher.Events:
-			for _, a := range args {
-				if cancel, ok := taskCancellers[a]; ok {
-					cancel()
-				}
-			}
+			cancel()
+			ctx, cancel = context.WithCancel(context.Background())
 			for _, a := range args {
 				// Use of goroutines means task order is not guaranteed.
 				// Not sure if task order is a requirement. (jackmordaunt)
 				go func(a string) {
-					ctx, cancel := context.WithCancel(context.Background())
-					taskCancellers[a] = cancel
 					if err := e.RunTask(ctx, Call{Task: a}); err != nil {
 						e.println(err)
 					}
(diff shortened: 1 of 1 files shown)


watch.go
```diff
@@ -13,12 +13,18 @@ import (
...
-	// run tasks on init
+	taskCancellers := map[string]context.CancelFunc{}
+
 	for _, a := range args {
-		if err := e.RunTask(context.Background(), Call{Task: a}); err != nil {
-			e.println(err)
-			break
-		}
+		// Use of goroutines means task order is not guaranteed.
+		// Not sure if task order is a requirement. (jackmordaunt)
```
smyrman: Running them in parallel makes sense to me. @andreynering?

watch.go
```diff
@@ -13,12 +13,18 @@ import (
...
+		go func(a string) {
+			ctx, cancel := context.WithCancel(context.Background())
+			taskCancellers[a] = cancel
```
smyrman: I suggest we create the context and update taskCancellers[a] _outside_ of the concurrent `go` function. Modifying a map concurrently without a mutex is subject to data-race conditions or worse.

watch.go
```diff
@@ -13,12 +13,18 @@ import (
...
+	taskCancellers := map[string]context.CancelFunc{}
+
 	for _, a := range args {
-		if err := e.RunTask(context.Background(), Call{Task: a}); err != nil {
-			e.println(err)
-			break
-		}
+		// Use of goroutines means task order is not guaranteed.
+		// Not sure if task order is a requirement. (jackmordaunt)
+		go func(a string) {
```
smyrman: You don't need to pass along a, you could capture it instead.
```go
for _, a := range args {
    a := a // capture range variable
    ...
}
```

This is relevant to change due to my next comment on line 23.

watch.go
```diff
@@ -41,11 +47,21 @@ loop:
 		select {
 		case <-watcher.Events:
 			for _, a := range args {
```
smyrman: could we just loop through taskCancellers directly? If it's set in a non-concurrent manner, we don't need as many checks as here.

watch.go
```diff
@@ -41,11 +47,21 @@ loop:
...
-				if err := e.RunTask(context.Background(), Call{Task: a}); err != nil {
```
smyrman: As long as taskCancellers is modified concurrently, this is subject to data-race, and some tasks might (theoretically) be started _after_ ok returns `false`.

watch.go
smyrman: After some thought... Could we also actually simplify this just use the same context (instead of taskCancellers map), and cancel all tasks through the same cancel function?

watch.go
JackMordaunt: Curious, are there any downsides to passing along the variable other than memory allocations?

watch.go
JackMordaunt: Great idea to use a single context. Updated the pull request to do just that.

JackMordaunt:
With the current commit, task outputs noisy 'error' messages when using watch flag:
`task: Failed to run task "build": context canceled`
Should this message be silenced, since it is actually intended behaviour, not an error?

andreynering:
@JackMordaunt This should do the trick. Print only if it's not a context error:

```go
switch err {
case context.Canceled, context.DeadlineExceeded:
        // supress error
default:
        e.println(err)
}
```

It's also missing args assigning:

```go
for _, a := range args {
        a := a // this is important to prevent goroutine using the value of the next iteration
```

andreynering:
@JackMordaunt Great work, thank you very much!

Issue #61, opened 2017-08-16
`env` appears broken as values are set with extra square braces


smyrman:
Minimal Taskfile.yml:

```yml
default:
  env:
    GOOS: linux
    GOARCH: amd64
    CGO_ENABLED: "0"
  cmds:
    - echo "GOOS='$GOOS' GOARCH='$GOARCH' CGO_ENABLED='$CGO_ENABLED'"
```

First, ensuring that task is up-to-date:
```sh
 % go get -u -v github.com/go-task/task/cmd/task
github.com/go-task/task (download)
```

When running task, I get extra curly braces and a space around the env value (`{`,` }`):
```sh
% task                                         
task: No argument given, trying default task
echo "GOOS='$GOOS' GOARCH='$GOARCH' CGO_ENABLED='$CGO_ENABLED'"
GOOS='{linux }' GOARCH='{amd64 }' CGO_ENABLED='{0 }'
```

I would have expected the following instead:
```sh
% task                                         
task: No argument given, trying default task
echo "GOOS='$GOOS' GOARCH='$GOARCH' CGO_ENABLED='$CGO_ENABLED'"
GOOS='linux' GOARCH='amd64' CGO_ENABLED='0'
```

PR #62, opened 2017-08-16
Fixes env: remove square braces and evaluate shell

smyrman:
Fixes bug #61, and makes `env` work more similar to `vars` by allowing
dynamic shell values to be evaluated as part of `CompiledTask`.


testdata/env/env.txt
```diff
@@ -0,0 +1 @@
+GOOS='linux' GOARCH='amd64' CGO_ENABLED='0'
```
smyrman: added by mistake. Add to .gitignore

andreynering:
Thanks :smile:

PR #64, opened 2017-09-03
Added suport for multiline variables from sh

### LINKED ISSUE #63: Allow multiline output in Dynamic vars
There is only one way of setting a multiline variable in task from shell output that I am aware of, and that is using the deprecated `set`.

Using the `sh: program` syntax will (deliberately) give an error. I don't know what the reasoning for that is, but there are many good use cases for _allowing_ multiline variables to be set from a shell command, such as fetching private keys from a keychain. In this case you would also prefer if you didn't need to store the private key in a temp file on disk.

smyrman:
Fixes #63.

~Instead of giving an error on multiline results from sh, the results are now concatenated to one line by default, separated by a single space. Trailing and leading white space are stripped like before.~
    
~In adition, some flags have been added for advanced use cases to turn off line concationaton and/or trimming off leading/trailing white-space.~

Multiline results are now allowed, while the last newline is stripped to allow the output of most commands to work better as variable values. Some helper functions are added to make it easier to deal with multi-line variables in various ways.

### DIFF
--- a/variables.go
+++ b/variables.go
@@ -15,19 +15,30 @@ import (
 )
 
 var (
-	// TaskvarsFilePath file containing additional variables
+	// TaskvarsFilePath file containing additional variables.
 	TaskvarsFilePath = "Taskvars"
-	// ErrMultilineResultCmd is returned when a command returns multiline result
-	ErrMultilineResultCmd = errors.New("Got multiline result from command")
 )
 
-// Vars is a string[string] variables map
+var (
+	// ErrCantUnmarshalVar is returned for invalid var YAML.
+	ErrCantUnmarshalVar = errors.New("task: can't unmarshal var value")
+)
+
+// Vars is a string[string] variables map.
 type Vars map[string]Var
 
-// Var represents either a static or dynamic variable
-type Var struct {
-	Static string
-	Sh     string
+func getEnvironmentVariables() Vars {
+	var (
+		env = os.Environ()
+		m   = make(Vars, len(env))
+	)
+
+	for _, e := range env {
+		keyVal := strings.SplitN(e, "=", 2)
+		key, val := keyVal[0], keyVal[1]
+		m[key] = Var{Static: val}
+	}
+	return m
 }
 
 func (vs Vars) toStringMap() (m map[string]string) {
@@ -43,12 +54,13 @@ func (vs Vars) toStringMap() (m map[string]string) {
 	return
 }
 
-var (
-	// ErrCantUnmarshalVar is returned for invalid var YAML
-	ErrCantUnmarshalVar = errors.New("task: can't unmarshal var value")
-)
+// Var represents either a static or dynamic variable.
+type Var struct {
+	Static string
+	Sh     string
+}
 
-// UnmarshalYAML implements yaml.Unmarshaler interface
+// UnmarshalYAML implements yaml.Unmarshaler interface.
 func (v *Var) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var str string
 	if err := unmarshal(&str); err == nil {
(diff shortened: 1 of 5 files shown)


andreynering:
Thanks again.

> In adition, some flags have been added for advanced use cases to turn off line concationaton and/or trimming off leading/trailing white-space.

I'm a bit in doubt about what approach to push.

I kind of dislike these boolean flags, they're a bit too verbose. Now we have two, in the future, if we have more use cases, then we'll have many?

This is why I thought about turning it in a slice. Then the use can programmatically do exactly what they want. No flag needed.

> Replacing the Var.Static attribute with a []string is a bigger and also a breaking change. E.g. it would break both set and cases where the end-users define multiline values in task via yaml

I think in these cases we would just push this string as the first element of the slice, to keep the current behavior.

---

I want your opinion if that would solve your use case from #63.

andreynering:
OK, let's simplest thing. One can just split the result if they want, as you [suggested before](https://github.com/go-task/task/issues/63#issuecomment-326799205)

- Don't change to slice, it's simpler as is :+1:
- Get rid of the multiline flag and keep line break by default. The user can replace it if they want
- What about trimming? Maybe just trim if there's a single line break in the end of the string?

smyrman:
> Get rid of the multiline flag and keep line break by default. The user can replace it if they want

Alright. 👍 I can add some convenience functions to make it easy to concat lines or split them (including Windows styled newlines).

> What about trimming? Maybe just trim if there's a single line break in the end of the string?

I guess that makes sense. That should make most command output work as expected (there is usually just one trailing newline), but leave some leeway for plausible use cases where you would want some of the whitespace kept, e.g. code generation.

andreynering:
>  I can add some convenience functions to make it easy to concat lines or split them (including Windows styled newlines).

Maybe something for the future, since even on Windows files often use `\n` instead of `\r\n`.

About the trimming, maybe always trim the last line break would be better because:

1) Unix tools like `cat` adds a trailing line break anyway
2) `strings.Join` adds a final empty string to slice with no trim: https://play.golang.org/p/K0clOebD-s

smyrman:
> Maybe something for the future, since even on Windows files often use \n instead of \r\n.

Yeah, but you just need to replace "\r\n"  with "\n" and then do the split. This is similar to how e.g. the Go line scanner works, although it uses "peek" to be more memory efficent.

I wrote some, but a bit unsure about the best way to name them, as go-taks appears to use `PascalCase` and `UPPERCASE`, where as sprig uses `camelCase` for functions.

variables.go
```diff
@@ -347,3 +317,41 @@ func (r *varReplacer) replaceVars(vars Vars) Vars {
...
+		"OS":   func() string { return runtime.GOOS },
+		"ARCH": func() string { return runtime.GOARCH },
+		"catLines": func(s string) string {
+			s = strings.Replace(s, "\r\n", " ", -1)
+			return strings.Replace(s, "\n", " ", -1)
+		},
+		"splitLines": func(s string) []string {
+			s = strings.Replace(s, "\r\n", "\n", -1)
+			return strings.Split(s, "\n")
+		},
```
smyrman: 👆

variables.go
```diff
@@ -15,19 +15,30 @@ import (
...
+		env = os.Environ()
+		m   = make(Vars, len(env))
+	)
+
+	for _, e := range env {
+		keyVal := strings.SplitN(e, "=", 2)
+		key, val := keyVal[0], keyVal[1]
+		m[key] = Var{Static: val}
+	}
+	return m
```
smyrman: Did some re-ordering in this file to get a better overview. Basically the ordering rule is to place methods and creator functions right below the type definitions.

Don't know if you agree with this ordering, or if you prefer a different style @andreynering

variables.go
```diff
@@ -43,12 +54,13 @@ func (vs Vars) toStringMap() (m map[string]string) {
...
-	ErrCantUnmarshalVar = errors.New("task: can't unmarshal var value")
-)
+// Var represents either a static or dynamic variable.
+type Var struct {
+	Static string
+	Sh     string
+}
 
-// UnmarshalYAML implements yaml.Unmarshaler interface
+// UnmarshalYAML implements yaml.Unmarshaler interface.
```
smyrman: The trailing period is recommended by [Effective Go](https://golang.org/doc/effective_go.html), or at least the way I interpret it:

> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

The standard lib uses a period consistently throughout all libraries as far as I have seen. Note that I am only talking about doc comments, not code comments.

variables.go
```diff
@@ -60,51 +72,20 @@ func (v *Var) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return nil
 	}
 
-	var sh struct {
-		Sh string
-	}
+	var sh struct{ Sh string }
```
smyrman: Should revert this.

variables.go
smyrman: If you prefer the `PascalCase`, I can rename them, or if you prefer consistency with `sprig`, we can rename all of them, but leave backwards compatibility:

```go
 	taskFuncs := template.FuncMap{
		(...)
		"fromSlash": func(path string) string {
			return filepath.FromSlash(path)
		},
		"toSlash": func(path string) string {
			return filepath.ToSlash(path)
		},
		"exeExt": func() string {
			if runtime.GOOS == "windows" {
				return ".exe"
			}
			return ""
		},
		// IsSH is deprecated.
		"IsSH": func() bool { return true },
	}
 	// Deprecated aliases for renamed functions.
	task["FromSlash"] = taks["fromSlash"]
	task["ToSlash"] = taks["toSlash"]
	task["ExeExt"] = taks["extExt"]
```

We can comment the backwards compatibility only very briefly in the README.

variables.go
smyrman: done

variables.go
```diff
@@ -215,19 +184,22 @@ func (e *Executor) handleShVar(v Var) (string, error) {
...
 
-	result := strings.TrimSuffix(stdout.String(), "\n")
-	if strings.ContainsRune(result, '\n') {
-		return "", ErrMultilineResultCmd
+	result := stdout.String()
+
+	// Trim a single trailing newline to make the result easier to use in shell
+	// commands.
+	if strings.HasSuffix(result, "\n") {
+		result = result[0 : len(result)-1]
```
```go
result = strings.TrimSuffix(result, "\n")
```

If also not required.

variables.go
andreynering: Agree with renaming them. The reason this is not consistent is because spring was added after some of the other funcs.

andreynering:
Two notes, then you can merge this. Thank you for your work and patience.

Feel free to push directly to master if/when you have small fixes and are confident. Specially for documentation and comments.

variables.go
smyrman: Yeah, I assumed it was historical reasons:-)

PR #71, opened 2017-09-29
An example covering passing values in the environment and value quoting

robinbryce:
The behavior when constructing and passing variables to processes invoked in cmds tripped me up. This example might save someone else some time.

I'm not entirely sure I have chosen a good location for the example in the repo, but I thought this would be clearer as a PR rather than an issue which isn't a bug.

### DIFF
--- a/example/cgo_envvars/Taskfile.yml
+++ b/example/cgo_envvars/Taskfile.yml
@@ -1,0 +1,24 @@
+# This applies generaly to environment variables passed in this way, not just
+# CGO_ xxx
+#
+right:
+  # Notice the quotes guarding the variable references on the second line of
+  # the cmds.
+  cmds:
+    - |
+      CGO_LDFLAGS="-L/some/path -laaa -lbbb -lccc"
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @robinbryce 

I think documentation is very important, and like the idea of having a set of advanced examples. I'm not sure where it would be better to have it: in a folder or in a new repo like https://github.com/go-task/examples

---

About this use case, I think you should consider use [the `env` attribute to set environment][env]. Like this:

```yml
show-env:
  cmds:
    - echo $CGO_LDFLAGS
  env:
    CGO_LDFLAGS: "-L/some/path -laaa -lbbb -lccc"
```

[env]: https://github.com/go-task/task#environment

robinbryce:
Yes, it didn't feel right putting them in task/example. The repo idea is much better, for one thing it then becomes possible to execute the examples against different release versions of task ...

Do I need to do anything to make that happen ?

andreynering:
@robinbryce That repo was just created. Pull Requests are welcome.

PR #76, opened 2017-11-19
Improve nested variables support

andreynering:
e.g. the situation where a variable concatenates another variable, either from the current or previous step.

/cc @smyrman 

Questions:
- What should be the definitive order of precedence of variables?
- What syntax is prefered for default values?


smyrman:
> I know go-task is not currently following semantic versioning, 

Some of this could be handled in #77 as well.

smyrman:
Also, the code itself looks like it's doing the right thing.

PR #80, opened 2017-11-29
Add indirections through functions

sascha-andres:
Pull request for discussion

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -96,6 +96,7 @@ func main() {
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
 	}
+
 	if err := e.ReadTaskfile(); err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


andreynering:
Agree with @smyrman 

Some thoughts:

- I don't think many people will want to customize how Task works. This would lead to fragmentation. If a use-case is common, a discussion for changes on Task is resonable to allow it by default;
- It's OK to just execute a Taskfile as a library, but that support is unnoficial currently;
- Changes like this would make refactoring and improvements harder (as @smyrman pointed above).

@sascha-andres Sorry for consuming your time, I think we should keep Task maintainable. I'm looking forward to improve those use cases, though.

smyrman:
IF anyone is interested though, splitting out `Taskfile` and creating an `Executor` interface would be a really useful PR though, if everyone agrees it's a good idea. Maybe when we reach `v2` we could guarantee semantic-versioning based backwards-compatibility, also for the library?

The most important aspect of that is mainly to keep as much as possible as private functions/methods/types.

PR #85, opened 2018-01-10
Added "passing variables to dependencies"

atar-axis:
Added "passing variables to dependencies"


andreynering:
Thanks for the patch!

Issue #106, opened 2018-03-22
Add a uniform way to override tasks  


skurtzemann:
Hi,

First thing : **thanks for this amazing tool !**

## Request

Ability to have a global `Taskfile` override using `Taskfile_override.yml`.

## Why that's 

For the moment the only way to override some tasks is to used `Taskfile_{{GOOS}}.yml` according your environment.

In my company we have define a workflow using a "global" `Taskfile.yml` to share tasks with each others.

This "global" file is generated by our CI and contains a lot of tasks.
So we need a way to simply override and improve some tasks present in the "global" `Taskfile` when we are in development.

Our dev team members use differents OS (Windows, Unix, MacOS) so we don't have a uniform way to override tasks.

### How to use it ?

If `Taskfile_override.yml` is present we override or amend tasks in `Taskfile.yml`
But we don't changes the current override using `Taskfile_{{GOOS}}.yml`.

So we have the follow possible "overrides"
```
Taskfile_{{GOOS}}.yml
        |
    (override)
        |
        v
Taskfile_override.yml
        |
    (override)
        |
        v
Taskfile.yml
```

## Finally

What do you think ?

andreynering:
Hi @skurtzemann!

I think this use case could be solved with #98. What do you think?

skurtzemann:
Yes you right @andreynering, this use case can be solved with the `imports` statement.
Have you an idea when this feature will be released ?

skurtzemann:
I already make a PR (#108) for this feature ;)

andreynering:
> Have you an idea when this feature will be released?

Unfortunately, I don't think I'll have time to work in it in the next couple of weeks.

PR #108, opened 2018-03-25
Ability to define an "Taskfile_override.yml"

skurtzemann:
Pull request for #106

### DIFF
--- a/taskfile.go
+++ b/taskfile.go
@@ -22,6 +22,19 @@ func (e *Executor) readTaskfile() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks for your PR.

But I think we should maybe just have #98 instead.

I mean, maybe we should even deprecated `Taskfile_{{GOOS}}.yml`, since I've could easily do this:

```yml
imports:
  - Taskfile_{{GOOS}}.yml
```

skurtzemann:
Ok no problem ;)

Issue #103, opened 2018-03-11
Add Bash, ZSH and Windows command line completion


harobed:
Add bash and zsh autocomplete feature.

andreynering:
@harobed I'd love to receive contributions in this one, since I don't actually know how that works. Let me know if you want to work in it.

sawadashota:
Simply, I use zsh completion.
https://github.com/sawadashota/go-task-completions

andreynering:
Edited title to include Windows as asked on #127

rmmcnulty9:
I am very interested in this feature. I have a very simple and not very complete command I use for bash:

```
# Taskfile.yml autocomplete - ASSUMES ALL TASKS ARE LOWERCASE
complete -W "\`test -e Taskfile.yml && grep -oE '^\s\s[a-z0-9_-]+:([^=]|$)' Taskfile.yml | sed 's/[^a-zA-Z0-9_-]*$//'\`" task
```

This does not work with:
- included files
- uppercase tasks, I went with lowercase otherwise it would return some of the variables I have defined

This also seems like a nice resource: https://iridakos.com/tutorials/2018/03/01/bash-programmable-completion-tutorial.html

smyrman:
I thing I have linked this previously, but another resource is:
- https://github.com/posener/complete

bfarayev:
I've made some work on bash autocompletion here: https://github.com/bfarayev/task/blob/feature/autocomplete/completion/task.bash

I've tested it on my fork of task, so need to check if it'd work with the latest release. Happy to submit a PR after testing.

andreynering:
@bfarayev Pull requests are more than welcome!

Chi-teck:
Can this be implemented as new Task command?
Example:
`task generate-completion --bash >> ~/.bashrc`

andreynering:
> Can this be implemented as new Task command?
> Example:
> `task generate-completion --bash >> ~/.bashrc`

Does other tools usually do that?

Chi-teck:
Drush 8 used to have `init` command that installed completion file.
https://github.com/drush-ops/drush/blob/8.3.2/commands/core/init.drush.inc#L83

Also https://github.com/bamarni/symfony-console-autocomplete is widely used tool for generating completions for Symfony Console commands.

Typically such approach is used in applications that distributed as a single file executable. If an application has installation/uninstallation procedure it's probably better to install/uninstall the completions automatically.

andreynering:
[...]

PR #111, opened 2018-04-27
Add zsh completion

sawadashota:
Related to #103 

Put file like `docker/cli`
https://github.com/docker/cli/blob/master/contrib/completion/zsh/_docker

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -1,0 +1,24 @@
+#compdef task
+
+# Listing commands from Taskfile.yml
+function __list() {
+    local -a scripts
+
+    if [ -f Taskfile.yml ]; then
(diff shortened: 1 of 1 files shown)


smyrman:
I think this looks like a nice start.

Longer term, it might be slightly better to go for a self-completing program (see e.g. https://github.com/posener/complete) so that the completion options and program flags does not need to be maintained several places.

andreynering:
Hi @sawadashota 

Thanks for your contribution! This will definitely improve the UX, specially if we can integrate it with Homebrew and Snapcraft.

@smyrman Yeah, we can consider that in the future, but until that happen I'll accept this one. 🙂

PR #118, opened 2018-06-13
This pull request remove vendor direcotry

SebastianPozoga:
Storage dependencies in project is bad practice

https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

> Should I commit my vendor directory?
> It's up to you:
> 
> Pros
> 
> It's the only way to get truly reproducible builds, as it guards against upstream renames, deletes and commit history overwrites.
> You don't need an extra dep ensure step to sync vendor/ with Gopkg.lock after most operations, such as go get, cloning, getting latest, merging, etc.
> Cons
> 
> Your repo will be bigger, potentially a lot bigger, though prune can help minimize this problem.
> PR diffs will include changes for files under vendor/ when Gopkg.lock is modified, however files in vendor/ are hidden by default on GitHub.
>


andreynering:
Hi @SebastianPozoga!

Thanks for the PR.

This was intended to have deterministic builds. Also, it makes it more convenient to users that install it using `go get`.

I'm closing this in favor of https://github.com/go-task/task/issues/119 if we can keep these two aspects with vgo.

PR #123, opened 2018-07-10
Add an IgnoreError option for Tasks and Cmds

Eun:
It happens that you want to run commands that may fail but your build should keep running.

For this case we need a way to skip errors during execution.

Example:

```yml
CreateOutputDirectory:
  cmds:
    # maybe bin does already exist, so skip any errors
    - cmd: mkdir bin
      ignoreError: true
```

> This PR also contains the possibility to ignore errors for complete tasks. See the `testdata/ignore_errors/Taskfile.yml` file


internal/taskfile/cmd.go
```diff
@@ -38,21 +40,25 @@ func (c *Cmd) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return nil
 	}
 	var cmdStruct struct {
-		Cmd    string
-		Silent bool
+		Cmd         string
+		Silent      bool
+		IgnoreError bool `yaml:"ignoreError"`
```
andreynering: YAML tag should be `ignore_error`

task.go
```diff
@@ -200,7 +200,7 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
 
 	switch {
 	case cmd.Task != "":
-		return e.RunTask(ctx, taskfile.Call{Task: cmd.Task, Vars: cmd.Vars})
+		return e.RunTask(ctx, taskfile.Call{Task: cmd.Task, Vars: cmd.Vars, IgnoreError: cmd.IgnoreError || call.IgnoreError})
```
andreynering: Do we really want to delegate this options to tasks being called, too? I think I'd expect `ignore_error: true` to be valid to the *current* task only

testdata/ignore_errors/Taskfile.yml
```diff
@@ -0,0 +1,15 @@
+CmdShouldPass:
```
andreynering: This file is out of convention of other fixtures. Task names should be lowercase, and we're missing space between tasks

Also, wrap them in a `tasks:` object and add `version: '2'` at the top

testdata/ignore_errors/Taskfile.yml
```diff
@@ -0,0 +1,15 @@
...
+  cmds:
+    - cmd: UnknownCommandThatNeverWillExist
+
+TaskShouldPass:
+  cmds:
+    - task: CmdShouldFail
+      ignoreError: true
+TaskShouldFail:
+  cmds:
+    - task: CmdShouldFail
```
andreynering: Missing empty line in the end of this file

smyrman:
> About the feature: I have almost decided to close this PR. The reason is that is pretty simple to ignore errors currently with no extra code necessary, you just need to add || true after commands, and they'll pass even if error

I was thinking the same when I saw the PR, but then I see that the code allows to ignore failing calls to _tasks_ as well, although not sure the semantics or code is right atm.

I.e.  allowing a _task_ to fail, would not be the same as allowing all cmds _inside_ the task to fail (you still want to abort the task at first non-allowed failure) This is in line with what @andreynering comments on line 203; allow-failures should not _propagate_, but instead returned errors from failing tasks should be ignored if allow failure is true. Might require some restructuring.

smyrman:
Just my 5 cents btw.

Eun:
@andreynering 
1. The idea was to not add `|| true` to a bunch of commands that are more less optional and might or might not fail
2. I know that some commands like `mkdir` or `rm` have switches that allow them to suppress errors, however I think task should not rely on options of other programs that its trying to call
3. I think that piping is totally fine for small tasks however It also opens the gate to more complex commands that might become hard to maintain (`ps aux | grep conky | grep -v grep | awk '{print $2}' | xargs kill`)

@smyrman 
Also an idea haven't thought about that

testdata/ignore_errors/Taskfile.yml
Eun: No problem

internal/taskfile/cmd.go
Eun: No problem

task.go
Eun: Related to @smyrman  answer

smyrman:
Regarding 3, there is also the multiline syntax and `set -e/set +e` to enable/disable exit status propagation within the command:
```yaml
version: "2"
tasks:
  mytask:
    cmds:
    - |
       mkdir foobar  # exit status ignored by default for multiline scripts/commands
       set -e        # care about errors from now on.
       ls foobar     # exit status returned if non-zero
       set +e        # ignore errors from now on (default)
       mkdir foobar  # exit status ignored again

**Comment by Eun** [comment]
@andreynering So what do you think should we close this?

**Comment by andreynering** [comment]
@Eun After some thought, I think we should proceed with this feature

Some thoughts:

- I think a task should have this property too, for consistency with other properties like `silent`. Good to add a test case as well
- Would you mind writing documentation for this? It can be short, just to inform users this exists
- Although I'm not 100% certain on the right behavior, I'm inclined to think we should NOT propagate this option to deps/sub-tasks. At least it would be consistent with other options

**Comment by andreynering** [comment]
@smyrman Maybe `set -e` should be made the default in a future release?

**Comment by smyrman** [comment]
> @smyrman Maybe set -e should be made the default in a future release?

The most obvious issue with that is backward-compatibility; if it's changes, some Taskfile somewhere will fail, so it probably deserves to be a v3 update, if it's considered.

Personally, I don't mind what ever the default is, as long as it's documented. People who are used to writing shell scripts, are also used to starting their script with `set -e` or `set -eu`, or which ever is the most suitable for their use-case.

So for now, maybe document it?

File: task.go [hunk 5]
```diff
@@ -188,7 +188,7 @@ func (e *Executor) runDeps(ctx context.Context, t *taskfile.Task) error {
 		d := d
 
 		g.Go(func() error {
-			return e.RunTask(ctx, taskfile.Call{Task: d.Task, Vars: d.Vars})
+			return e.RunTask(ctx, taskfile.Call{Task: d.Task, Vars: d.Vars, IgnoreError: d.IgnoreError})
```
smyrman:
```go
	if err := e.RunTask(ctx, taskfile.Call{Task: d.Task, Vars: d.Vars}); !d.IgnoreError {
		return err
	}
	return nil
```

task.go
smyrman:
```go
case cmd.Task != "":
	var err error
	if err := e.RunTask(ctx, taskfile.Call{Task: cmd.Task, Vars: cmd.Vars}); !cmd.IgnoreError {
		return err
	}
	return nil

File: task.go [hunk 6]
```diff
@@ -212,13 +212,14 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
...
-			Stdout:  stdOut,
-			Stderr:  stdErr,
+			Context:         ctx,
+			Command:         cmd.Cmd,
+			Dir:             t.Dir,
+			Env:             getEnviron(t),
+			Stdin:           e.Stdin,
+			Stdout:          stdOut,
+			Stderr:          stdErr,
+			IgnoreErrorCode: cmd.IgnoreError || call.IgnoreError,
```
smyrman:
```go
if err := execext.RunCommand(&execext.RunCommandOptions{
 	Context: ctx,	
 	Command: cmd.Cmd,
 	Dir:     t.Dir,
 	Env:     getEnviron(t),
 	Stdin:   e.Stdin,
	Stdout:  stdOut,
 	Stderr:  stdErr,
}; !cmd.IgnoreError {
 	return err
}
return nil
``

File: internal/taskfile/call.go [hunk 7]
```diff
@@ -2,6 +2,7 @@ package taskfile
 
 // Call is the parameters to a task call
 type Call struct {
-	Task string
-	Vars Vars
+	Task        string
+	Vars        Vars
+	IgnoreError bool
```
smyrman: Not needed here.

internal/execext/exec.go
```diff
@@ -62,5 +63,12 @@ func RunCommand(opts *RunCommandOptions) error {
...
 	}
-	return r.Run(p)
+	if err = r.Run(p); err != nil {
+		if opts.IgnoreErrorCode {
+			if _, ok := err.(interp.ExitCode); ok {
+				return nil
+			}
+		}
+	}
+	return err
```
smyrman: Can be reverted to `return r.Run(p)` -- handle one level up instead.

internal/taskfile/cmd.go
```diff
@@ -38,21 +40,25 @@ func (c *Cmd) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-		Task string
-		Vars Vars
+		Task        string
+		Vars        Vars
+		IgnoreError bool `yaml:"ignoreError"`
 	}
 	if err := unmarshal(&taskCall); err == nil {
 		c.Task = taskCall.Task
 		c.Vars = taskCall.Vars
+		c.IgnoreError = taskCall.IgnoreError
```
smyrman: not needed.

internal/taskfile/cmd.go
```diff
@@ -66,12 +72,14 @@ func (d *Dep) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-		Task string
-		Vars Vars
+		Task        string
+		Vars        Vars
+		IgnoreError bool `yaml:"ignoreError"`
 	}
 	if err := unmarshal(&taskCall); err == nil {
 		d.Task = taskCall.Task
 		d.Vars = taskCall.Vars
+		d.IgnoreError = taskCall.IgnoreError
```
smyrman: not needed.

internal/taskfile/cmd.go
```diff
@@ -66,12 +72,14 @@ func (d *Dep) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return nil
 	}
 	var taskCall struct {
-		Task string
-		Vars Vars
+		Task        string
+		Vars        Vars
+		IgnoreError bool `yaml:"ignoreError"`
```
smyrman: not needed.

task.go
smyrman: Maybe this is more correct...

```go
err := execext.RunCommand(&execext.RunCommandOptions{
 	Context: ctx,	
 	Command: cmd.Cmd,
 	Dir:     t.Dir,
 	Env:     getEnviron(t),
 	Stdin:   e.Stdin,
	Stdout:  stdOut,
 	Stderr:  stdErr,
}
if _, ok := err.(interp.ExitCode); ok && cmd.IgnoreError {
	return nil
}
return err
```

Eun:
@andreynering Please review the changes.

andreynering:
Thanks @Eun!

I manually merged this with few improvements:

- `ignore_error` also available on a task (no propagation!)
- Did @smyrman's suggestion of handling the logic one level up on the code

Issue #125, opened 2018-07-16
Dry Run Ability


bobhenkel:
- Task version - Task version: 2.0.2
- OS - OSX and Linux

Awesome app!  Is there anyway to have the ability to do a dry run.  For example say I have a tast called build_and_publish

```
task --dry-run build_and_publish
```
By adding dry run all my task place holders would be filled in and I could print out the commands that would be executed if I removed --dry-run. This would help debug a task file.

Thanks
Bob

andreynering:
Hi @bobhenkel,

Thanks for the suggestion

I think this feature request makes sense, and should be relatively easy to implement

andreynering:
Implemented on #126 

See https://github.com/go-task/task#dry-run-mode

PR #126, opened 2018-07-28
Implemented dry run mode

bebbs:
Added a --dry-run flag that compiles and steps through each task, but does not execute them. The commands that would have been run are printed. See #125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,6 +635,11 @@ tasks:
       - echo "This will print nothing" > /dev/null
 ```
 
+## Dry Run Mode
+
+Dry run mode (`--dry-run`) compiles and steps through each task, printing the commands
+that would be run without executing them. This is useful for debugging your Taskfiles.
+
 ## Output syntax
 
 By default, Task just redirect the STDOUT and STDERR of the running commands
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -12,6 +12,7 @@ function __list() {
 
 _arguments \
     '(-d --dir)'{-d,--dir}': :_files' \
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -635,6 +635,11 @@ tasks:
       - echo "This will print nothing" > /dev/null
 ```
 
+## Dry Run Mode
+
+Dry run mode (`--dry-run`) compiles and steps through each task, printing the commands
+that would be run without executing them. This is useful for debugging your Taskfiles.
```
andreynering: 👍 for adding documentation for it

completion/zsh/_task
```diff
@@ -12,6 +12,7 @@ function __list() {
 
 _arguments \
     '(-d --dir)'{-d,--dir}': :_files' \
+    '(--dry-run)'--dry-run \
```
andreynering: 👍 for remembering to update this file

bebbs:
No problem!

```
What about calling it --dry instead of --dry-run?
```

Done.

```
A simple test for it would be nice, but if you don't want to write it, I can do that after merging this PR
```

I couldn't see where to add tests for the effects of flags, apologies if I've missed it. Agree it would be nice to have them, happy for you to merge and add the test if you'd prefer, or let me know where you would like to see one and I'll add it to this PR?

andreynering:
Indeed there's no test for flags currently, but we can have a test on `task_test.go` that test the behavior with the `DryRun` prop set on an `Executor`

Optional in your part, but you can do if you're feeling adventurous 🙂

PR #142, opened 2018-10-26
Fix broken URL to installation doc

aseembajaj:
The links to installation doc were broken, perhaps it was a typo or maybe installation document was relocated from "installation" directory to "installation.md". In either case, making this change fixes the problem.


andreynering:
Hey @aseembajaj, thanks for your contribution!

This was actually intentional, because these docs are optimized for reading on https://taskfile.org and not GitHub.

I just tried and discovered that Docsify just strips `.md` from the links, though. So you patch makes sense, and I'll also apply them to other links not covered by this PR. 🙂

PR #152, opened 2018-12-06
Add Scoop

lambdalisue:
Task is now available on Scoop for Windows users.
https://github.com/lukesampson/scoop-extras/pull/1485

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,8 +28,8 @@ guide to check the full schema documentation and Task features.
 ## Features
 
 - [Easy installation](installation.md): just download a single binary, add to
-  $PATH and you're done! Or you can also install using [Homebrew][homebrew] or
-  [Snapcraft][snapcraft] if you want;
+  $PATH and you're done! Or you can also install using [Homebrew][homebrew],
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @lambdalisue, thanks for opening this PR!

Who will be responsible to update the version on each release? I could do that, but I wouldn't be properly testing if that install method is correctly working on each release. Ideally should be made by someone else.

In that case, I'd also like to mention in the documentation that is this a community owned install method.

lambdalisue:
> Who will be responsible to update the version on each release? I could do that, but I wouldn't be properly testing if that install method is correctly working on each release. Ideally should be made by someone else.

Well, actually I use Windows just for testing so I don't think I notice that the version in scoop-extra is outdated on each release unfortunately.

andreynering:
Hey @lambdalisue, it seems that the owners are actually updating versions: https://github.com/lukesampson/scoop-extras/commit/d7f1a76a15acf8f224ce29015bc2628a845fbdc1

That's great! I'll merge this and do a small change on the wording to make it clear this is a community owned install method.

PR #170, opened 2019-01-30
Passing tests on Windows (GitBash)

mikynov:
This PR is trying to fix two failing tests when running from GitBash on Windows.

1. **TestGenerates**: I believe backslashes must be replaced prior to call [execext.Expand()](https://github.com/mikynov/task/blob/468ff1824380169690441183cd392c51a479f003/internal/execext/exec.go#L74). Otherwise they are just removed what makes wrong filename (path) on Windows (```C:\Users``` becomes ```C:Users```). Windows are fine with slashes in filenames, thus this shouldn't break anything.
    This was an issue on Windows where generated files were not detected correctly and always regenerated (```sources``` and ```generates``` in  [status.IsUpToDate()](https://github.com/mikynov/task/blob/468ff1824380169690441183cd392c51a479f003/internal/status/timestamp.go#L17) were empty).

2. **TestExpand**: Tilda is not expanded properly on Windows. OS-specific Taskfile with ```USERPROFILE``` env. variable make a test to pass.

### DIFF
--- a/internal/status/glob.go
+++ b/internal/status/glob.go
@@ -14,7 +14,7 @@ func glob(dir string, globs []string) (files []string, err error) {
 		if !filepath.IsAbs(g) {
 			g = filepath.Join(dir, g)
 		}
-		g, err = execext.Expand(g)
+		g, err = execext.Expand(filepath.ToSlash(g))
 		if err != nil {
 			return nil, err
 		}
--- a/testdata/expand/Taskfile_windows.yml
+++ b/testdata/expand/Taskfile_windows.yml
@@ -1,0 +1,8 @@
+version: '2'
+
+tasks:
+  pwd:
+    cmds:
+      - pwd
+    dir: '$USERPROFILE'
+    silent: true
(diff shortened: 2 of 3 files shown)


andreynering:
Hey @mikynov,

Thanks a lot for opening this pull request and explaining in detail what you did and why.

I believe these issues are bugs on the [mvdan/sh](https://github.com/mvdan/sh) project, and I just opened an issue there: https://github.com/mvdan/sh/issues/356. I'll wait a bit to check if this PR will still be necessary.

Also, you reminded me that I need to setup a proper CI for Windows. I'll open another issue for it.

andreynering:
Hi again @mikynov,

OK, since the discussion on the other project will be longer, I just pushed your fix (with changes) to master: https://github.com/go-task/task/commit/5e49b38c333a48d52e337ea21793e31851cb8fb8

Not all tests will pass yet (since we still have the `~` not being expanded issue), but this should prevent some other issues on Windows.

Issue #107, opened 2018-03-22
Add help informations for a task 


skurtzemann:
## Request

Ability to add help informations for a task.

## Example

```
version: "2"

tasks:
  "gcloud:docker:push":
    desc: Push Docker image to Google cloud registry
    help: |
        Push Docker image to Google cloud registry.

        The env variable "DOCKER_IMAGE_FQN" define the Docker image name to push.
    silent: true
    cmds:
    - cmd: gcloud docker -- push {{.DOCKER_IMAGE_FQN}}
```

```
$ task -l
task: Available tasks for this project:
* gcloud:docker:push: 	Push Docker image to Google cloud registry
```

```
$ task help gcloud:docker:push

Push Docker image to Google cloud registry.

The env variable "DOCKER_IMAGE_FQN" define the Docker image name to push
```

## Finally

What do you think ?

andreynering:
Hi @skurtzemann! Thanks for the feature request.

I think this proposal makes sense. Since this would be easy to implement, I'll keep this open for potential contributors.

stephenwithav:
Simplest case: use [cobra](https://github.com/spf13/cobra) to add support for the help subcommand.

What do you think, @andreynering?

andreynering:
@skurtzemann I think I'd prefer to keep using `pflag` as the cli library

I think this should work the same as `-l`/`--list` flag, so beside it being a flag, it will work like a command:

```sh
task --help my:task:name
```

I'm not sure `pflag` allow overwrite the `--help` flag, though. Otherwise, we'd have to use [`pflag.Usage`](https://github.com/go-task/task/blob/master/cmd/task/task.go#L44) instead

stephenwithav:
`-h/--help` can be overridden with pflag.  See: [L102](https://github.com/spf13/pflag/blob/master/flag.go#L1002), [L949](https://github.com/spf13/pflag/blob/master/flag.go#L949).

Do we use another flag instead?

andreynering:
I think we can overwrite it then, but if given without any other argument, the behavior should be the same as today

Or, if someone have a better flag name ideia other than `--help`, I'm open to suggestions, but I couldn't think in any

smyrman:
I think sticking with `-h` and `--help` and the semantics that @andreynering suggests sounds like a good suggestion UX wise.

stephenwithav:
Look for a PR later today.  (-0400 GMT.)

stephenwithav:
[...]

PR #180, opened 2019-02-24
Display task summary

jaedle:
Adding a ``--summary`` flag for task to get more a summary of the task

See #107 

New field in task called `summary`
```yaml
version: '2'
tasks:
  release:
    deps: [build]
    summary: |
      Release your project to github

      It will build your project before starting the release it.
      Please make sure that you have set GITHUB_TOKEN before starting.
    cmds:
      - your-release-tool
  build:
    cmds:
      - your-build-tool
```

Output for `task --summary release`:

```
task: release

Release your project to github

It will build your project before starting the release it.
Please make sure that you have set GITHUB_TOKEN before starting.

dependencies:
 - build

commands:
 - your-release-tool
```

Behaviour:
- `--summary` will not execute the task
- `--summary` will display the summary of a task (see above)
- `--summary` will use description as fallback, if summary does not exist
- `--summary` will output a warning (or information), if no description/summary is present
- `--summary` will output the dependencies if there are any
- `--summary` will output the commands if there are any

Tests:
- Because there are a lot of edge-cases I decided to extract out a package (summary), which is completly unit tested with above named edge cases.
- There is an integration like test in `task_test.go`

Documentation is present for that feature


jaedle:
Usage:
```task --details some-task```

There are still some questions to me:
What should happen, if there is no detailed description for this task? 

- Outputting an error message makes perfect sense for me. 
- Should there be a non-zero exit value? I would guess no..

**Comment by tonobo** [comment]
Wouldn't it be cooler to have kind of inspect/describe command, which shows long description and the whole command specification?

**Comment by jaedle** [comment]
> Wouldn't it be cooler to have kind of inspect/describe command, which shows long description and the whole command specification?

I think there is a difference. Sometimes it makes sense, if you have preconditions or something like that, to document it. I think for that a ``details`` section makes perfect sense.
For other purposes I think your suggestion is also very good - but maybe in a different use case.

**Comment by tonobo** [comment]
I meant this in addition to your newly introduced keyword, it's pretty cool to give tasks a more detailed description. But the details context could print also the whole task in detail.

**Comment by jaedle** [comment]
Could you propose a sample output for this?

**Comment by tonobo** [comment]
I don't know maybe something like this ?

```
task: build

Pretty long description comes here Lorem ipsum dolor sit amet, 
consectetur adipiscing elit, sed do eiusmod tempor incididunt ut 
labore et dolore magna aliqua.

Dependencies:
  - lint
  - format

Commands:
  - sh -c "go build ..."
```

**Comment by jaedle** [comment]
My only concern about that is that the content of the taskfile would be duplicated. What is the use case for that instead of looking in the taskfile itself?

**Comment by tonobo** [comment]
Of course, same for the details. But it the taskfile might be included, so there is no direct way to see what happens when running this task.

**Comment by jaedle** [comment]
> Of course, same for the details. But it the taskfile might be included, so there is no direct way to see what happens when running this task.

That is convincing for me. 😊

**Comment by tonobo** [comment]
Great! :grin:

**Comment by andreynering** [comment]
Hi @jaedle,

Thanks for working on this.

I agree on having another flag and also on showing more information (like dependencies).

I'll take a look at the code in another moment.

In the meantime, what do you think of `--summary` instead of `--details`?

**Comment by andreynering** [comment]
Same for the attribute name, could be `summary:` instead of `details:`

**Comment by jaedle** [comment]
> Same for the attribute name, could be `summary:` instead of `details:`
Perfect, I'm gonna change this today/tomorrow.

> I agree on having another flag and also on showing more information (like dependencies).
Do you mind if this PR would add both flags/commands?
I liked the ``--inspect`` idea, it's quite close to docker.

**Comment by andreynering** [comment]
> Thanks for this great tool! For my hobby project and at my organization we're heavily using task. It's a great tool for automating stuff!

😄 

> Do you mind if this PR would add both flags/commands? I liked the `--inspect` idea, it's quite close to docker.

When I said, "having another flag", I meant other than `--help`(as initially suggested on the original issue).

IMHO we don't need separated `--summary` and `--inspect` flags. We can just show the additional info while calling `--summary`. Let me know if you disagree.

**Comment by jaedle** [comment]
Ready for review! :)

File: docs/usage.md [hunk 1]
```diff
@@ -542,6 +542,50 @@ would print the following output:
 * test:    Run all the go tests.
 ```
 
+## Display summary of task
```
andreynering: Thanks for writing documentation! 🙂

docs/usage.md
```diff
@@ -542,6 +542,50 @@ would print the following output:
...
+  release:
+    deps: [build]
+    summary: |
+      Release your project to github
+
+      It will build your project before starting the release it.
+      Please make sure that you have set GITHUB_TOKEN before starting.
+    cmds:
+      - your-release-tool
+  build:
```
andreynering: Missing an empty line here.

```suggestion

  build:
```

docs/usage.md
```diff
@@ -542,6 +542,50 @@ would print the following output:
...
+dependencies:
+ - build
+
+commands:
+ - your-release-tool
+```
+If a summary is missing, the description will be printed. 
+If the task does not have a summary or a description, a warning is printed.
+
+Please note: *showing the summary will not execute the command*
```
andreynering:
```suggestion
Please note: *showing the summary will not execute the command*.
```

internal/summary/summary.go
```diff
@@ -0,0 +1,79 @@
+package summary
+
+import (
+	"github.com/go-task/task/v2/internal/logger"
+	"github.com/go-task/task/v2/internal/taskfile"
+	"strings"
+)
```
andreynering: Please, separate imports by groups (stdlib, project packages, external libs):

```go
import (
	"strings"

	"github.com/go-task/task/v2/internal/logger"
	"github.com/go-task/task/v2/internal/taskfile"
)
```

internal/summary/summary_test.go
```diff
@@ -0,0 +1,170 @@
+package summary_test
+
+import (
+	"bytes"
+	"github.com/go-task/task/v2/internal/logger"
+	"github.com/go-task/task/v2/internal/summary"
+	"github.com/go-task/task/v2/internal/taskfile"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
```
andreynering: Same here:

```go
import (
	"bytes"
	"testing"

	"github.com/go-task/task/v2/internal/logger"
	"github.com/go-task/task/v2/internal/summary"
	"github.com/go-task/task/v2/internal/taskfile"

	"github.com/stretchr/testify/assert"
)
```

task.go
```diff
@@ -3,6 +3,7 @@ package task
 import (
 	"context"
 	"fmt"
+	"github.com/go-task/task/v2/internal/summary"
```
andreynering: Same here

internal/summary/summary.go
```diff
@@ -0,0 +1,79 @@
...
+	for i, line := range lines {
+		notLastLine := i+1 < len(lines)
+		if notLastLine || line != "" {
+			l.Outf(line)
+		}
+	}
+}
+
+func printTaskName(l *logger.Logger, t *taskfile.Task) {
+	l.Outf("task: " + t.Task)
```
andreynering:
```suggestion
	l.Outf("task: %s", t.Task)
```

internal/summary/summary_test.go
```diff
@@ -0,0 +1,170 @@
...
+		Deps: []*taskfile.Dep{
+			{Task: "dep1"},
+			{Task: "dep2"},
+			{Task: "dep3"},
+		},
+	}
+
+	summary.Print(&l, task)
+
+	assert.Contains(t, buffer.String(), "\ndependencies:\n"+" - dep1\n"+" - dep2\n"+" - dep3\n")
```
andreynering:
```suggestion
	assert.Contains(t, buffer.String(), "\ndependencies:\n - dep1\n - dep2\n - dep3\n")
```

task.go
```diff
@@ -62,6 +64,12 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 		}
 	}
 
+	if e.Summary {
+		firstTask := calls[0].Task
```
andreynering: Hey @jaedle,

The way it's implemented now, the `--summary` flag allows multiple tasks to be specified, but it'll only print the summary for the first one. E.g. `task --summary foo bar` will only print summary for `foo`.

IMHO we should either disallow more than one non-flag arguments or print the summary for all the specified tasks, with a two-empty-line break between then. I prefer the latter.

testdata/summary/Taskfile.yml
```diff
@@ -0,0 +1,26 @@
+version: 2
+
+tasks:
+
```
andreynering: Remove this empty line.

```suggestion
```

testdata/summary/Taskfile.yml
```diff
@@ -0,0 +1,26 @@
...
+  task-with-summary:
+    deps: [dependend-task-1,dependend-task-2]
```
andreynering:
```suggestion
    deps: [dependend-task-1, dependend-task-2]
```

task.go
jaedle: Hey @andreynering,

I completly agree on having multiple summaries. I‘ll change this.

jaedle:
> Hey @jaedle, great work with the code, tests and documentation! 👏 👏👏
> 
> Thanks for spending the time on work on this. I did just a few comments, but once addresses this is good to go.

Hey @andreynering,

thanks for the review. I‘m quite new to golang, so I really appreciate it! Fixes are comming next week 😊

task.go
```diff
@@ -62,6 +64,11 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 		}
 	}
 
+	if e.Summary {
+		summary.PrintAll(e.Logger, e.Taskfile, calls)
```
jaedle: I've been thinking hard to do this.
I think the task component (high level abstraction, only delegates to other components) should not care about a low level detail like spacing for the output of the summary.
To fix this, I added a `PrintAll` Method for the `summary`-package, which deals with it.

jaedle:
I really hope this is good to go! :)

andreynering:
@jaedle Thanks again!

If I can give you a friendly tip, next time try keeping a smaller number of commits.

jaedle:
> @jaedle Thanks again!
> 
> If I can give you a friendly tip, next time try keeping a smaller number of commits.

Thanks for the recommandation. I usally tend to squash merge, if you don‘t, I will squash the branch before raising the PR.

PR #182, opened 2019-03-02
Fix the zsh completion with sub-tasks

GuillaumeAmat:
When the taskfile has sub-tasks (eg: `task foo:bar`), the completion does not work correctly. That PR fixes that.

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -5,7 +5,7 @@ function __list() {
     local -a scripts
 
     if [ -f Taskfile.yml ]; then
-		scripts=($(task -l | sed '1d' | sed 's/^\* //' | awk '{ print $1 }' | sed 's/:$//' | sed 's/:/\\:/'))
+        scripts=($(task -l | sed '1d' | sed 's/^\* //' | awk '{ print $1 }' | sed 's/:$//' | sed 's/:/\\:/'))
         _describe 'script' scripts
     fi
 }
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -5,7 +5,7 @@ function __list() {
     local -a scripts
 
     if [ -f Taskfile.yml ]; then
-        scripts=($(task -l | sed '1d' | sed 's/://' | awk '{ print $2 }'))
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @GuillaumeAmat, thanks for opening this PR.

Hey @sawadashota, sorry for pinging you, but do you mind reviewing this change? (Don't be afraid of saying "no" if you don't want to).

completion/zsh/_task
```diff
@@ -5,7 +5,7 @@ function __list() {
     local -a scripts
 
     if [ -f Taskfile.yml ]; then
-        scripts=($(task -l | sed '1d' | sed 's/://' | awk '{ print $2 }'))
+        scripts=($(task -l | sed '1d' | sed 's/^* //' | sed 's/\:\s.*//' | sed 's/\(\w\):\(\w\)/\1\\\:\2/'))
```
sawadashota: About ``sed 's/\:\s.*//'``,

* It's needless `\` of `\:` because `:` is not special character. 
* `\s` doesn't work in my MacOS because sed doesn't support expand expression.

So if you want remove `:` and space,  `sed 's/: //'`.
Or in addition, want remove tab, `sed s/:$' '$'\t'//`.

sawadashota: About `sed 's/\(\w\):\(\w\)/\1\\\:\2/'`,
sorry but can you explain how works this with example?

To fixing your described bug, how about following?

```
task -l | sed '1d' | sed 's/^\* //' | awk '{ print $1 }' | sed 's/:$//'
```

GuillaumeAmat: Thanks for the review @sawadashota :+1: 

I've just updated the PR with your proposal and add a required replacement.

About my use case, I use included Taskfiles so my commands look like that: `$ task file:command`
But with the previous completion script the `:` break the proposed commands as it is used as a description separator by zsh.

sawadashota: I think it's almost OK but can you adjust indent?

GuillaumeAmat: Sure! My bad :hand:

GuillaumeAmat: It is done

andreynering:
@GuillaumeAmat @sawadashota Thanks!

Issue #183, opened 2019-03-07
Add Taskfile.yaml default file name alternative


lalloni:
Currently Task searches for Taskfile.yml as the default taskfile name.

I, as a developer, would like Task to use Taskfile.yaml as a valid alternative Taskfile name so I can have uniformity with other YAML files used by other tools and name all with the same .yaml extension instead of using .yaml for all except Taskfile.yml.

andreynering:
Closing this issue.

Reasoning here: https://github.com/go-task/task/pull/184#issuecomment-552246185

andreynering:
FYI: We are about to allow `.yaml` as extension soon as part of #584 and #621.

PR #184, opened 2019-03-07
Allow to use Taskfile.yaml a alternative name

lalloni:
When reading the taskfile first try using Taskfile.yml and then try Taskfile.yaml if the former was not found.

Implements #183

### DIFF
--- a/internal/taskfile/read/taskfile.go
+++ b/internal/taskfile/read/taskfile.go
@@ -22,8 +22,15 @@ var (
 
 // Taskfile reads a Taskfile for a given directory
 func Taskfile(dir string) (*taskfile.Taskfile, error) {
-	path := filepath.Join(dir, "Taskfile.yml")
-	if _, err := os.Stat(path); err != nil {
+	var path string
+	for _, name := range []string{"Taskfile.yml", "Taskfile.yaml"} {
+		p := filepath.Join(dir, name)
+		if _, err := os.Stat(p); err == nil {
+			path = p
+			break
+		}
+	}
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @lalloni,

Thanks for the idea and for opening a pull request for it!

If we're going to allow `.yaml` extensions we should be consistent and allow it everywhere. This means there's at least three more places we need to update:

- When [including a Taskfile by a directory](https://taskfile.dev/#/usage?id=including-other-taskfiles), Task currently look for the `Taskfile.yml` file in that directory. It should also look for `Taskfile.yaml`.
- Task automatically includes `Taskfile_{{OS}}.yml` when it exists. Should also look for `Taskfile_{{OS}}.yaml`.
- Task should also look for `Taskvars.yaml` and not just `Taskvars.yml`.

We also need tests to ensure these new file extensions are working correctly.

lalloni:
Hey @andreynering! Thanks for taking your time for looking into this.

I will get into implementing the points you brought up ASAP.

Cheers!

andreynering:
@lalloni No problem, take your time! And feel free to ping me on any doubt.

andreynering:
Hi @lalloni,

I decided to close this PR. Most YAML files out there are `.yml` and not `.yaml` anyway, and I think it's fine to just keep support for one of them. This also means there's a smaller chance of having bugs related to this on included Taskfiles, etc.

andreynering:
FYI: We are about to allow `.yaml` as extension soon as part of #584 and #621.

Issue #53, opened 2017-07-15
Multiple tasks with the same dependency should run dependency once!


smyrman:
UPDATE: This is a general feature request to not run _equal_ dependencies more than once. E.g. if `task A` and `task B` both depend on `taskC`, running _both_ `task A` and `task B` in parallel (witin one instance of task), should be able to _resolve_ the dependencies so that `taskC` is only run once, and this is the interesting port, _if it's equal_, which we will get to next.

### what makes a task equal?

Since you can _pass variables_ to tasks, it's not enough to just look at the name. The _best_ option is probably to rely on a hash of the name + _sorted_ and _resolved_ input variables. A close second could be to rely on _unresolved_ but still _sorted_ input variables, which is not perfect, but could be an acceptable trade-off if it's somehow easier in code.

---
# Original description

I have multiple tasks in my `Taskfile` that depend on a single task called `go-dep` that runs the command `glide install`. When I run `task` to build multiple targets, I would expect go-dep to be called at most 1 time (zero if it's found to be up-to-date). Instead, I see it's called once for each task that depend on it:

```go
$ task
task: No argument given, trying default task
glide install
glide install
glide install
glide install
```

PS! `go-deps` in this case is called with no vars, but the rule should really apply to run a dep at most 1 time when the vars are _exactly_ the same.

andreynering:
Hmm... I'm not sure I agree with this. Some users may want to re-run a dependency each time (for example, if another task changed any file and something has to be re-generated).

Wouldn't this work for you?

```yml
go-dep:
  cmds:
    - glide install
  sources:
    - glide.yaml
    - glide.lock
  generates:
    - vendor/**/*
```

I'll keep this open to see if anyone have an opinion on this.

smyrman:
Thats what I have, but when multiple tasks  (say 4) call it simultaniously, it's launcher 4 rimes concurrently. If a dependent can not be resolved to one call inn this case, it's not really a dependent at all, just a concurrent-task.

andreynering:
Maybe we get a solution to this someday. I'm don't have anything in mind now.

By now, you can workaround it in at least two ways:

```yml
# remove `go-deps` as a dependency of each build
build-all:
  deps:
    - go-deps
    - cmd/go-binary1
    - cmd/go-binary2
    - cmd/go-binary4
    - cmd/go-binary3
```

```yml
# run serially instead of concurrently
# e.g.: use `cmds` instead of `deps`
build-all:
  cmds:
[...]

PR #193, opened 2019-03-29
Force tasks that generate the same files to not clobber each other

Evertras:
Potential solution to #53 

This will create a lock for every `generates` entry in the list.  This means if that if multiple tasks would generate the same files as according to the task, the tasks cannot run in parallel.  This seems useful regardless of whether #53 ends up with task-specific configuration; tasks that touch the same files should never run in parallel.

I've tested this locally and it allows me to safely declare dependencies multiple times without issue.  After the first run ends, the rest will say 'up to date' and everything works properly.

### DIFF
--- a/task.go
+++ b/task.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"sync/atomic"
 
 	"github.com/go-task/task/v2/internal/compiler"
@@ -51,6 +52,8 @@ type Executor struct {
 	taskvars taskfile.Vars
 
 	taskCallCount map[string]*int32
+	generating    map[string]*sync.Mutex
+	generateLock  sync.Mutex
 }
 
 // Run runs Task
@@ -166,6 +169,8 @@ func (e *Executor) Setup() error {
 		}
 	}
 
+	e.generating = make(map[string]*sync.Mutex)
+
 	e.taskCallCount = make(map[string]*int32, len(e.Taskfile.Tasks))
 	for k := range e.Taskfile.Tasks {
 		e.taskCallCount[k] = new(int32)
@@ -186,6 +191,21 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 	if err := e.runDeps(ctx, t); err != nil {
 		return err
 	}
+
+	e.generateLock.Lock()
+	for _, g := range t.Generates {
+		m, ok := e.generating[g]
+
+		if !ok {
+			m = &sync.Mutex{}
+			e.generating[g] = m
+		}
+
+		m.Lock()
+
+		defer m.Unlock()
+	}
+	e.generateLock.Unlock()
 
 	if !e.Force {
 		upToDate, err := e.isTaskUpToDate(ctx, t)


andreynering:
Hi @Evertras,

I kinda like the proposal. But that's a potential breaking change, since some user that intentionally call these tasks in parallel may really want them to run in parallel.

This is something we can discuss including on the next major Task release (version 3), though.

Evertras:
I understand wanting to be conservative here, and will defer to you in the end.

However, I would strongly argue that any task that lists the same file(s) under 'generate' runs a very serious risk of hitting race conditions and general undefined behavior with things as they are right now.  If a task has files listed under 'generate', it is expected that those files are in some way created/modified by the task.  If multiple tasks are running in parallel and they are modifying the same files, there is no sane way to avoid a race condition without locks; which this would do for you.  I would be very surprised to know that someone is using unpredictable behavior as part of their build process.  Even if they were taking advantage of running them in parallel, this change would only slow down their build, not break it.  To me this is a bug fix, not a feature.

I think the idea of forcing dependencies to run only once is a separate discussion that is more apt for v3.  This just has a lot of overlap.

task.go
```diff
@@ -187,6 +192,21 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 		return err
 	}
 
+	e.generateLock.Lock()
+	for _, g := range t.Generates {
```
andreynering: Keep in mind that `t.Generates` can contains a glob instead of a file path.

In an ideal implementation, you'd expand this glob and lock each of the returned files.

andreynering:
@Evertras I also think that extracting this behavior in a separated function would be ideal.

Globbing is currently implemented [here](https://github.com/go-task/task/blob/master/internal/status/glob.go).

task.go
Evertras: This initial pass was intended to work with globs since the user is declaring what the task is going to modify.  The problem with checking for any found files for the glob is that any files being generated may not exist yet, so nothing will be returned and either nothing will lock or everything will lock.

That being said, there are definitely some holes.  If Task A generates `output/file.data` and Task B generates `output/**/*.data`, then we likely have a problem.  If Task A generates `output/**/*.jpg` and Task B generates `output/*.jpg` there could be some overlap as well.

One improvement might be to double down on the locks; basically use the method already in the PR, and then *also* lock on glob matches.  But that's inconsistent depending on whether the files exist already or not, and to me that's a large problem and makes that approach unappealing.

Another approach would be to try to be smart about the generated field matches.  I could imagine some basic logic being able to tell that there's overlap between `output/**/*.data` and `output/file.data`, for example.  This would give us consistency regardless of the state of generation and also give the user a lot more safety without having to worry about their config being a certain way.

Evertras:
Happy to move it to its own section, especially if it grows.  But see the other comment for my concerns with globbing.

Evertras:
Also sorry for the delay, dealing with travel. :D

Evertras:
I took a look at the globbing code and I'm curious for your thoughts on an idea.  I'm typing this out and thinking at the same time to see if it would even work, but I think it might.

What if we took all the 'generated' files and created a simple virtual filesystem that actually 'created' a file for each entry in the list.  It would start with non-patterned/explicit files, then move to more complicated patterns and each run would record what other patterns it would clobber with.  Patterned globs would create a simple 'star' file for bookkeeping purposes.  We could then run the same code against the virtual filesystem as we do against the real filesystem.

For reference, my initial thought is to try using this: https://github.com/spf13/afero .  We should just be able to drop this into fastwalk and be good to go.  It would also make unit testing easier as a bonus.

So let's say we have the following tasks:

Task A - Generates `output/file.data` and `output/pkg/*.pkg`

Task B - Generates `output/*.data` and `output/**/*.pkg`

Task C - Generates `output/pkg/c.pkg` and `output/**/*` (for good measure)

So while none of these tasks have the same exact 'generates' entries, they all have the potential to clobber each other!  The algorithm would then work as follows.

First, sort the generated entries by specific -> nonspecific; the `**` patterns go after the `*` ones, and less nested globs go after more nested ones.

- `output/file.data`
- `output/pkg/c.pkg`
- `output/pkg/*.pkg`
- `output/*.data`
- `output/**/*.pkg`
- `output/**/*`

Now that we have a list of globs to check, we start with the first that are actually just filenames.  The glob is run against the empty virtual filesystem and `output/file.data` is 'created' in memory.  Then `output/pkg/c.pkg` is created.

Then we get to our patterned lists.  We start with the simpler globs.

The `output/pkg/*.pkg` entry finds `output/pkg/c.pkg` using the existing fastwalk code, so we note that `output/pkg/*.pkg` clobbers with `output/pkg/c.pkg` entries.  We also create `output/pkg/*.pkg` as an actual file.

Now `output/*.data` runs.  It finds `output/file.data` and creates `output/*.data` as an actual file.  Like above, we note that `output/*.data` will clobber with any `output/file.data` entries.

Then we get `output/**/*.pkg`.  It runs and discovers `output/pkg/c.pkg` and `output/pkg/*.pkg`.  Since we used the asterisk for the filename, we don't have to do anything special!  We just note that we found `output/pkg/c.pkg` and `output/pkg/*.pkg` as paths that will clobber.  Then we write `output/**/*` which is literally the file `*` in the directory `**` in `output`.

Finally we run `output/**/*`.  It runs and discovers the following:

- `output/file.data`
- `output/pkg/c.pkg`
- `output/pkg/*.pkg`
- `output/*.data`
- `output/**/*.pkg`

So we know that `output/**/*` will clobber with any of the above matches.

After all that, what we end up with is a list of glob patterns that have a generated list of glob patterns tied to them that they will also lock on.  So when Task A runs, which has the patterns `output/file.data` and `output/pkg/*.pkg`, it will lock on `output/file.data`, `output/pkg/*.pkg`, and `output/pkg/c.pkg`.  When Task C tries to run in parallel it will lock on `output/pkg/c.pkg` (along with others) which will properly wait for Task A to be done to ensure it isn't stepping on anything.  In fact in this case all tasks will run serially because they all might clobber each other somehow; but that's exactly what we want for safety.  I'd argue that this particular example is insane and the user should wonder what they're doing with themselves, but it's an example purely to make our lives miserable and figure out how to solve this.

andreynering:
Hi @Evertras,

Sounds a bit too complex for a solution that intended to be simple.

Evertras:
I agree it's complex, but unfortunately I'm not seeing a simpler way that would handle clobbering without just matching on the user-provided strings.

I think there's still a lot of benefit in the simple approach that's currently implemented, but to make it airtight there's a lot more to think about.

andreynering:
Hi @Evertras,

Since it's been a long time, I'm going to close this PR.

That doesn't mean I don't plan to make improvements in this regard. Work on this by happen again in the future.

PR #175, opened 2019-02-11
Try out Windows builds in Travis

emirb:
Closes #171

PR #199, opened 2019-04-13
Try out Windows builds in Travis (again)

andreynering:
We had some random issues on the first time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: go
 
+os:
+  - linux
+  - windows
+
 go:
   - 1.10.x
   - 1.11.x


marco-m:
Ah, I didn't know that Travis supported Windows! @andreynering would you be interested in my giving it a try ? Should it come as a new PR ?

andreynering:
Hi @marco-m,

Yes, Travis supports Windows, but tests passed only once here: https://github.com/go-task/task/pull/175.

Once I merged that, it started to break in master branch. This PR was a second try.

I have no idea what the problem is. Travis logs aren't useful.

So we have two options here:

1. Figure it out what the problem is with our Travis setup on Windows and fix it;
2. Give up on Travis for Windows and just use AppVeyor instead. I've used AppVeyor before [here](https://github.com/go-testfixtures/testfixtures/blob/master/appveyor.yml) and [here](https://github.com/go-gormigrate/gormigrate/blob/master/appveyor.yml).

andreynering:
Seems that Windows support on Travis is weak.

The Task organization is on the new GitHub Actions beta. I'm just waiting GitHub migrate us to v2 (which supports Windows and macOS) to try it as a CI.

PR #205, opened 2019-05-17
Add Preconditions to Tasks

stephenprater:
## What changes does this PR introduce?

This adds a `preconditions` stanza to the `task` stanza.  A precondition is _basically_ the inverse of a `status` in that it's a condition that _must_ succeed before the task can run.

## Any background context you want to provide?

Preconditions on file statuses doesn't make much sense, so preconditions are limited to `sh` interpreter invocations.  Furthermore - preconditions often require the user to do something in order to proceed, so they have a `msg` field that can be used to report next steps to the user.

I've added an "ignore_error" feature to preconditions, which will ignore errors in preconditions if the task is being executed as either a dependency or as a sub-task.  I'm not 100% this isn't a misfeature, but it was requested by our team, so it's there.   Normally a failure in a precondition will fail the task that it's on, and thus any "upstream" tasks as well.

## Where should the reviewer start?

`TestPrecondition` enumerates all of the situations I could think of that encounter a conditional - it's a pretty good place to see how to use `preconditions` - they are executed in the same context as `status` commands in `areTaskPreconditionsMet`

## Has this been manually tested? How?

We run our own fork of go-task with this feature for our automation.  So far it's working well.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -344,6 +344,53 @@ up-to-date.
 Also, `task --status [tasks]...` will exit with a non-zero exit code if any of
 the tasks are not up-to-date.
 
+If you need a certain set of conditions to be _true_ you can use the
+`preconditions` stanza.  `preconditions` are very similar to `status`
+lines except they support `sh` expansion and they SHOULD all return 0
+
+```yaml
+version: '2'
+
+tasks:
+  generate-files:
+    cmds:
+      - mkdir directory
+      - touch directory/file1.txt
+      - touch directory/file2.txt
+    # test existence of files
+    preconditions:
+      - test -f .env
+      - sh: "[ 1 = 0 ]"
+        msg: "One doesn't equal Zero, Halting"
+```
+
+Preconditions can set specific failure messages that can tell
+a user what to do using the `msg` field.
+
+If a task has a dependency on a sub-task with a precondition, and that
+precondition is not met - the calling task will fail.  Adding `ignore_errors`
+to the precondition will cause parent tasks to execute even if the sub task
+can not run.  Note that a task executed directly with a failing precondition
+will not run unless `--force` is given.
+
+```yaml
+version: '2'
+tasks:
+  task_will_fail:
+    preconditions:
+      - sh: "exit 1"
+        ignore_errors: true
+
+  task_will_succeed:
+  deps:
+    - task_will_fail
+
+  task_will_succeed:
+  cmds:
+    - task: task_will_fail
+    - echo "I will run"
+```
+
 ## Variables
 
 When doing interpolation of variables, Task will look for the below.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -346,7 +346,7 @@ the tasks are not up-to-date.
 
 If you need a certain set of conditions to be _true_ you can use the
 `preconditions` stanza.  `preconditions` are very similar to `status`
-lines except they support `sh` expansion and they SHOULD all return 0
+lines except they support `sh` expansion and they SHOULD all return 0.
 
 ```yaml
 version: '2'
@@ -373,7 +373,7 @@ executed with a failing precondition will not run unless `--force` is
 given.
 
 Unlike `status` which will skip a task if it is up to date, and continue
-executing tasks that depenend on it a `precondition` will fail a task, along
+executing tasks that depend on it, a `precondition` will fail a task, along
 with any other tasks that depend on it.
 
 ```yaml
@@ -384,13 +384,13 @@ tasks:
       - sh: "exit 1"
 
   task_will_also_fail:
-  deps:
-    - task_will_fail
+    deps:
+      - task_will_fail
 
   task_will_still_fail:
-  cmds:
-    - task: task_will_fail
-    - echo "I will not run"
+    cmds:
+      - task: task_will_fail
+      - echo "I will not run"
 ```
 
 ## Variables
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -365,13 +365,16 @@ tasks:
 ```
 
 Preconditions can set specific failure messages that can tell
-a user what to do using the `msg` field.
+a user what steps to take using the `msg` field.
 
 If a task has a dependency on a sub-task with a precondition, and that
-precondition is not met - the calling task will fail.  Adding `ignore_errors`
-to the precondition will cause parent tasks to execute even if the sub task
-can not run.  Note that a task executed directly with a failing precondition
-will not run unless `--force` is given.
+precondition is not met - the calling task will fail.  Note that a task
+executed with a failing precondition will not run unless `--force` is
+given.
+
+Unlike `status` which will skip a task if it is up to date, and continue
+executing tasks that depenend on it a `precondition` will fail a task, along
+with any other tasks that depend on it.
 
 ```yaml
 version: '2'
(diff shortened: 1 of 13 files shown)


andreynering:
Hi @stephenprater,

Thanks! This looks useful, indeed.

I'll review it soon.

internal/taskfile/task.go
```diff
@@ -5,19 +5,20 @@ type Tasks map[string]*Task
...
-	IgnoreError bool `yaml:"ignore_error"`
+	Task         string
+	Cmds         []*Cmd
+	Deps         []*Dep
+	Desc         string
+	Summary      string
+	Sources      []string
+	Generates    []string
+	Status       []string
+	Precondition []*Precondition
```
smyrman: Since other names are short (Cmds, Deps, Desc etc.), maybe this name can be just `Pre` (`"pre"` in YAML).

internal/taskfile/version/version.go
```diff
@@ -10,6 +10,8 @@ var (
 	v21 = mustVersion("2.1")
 	v22 = mustVersion("2.2")
 	v23 = mustVersion("2.3")
+	v24 = mustVersion("2.4")
+	v25 = mustVersion("2.5")
```
smyrman: @andreynering. please correct me if I am wrong.

2.4 and 2.5 have been released, so when adding a new feature to the taskfile here, it should probably come as a 2.6 version.

Since there are not checks for 2.4 and 2.5 today, I presume these two releases does not provide any updated to the Taskfile format, and does not need the `v2X` and `Isv2X()` definitions?

task.go
```diff
@@ -192,7 +193,13 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 		if err != nil {
 			return err
 		}
-		if upToDate {
+
+		preCondMet, err := e.areTaskPreconditionsMet(ctx, t)
```
andreynering: I think this should run before `upToDate, err := e.isTaskUpToDate(ctx, t)` because that function has side-effects when using the `checksum` method.

variables.go
```diff
@@ -73,6 +73,7 @@ func (e *Executor) CompiledTask(call taskfile.Call) (*taskfile.Task, error) {
 				IgnoreError: cmd.IgnoreError,
 			}
 		}
+
```
```suggestion
```

stephenprater:
@andreynering We discussed the `ignore_errors` behavior this morning.  We added this as part of adding preconditions to our fork, but we don't actually use it in practice.  We're all of the opinion now that it's a misfeature.  (IE - A "feature" that works in surprising, confusing or bug-like ways.)  I am just going to remove it.

The TL;DR of the discussions was that nobody could think of a single situation in which you would want the behavior of a task _not to run_ because it's preconditions weren't met AND where that failure wouldn't bubble up to a calling task.

I'll make these changes today.

docs/usage.md
```diff
@@ -344,6 +344,53 @@ up-to-date.
 Also, `task --status [tasks]...` will exit with a non-zero exit code if any of
 the tasks are not up-to-date.
 
+If you need a certain set of conditions to be _true_ you can use the
+`preconditions` stanza.  `preconditions` are very similar to `status`
+lines except they support `sh` expansion and they SHOULD all return 0
```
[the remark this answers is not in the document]
stephenprater: That is correct.

task.go
```diff
@@ -119,16 +119,17 @@ func (e *Executor) Setup() error {
 			Vars:   e.taskvars,
 			Logger: e.Logger,
 		}
-	case version.IsV2(v), version.IsV21(v), version.IsV22(v):
+	case version.IsV2(v), version.IsV21(v), version.IsV22(v), version.IsV23(v):
```
stephenprater: @andreynering Should this be changed to `version.isV26(v)`?   I'm a little confused about how this version detection works - since if I change the point version in my actual Taskfiles I get "X is only available on <older version>"

stephenprater: minimal repro:

```
---
version: '2.3'
output: prefixed
```

```
task: Taskfile option "output" is only available starting on Taskfile version v2.1
```

This _seems_ like a bug to me in that '2.3' is obviously > 2.1 - but maybe the version checking is only against the major version?

stephenprater:
@andreynering Anything more for me to do on this at this time?

andreynering:
@stephenprater Sorry, I'll review this again once I have available time

stephenprater:
No worries!  Just wanted to make sure you weren't waiting on me.

task.go
andreynering: @stephenprater Indeed, the version code needs some love.

In theory this should actually be:

```go
case version.IsV2(v), version.IsV21(v), version.IsV22(v), version.IsV23(v), version.IsV24(v), version.IsV25(v), version.IsV26(v):
```

But there's likely a way to create a specific constraint for `>= 2, < 3`.

I'd say, let this as is, and I'll add a reminder to fix this once I merge this PR to master. 🙂

task.go
```diff
@@ -224,7 +231,12 @@ func (e *Executor) runDeps(ctx context.Context, t *taskfile.Task) error {
 		d := d
 
 		g.Go(func() error {
-			return e.RunTask(ctx, taskfile.Call{Task: d.Task, Vars: d.Vars})
+			err := e.RunTask(ctx, taskfile.Call{Task: d.Task, Vars: d.Vars})
+			if err != nil {
+				e.Logger.Errf("%s", err)
```
andreynering: This log isn't needed. If an error happen, it'll be returned and logged by the caller.

task.go
```diff
@@ -236,7 +248,12 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
 
 	switch {
 	case cmd.Task != "":
-		return e.RunTask(ctx, taskfile.Call{Task: cmd.Task, Vars: cmd.Vars})
+		err := e.RunTask(ctx, taskfile.Call{Task: cmd.Task, Vars: cmd.Vars})
+		if err != nil {
+			e.Logger.Errf("%s", err)
```
andreynering: Same here.

This log isn't needed. If an error happen, it'll be returned and logged by the caller.

task_test.go
```diff
@@ -273,6 +273,48 @@ func TestStatus(t *testing.T) {
...
 
+func TestPrecondition(t *testing.T) {
+	const dir = "testdata/precondition"
+
+	var buff bytes.Buffer
+	e := &task.Executor{
+		Dir:    dir,
+		Stdout: &buff,
+		Stderr: &buff,
+		Silent: false,
```
andreynering: `silent` defaults to `false`.

```suggestion
```

docs/usage.md
```diff
@@ -344,6 +344,55 @@ up-to-date.
...
```
andreynering:
```suggestion
lines except they support `sh` expansion and they SHOULD all return 0.
```

internal/taskfile/precondition.go
```diff
@@ -0,0 +1,45 @@
+package taskfile
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrCantUnmarshalPrecondition is returned for invalid precond YAML.
+	ErrCantUnmarshalPrecondition = errors.New("task: can't unmarshal precondition value")
```
andreynering:
```suggestion
	ErrCantUnmarshalPrecondition = errors.New("task: Can't unmarshal precondition value")
```

stephenprater:
@andreynering Changing the extra error logging in task.go to only log at the caller does have a slight behavioral change.

From:

```
1 != 0 obviously!
task: precondition not met
```

To:

```
task: 1 != 0 obviously!
```

I went ahead and made the change - I think the original output is a _little_ clearer in that it's more obvious why the task failed - but it's also noisier, so 🤷‍♂

andreynering:
@stephenprater Merging this now, and going to fix the version check code.

Thanks again for your work and patience. 🙂

PR #207, opened 2019-05-26
Use colors for some output messages

andreynering:
This makes identifying specific types of messages easier.

- Green: commands
- Yellow: waning messages
- Red: error messages
- Magenta: debug information
- Etc

Output from commands will be kept untouched.

TODO:

- [x] Add `--color=false` flag to disable this
- [x] Disable color for version < 3
- [ ] ~Prettify `task --list`~(skipping for now: tabwriter don't play nicely with colors)?

Screen Shot 2019-05-26 at 6 41 57 PM

Screen Shot 2019-05-26 at 6 42 34 PM

Screen Shot 2019-05-26 at 6 43 09 PM


stephenprater:
I'd like to see the colors exposed in Templater so that you we can control color in echoed strings.   For instance, we have a task that does this:

```
<snip>
cmds:
  - echo "$(tput setaf 1) Press Ctrl-C now to cancel the deploy!"
  - sleep 5
  - task: deploy
``` 

It would be nice to expose the colors as format functions so you could do something similar to this:

```
  - echo "{{.RED}} Press Ctrl-C {{.BOLD}}NOW{{.REGULAR}} to cancel the deploy"
```

andreynering:
@stephenprater I think I'd prefer to have a different flag instead:

```yaml
print:
  cmds:
    - cmd: echo "Something"
      color: red
```

This wouldn't solve the use case of having different colors on the same command, though.

stephenprater:
Hmm - that's good point.   Our use case was actually for the prefix string so it's easier to differentiate between tasks.

```
tasks:
   red_task:
      prefix: '{{.RED}} RED'
      cmds:
         - echo "something"
```

output would be like
```
[ RED ] something
```

Where "RED" is in red.

But the more that I think about it, the less I want to expose a plethora of special variables.  Maybe a function that does something like:  `{{ color('red') }}` instead?

andreynering:
@stephenprater Yeah, we need to mature that idea first.

I gonna finish this PR at some point, and you're free to open another issue with a proposal if you come with a good one. 🙂

wdscxsj:
Color not working on Windows with 3.0 preview 4?

andreynering:
@wdscxsj Have you changed your Taskfile from `version: 2` to `version: 3`?

wdscxsj:
@andreynering Sure. Here is the effect on my machine (Windows 10 64-bit):

task

andreynering:
@wdscxsj It may be a Windows-specific issue. Can you open another issue for tracking?

Also, does it also happens on v2?

wdscxsj:
@andreynering Sure, please see #343.

PR #211, opened 2019-06-04
#209 When "dir:" attribute points to a non-existing dir, create it

### LINKED ISSUE #209: Task directory: allow to create directory if missing
Hello,
consider the following taskfile:
```
tasks:
  build:
    dir: bin
    cmds: [[go build ../cmd/foo]]
```

The reason why I find it handy to specify `dir:` is because this will always put below `bin/` the OS-correct executable name.

If I where to fiddle with `go build -o`, I would have to do something like

```
tasks:
  build:
    cmds: [[go build -o bin/foo.{{exeExt}} ../cmd/foo]]
```
which is uglier, especially if I have multiple executables to build.

And now the problem: `bin` doesn't exist in the git source repo, so I have to add the classic `bin/.git_keepme` just to keep the directory existing.

To avoid the `.git_keepme`, one would have to make the command heavier with something like

```
tasks:
  build:
    cmds: [[mkdir -p bin && cd bin && go build ../cmd/foo]]
```

and this would still break on Windows because I think that `-p` is not supported.

I was thinking: what about adding a cousin of `dir:`, for example called `mkdir:` that creates the directory if it doesn't exist, or changing the `dir:` behavior to create the directory ?

Or am I missing something and there exist already an easy way to do what I want?

marco-m:
Also:

* Task directory: test when "dir:" attribute points to an existing dir
* Task directory: test default case (no "dir:" attribute)
* FIx spelling

Closes #209

### DIFF
--- a/task.go
+++ b/task.go
@@ -200,12 +200,14 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 		}
 	}
 
-	// When using the dir: attribute it can happen that the directory doesn't exist.
+	// When using the "dir:" attribute it can happen that the directory doesn't exist.
 	// If so, we create it.
-	if _, err := os.Stat(t.Dir); os.IsNotExist(err) {
-		if err := os.MkdirAll(t.Dir, 0755); err != nil {
-			e.Logger.Errf("cannot make directory %v: %v", t.Dir, err)
-			return err
+	if t.Dir != "" {
+		if _, err := os.Stat(t.Dir); os.IsNotExist(err) {
+			if err := os.MkdirAll(t.Dir, 0755); err != nil {
+				e.Logger.Errf("task: cannot make directory %q: %v", t.Dir, err)
+				return err
+			}
 		}
 	}
 
--- a/task.go
+++ b/task.go
@@ -202,13 +202,8 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 
 	// When using the "dir:" attribute it can happen that the directory doesn't exist.
 	// If so, we create it.
-	if t.Dir != "" {
-		if _, err := os.Stat(t.Dir); os.IsNotExist(err) {
-			if err := os.MkdirAll(t.Dir, 0755); err != nil {
-				e.Logger.Errf("task: cannot make directory %q: %v", t.Dir, err)
-				return err
-			}
-		}
+	if err := t.Mkdir(); err != nil {
+		e.Logger.Errf("task: cannot make directory %q: %v", t.Dir, err)
 	}
 
 	for i := range t.Cmds {
--- a/task.go
+++ b/task.go
@@ -200,6 +200,15 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


marco-m:
Ah, I didn't realize that CI failed.

marco-m:
@andreynering I fixed the embarrassing CI failure ;-)
This is now ready for review.

task.go
```diff
@@ -200,6 +200,17 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 		}
 	}
 
+	// When using the "dir:" attribute it can happen that the directory doesn't exist.
+	// If so, we create it.
+	if t.Dir != "" {
+		if _, err := os.Stat(t.Dir); os.IsNotExist(err) {
+			if err := os.MkdirAll(t.Dir, 0755); err != nil {
```
andreynering: Perhaps move this code to `Executor.Setup()`?

That would prevent race conditions if the same task is ran in parallel.

marco-m: Mhh, I am not very familiar with the internals of task, but here is what I found:

- To create the directory, we need attribute `t.Dir`. At the very beginning of `RunTask()`, we call `CompiledTask()` https://github.com/go-task/task/blob/9c475c36e7f39bca76414c46f783fcb427f905d3/task.go#L178 which can also change the value of `t.Dir`

If we were to move this code to `Setup()`, we would have to do the following:
- Add a loop that goes through all the tasks, so that it can look at each `call` of type `taskfile.Call`, call `t := CompiledTask(call)` and then create the directory based on `t.Dir` ? This would mean also that it would create directories also for tasks that are not required on the command line (directly or indirectly via dependencies).

If my understanding is right, then I need some more guidance on how to do this properly :-)

marco-m: I was also thinking: if there is a race condition in calling `RunTask()`, then this race condition is also there independently from this PR, or am I missing something?

andreynering: Hi @marco-m,

You're right that it'd be more complicated. I totally forgot that we'd need to compile each task.

I think we can then just have a mutex/lock for that. Extracting into another function would be nice, too.

```go
func (t *Task) mkdir() error {
	if t.Dir == "" {
		return nil
	}

	t.mkdirMutex.Lock()
	defer t.mkdirMutex.Unlock()

	// ...
}
```

marco-m: @andreynering I can do that. But what I am missing is this: the mutex would protect for the specific race when creating a directory, but it looks like there is an inherent race when task run in parallel? Wouldn't make more sense to put the mutex outside this new mkdir() function ?

marco-m:
@andreynering I added the mutex, but please see my questions above about the inherent race.

marco-m:
ping

andreynering:
Hi @marco-m,

This is merged. I did some small changes on fe2b8c8afa4642ba1614a7e7155e03daad7f9774.

About your question above: tasks are supposed to run in parallel, and there's no race condition in Task itself, unless the user programs a task that won't work if ran in parallel.

But since the dir creation is Task work, let's keep the mutex to keep things safe. 🙂

marco-m:
Thank you!

marco-m:
I see you moved the mutex code. Yes, I was unsure where to put it.

PR #213, opened 2019-06-16
Document dir: creation (see PR #211)

marco-m:
Forgot to document that now we create the directory if it doesn't exist.


andreynering:
Thanks!

Issue #195, opened 2019-03-31
Roadmap to version 3


andreynering:
This is a general issue to track proposals with small or big breaking changes, that *could* be released in the next major version.

*This list is a work in progress and can change at any time*.

Development is happening on the [v3 branch](https://github.com/go-task/task/pull/220).

## Cleanup

- [x] Change module name from `github.com/go-task/task/v2` to `v3`
- [x] Remove all code related support of version 1 (#237)
- [x] Remove automatic inclusion of `Taskvars.yml` and `Taskfile_{{OS}}.yml`. This was done before we had a [generic inclusion of Taskfiles](https://taskfile.dev/#/usage?id=including-other-taskfiles) implemented. We should add options to
  - [ ] Don't include the file if it doesn't exist
  - [x] Allow interpolation on the file name to make it possible to include a OS specific file
- [x] Finally delete the `vendor` directory? (#214)

## Breaking changes

- [x] Added global `method:` option. Default changed from `timestamp` to `checksum` (#246)
- [x] Refactor vars and templates to reduce the number of bugs and undesired behavior (#218)
- [ ] Make CLI arguments override everything (#181) (EDIT: What about #336?)
- [ ] Make up-to-date resolution work when only "generates" is given (#187)
- [ ] Don't run the same dependency (with the same args) more than once by default (#53, #193)
- [ ] Allow forcing specific tasks (#189)

## New stuff

- [x] Colored output messages (#207)
- [x] Expose `.TIMESTAMP` and `.CHECKSUM` to `status:` (#216)
- [x] Allow shorter syntax for tasks with default configuration (#194)
- [ ] Allow use as a library (#121)

## Documentation

- [x] Mention changes related to `method:` (shouldn't also check the main guide)
- [x] Update "Taskfile versions"  article

---

There are more [issues and feature request](https://github.com/go-task/task/issues) that could be added to this release.

andreynering:
/cc @smyrman @sheerun @jamestenglish @jaxxstorm @Evertras

---

As you may have noticed, I've being very slow on implementing new stuff for this project. So don't expect this release to happen soon. This is mostly to grab feedback.

Evertras:
Just a big +1 to the 'run a dependency with the same input only once' by default.  That's my only pain point right now, so I'm a little biased. :)

smyrman:
It's a great list, agree to all points; take your time:-)

edrd-f:
[...]

PR #220, opened 2019-06-23
v3 base branch

andreynering:
You can build from this branch if you want to anticipate v3 changes.

Backward incompatible changes may happen until this is released, though.

See main issue: https://github.com/go-task/task/issues/195


kovetskiy:
```
task: The "expansions" setting is not available anymore on v3.0
```

What is the expansions setting? Could *task* include line and filename where the error occurred?

Btw, here is my taskfile: https://github.com/reconquest/snake-runner (testing with `version: '3'` locally)

kovetskiy:
Ok, my first successful run failed of task failed due to a failed command in the Taskfile and here is an that I got from task:

```
task: error cleaning status on error: remove .task/checksum/rr: no such file or directory
```
(rr is the name of task)

PR #224, opened 2019-06-28
Added TASKFILE_DIR template variable

### LINKED ISSUE #215: dirname of current Taskfile
I often want to use the directory of where the Taskfile is located as an input to other commands (not the current working directory), but it seems like `sprig` really locks down access to the filesystem.

Is this possible to do with `Task` as a first-class idea?

For example, something like:

```
tasks:
  show-current-dir:
    desc: "show where Task is executing from"
    cmds:
      - echo {{ .TASKFILE_LOCATION }}
```

Note that this isn't the same as just using `$PWD` or `-d` because the working directory might be different; Task should probably provide this as a first-class feature if it doesn't already.

Shivakishore14:
Added TASKFILE_DIR template variable to access the directory where the task file is located.

### example:
```
version: '2'

tasks:
    build: 
        cmds:
            - echo '{{TASKFILE_DIR}}'
```

### output 
```
$ task build
echo '/Users/coda/go/src/github.com/go-task/task/cmd/task'
/Users/coda/go/src/github.com/go-task/task/cmd/task
```
closes #215

### DIFF
--- a/internal/templater/funcs.go
+++ b/internal/templater/funcs.go
@@ -1,6 +1,7 @@
 package templater
 
 import (
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @Shivakishore14, and thanks for opening this PR.

Unfortunately, there are use cases which won't properly work with this patch.

1) If someone use `--dir` to run a Taskfile in another directory, this will return the wrong directory
2) This will also return the wrong directory for included Taskfiles
3) Probably other similar use cases

I think this will be more easily done correctly if added as a variable instead of a function.

So, since I realized this isn't as simple, I just removed the `good first issue` from https://github.com/go-task/task/issues/215 and will close this for now.

Feel free to ping me if you have any questions or comments. 🙂

PR #226, opened 2019-07-30
allow skipping dependencies

dtomasi:
In some special cases within our CI/CD pipelines we have to run a task but not its dependencies. Therefore I added a option to skip them using `task blabla --skip-deps`.

PR #227, opened 2019-07-30
allow running dependencies serially

dtomasi:
In some cases we have to run dependency tasks serially. I added a option to the taskfile to allow this.

### DIFF
--- a/internal/taskfile/task.go
+++ b/internal/taskfile/task.go
@@ -5,20 +5,21 @@ type Tasks map[string]*Task
 
 // Task represents a task
 type Task struct {
-	Task         string
-	Cmds         []*Cmd
-	Deps         []*Dep
-	Desc         string
-	Summary      string
-	Sources      []string
-	Generates    []string
-	Status       []string
+	Task          string
+	Cmds          []*Cmd
+	Deps          []*Dep
+	Desc          string
+	Summary       string
+	Sources       []string
+	Generates     []string
+	Status        []string
 	Preconditions []*Precondition
-	Dir          string
-	Vars         Vars
-	Env          Vars
-	Silent       bool
-	Method       string
-	Prefix       string
-	IgnoreError  bool `yaml:"ignore_error"`
+	Dir           string
+	Vars          Vars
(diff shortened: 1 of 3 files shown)


andreynering:
Hi again @dtomasi,

Before accepting this feature request, I'd like to understand if there's real need for it.

Nowadays, it should already possible to run things serially: you just need to declare it directly in `cmds` instead of in `deps`:

```yaml
version: '2'

tasks:
  default:
    cmds:
      - task: foo
      - task: bar
      - echo "..."
```

Instead of:

```yaml
version: '2'

tasks:
  default:
    deps: [foo, bar]
    cmds:
      - echo "..."
```

Is there any reason why the above isn't enough to satisfy your use case?

dtomasi:
Hi @andreynering 

As described in #226 

```yaml
tasks:
  default:
    cmds:
      - task: foo
      - task: bar
      - echo "..."
```

will not work for me, as I want to skip the dependencies. Skipping the first 3 `cmds` in the list is not possible ;-)

smyrman:
I think I hold with @andreynering's suggestion in #226 that you should rely on two tasks for what you want. Then you don't need neither of these two features:

```yaml
version: "2"

tasks:
  build:
    cmds:
    - task: foo
    - task: bar
    - task: build-no-deps
  build-no-deps:
    cmds:
    - echo "..."
```

The benefit for relying on the Taskfile over adding new flags, is that you state what you want in a _declarative_ way, rather than adding flags that _alter_ how things are run.

dtomasi:
> Prevent unnecessary work 

Coming from your documentation ;-) 
I´m not really happy with this suggested workaround. However, will keep this feature in my fork then, as I really want to use it as described. 

Cheers

andreynering:
Hi @dtomasi,

I'm going to reject both proposals for now as IMO this use case is already possible today with changes in your Taskfile (my making tasks less coupled / dependent on each other).

I'm sorry, but thanks anyway for opening these pull requests!

I can revisit this decision in the future if more people show interest or need in these features.

PR #228, opened 2019-08-01
Re-run the task if generated files do not exist

topecongiro:
When using `method: checksum`, I was surprised to find out that even when files specified by `generates` do not exist the task won't get triggered as long as the source files are not modified. IMHO this behavior is unintuitive, especially considering how Makefile works. This PR adds a logic to re-run the task when files specified in generates do not exist.

### DIFF
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -34,7 +34,10 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 	}
 
 	generates, err := glob(c.Dir, c.Generates)
-	if err != nil || len(generates) == 0 {
+	if err != nil {
+		return false, err
+	}
+	if len(generates) == 0 {
 		return false, err
 	}
 	for _, generate := range generates {
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -28,7 +28,7 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 	data, _ := ioutil.ReadFile(checksumFile)
 	oldMd5 := strings.TrimSpace(string(data))
 
-	sources, err := glob(c.Dir, c.Sources)
+	sources, err := globs(c.Dir, c.Sources)
 	if err != nil {
 		return false, err
 	}
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -14,10 +14,11 @@ import (
 // Checksum validades if a task is up to date by calculating its source
 // files checksum
 type Checksum struct {
-	Dir     string
-	Task    string
-	Sources []string
-	Dry     bool
+	Dir       string
+	Task      string
+	Sources   []string
+	Generates []string
+	Dry       bool
 }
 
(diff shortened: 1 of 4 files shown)


internal/status/checksum.go
```diff
@@ -32,6 +33,16 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 		return false, err
 	}
 
+	generates, err := glob(c.Dir, c.Generates)
+	if err != nil || len(generates) == 0 {
```
andreynering: Even if the execution is equal, I think that for code clarity this should be two `if` checks:

```go
if err != nil {
	return false, err
}
if len(generates) == 0 {
	return false, nil
}
```

These checks can also be moved as the very first thing to run in this function.

andreynering: Additionally, since `glob` also returns directories, the second check would not work as is if only directories are returned, but no files.

I'm asking myself if we should only return files (and not directories) on `glob` [here](https://github.com/go-task/task/blob/4c295b564a7da0281b70652df09d429eab77684c/internal/status/glob.go#L25). If we do that, we could also removed [these lines](https://github.com/go-task/task/blob/4c295b564a7da0281b70652df09d429eab77684c/internal/status/checksum.go#L61-L63).

internal/status/checksum.go
```diff
@@ -32,6 +33,16 @@ func (c *Checksum) IsUpToDate() (bool, error) {
...
+		return false, err
+	}
+	for _, generate := range generates {
+		if _, err := os.Stat(generate); err != nil {
```
andreynering: I think this `for` loop + check is not necessary at all. Let me know if I'm missing something.

internal/status/checksum.go
```diff
@@ -32,6 +33,16 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 		return false, err
 	}
 
+	generates, err := glob(c.Dir, c.Generates)
```
andreynering: Hi @topecongiro,

I just reminded that have no globs listed at `generates:` with `method: checksum` is a totally valid use case.

For this reason, we should only do the checks below if the user has informed any glob on `generates:`:

```go
if len(c.Generates) > 0 {
	// do all the stuff
}
```

internal/status/checksum.go
topecongiro: Thank you for the reviews. Addressed them in the new commits.

internal/status/checksum.go
topecongiro: Yes this is unnecessary indeed. Thank you for the catch!

internal/status/checksum.go
topecongiro: Done!

andreynering:
Hi @topecongiro,

Thanks again! I merged and did small code improvements here: https://github.com/go-task/task/commit/0d9fdbaac1e8d8a04331edeedbed64daac361ac5.

Issue #235, opened 2019-08-09
Add interactive input system (feature++ awesomeness++++)


JulienBreux:
Hi awesome team!

# Request

Possibility to declare interactive inputs. Yeah budy! No less no more! #humor

# Simple example

## File: `Taskfile.yaml`

```yaml
version: "2"

task:
  aTaskWithInput:
    inputs:
      - foo
      - bar
      - baz
```

## How to use?

```shell
$ task aTaskWithInput
$ foo
$ <input>
$ bar
$ <input>
$ baz
$ <input>
```

# Advanced example

## File: `Taskfile.yaml`

```yaml
version: "2"

task:
  aTaskWithInput:
    inputs:
      foo:
        required: true
      bar:
        default: "🍻"
        desc: "Do you love beer?"
      baz:
        validator: "([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))"
```

## How to use?

```shell
$ task aTaskWithInput
$ foo (Is required)
$ <input>
$ Do you love beer? (default: 🍻)
$ <input>
$ baz (must validate: "([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))")
$ <input>
```

# Feedback need!

👍 OMG! I want this now!
😕 I don't know guy, it's cool but I can leave without!
👎 Please... I've a lot of work!

# Implem' status
1. Fri 9 Aug: Started (https://github.com/JulienBreux/task/tree/features/inputs)

Issue #92, opened 2018-02-16
Support user input at runtime


ebr:
It would be great to pause and wait for user input, for subsequent use in the script:

```
test:
  cmds:
    - read name
    - echo "hi ${name}"
```

```
$ task test 
read name
task: Failed to run task "test": exit status 127
```

I also tried moving the `read` command to Taskvars, but still no luck. I suspect this is because execution of go-task does not allocate a TTY? Is there any way to do this currently?

andreynering:
I never thought about this possibility, and I don't know why it doesn't work. I'll think about it.

andreynering:
This currently possible as follows:

```yml
read:
  cmds:
    - |
      echo "Type value of FOO:"
      read FOO;
      echo "You said: ${FOO}";
  silent: true
```

And for allowing `read` in variables, I don't know if that makes sense. One can pass parameters to tasks using environment variables or the special syntax:

```sh
task foo BAR=baz
```

tommyknows:
Hi, 
Stumbled upon this issue while trying to work out the exact same thing.
I want to ask the user for input for a few variables that are going to be used throughout the Taskfile.

Obviously, as you noted, something like this panics:
```yaml
vars:
  MY_VAR:
    sh: read
```

At the same time, I can't use this:
```yaml
env:
  MY_VAR: "{{.MY_VAR}}"
tasks:
  get-vars:
    cmds:
      - {{if not .MY_VAR}}read MY_VAR; export MY_VAR{{end}}
  show-vars:
    cmds:
      - echo {{.MY_VAR}}
```

Is there any way to export variables from the `cmds` to the taskfile?

andreynering:
@tommyknows Isn't passing variables via CLI enough for you? (See my comment [here](https://github.com/go-task/task/issues/92#issuecomment-370253942))

tommyknows:
Well, as I've got quite a few variables to ask for, no (or only partially).
I wanna allow the user to supply some variables via CLI (if he wishes to automate the call), and ask for every variable that is missing.

andreynering:
I think doing that via `read` results in a bad UX experience, because the user won't know what variable they are typing.

If we implement something to allow that, that should have be handled by Task, that would print something like this:

```
task: Please type the value for VAR (default: foobar): <cursor>
```

I still think using the CLI a better on most use cases, though.

andreynering:
Some discussion about adding this is happening at https://github.com/go-task/task/pull/236.

PR #236, opened 2019-08-09
Add interactive input system

JulienBreux:
## Link

🔗 https://github.com/go-task/task/issues/235

## Todo

- [x] Feature MVP (required, default, validator regexp)
- [x] Get feedback
- [ ] Add verbose logs
- [ ] Tests
- [ ] Commits clean/rebase

## Description

The need is to be able to request inputs.

### File: `Taskfile.yaml`

```yaml
version: "2"

task:
  aTaskWithInput:
    inputs:
      foo:
        required: true
      bar:
        default: "🍻"
        desc: "Do you love beer?"
      baz:
        validator: "([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))"
```

### How to use?

```shell
$ task aTaskWithInput
$ foo (required: yes)
$ <input>
$ Do you love beer? (default: 🍻)
$ <input>
$ baz (must validate: "([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))")
$ <input>
```

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -18,7 +18,7 @@ var (
 	version = "master"
 )
 
-const usage = `Usage: task [-ilfwvsd] [--init] [--list] [--force] [--watch] [--verbose] [--silent] [--dir] [--dry] [--summary] [task...]
+const usage = `Usage: task [-ilfwvsd] [--init] [--list] [--force] [--watch] [--verbose] [--silent] [--dir] [--dry] [--summary] [--interactive] [task...]
 
 Runs the specified task(s). Falls back to the "default" task if no task name
 was specified, or lists all tasks if an unknown task name was specified.
@@ -58,6 +58,7 @@ func main() {
 		silent      bool
 		dry         bool
 		summary     bool
+		interactive bool
 		dir         string
 		output      string
 		color       bool
@@ -73,6 +74,7 @@ func main() {
 	pflag.BoolVarP(&silent, "silent", "s", false, "disables echoing")
 	pflag.BoolVar(&dry, "dry", false, "compiles and prints tasks in the order that they would be run, without executing them")
 	pflag.BoolVar(&summary, "summary", false, "show summary about a task")
+	pflag.BoolVar(&interactive, "interactive", true, "interactive prompt inputs")
 	pflag.StringVarP(&dir, "dir", "d", "", "sets directory of execution")
 	pflag.StringVarP(&output, "output", "o", "", "sets output style: [interleaved|group|prefixed]")
 	pflag.BoolVarP(&color, "color", "c", true, "colored output")
@@ -95,14 +97,16 @@ func main() {
 	}
 
 	e := task.Executor{
-		Force:   force,
-		Watch:   watch,
-		Verbose: verbose,
-		Silent:  silent,
-		Dir:     dir,
-		Dry:     dry,
-		Summary: summary,
-		Color:   color,
+		Force:       force,
+		Watch:       watch,
+		Verbose:     verbose,
+		Silent:      silent,
+		Dir:         dir,
+		Dry:         dry,
+		Entrypoint:  entrypoint,
+		Summary:     summary,
+		Color:       color,
+		Interactive: interactive,
 
 		Stdin:  os.Stdin,
 		Stdout: os.Stdout,
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -18,6 +18,7 @@ _arguments \
     '(-l --list)'{-l,--list} \
     '(-s --silent)'{-s,--silent} \
     '(--status)'--status \
+    '(--interactive)'--interactive \
     '(-v --verbose)'{-v,--verbose} \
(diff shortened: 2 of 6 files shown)


andreynering:
Hi @JulienBreux,

Let's discuss this a bit before continuing...

What's your use case for it? Why do you prefer it to passing [variables](https://taskfile.dev/#/usage?id=variables) via CLI or environment variables?

(Also, if we're going to have this feature, I'd prefer to target it to the `v3` branch instead of `master`).

JulienBreux:
Hi @andreynering,

> What's your use case for it? Why do you prefer it to passing [variables](https://taskfile.dev/#/usage?id=variables) via CLI or environment variables?

My use case is very simple ; add more user interaction with go-task.

Variables are very cool but a bit unsecure (history in your shell) and more static (only on running command).

The idea is to have the capability of pompt the user and re-use command with go-task.

## Possible use case

Request something to user (password, token, etc).

smyrman:
https://github.com/AlecAivazis/survey has some nice options for input; a go-task interactive mode could potentially use that for setting task and/or global variables. Require a yaml validaition model that maps to the different input options one would want to support.

andreynering:
Hi @JulienBreux,

Indeed, I think passwords are a great use case for it!

As an additional feedback:

1) For a matter of simplicity, I think I'd get rid of the "validator" stuff. If the most common use case will be for passwords and tokens, this wouldn't be much useful (since passwords doesn't have a pettern)
2) Do we really need the `--interactive` flag? Shouldn't we always ask for input if the task requires it?

andreynering:
> https://github.com/AlecAivazis/survey has some nice options for input; a go-task interactive mode could potentially use that for setting task and/or global variables. Require a yaml validaition model that maps to the different input options one would want to support.

That seems an interesting project. I like how it allows to ask to select a value from a list of possible values (as an alternative to free text).

JulienBreux:
Nice, I work on it this night (UTC+1).

Thanks for your feedbacks 🤗

marco-m:
@andreynering 

> Indeed, I think passwords are a great use case for it!

I am a minimalist and I like that Task tries to stay minimal. In this optics, I would like to suggest an alternative to the proposal in this PR that, I think, doesn't require to change Task at all.

It is possible to use the OS secure storage and make the secrets available to Task as environment variables via a wrapper. I use https://github.com/sorah/envchain with Task and it works flawlessly.

For example:

```
envchain FOO task ...
```

will run `task ...`, with all the environment variables for the `FOO` "namespace" extracted from the OS secure storage. On Mac it is the Keychain, on Linux it is Gnome Keyring.

Would this be a viable alternative ?

andreynering:
@marco-m Indeed, accepting or rejecting proposals is hard.

But while I definitely care about simplicity and maintainability, I also see how this could be useful. This was also asked before more than once (see #92), so I think this is more than worth discussing. I still think the design and code should be as simple as possible.

Another important consideration is that envchain doesn't seem to support Windows, and it's a core Task selling point that it tries to work the best it can on every supported OS.

I'd say this isn't a target for v3, though. Perhaps v3.1. v3 has lot of target features already, and I think I'll even move some to v3.x in order to be able to release them sooner.

marco-m:
@andreynering I understand, and I appreciate you taking the time to give your rationale.
I think it is possible to make envchain work in a Windows environment, but I agree that somebody would have to do the work.

JulienBreux:
After long time, I think that the more efficient experience maybe a "command provider".
For example:

**Short**
```
tasks:
  myTask:
    cmds:
      - cmd: echo "Hello"
      - input: yyy
      - cmd: echo "Hi {{ .yyy }}!"
```

**Long**
```
tasks:
  myTask:
    cmds:
      - cmd: echo "Hello"
      - input:
         name: yyy
         default: xxx
         desc: "Do you love beer?"
         validator: "([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))"
      - cmd: echo "Hi {{ .yyy }}!"
```

andreynering:
> After long time, I think that the more efficient experience maybe a "command provider".

Sorry for the long wait, I've been relatively busy recently.

It's not easy to address proposal sometimes, but I decided to close this for now, since other stuff is higher in the priority list. Also, seems it's still not totally clear what path we should take here. I may reconsider adding this in the future.

@JulienBreux I hope you undertand.

Jarmos-san:
I stumbled across [#92](https://github.com/go-task/task/issues/92) through a Google Search, so I was wondering whether this feature implementation is still in the pipeline?

JulienBreux:
Nope, this proposal was closed by @andreynering :)

andreynering:
Hey guys, I just opened some issues about adding more interactivity to Task.

- https://github.com/go-task/task/issues/2077
- https://github.com/go-task/task/issues/2078
- https://github.com/go-task/task/issues/2079

plutonium-239:
`bubbletea` is amazing!

PR #237, opened 2019-08-13
Remove all code related support of version 1

jaedle:
Pull request for cleanup task `Remove all code related support of version 1`

Changes:
- versions prior to version `2` are not supported anymore ==> error on parsing of those files
- files without any version information are considered as version `1`
- all tests specific related to version `1` are removed
- all test fixtures are converted to version `2`
- documentation states that version `1` is deprecated
- inline example on task cli help is updated to version 2

Ready for review :)

If I missed anything, please tell me!


cmd/task/task.go
```diff
@@ -28,12 +28,14 @@ Example: 'task hello' with the following 'Taskfile.yml' file will generate an
 'output.txt' file with the content "hello".
 
 '''
-hello:
-  cmds:
-    - echo "I am going to write a file named 'output.txt' now."
-    - echo "hello" > output.txt
-  generates:
-    - output.txt
+version: '2'
```
andreynering: Actually, this should be version 3 since this PR is targeting the `v3` branch 🙂 

Good catch, though. This example was forgotten for a long time.

```suggestion
version: '3'
```

docs/taskfile_versions.md
```diff
@@ -15,21 +15,7 @@ available, but not `3.0.0+`.
 
 ## Version 1
 
-In the first version of the `Taskfile`, the `version:` key was not available,
```
andreynering: Actually, I think we should probably still keep this documentation, but add a notice right after the title informing that this is not available anymore on Task >= v3.0.0.

task.go
```diff
@@ -122,14 +121,16 @@ func (e *Executor) Setup() error {
 	if err != nil {
 		return fmt.Errorf(`task: Could not parse taskfile version "%s": %v`, e.Taskfile.Version, err)
 	}
+
+	if v < 2 {
+		return fmt.Errorf(`task: Taskfile version prior to v2 are not supported anymore`)
```
```suggestion
		return fmt.Errorf(`task: Taskfile versions prior to v2 are not supported anymore`)
```

task_test.go
```diff
@@ -678,3 +640,12 @@ func TestWhenDirAttributeItCreatesMissingAndRunsInThatDir(t *testing.T) {
...
 	_ = os.Remove(toBeCreated)
 }
+
+func TestDisplaysErrorOnUnsupportedVersion(t *testing.T) {
+	e := task.Executor{
+		Dir:    "testdata/version/v1",
+		Stdout: ioutil.Discard,
+		Stderr: ioutil.Discard,
+	}
+	assert.Error(t, e.Setup())
```
andreynering: We should also assert the output message here, otherwise this test will pass if another error happen.

jaedle:
I hope that is now good to go! :)

I also bumped the version of the auto generated taskfile to version `3`. I hope that was fine! :)

andreynering:
Awesome!

Thanks again @jaedle!

Issue #194, opened 2019-03-30
Allow shorter syntax for tasks with default configuration


andreynering:
Instead of requiring tasks to be declared as:

```yaml
version: '2'

tasks:
  foobar:
    cmds:
      - echo "something"
```

We could allow:

```yaml
version: '3'

tasks:
  foobar:
    - echo "some"
    - echo "thing"
```

Or even:

```yaml
version: '3'

tasks:
  foobar: echo "something"
```

If users want to change any task configuration, they'd have to use the original form.

Ilyes512:
I don’t actually mind having more verbose task. What I do miss is the option to configure `silent: true` globally instead of having to repeat it everywhere.

andreynering:
This was just implemented on the `v3` branch in b7b752b92f5cb2b38a929eb447b36cf5fc35c80e.

andreynering:
> What I do miss is the option to configure `silent: true` globally instead of having to repeat it everywhere.

@Ilyes512 Sorry, I forgot about your suggestion for a long time, but this is now done at 4bdfe64afb6cf96c4daa1faacf06b036864506e7.

For future feature requests, it's better to open new issues so it's less likely it'll be forgotten.

PR #240, opened 2019-08-23
[WIP] Allow shorter syntax for tasks with default configuration

jaedle:
See #194

After doing some coding/research how to implement this, I have a few questions:

1. If I got it right, this is a new feature only for version 3 taskfiles?
2. What should happen if an older version (i.e. version 2) is used with that notation? Should that result in an parsing error? - This would mean like having a switch inside of `internal/taskfile/taskfile.go` for disallow a shorter syntax for older taskfile versions?
3. Because the unmarshalling has now different source structures (unmarshal from task structure, umarshal an array of strings, unmarshal a string) how should be dealt with the possible errors? I think giving a good error message may be complicated, because we can't guess which synrax should be used. For an approach to unmarshal different types within go-yaml see https://github.com/go-yaml/yaml/issues/100#issuecomment-110505747

(I will squash this branch before review)


andreynering:
Hi @jaedle, thanks for opening this PR!

Regarding 1) and 2), I think would not be a *that* big deal to allow the shorter syntax on v2, too. (Even then, this should still target the `v3` branch). That's what we've been done with other syntax alternatives, too.

That said, if we find a easy/right way to only allow it on v3, it's be the ideal scenario. If you're unsure about how to do that, it could be pulled to a future fix/PR.

Regarding 3), we already do that for other data structures. In short, what you should do is:

1) If the value is a string slice, assign it to `task.Cmds` and keep other attributes zeroed/default;
2) The same if the value is a string, just wrap it into a string slice and assign it to `task.Cmds`.

Does that makes sense?

internal/taskfile/taskfile.go
```diff
@@ -40,3 +40,61 @@ func (tf *Taskfile) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 	return nil
 }
+
+func (t *Task) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var task struct {
```
andreynering: Hey @jaedle,

IMHO you don't need to copy the entire task struct into this function. Just try to marshal the `string` and `[]string` types first, and none work then unmarshal to the `t` variable directly. 🙂

testdata/short_task_notation/Taskfile.yml
```diff
@@ -0,0 +1,16 @@
+version: '3'
+
+tasks:
+
```
andreynering:
```suggestion
```

internal/taskfile/task.go
```diff
@@ -20,5 +20,5 @@ type Task struct {
 	Silent       bool
 	Method       string
 	Prefix       string
-	IgnoreError  bool `yaml:"ignore_error"`
```
andreynering: Given the suggestion made on `func (t *Task) UnmarshalYAML`, we should re-add this reflect tag.

internal/taskfile/taskfile.go
andreynering: Oops, I just remembered that we also have alternative command syntaxes, like `{ cmd: "echo foo", silent: true }` for example.

That's easy to do, though. Just check for `taskfile.Cmd` and `[]taskfile.Cmd` instead of for `string` and `[]string`. All variations should just work.

andreynering:
Hi @jaedle,

I just manually merged your work on `v3` with some changes on b7b752b92f5cb2b38a929eb447b36cf5fc35c80e.

Thanks a lot for contributing!

PR #266, opened 2019-11-13
Add flag to allow tasks provided on the command line to be run in parallel

RossHammer:
I have a Taskfile with many different targets and the targets I want to run vary often enough I cannot just put them under another task's `dep`. I'm looking for when I run I supply the list of targets via the command line they will execute in parallel rather than sequentially.

Example of what I looking to
```
$ time task -p sleep1 sleep2
sleep 10
sleep 10
echo sleep 1
echo sleep 2
sleep 1
sleep 2

real	0m10.986s
user	0m0.755s
sys	0m0.530s
```

```yaml
version: '2'

tasks:
  default:
    deps:
      - task: sleep1
      - task: sleep2
      - task: sleep3
  sleep1:
    cmds:
      - sleep 10
      - echo sleep 1
  sleep2:
    cmds:
      - sleep 10
      - echo sleep 2
  sleep3:
    cmds:
      - sleep 10
      - echo sleep 3
```

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -58,6 +58,7 @@ func main() {
 		silent      bool
 		dry         bool
 		summary     bool
+		parallel    bool
 		dir         string
 		entrypoint  string
 		output      string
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @RossHammer,

That's fair enough. Thanks for contributing! 🙂

PR #269, opened 2019-11-15
fix: tab indent 4 space

klauern:


andreynering:
Hi @klauern,

Thanks for opening this pull request!

I understand that 8 spaces tabs is a more common convention in the Go world than 4 or 2. That's the reason it's currently set as is.

But giving it a bit of thought... what if we just removed this line? This way, everybody can set their preferred tab size in their editor and be happy. 🙂 

Since this is personal preference, perhaps we shouldn't enforce any tab size.

andreynering:
Closed at ec934ba3c0f06870da50f3408dc09e7e8c9b00e3

PR #281, opened 2020-01-20
Add PowerShell completion script

paulvarache:
This PR adds a script to register a PowerShell completion helper

Relates to https://github.com/go-task/task/issues/103


andreynering:
Thanks @paulvarache! ❤️

PR #285, opened 2020-01-24
added support for docker images in tasks

no0dles:
Hi, really like your taskfile project and use it almost every day. I created a small proof of concept for running tasks inside docker containers. 

This feature mounts the current working directory into the container, executes the commands inside the container environment and forwards the log outputs. 

Personally I do not have a lot of go experience, maybe it could be done in a better way. Open for feedback.


internal/execext/exec.go
```diff
@@ -3,6 +3,14 @@ package execext
...
 	"context"
 	"errors"
+	"fmt"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/mount"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/docker/docker/pkg/term"
```
andreynering: All imports from packages outside the standard library should be moved to the second block.

andreynering: All imports from packages outside the standard library should be moved to the second block.

internal/execext/exec.go
```diff
@@ -64,6 +77,86 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
...
+		return err
+	}
+
+	pullReader, err := cli.ImagePull(context.Background(), image.String(), types.ImagePullOptions{})
+	if err != nil {
+		return err
+	}
+	defer pullReader.Close()
+
+	termFd, isTerminal := term.GetFdInfo(os.Stderr)
```
andreynering: Should probably use `opts.Stderr` instead of `os.Stderr`.

internal/execext/exec.go
```diff
@@ -64,6 +77,86 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
...
+	err = jsonmessage.DisplayJSONMessagesStream(pullReader, os.Stderr, termFd, isTerminal, nil)
```
andreynering: Same here

internal/execext/exec.go
```diff
@@ -64,6 +77,86 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
...
+		ShowStderr: true,
+		Follow:     true,
+	})
+	if err != nil {
+		return err
+	}
+	defer reader.Close()
+
+	go func() {
+		io.Copy(opts.Stdout, reader)
```
andreynering: Question: how to handle any errors returned by `io.Copy`?

no0dles:
Hey @andreynering 
Thank you very much for your feedback.

Agree with the --docker-image flag, only question would be if that flag applies to deps and sub-tasks.

Tried to fix your feedback and discovered another issue. Currently the context gets lost, because each step is in a separat container. This is bad for multiple use cases, for example this:

```
tasks:
  build:
    image: node:12
   cmds:
     - npm i -g typescript
     - tsc
```

Output: `in/sh: 1: tsc: not found`

no0dles:
Another thing would be changing directory, currently the behaviour with and without an image is the same (see example below). But as soon as I would change the docker implementation to execute all the commands in one container, this would differ from the default way without an image.

```
tasks:
  test-pwd:
    dir: packages
    cmds:
      - pwd
      - cd ..
      - pwd
```

Output
```
test $ task test-pwd
task: pwd
/Users/pascal/test/dir
task: cd ..
task: pwd
/Users/pascal/test/dir

```

no0dles:
what's the reason for closing this issue? never heard back from you.. would be sad to close this feature without any response

andreynering:
Hi @no0dles,

Sorry, this was a mistake. GitHub closed many PRs automatically after merging #220

Reopening...

Also, I think you deserve an explanation on why I never came back to this PR: I was just busy, and focusing on features that will be released on v3 (hopefully very soon). I want to back to this at some point, but I can't promise when.

orbatschow:
@andreynering  Are there any plans to merge this feature? (IMHO Docker images inside a Taskfile would be pretty awesome.)

andreynering:
@orbatschow Sorry, but this may take some time.

I have been having little time to work on Task, and bug fixes are a priority at the moment.

max-sixty:
This would be really useful for some tasks, where there are complex dependencies and a simple result. Some CI pipelines act similarly — select an image and a command — and this could be a cross-platform, abstract version of this. (Alternatively I'd be interested if there are alternatives that are better suited to this sort of task). Thanks.

no0dles:
I am closing this PR since I created another [tool](https://github.com/no0dles/hammerkit) that solves that issue and it's easier for me to contribute there in a language I am familiar with. Thanks @andreynering

andreynering:
Hi @no0dles,

You're did a great job on hammerkit, congrats! I saw other build tools inspired by Task out there, but I think this is the most interesting given the simplicity and Docker support. (I haven't tried it yet, though).

I was very in doubt on how an integration with Docker would look like and yours seems like great, simple and easy to understand approach. Hope you don't mind if I happen to take some inspiration from there 😉. That shouldn't happen too soon, though.

no0dles:
@andreynering no worries, I did take some inspiration from taskfile as well 😄. Good luck, looking forward to a new release with this.

Issue #260, opened 2019-11-05
Task directory for included Taskfiles


wtcross:
**Task Version**: 2.7.0

It appears that a Task's directory is set to that of the root Taskfile if it is included and doesn't specify its own directory.

https://taskfile.dev/#/usage?id=task-directory

This was unexpected to me since it affects how file paths can be structured in included Taskfiles. The scenario that caused me to look at it this way was to use Task to replace Make for a monorepo. Right now it's working great, but I have to run all tasks from the root of the project if I want it to work. Ideally I'd be able to run the included Taskfiles, which are local to each package in the monorepo.

Is this something on the roadmap? After looking at the code it looks like adding a `Dir` field to the `Taskfile` struct and massaging how tasks are run could enable a Task to have the following level of precedence for its directory:
1. Set at Task level
2. Parent Taskfile (might not be the root directory of the project)
3. Root Taskfile directory

wtcross:
I'd try to work on this if you're interested in this sort of contribution. It could break Taskfiles using v2 so I understand it would need to be carefully considered.

andreynering:
Hi @wtcross, thanks for opening this issue!

There's [an open issue to track the ability of making included Taskfiles run on its own directory](https://github.com/go-task/task/issues/144). Please let me know of your thoughts (would it suffice your use case?).

Also, [work on v3](https://github.com/go-task/task/issues/195) (on the [v3 branch](https://github.com/go-task/task/pulls)), although slow, is ongoing. So this is the right time to consider adding that.

Contribution is always welcome, but I'll probably do that myself at some point otherwise. In case you want to try, feel free to open a WIP PR and ask for direction/help if needed.

andreynering:
@wtcross As this is a duplicate of #144, I'll close this.

And FYI, this is a WIP right now: https://github.com/go-task/task/pull/292

Issue #144, opened 2018-11-04
Add option to make included Taskfile run commands on its own directory


andreynering:
Currently, when you include a Taskfile, its commands are ran on the main Taskfile directory.

We should allow users to set a Taskfile to run its commands on the included Taskfile directory.

---

Syntax brainstorming:

```yaml
includes:
  docs:
    taskfile: ./docs/Taskfile.yml
    dir: ./docs
```

Or, alternatively, we can introduce a different keyword for it: `imports` instead of `includes`, so:

- `includes`: runs on the main Taskfile directory;
- `imports`: runs on imported Taskfile directory.

I think I prefer the latter.

andreynering:
We should write a test case to make sure this snippet returns the expected directory:

```yaml
vars:
  DIRECTORY: {sh: pwd}
```

tonobo:
This should be pretty easy after merging #145 :smile:

But what happens on tasks which already have `dir` option. Should it be overwritten, alternatively we could ignore it or prepend the scrope if the task path seems to be a relative one.

andreynering:
> But what happens on tasks which already have dir option. Should it be overwritten, alternatively we could ignore it or prepend the scope if the task path seems to be a relative one.

@tonobo We should [`filepath.Join()`](https://golang.org/pkg/path/filepath/#Join) them if the second one is relative. If absolute, keep as is.

smyrman:
Alternate solution could be #215.

andreynering:
FYI, this should soon be release to the v3 preview channel: https://github.com/go-task/task/pull/292

andreynering:
Implementation in the preview channel.

dhax:
> We should write a test case to make sure this snippet returns the expected directory:
> 
> ```yaml
> vars:
>   DIRECTORY: {sh: pwd}
> ```

experiencing this problem with included taskfiles not running sh command in their specified directory, making e.g. the example fail because it's now run in the root dir if not prepended with cd my/dir &&:
```
vars:
  GIT_COMMIT:
    sh: cd my/dir && git log -n 1 --format=%h
```

andreynering:
Hi @dhax,

Ty thanks for pinging me. I opened an issue to not forget about it: https://github.com/go-task/task/issues/384

PR #292, opened 2020-02-09
Added option to make included Taskfile run commands on its own directory

evg4b:
Hi, this pull request relates to [#260](https://github.com/go-task/task/issues/260) and [#144](https://github.com/go-task/task/issues/144) issues. This syntax was implemented:

``` yaml
includes:
  docs:
    taskfile: ./docs/Taskfile.yml
    dir: ./docs
```

Also added support `Taskfile_{GOOS}.yaml` for included task files.

### DIFF
--- a/internal/taskfile/included_taskfile.go
+++ b/internal/taskfile/included_taskfile.go
@@ -1,0 +1,38 @@
+package taskfile
+
+import "errors"
+
+var (
+	// ErrCantUnmarshalIncludedTaskFile is returned for invalid var YAML.
+	ErrCantUnmarshalIncludedTaskFile = errors.New("task: can't unmarshal included value")
+)
+
+// IncludedTaskFile represents information about included tasksfile
+type IncludedTaskFile struct {
+	Taskfile string
+	Dir      string
+}
+
+// IncludedTaskFiles represents information about included tasksfiles
+type IncludedTaskFiles = map[string]IncludedTaskFile
+
+// UnmarshalYAML implements yaml.Unmarshaler interface
+func (it *IncludedTaskFile) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var str string
+	if err := unmarshal(&str); err == nil {
+		it.Taskfile = str
+		return nil
+	}
+
+	var includedTaskfile struct {
+		Taskfile string
+		Dir      string
+	}
+	if err := unmarshal(&includedTaskfile); err == nil {
+		it.Dir = includedTaskfile.Dir
+		it.Taskfile = includedTaskfile.Taskfile
+		return nil
+	}
+
+	return ErrCantUnmarshalIncludedTaskFile
+}
--- a/internal/taskfile/included_taskfile.go
+++ b/internal/taskfile/included_taskfile.go
@@ -9,8 +9,9 @@ var (
 
 // IncludedTaskfile represents information about included tasksfile
 type IncludedTaskfile struct {
-	Taskfile string
-	Dir      string
+	Taskfile       string
+	Dir            string
+	AdvancedImport bool
 }
 
 // IncludedTaskfiles represents information about included tasksfiles
@@ -31,6 +32,7 @@ func (it *IncludedTaskfile) UnmarshalYAML(unmarshal func(interface{}) error) err
 	if err := unmarshal(&includedTaskfile); err == nil {
 		it.Dir = includedTaskfile.Dir
 		it.Taskfile = includedTaskfile.Taskfile
+		it.AdvancedImport = true
 		return nil
 	}
 
--- a/internal/taskfile/included_taskfile.go
+++ b/internal/taskfile/included_taskfile.go
@@ -3,21 +3,21 @@ package taskfile
 import "errors"
 
 var (
-	// ErrCantUnmarshalIncludedTaskFile is returned for invalid var YAML.
-	ErrCantUnmarshalIncludedTaskFile = errors.New("task: can't unmarshal included value")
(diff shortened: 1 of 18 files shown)


internal/taskfile/included_taskfile.go
```diff
@@ -0,0 +1,38 @@
+package taskfile
+
+import "errors"
+
+var (
+	// ErrCantUnmarshalIncludedTaskFile is returned for invalid var YAML.
+	ErrCantUnmarshalIncludedTaskFile = errors.New("task: can't unmarshal included value")
```
andreynering: `TaskFile` -> `Taskfile`

internal/taskfile/included_taskfile.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// IncludedTaskFile represents information about included tasksfile
+type IncludedTaskFile struct {
```
andreynering: `IncludedTaskFile` -> `IncludedTaskfile`

internal/taskfile/included_taskfile.go
```diff
@@ -0,0 +1,38 @@
...
+	Taskfile string
+	Dir      string
+}
+
+// IncludedTaskFiles represents information about included tasksfiles
+type IncludedTaskFiles = map[string]IncludedTaskFile
```
andreynering: `IncludedTaskFiles` -> `IncludedTaskfiles`

internal/taskfile/read/taskfile.go
```diff
@@ -44,6 +45,24 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
 		if len(includedTaskfile.Includes) > 0 {
 			return nil, ErrIncludedTaskfilesCantHaveIncludes
 		}
+
+		includedTaskDirectory := filepath.Dir(path)
+		includedTaskfileName := strings.TrimSuffix(filepath.Base(path), filepath.Ext(path))
+		path = filepath.Join(includedTaskDirectory, fmt.Sprintf("%s_%s.yml", includedTaskfileName, runtime.GOOS))
```
andreynering: Hey @evg4b,

Actually, including these files automatically going in the reverse direction I had in mind.

I planned to deprecate the automatic inclusion of Taskfiles by OS in v3, and support templating to something like this would work:

```yaml
includes:
  taskfile: Taskfile_{{OS}}.yml
  namespace: false
```

(The `namespace: false` part needs thought, but we would need a way allow overwriting instead of namespacing).

I propose that you just remove the automatic inclusion for this PR. No need to implement the above now, I can do that in another PR.

What do you think?

internal/taskfile/read/taskfile.go
```diff
@@ -44,6 +45,24 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			if err = taskfile.Merge(includedTaskfile, osIncludedTaskfile); err != nil {
+				return nil, err
+			}
+		}
+
+		for _, task := range includedTaskfile.Tasks {
+			task.Dir = filepath.Join(includedTask.Dir, task.Dir)
```
andreynering: We need to skip joining if the path in the task is absolute:

```go
if !filepath.IsAbs(task.Dir) {
	task.Dir = filepath.Join(includedTask.Dir, task.Dir)
}
```

testdata/includes/module1/Taskfile.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+tasks:
+  gen_dir:
+    cmds:
+      - echo included_directory_without_dir > included_directory_without_dir.txt
+  gen_file:
```
andreynering: Add an empty line to separate tasks

testdata/includes/module2/Taskfile.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+tasks:
+  gen_dir:
+    cmds:
+      - echo included_directory_with_dir > included_directory_with_dir.txt
+  gen_file:
```
andreynering: Add an empty line to separate tasks

testdata/includes/module1/Taskfile.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
```
andreynering: Those test files should all be in version 3. Also, we should disallow setting a dir if version is 2. Just returning an error if version is 2 and dir is not `""` is fine. This should be on `Setup()` like the other version validations.

andreynering:
If you don't mind, adding an entry in the CHANGELOG and updating documentation would also be nice.

If your time is short, I can do that once I merge this.

internal/taskfile/read/taskfile.go
evg4b: Sounds good.
Sorry, I did not carefully read the roadmap. I removed this functionality.

internal/taskfile/read/taskfile.go
evg4b: You're right, also I fixed this point for included files (`internal\taskfile\read\taskfile.go:32`), it works incorrectly in some cases.
Example:
image

evg4b:
> If you don't mind, adding an entry in the CHANGELOG and updating documentation would also be nice.
> 
> If your time is short, I can do that once I merge this.

I fixed all your comments and added an entry in the CHANGELOG.md, but updating documentation can be difficult for me because my English isn't good.

andreynering:
@evg4b Seems great, thanks!

I'll add the documentation myself, no problem.

PR #298, opened 2020-02-19
Remove trailing colon from task listing

thejray:
This fixes bash autocompletion that fails by producing an extra colon at the end of completions.  Note that there is a similar sed for colon removal already present in the zsh autocomplete

It also removes an unnecessary `sed 's/^\* //'` in favor of just using `print $2` in the awk command.

### DIFF
--- a/completion/bash/task.bash
+++ b/completion/bash/task.bash
@@ -6,7 +6,7 @@ _task_completion()
   # Remove colon from word breaks
   COMP_WORDBREAKS=${COMP_WORDBREAKS//:}
 
-  scripts=$(task -l | sed '1d' | sed 's/^\* //' | awk '{ print $1 }');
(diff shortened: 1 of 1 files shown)


iquiw:
How about using one `awk` command?
```
task -l  | awk 'NR > 1 { sub(/:$/, "", $2); print $2 }'
```
Or one `sed` ?
```
task -l  | sed -n '1d; s/^\* \([^ ]*\):.*/\1/p'
```

(These are not equivalent. The latter needs `:` to exist, the former does not.)

thejray:
@iquiw  I leave it up to the maintainers.  The PR I have is a small delta to fix a specific bug, and I do not think performance matters here.

andreynering:
Hi @thejray,

Thanks a lot!

Issue #225, opened 2019-06-29
Extended globbing support


Yottum:
| Task version 	| OS                    	|
|:--------------	|:-----------------------	|
| master       	|  macOS Mojave 10.14.5 	|

Using extglob patterns in `sources` entries crashes Task. I tried using this to exclude some files from being picked up. Is this known? If yes, are there any plans to add this or the ability to exclude files from `sources` entries in the future?

---

Given the following `Taskfile.yml`:

```yml
version: '2'

tasks:
  foo:
    cmds:
      - printf %s Test
    sources:
      - ./!(vendor)/**/*.go
    method: checksum
```

Task crashes after running `task foo`:

    panic: unhandled word part: *syntax.ExtGlob

    goroutine 1 [running]:
    github.com/go-task/task/vendor/mvdan.cc/sh/expand.(*Config).wordFields(0xc00019e140, 0xc00015a540, 0x2, 0x2, 0x1, 0x1, 0x203000, 0x0, 0xc0001a0010)
        /Users/thierry/go/src/github.com/go-task/task/vendor/mvdan.cc/sh/expand/expand.go:546 +0x1292
    github.com/go-task/task/vendor/mvdan.cc/sh/expand.Fields(0xc00019e140, 0xc0001a0010, 0x1, 0x1, 0x0, 0x0, 0x1, 0x0, 0xc000194020)
        /Users/thierry/go/src/github.com/go-task/task/vendor/mvdan.cc/sh/expand/expand.go:323 +0x19a
    github.com/go-task/task/vendor/mvdan.cc/sh/shell.Fields(0xc0001940a0, 0x11, 0x0, 0x1, 0x13852c8, 0x2, 0xffffffffffffffff, 0xc0001940a0)
        /Users/thierry/go/src/github.com/go-task/task/vendor/mvdan.cc/sh/shell/expand.go:62 +0x162
    github.com/go-task/task/internal/execext.Expand(0xc0001940a0, 0x11, 0x2, 0xc0001940a0, 0x11, 0x10e1548)
        /Users/thierry/go/src/github.com/go-task/task/internal/execext/exec.go:78 +0x8e
    github.com/go-task/task/internal/status.glob(0x0, 0x0, 0xc000146010, 0x1, 0x1, 0x21, 0x0, 0x2341460, 0x0, 0x0)
        /Users/thierry/go/src/github.com/go-task/task/internal/status/glob.go:17 +0xca
[...]

PR #309, opened 2020-03-20
Add support for exclude rules for sources

lunny:
When I use task to package a directory which includes `node_modules`, I found it will spend about 7 seconds even if I have set `sources`. And I found that's because `task` needs exclude rules, so I write this PR and now I can set as below:

```yaml
sources:
      - "!web/.*"
      - "!web/node_modules/**"
      - "!web/tests/**"
      - "!web/dist/**"
      - "!web/mock/**"
      - "!web/LICENSE"
      - "!web/*.md"
      - "!web/**/*.md"
      - web/**
method: checksum
```

After the change, it will spend only `0.29s`. Yeah!


lunny:
@andreynering 
1. I will change the base branch to v3
2. I will not update vendor
3. I found https://github.com/mattn/go-zglob/blob/a8912a37f9e79ee444e77452cba926b3e5b09a65/zglob.go#L27 , that `zenv` is unexported, so that it could not be saved. If we have multiple rules, it's low performance. So I suggest to drop zglob and keep the new one. But for this PR, I won't to touch other place which depends on zglob. So that it's easier to review. On further PRs, zglob could be replaced totally.

lunny:
@andreynering done.

andreynering:
Closed by mistake, sorry

marco-m:
Hello, what is the status of this PR? From #225 I understand it is complicated, but I have hope :-)

butuzov:
@andreynering is there any updates on this issue? Having something to exclude is quite a nice feature to have.

tamasfe:
@andreynering is there anything I could do to help with driving this forward? It would be a crucial feature for me as currently code generation patterns (e.g. `generating **/*.gen.go` based on `**/*.go`) do not seem possible without workarounds.

kevpie:
> @andreynering is there anything I could do to help with driving this forward? It would be a crucial feature for me as currently code generation patterns (e.g. generating `**/*.gen.go` based on `**/*.go`) do not seem possible without workarounds.

How promising does helping getting https://github.com/mattn/go-zglob/pull/30 landed go towards solving this?

PR #310, opened 2020-03-21
Enhance zsh/bash shell tab-completion

ardnew:
The tab-completion scripts for bash/zsh are very slow. The bash tab-completion would also not actually auto-complete unambiguous task names and would just display them (I didn't verify zsh/powershell).

The problem is that `cmd/task` was parsing the YAML, compiling tasks, initializing the `Logger` instances, etc. just to print out the task names. 

Instead of adding a new command-line flag, I've just combined two incompatible flags `--list` and `--silent` that will perform only as much overhead as required to print just the available task names, e.g. `task -l -s`

Also updated the bash/zsh completion scripts to use these flags instead of running through a bunch of sed/awk pipes. Added auto-completion for tasks and command-line options too (bash only).

The completion scripts could still use work to perform context-aware suggestions.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -124,6 +124,12 @@ func main() {
 
 		OutputStyle: output,
 	}
+
+	if list && silent {
+		e.PrintTaskNames() // silently ignore any errors
+		return
+	}
+
 	if err := e.Setup(); err != nil {
 		log.Fatal(err)
 	}
--- a/help.go
+++ b/help.go
@@ -2,7 +2,10 @@ package task
 
 import (
 	"fmt"
+	"io"
+	"os"
 	"sort"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/go-task/task/v2/internal/taskfile"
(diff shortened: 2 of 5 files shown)


help.go
```diff
@@ -39,3 +42,33 @@ func (e *Executor) tasksWithDesc() (tasks []*taskfile.Task) {
 	sort.Slice(tasks, func(i, j int) bool { return tasks[i].Task < tasks[j].Task })
 	return
 }
+
+// PrintTaskNames prints only the task names in a taskfile.
```
andreynering:
```suggestion
// PrintTaskNames prints only the task names in a Taskfile.
```

help.go
```diff
@@ -39,3 +42,33 @@ func (e *Executor) tasksWithDesc() (tasks []*taskfile.Task) {
...
+func (e *Executor) PrintTaskNames() error {
+	// if called from cmd/task.go, e.Taskfile has not yet been parsed
+	if nil == e.Taskfile {
```
andreynering:
```suggestion
	if e.Taskfile == nil {
```

help.go
```diff
@@ -39,3 +42,33 @@ func (e *Executor) tasksWithDesc() (tasks []*taskfile.Task) {
...
+		err := e.readTaskfile()
+		if nil != err {
+			return err
+		}
+	}
+	var w io.Writer = os.Stdout
+	if nil != e.Stdout {
```
andreynering:
```suggestion
	if e.Stdout != nil {
```

cmd/task/task.go
```diff
@@ -124,6 +124,12 @@ func main() {
 
 		OutputStyle: output,
 	}
+
+	if list && silent {
+		e.PrintTaskNames() // silently ignore any errors
```
andreynering:
```suggestion
		_ = e.PrintTaskNames()
```

help.go
ardnew: I'm new to Go, but often see your change here as _the convention_ when comparing equality with literals and variables. 

I was taught (with idioms from certain other barbaric languages) to put literals/constants on the left-hand side, as to avoid accidental assignment (`=` vs `==`), which isn't an issue with Go (for a couple reasons). 

Anyway, is this just style/preference or is there a pragmatic difference here I should be aware of?

ardnew:
Oops I didn't mean to edit your comment, and now I can't figure out how I did (to restore it)! Was using the new Github mobile app, now I see my mistake from the desktop

andreynering:
> Oops I didn't mean to edit your comment, and now I can't figure out how I did (to restore it)! Was using the new Github mobile app, now I see my mistake from the desktop

This seems to be a bug on GitHub. I just reported it, together with the pending Travis status which shouldn't be shown.

ivotron:
should this be closed, now that https://github.com/go-task/task/pull/671 got merged?

andreynering:
@ivotron Yeah, like we could. Thanks for pointing out.

Issue #313, opened 2020-04-20
occur context canceled if use --watch


hatappi:
- Task version: 2.8.0
- OS: macos 10.15.3

When I launch task with `--watch` option, occur `context canceled` occurs.

I prepared a project to reproduce.
https://github.com/hatappi/go-http-server

first step, launch http server

```sh
$ git clone https://github.com/hatappi/go-http-server
$ cd go-http-server
$ make install-task

# run http server using task
$ make run-task
./bin/task -w run
task: Started watching for tasks: run
go build -v -o ./dist/app main.go
command-line-arguments
./dist/app
2020/04/21 00:19:57 start http server
```

second step, change `main.go`.
The following will be output error

```sh
task: Failed to run task "run": context canceled
go build -v -o ./dist/app main.go
command-line-arguments
./dist/app
2020/04/21 00:23:15 start http server
```

Is this the behavior  expecting?

andreynering:
Hi @hatappi,

Yeah, I think we can probably just suppress context errors from being printed in the terminal.

hatappi:
HI @andreynering 
Thank you for response.
I tried to resolve it and created PR.
https://github.com/go-task/task/pull/317

Could you review it if you have a time?

andreynering:
Fixed on #317.

hatappi:
Thank you for merge my PR and add new release tag.

PR #317, opened 2020-04-29
suppress context errors when use watch option

hatappi:
When I launch `task` with --watch option, occur context canceled occurs.
issue is here. https://github.com/go-task/task/issues/313

The cause was that it couldn't check custom error struct ([taskRunError](https://github.com/go-task/task/blob/master/errors.go#L21)).
this pr resolve it.


andreynering:
Hi @hatappi,

Thanks for your pull request! Seems perfect.

Issue #325, opened 2020-05-11
name_template in .goreleaser.yml is deprecated


danquah:
I just attempted to run the `test-release` target in the root of the project but got the following output

```shell
goreleaser --snapshot --rm-dist

   • releasing...
   • loading config file       file=.goreleaser.yml
   ⨯ release failed after 0.00s error=yaml: unmarshal errors:
  line 43: field name_template not found in type config.NFPM
task: Failed to run task "test-release": exit status 1
```

It seems `name_template` is now `file_name_template` as pr https://goreleaser.com/deprecations/#nfpmsname_template

Should I open a PR or do you have a reason for the current use of name_template

andreynering:
Hi @danquah,

Thanks for informing me!

There's no reason to keep the old name, I just haven't noticed this was deprecated.

A PR would be welcome 😉

PR #328, opened 2020-05-14
Fix name_template in Go Releaser

danquah:
As mentioned in #325  config name_template is now file_name_template.

See https://goreleaser.com/deprecations/#nfpmsname_template for details.


andreynering:
@danquah Thanks! ❤️

PR #330, opened 2020-05-14
Have the --help/-h flag print usage.

danquah:
pflag will output a "pflag: help requested" message after displaying the usage
if we don't explicitly  handle the flag.


andreynering:
@danquah Thanks! ❤️

PR #334, opened 2020-05-26
Eliminate use of single quotes for version string.

achton:
According to the [Styleguide](https://taskfile.dev/#/styleguide), the `version` string should not use single quotes for its value.

This PR attempts to consistently fix this in examples and tests across the codebase.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: 2
 
 # silent: true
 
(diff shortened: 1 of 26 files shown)


andreynering:
Hi @achton,

This is a good patch, but I'll conflict with #220 as we already bumped some version to `3` there and some more will be updated soon.

So, if you're still willing to work on this, I advice you to open a new PR targeting `v3` instead of `master`, and also update the versions on `docs/usage.md` to `3` in the same commit.

achton:
Gotcha - will do :+1:

Issue #321, opened 2020-05-07
Allow overriding the task name in the logger output


adamwasila:
Given example Taskfile:

```yaml
version: '2'

tasks:
  a:
    cmds:
      - task: b
      - task: b
      - task: b

  b:
    cmds:
      - echo "Hello b"
    sources:
      - "source.txt"
```
Then if `source.txt` is present I see following output:

```
task a
task: Task "b" is up to date
task: Task "b" is up to date
task: Task "b" is up to date
```

Now, the problem is that I cannot customize this output to parametrize task name for each execution separately. This makes output quite unclear of what has actually happened and it could be easily solved using following syntax:

```
tasks:
  a:
    cmds:
      - task: b
        alias: firstb
      - task: b
        alias: secondb
      - task: b
        alias: thirdb

```

```
task a
task: Task "firstb" is up to date
task: Task "secondb" is up to date
task: Task "thirdb" is up to date
```
Now, the real use case is I have kind of "template" task to build single go binary which I don't want to repeat for every command for obvious reasons (DRY):

```
  build_target:
    cmds:
      - "go build {{.BUILD_FLAGS}} -o {{.PREFIX}}{{.TARGET}} ./cmd/{{.TARGET}}"
    sources:
      - "**/*.go"
    generates:
      - "{{.PREFIX}}{{.TARGET}}"
```

Now, project has multiple commands to build, but all I see is: 

```
task build
task: Task "build_target" is up to date
task: Task "build_target" is up to date
task: Task "build_target" is up to date
```

instead of desired:

```
task build
task: Task "cmd/command1" is up to date
task: Task "cmd/othercommand" is up to date
task: Task "cmd/supercommand" is up to date
```

andreynering:
Hi @adamwasila,

Sounds like s reasonable feature request

adamwasila:
I made a quick hack to check how it could work. You may see it in this branch: https://github.com/adamwasila/task/tree/subtask-alias
What do you think about it? Should I add support for tasks called by deps and send you pull request or should I rewrite/refactor it first?

Example `Taskfile.yml` to test against my branch:

```
version: '2'

tasks:
  default:
    deps:
      - test
    cmds:
      - task: test
        alias: "foo"
      - task: test
        alias: "bar"
      - task: test
  test:
    desc: lorem ipsum
    cmds:
      - echo "hello world"
    sources:
      - Taskfile.yml
    generates:
      - Taskfile.yml
```

andreynering:
[...]

Issue #268, opened 2019-11-15
Task identifier shorthand notation / aliases


aegershman:
In order to allow shorthand notation for tasks, consider allowing a task to be referenced by multiple values. For example, and equivalent in `shell` would be...

```sh
#!/usr/bin/env bash

case "$1" in
dt | do | do-thing | do-the-thing )
  ...
;;
```

perhaps in a Taskfile.yml it _could_ be...

```yml
tasks:
  do-thing:
    cmds:
      - ...
    aliases:
      - dt
      - do
      - do-the-thing
```

Just an idea, just putting it out there. Feel free to close whenever, etc. Thanks for your time 👍

andreynering:
Hi @aegershman,

That's kinda already possible, just more verbose:

```yaml
version: 2

tasks:
  do-thing:
    cmds:
      - ...

 dt:
    cmds:
      - task: do-thing

 do:
    cmds:
      - task: do-thing

 do-the-thing:
    cmds:
      - task: do-thing
```

I liked your idea though, and it's probably easy to implement.

tommyalatalo:
> Hi @aegershman,
> 
> That's kinda already possible, just more verbose:
> 
> ```yaml
> version: 2
> 
> tasks:
>   do-thing:
>     cmds:
>       - ...
> 
>  dt:
>     cmds:
>       - task: do-thing
> 
>  do:
>     cmds:
>       - task: do-thing
> 
>  do-the-thing:
>     cmds:
>       - task: do-thing
> ```
> 
> I liked your idea though, and it's probably easy to implement.

While you're right about this being kind of possible with your examples, it creates a very cluttered output when doing "task --list". It would be really great to have an "alias:" property to set for tasks, and probably a good way of showing the aliases in the "--list" output would woulbe to just chain them, or have a column for aliases separately.

"chained"
`
* clean c cl: 	clean temporary files and folders
* build b: 	clean temporary files and folders
* test t: 	clean temporary files and folders
`

alias column:

`
name      alias     description
* clean    c cl 	   clean temporary files and folders
* build     b 	   build binary
* test       t 	   run tests
`

Something like the above.

andreynering:
@tommyalatalo That makes sense 👍

andreynering:
Closed by mistake, sorry.

slashsbin:
Here is another idea to achieve shortcuts automatically:
If task names are namespaced with `:`(the way style guide recommends), then we could support specifying task names using the shortest unambiguous text for each part.

[...]

PR #337, opened 2020-06-01
Allow overriding the task name in the logger output

adamwasila:
My attempt to solve issue described in #321 . Should work correctly though it needs review.

Simplest `Taskfile.yml` to demonstrate effects of this change:

```
version: '2'
tasks:
  default:
    deps:
      - task: test
        alias: "dep-foo"
    cmds:
      - task: test
        alias: "foo-bar"
      - task: test
  test:
    sources:
      - Taskfile.yml
    generates:
      - Taskfile.yml
```

Output before:

```
task: Task "test" is up to date
task: Task "test" is up to date
task: Task "test" is up to date
```

Expected output with change:

```
task: Task "dep-foo" is up to date
task: Task "foo-bar" is up to date
task: Task "test" is up to date
```

Also there were some concerns if name "alias" is a good one for such property. Of the top of my head I have two other options: "name-override" or "label". Are these any better?

### DIFF
--- a/help.go
+++ b/help.go
@@ -21,7 +21,7 @@ func (e *Executor) PrintTasksHelp() {
 	// Format in tab-separated columns with a tab stop of 8.
 	w := tabwriter.NewWriter(e.Stdout, 0, 8, 0, '\t', 0)
 	for _, task := range tasks {
-		fmt.Fprintf(w, "* %s: \t%s\n", task.Task, task.Desc)
+		fmt.Fprintf(w, "* %s: \t%s\n", task.Name(), task.Desc)
 	}
 	w.Flush()
 }
--- a/internal/summary/summary.go
+++ b/internal/summary/summary.go
@@ -56,7 +56,7 @@ func printTaskSummary(l *logger.Logger, t *taskfile.Task) {
 }
 
 func printTaskName(l *logger.Logger, t *taskfile.Task) {
-	l.Outf(logger.Default, "task: %s", t.Task)
+	l.Outf(logger.Default, "task: %s", t.Name())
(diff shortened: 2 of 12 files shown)


adamwasila:
I think I found sth that may be much more elegant solution to the problem I have. Here, I use alias (or label) on task itself and I can use variables to simply substitute it with any name I want when it is executed using deps or direct call. Example Taskfile with the syntax:

```
version: '3'
tasks:
  default:
    deps:
      - task: test
        vars: 
          ALIAS: IAMDEPTASK
    cmds:
      - task: test
        vars: 
          ALIAS: IAMSUBTASK
      - task: test
  test:
    alias: "{{.ALIAS}}"
    sources:
      - Taskfile.yml
    generates:
      - Taskfile.yml
```

As you may see, implementation is minimal and modifies just task struct but nothing else. Another advantage is it can be used to customize task name in every context, even if task is called directly.

Problems I found so far: in example above, last task currently generates message `task: Task "<no value>" is up to date` while it should use original name of the task in such case. But that seems trivial to fix.

I leaved all changes in this branch so far waiting for any comments, but if accepted only last commit will be one proposed to merge (updated with tests, documentation etc. if required).

andreynering:
Hi @adamwasila,

I liked this approach, it seems really simple.

I like `label` much more than `alias`, though, specially because there's another feature request that would be called "alias" (https://github.com/go-task/task/issues/268) and this name makes more sense there than here.

adamwasila:
Thanks for your comments. I have implemented it accordingly and it is ready for review now.

andreynering:
Hi @adamwasila,

Seems great, thanks! ❤️

PR #341, opened 2020-06-09
always treat command line arguments as global

### LINKED ISSUE #336: v3: expansion doesn't work as expected
I'm using latest version of v3 branch

Taskfile.yml
```
version: '3'

vars:
  root: 'x'
  exp: 'exp-{{.root}}'

tasks:
  test:
    desc: a
    cmds:
      - echo root is '{{.root}}'
      - echo but exp is '{{.exp}}'
```

Current output:
```
$ task test root=y
task: echo root is 'y'
root is y
task: echo but exp is 'exp-x'
but exp is exp-x

$ task root=y test
task: echo root is 'y'
root is y
task: echo but exp is 'exp-y'
but exp is exp-y
```

Expected: both variants should output the same — exp variable should be `exp-y` because `root` is `y`.

kovetskiy:
This pull request converts all command line arguments to global variables.
Fixes #336

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -16,9 +16,7 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var (
-	version = "master"
-)
+var version = "master"
 
 const usage = `Usage: task [-ilfwvsd] [--init] [--list] [--force] [--watch] [--verbose] [--silent] [--dir] [--taskfile] [--dry] [--summary] [task...]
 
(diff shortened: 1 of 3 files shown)


andreynering:
Hi @kovetskiy,

I didn't have a chance to try or think about it yet, but I'd like to make it clear that this is a backward incompatible change, and so if we proceed with this we'd need to support both behaviors (the old one for v2 and the new one for v3, based on what version the user set on his Taskfile.yml).

kovetskiy:
@andreynering true, I'll make it backward compatible

andreynering:
Hi @kovetskiy,

I merged this manually with changes at 6ff3c9015b7094cd82c873e862a0be16e7001911

Thanks for contributing!

PR #345, opened 2020-06-12
Adding a jobs flag to limit the number of tasks running in goroutines

RossHammer:
We are using a single taskfile to build all of our application which consists of many different docker images. This can overwhelm our computer when we start building it and it actually runs faster if limit how many things run at once. There are not many dependencies between the images so a lot is run in parallel and we want to limit it based off the developer's system rather than coding it in the taskfile directly.

I used the term jobs for everything using make as an example people may be familiar with.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -65,6 +65,7 @@ func main() {
 		dry         bool
 		summary     bool
 		parallel    bool
+		concurrency int
 		dir         string
 		entrypoint  string
 		output      string
@@ -87,6 +88,7 @@ func main() {
 	pflag.StringVarP(&entrypoint, "taskfile", "t", "", `choose which Taskfile to run. Defaults to "Taskfile.yml"`)
 	pflag.StringVarP(&output, "output", "o", "", "sets output style: [interleaved|group|prefixed]")
 	pflag.BoolVarP(&color, "color", "c", true, "colored output")
+	pflag.IntVarP(&concurrency, "concurrency", "C", 0, "limit number tasks to run concurrently")
 	pflag.Parse()
 
 	if versionFlag {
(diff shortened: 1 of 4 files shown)


cmd/task/task.go
```diff
@@ -81,6 +82,7 @@ func main() {
 	pflag.StringVarP(&dir, "dir", "d", "", "sets directory of execution")
 	pflag.StringVarP(&entrypoint, "taskfile", "t", "", `choose which Taskfile to run. Defaults to "Taskfile.yml"`)
 	pflag.StringVarP(&output, "output", "o", "", "sets output style: [interleaved|group|prefixed]")
+	pflag.IntVarP(&jobs, "jobs", "j", 0, "limit number of jobs (tasks) to run simultaneously")
```
andreynering, line 85: What about calling this "threads"? 🤔 

Probably not the best suggestion, but I think that "jobs" is too generic and it may not be clear to the user we're talking about concurrency here.

Let me know if you have more suggestions.

andreynering: `--concurrency` feels good despite being a bit long (perhaps we could use `-C` as an alias).

RossHammer: Yea I'm alright with that! I was mostly following the convention of `make` I wasn't sure what the best name was.

RossHammer: Everything has been updated when you have time to look again.

andreynering: Hey @RossHammer,

Just commenting since it has been some time...

I have a few concerns and I think I'll postpone to be re-reviewed after the v3 release (which is more than late already).

Sorry, been a but busy recently...

RossHammer:
I understand that now is not the best time to review/merge this but I'm wondering why it was just closed?

andreynering:
Hi @RossHammer,

I'm sorry, this was a mistake. GitHub closed many PRs automatically when I merged #220 sooner today.

I'm now reopening this...

RossHammer:
I have updated this the test to use v3 of the taskfile instead

RossHammer:
@andreynering Do you have any updates when this will get reviewed and potentially released? I understand you are busy and if there is anything I can do to help I'm willing to assist.

This starting to move to becoming a blocker for us as we have continued to add and are seeing network calls fail because too much is running in parallel. We would really rather not have to build and distribute a fork to our team.

RossHammer:
I have re-based this to the latest master as we are still desperate for this change

RossHammer:
No worries, that is kinda the impression I have gotten as I have seen a lot more activity here over time as well!

andreynering:
This was just manually merged at c6ecf7037794f509b1187bb364248c42db30fb2b.

@RossHammer Thanks a lot for your work and patience! ❤️ 

This is for sure a very useful feature.

Issue #100, opened 2018-03-10
A warning/confirm field that shows a confirm prompt before proceeding 


ianwalter:
I think it would be cool if there was a way to specify a warning before executing commands in which the user may not fully understand the consequences. It could look like:

```yml
reset:
  desc: Reset environment from scratch
  warning: This will remove all data and environment customization. Are you sure you want to do this?
```

And then before executing `task reset` it would prompt the user with a confirm prompt that looks like:

```
This will remove all data and environment customization. Are you sure you want to do this? [y/N]
```

andreynering:
Hi @ianwalter,

This is probably already possible somehow using bash inside a command.

I'll keep this open for discussion if this is worth implementing. It would not be hard.

ianwalter:
@andreynering That's a good point. I could use `read` and then conditionally execute a shell command based on the user's response. The only problem is that on confirm I want to execute a series of tasks. I could just create a new task and then execute something like `task resetConfirmed` but that would expose the command in help which is not really a good UX. Feel free to close this though if you want to prioritize being lean over UX.

andreynering:
@ianwalter Take a look at the [silent](https://github.com/go-task/task#silent-mode) option, you can prevent a command to be printed.

I still consider to accept a patch for this, though.

parkerduckworth:
Hey @andreynering is something still being considered?

Having this command in a `cmds` block fails to run: 
```yml
cmds:
  - read -p "Display me as a prompt. Continue? (y/n)"
```

And produces this oddity:

```bash
read: invalid option "-p"
```

If you still think adding a `warning` or `prompt` block would be a good addition to this project I would gladly take on the `task` (😉) to write the patch.

andreynering:
@parkerduckworth That should be possible once https://github.com/mvdan/sh/issues/551 is implemented upstream.

AnonLibre:
Here an example prompt implemented with the dialog command in case someone needs a confirmation prompt until this feature is discussed.

```yaml
version: '3'

tasks:

  confirm:
    desc: "Executes a dangerous command.."
    cmds:
      - |
        dheight=10 \
        ; dwidth=60 \
        ; dialog \
          --defaultno \
          --backtitle "Dangerous Command Title" \
          --yesno \
            "Do you really want to execute the dangerous command?" \
          ${dheight} \
          ${dwidth}
      - echo "Executing dangerous command."
```

core-dev-crypto:
Hello @andreynering ,

[...]

PR #352, opened 2020-07-20
Add command-level and task-level warning

parkerduckworth:
Hey @andreynering, just saw your comment on issue #100 regarding the pending upstream ability to pass a `-p` flag to `read`. Coincidentally, I just finished writing this patch before your reply.

I understand that [mdvan/sh#551](https://github.com/mvdan/sh/issues/551) will clear up the ability to add a prompt, but I think that having an explicit `warning` option at both task and command levels may improve this tool.  What do you think?

Here is how `warning` can be used at both levels:

```yml
version: '3'

tasks:
  example-task:
    desc: An example
    warning: Are you sure you want to run this task?
    cmds:
      - cmd: rm -rf /important-files
        warning: Are you sure you want to run this command?
      - rm -rf /unimportant-files
```
Declining to run a command with a command-level warning results in the command being skipped, and the rest of the commands in the task (without warnings) will run unhindered.

I know you will have this project's best interest in mind.  Thanks for considering this feature!

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -68,6 +68,7 @@ func main() {
 		entrypoint  string
 		output      string
 		color       bool
+		yes         bool
 	)
 
 	pflag.BoolVar(&versionFlag, "version", false, "show Task version")
@@ -86,6 +87,7 @@ func main() {
 	pflag.StringVarP(&entrypoint, "taskfile", "t", "", `choose which Taskfile to run. Defaults to "Taskfile.yml"`)
 	pflag.StringVarP(&output, "output", "o", "", "sets output style: [interleaved|group|prefixed]")
 	pflag.BoolVarP(&color, "color", "c", true, "colored output")
+	pflag.BoolVarP(&yes, "yes", "y", false, "automatic yes to warning prompts")
 	pflag.Parse()
 
 	if versionFlag {
@@ -131,6 +133,7 @@ func main() {
 		Summary:    summary,
 		Parallel:   parallel,
 		Color:      color,
+		Yes:        yes,
 
 		Stdin:  os.Stdin,
 		Stdout: os.Stdout,
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -920,3 +920,47 @@ so task know which files to watch.
 
 [gotemplate]: https://golang.org/pkg/text/template/
 [minify]: https://github.com/tdewolff/minify/tree/master/cmd/minify
+
+## Warning prompts
+
+You can add both task level and command level warning prompts to your Taskfile.
+
+Here is how `warning` can be used at both levels:
+
+```yaml
+version: '3'
+
+tasks:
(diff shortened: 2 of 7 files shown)


andreynering:
Hi @parkerduckworth, thanks for the PR! 🙂 

The code seems simple enough to be worth adding this. I'll probably check this in another day, though...

In the meantime, can you change the base branch from `master` to `v3` and rebase?

parkerduckworth:
Absolutely!

parkerduckworth:
Ok @andreynering , everything looks to be in order!

parkerduckworth:
@andreynering no worries, life is to be lived!

---

> I think that prompt may be a better name than warning. What do you think?

I think that `prompt` offers more flexibility! At the moment, these changes append a yes/no option at the end of the supplied `warning`, i.e. `fmt.Printf("%s (y/N): ", warning)`. If we go with `prompt`, this should probably change to print only the provided prompt (`fmt.Printf("%s: ", prompt`), as a prompt may often require more than a yes or no answer.

So maybe an example would be like this:

```yaml
tasks:
  example-task:
    desc: An example
    prompt: Are you sure you want to run this task? (y/N)
```

I think the answer depends on whether or not we want more flexibility here.  You have obviously spent much more time than I working with this codebase, so I would happily agree with your preference.

---

> I think it's a bit inconsistent that command prompts skips the command while task prompts makes it error. I think both should have the same behavior, and I believe the right behavior would be to skip the task. Sounds good?

Excellent! So shall it be.

---

> Many CLI tools have a -y argument to bypass prompts and execute then without asking. This is important to allow executing these commands on CI, for example.

I agree!  We should provide this affordance as well.  

---

> [Optional] A small section on the documentation about this would be nice. But if you don't want to do it, I'll after the merge

Ha! I would be more than happy to write up the docs for this!!

---

If you are good with all my above remarks, I will make these changes and push them up.

andreynering:
Actually, let's keep `warning` for a matter of clarity since it'll only have two options (yes and no, not sure how we'd support more)

On the other points, I think we agree about them, so feel free to proceed =)

parkerduckworth:
Requested changes have been made 👍

parkerduckworth:
Hey @andreynering, you mind sharing why this was closed?

andreynering:
@parkerduckworth Sorry, this was a mistake

For some reason GitHub closed many PRs after merging https://github.com/go-task/task/pull/220

Reopening...

parkerduckworth:
Hello @andreynering ! Is this feature still going to be merged?

andreynering:
@parkerduckworth Sorry for not responding to you all this time. Life happens.

Good to hear that the `read -p` works for you. Thanks for your contribution on https://github.com/mvdan/sh/pull/722 as well.

parkerduckworth:
@andreynering apology not necessary 🙂 Hope all is well mate!

Issue #121, opened 2018-06-28
Stabilise the package API


jaxxstorm:
I have a use case I'd like to have considered.

I'm building an internal application which will:

- Traverse through a directory hierarchy and fine all `Taskfile.yml` instances
- Run the equivalent of `task -d <dir>/Taskfile.yml` on all those taskfiles concurrently

As pointed out in slack, I can use `os/exec` for this, but ideally I want to be able to just import `github.com/go-task/task` and then use those functions/methods from there. This will ensure that user's don't need to download a binary to run the internal tool

Would this be something that might be considered?

smyrman:
While task is a great CLI tool, my personal opinion is that the underlying code may not be among the cleanest,  and not all packages has the right separation of public vs private members. I should add as a disclaimer that there is some time since I last looked at the code base.

Still, I think the best approach for this issue may be to first suggest/dream up a good spec for how a _minimal public_ interface should look like, and then those public functions can either be reimplemented or call code living in packages  under the `internal` namespace to get the job done, which ever is easiest in each case.

Basically hide away most of the code under `internal/` (or make things private within each particular package) to allow for refactoring intenally, while keeping a clean, minimal and backwards-compatible public interface.

smyrman:
PS! I think that the code base _has_ changed significantly since my last contribution, so a lot of code does already live in the `internal/` packages.

andreynering:
@smyrman Yes, I have been trying to refactor things step-by-step, specially splitting code to packages and declaring interfaces to abstract code

I agree that once we declare the use as a lib as "officially supported" we should worry about backwards compatibility

@jaxxstorm By now, if you don't want to wait, feel free to open a pull request moving the `args` and `taskfile` packages from `internal/` to the root directory. I think that will be enough to allow you doing what you need. That said, I suggest you to keep this lib in the `vendor/` directory for now

gedw99:
Agree that stabilising the API is a good idea.

I think worrying about not breaking people in the future is fools gold. Hey ? It's a tool not an APP delivered over Google play with a server using grpc that needs to support Schema Evolution. Being a tool means yes breaking the API.
Instead support a version tag at the top and then have public API's that are for each version.
That's probably easily good enough.
But also that's slot of work to support. So when you get down to the fact that it's a tool and a go get -u will get you to the latest then maybe that's ok.

I just worry that stable API / schema is too much work 

[...]

PR #354, opened 2020-07-29
Don't overwrite Executor's Output if it is already configured

aidansteele:
When using `task` as a Go package (which works extremely well, it's a great public interface!) the `Executor`'s `Output` property gets overwritten in `(*Executor).Setup()`. This PR changes the behaviour to not override the `Output` if it has already been configured (e.g. by a different app).

I'm happy to make any changes to this if you want. Thanks!

**EDIT**: So I just realised that `taskfile` is actually in the `internal` directory, so it can't be used outside of this project. I didn't realise until now because I was experimenting on a fork. Would you be open to exposing this?

### DIFF
--- a/task.go
+++ b/task.go
@@ -172,15 +172,18 @@ func (e *Executor) Setup() error {
 	if e.OutputStyle != "" {
 		e.Taskfile.Output = e.OutputStyle
 	}
-	switch e.Taskfile.Output {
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @aidansteele!

There was some discussion about allowing the use as a library in #121, but this will probably not be made before v3 is released.

PR #356, opened 2020-08-03
Add support for ".env" like files

chris-garrett:
Here is the first take at dotenv. I wasn't sure about updating the CHANGELOG.


chris-garrett:
I forgot to add that I made a build using github.com/joho/godotenv about three months ago. It's been used on a dozen or so projects with three devs running and on linux, mac and windows. No issues reported yet.

andreynering:
Hi @chris-garrett,

I just merged this with a few improvements: eab14b6c49957d8daec7a069f6a90581da01cd9b

Thanks a lot! ❤️

PR #358, opened 2020-08-12
Allow usage as an external library

damian-szulc:
The scope of this PR is to just move `args` and `taskfile` packages to the root directory. With those changes, as mentioned by @andreynering  in [this issue](https://github.com/go-task/task/issues/121), this is a minimal effort to enable usage as an external library. This still won't make usage a lib officially supported. It will only make it possible to use it on one's own risk (that some people - like me - are willing to take).


andreynering:
Hi @damian-szulc, thanks for opening this Pull Request!

I'm closing this because we aren't touching v2 (currently in `master`) often anymore.

You should open another PR targeting `v3` (which will hopefully reach `master` and be released soon 🙏)

andreynering:
UPDATE: `v3` was merged into `master` sooner today, so this just needs to be updated...

damian-szulc:
Done. Updated 👌

andreynering:
Thanks @damian-szulc!

Issue #338, opened 2020-06-05
task 2.8.1 panic with empty task


marco-m:
Both these Taskfiles cause task to panic:

```
version: '2'

tasks:
  foo:
```

and (so `default` is not a special case)

```
version: '2'

tasks:
  default:
```

Panic is:

```
$ task
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x12cf3e1]

goroutine 1 [running]:
github.com/go-task/task/v2/internal/taskfile/read.Taskfile(0x0, 0x0, 0x139e0cb, 0xc, 0xc000060460, 0x5, 0x5)
        /home/runner/work/task/task/internal/taskfile/read/taskfile.go:64 +0x5c1
github.com/go-task/task/v2.(*Executor).Setup(0xc00007c2c0, 0xc00007c2c0, 0x0)
        /home/runner/work/task/task/task.go:109 +0x89
main.main()
        /home/runner/work/task/task/cmd/task/task.go:135 +0x989
```

Happening with task 2.8.1 on macOS.

andreynering:
I'm not sure why someone would write an empty task.

Seems worth fixing it though, since it'd be pretty easy.

marco-m:
> I'm not sure why someone would write an empty task.

By mistake :-)

I had a partially commented-out task, which resulted in the equivalent of the reproduction case I added in the issue.

To me, any crash, especially with a so simple input, should be fixed :-)

andreynering:
Fixed on d78f78bb5cdd94fec59244121bdc1130547beb2b

marco-m:
thanks!

PR #362, opened 2020-08-21
#338: Handle an empty task in a file

bharathshetty4:


andreynering:
Hi @bharathshetty4, thanks for opening this pull request!

I think I have a slightly better idea: I think we can just assign an empty task if the pointer is `nil`:

```go
if t == nil {
        t = &taskfile.Task{}
}
```

This should be enough to prevent panics without having to keep checking for `nil` on many places...

bharathshetty4:
> Hi @bharathshetty4, thanks for opening this pull request!
> 
> I think I have a slightly better idea: I think we can just assign an empty task if the pointer is `nil`:
> 
> ```go
> if t == nil {
>         t = &taskfile.Task{}
> }
> ```
> 
> This should be enough to prevent panics without having to keep checking for `nil` on many places...

Hi @andreynering checked by ``` t = &taskfile.Task{}``` this doesn't solve the nil problem for an empty task, as the task uses task.Var while running that task. We had use these nil checks to get rid.

bharathshetty4:
> Hi @bharathshetty4, thanks for opening this pull request!
> 
> I think I have a slightly better idea: I think we can just assign an empty task if the pointer is `nil`:
> 
> ```go
> if t == nil {
>         t = &taskfile.Task{}
> }
> ```
> 
> This should be enough to prevent panics without having to keep checking for `nil` on many places...

this is done.

andreynering:
Hi @bharathshetty4,

I manually merged that here: d78f78bb5cdd94fec59244121bdc1130547beb2b

Thanks for contributing!

PR #370, opened 2020-08-30
Run dependencies only once and in the correct order

minitauros:
This PR adds that dependencies, which before this change could run multiple times, concurrently, are now still run concurrently, but only once, and in the correct order.

Example:
a depends on b and c
b depends on c
c depends on d
d depends on nothing

When getting the dependencies of a, the following is found:
[0]{"b","c"} - because a depends on b and c
[1]{"c","d"} - because b (from level 0) depends on c, and c (from level 0) depends on d
[2]{"d"} - because c (from level 1) depends on d

Knowing now the dependency tiers, we can start running concurrently the dependencies at the deepest level (2) and move up one level at a time until we finish at the highest level (0).

If we see that a dependency has already run, we don't need to run it again. In the example above, d and c are only run once, even though they appear in the dependency analysis twice. Dependencies still run concurrently per level.

A mutex is used to ensure that only one instance of a dependency can run at a time, ensuring that even when running tasks in parallel, the same dependency (same name) will be able to block itself from being run twice if it has the correct `status`, `generates`, etc.

For example these tasks

```yml
tasks:
  a:
    cmds:
      - echo 'a' >> a.txt
    deps:
      - b
      - c
    status:
      - test -f a.txt

  b:
    cmds:
      - echo 'b' >> b.txt
    deps:
      - c
      - d
    status:
      - test -f b.txt

  c:
    cmds:
      - echo 'c' >> c.txt
    deps:
      - d
      - e
    status:
      - test -f c.txt

  d:
    cmds:
      - echo 'd' >> d.txt
    status:
      - test -f d.txt

  e:
    cmds:
      - echo 'e' >> e.txt
    deps:
      - d
    status:
      - test -f e.txt
```

ran with

```
task -p e c e b d a b c d a e b
```

will produce something like

```
task: echo 'd' >> d.txt
task: Task "d" is up to date
task: echo 'e' >> e.txt
task: Task "d" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: echo 'c' >> c.txt
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "d" is up to date
task: Task "c" is up to date
task: echo 'b' >> b.txt
task: Task "e" is up to date
task: Task "c" is up to date
task: Task "b" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "c" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "d" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "d" is up to date
task: Task "e" is up to date
task: Task "c" is up to date
task: Task "b" is up to date
task: Task "e" is up to date
task: Task "c" is up to date
task: Task "b" is up to date
task: Task "c" is up to date
task: echo 'a' >> a.txt
task: Task "b" is up to date
task: Task "a" is up to date
```

and result in a.txt, b.txt, c.txt, d.txt, e.txt being created, all with only one time their task name as file content.

After the initial PR I found a bug in this code where when having 
a depends on b
b on c
c on nothing, but c calls a directly

What happens there is that we end up in a deadlock where the running of task a is waiting on the finishing of running its previous dependencies. Because only one dependency with the same name can run at the same time, as soon as - in the example above - dependency c calls a directly, a must via b run c again as a dependency, but that run of dependency c is being blocked by the previous run of c, which cannot finish, leaving us in a deadlock.

This is fixed by inspecting the full call stack before running a task. Every time `runTask` is called, the call is appended to the call stack. If we discover that the call stack already contains the call that is being added, we return an infinite loop error. This way we prevent an infinite loop from happening, for example when a runs b, b runs a. The stack would be "a,b,a", which is not allowed (two times "a").

## Comment on original PR description

I can see now that this PR relates to https://github.com/go-task/task/issues/53.

I was running sort of into the same issue. I was building a Taskfile with some dependency chains that would run some dependencies more than once and started reading [the documentation](https://taskfile.dev/#/usage?id=task-dependencies), that says 

> Dependencies run in parallel, so dependencies of a task shouldn’t depend one another. If you want to force tasks to run serially take a look at the Calling Another Task section below.

I came up with the idea for this PR and spent a bit of time on it. Then I saw that https://github.com/go-task/task/pull/359 does something very similar (although not the same).

https://github.com/go-task/task/issues/53 seems to request that dependencies are run only once when running multiple tasks in parallel. This PR does not fix that directly, although it can provide a way of fixing that.

When running multiple tasks in parallel, dependencies will still only be run once if there is a correct `status` (or something else that prevents a task from running) property set on each dependency. Because mutex is used for each dependency that runs, there can only be one run of a dependency (with the same name) at a given time. So if both a and b depend on c, and a and b are run in parallel, then if a starts running c first, and c has for example a `status` property preventing it from being run again, b will not run c.

I can see that @andreynering commented

> Hmm... I'm not sure I agree with this. Some users may want to re-run a dependency each time (for example, if another task changed any file and something has to be re-generated).

My PR also prevents the same dependency from being run twice (if the name and vars are the same), so the same comment goes for this PR. This functionality could be removed though; if users use the correct `status` and `generate` properties of a task, that could prevent the unnecessary work, too. The code that prevents dependencies from running twice can easily be removed (it's just a few lines), and we'd still get the other benefits of this PR (dependencies run in correct order, and cannot run at the same time, so duplicate runs can be prevented with `status` etc.). On the other hand, one might argue that if users really want to re-run dependencies, they should be part of the `cmds` instead of being a dependency. Also, if two tasks have the same dependency (to be run with the same vars), it doesn't sound very useful to me to run multiple of them at the same time, and also there I think putting it in the `cmds` might actually make more sense (what do you think?).


task.go
```diff
@@ -54,8 +60,10 @@ type Executor struct {
 
 	taskvars *taskfile.Vars
 
+	callStackMux  *sync.Mutex
```
minitauros, line 63: I see that I forgot to remove this one. It's not being used. Will do it later.

andreynering:
Well, now we have two PRs that attempt to fix the same thing. /cc @RossHammer 

I've been a bit busy recently, so I may take time to review these PRs and decide how to proceed. I hope you understand.

minitauros:
Definitely, no worries! Although perhaps it's good to note that the original purpose of this PR was to not run dependencies multiple times while running **not** in parallel, while the other PR (referenced in my PR description) seems an attempt to prevent it from happening especially while running in parallel.

There is definitely some overlap. In @RossHammer 's PR users will get the ability to explicitly say if a task may be run more than once, while in my PR dependencies (not tasks) will more or less always run only once. For as far as I can see that's the big difference.

andreynering:
Hi @minitauros,

I decided to proceed with #359, so I'll close this one as duplicate. Let me know if you have questions.

Hopefully we'll have merged soon.

PR #371, opened 2020-09-03
Make fish completions compatible with MacOS

onedr0p:
MacOS by default does not use gnu sed, which makes it not work. This patch will work with MacOS and also Linux


andreynering:
Thanks @onedr0p!

PR #374, opened 2020-09-07
[fish-completions] use expressions from zsh auto-completions

onedr0p:
Currently, while my last PR worked on mac. There still was an issue of tasks that were named with colons e.g. `ansible:playbook:common` breaking the auto-completions

Instead of fussing around creating a new expression, I updated it use what the ZSH script is using. This works on Mac as well.


onedr0p:
I would like to add descriptions back to the completions, closing for now

PR #385, opened 2020-10-03
Resolves #378 - allow for missing env files as they may be bootstrapped.

### LINKED ISSUE #378: Allow dotenv files that do not exist
A typical workflow involving `.env` files is:

1. `.env` is in `.gitignore`
2. A project repository includes a `.env.example` file
3. A "first run" script creates a `.env` from `.env.example` in the developers environment

I would like to use task instead of make, and as part of that I need to use task for my project's "first run" script -- which creates the `.env` file. Unfortunately, task throws an error because `.env` doesn't yet exist. 

```yaml
version: '3'

dotenv: ['.env']

tasks:
  default:
    cmds:
      - cp -n .env.example .env # contrived example
```

```console
open .env: no such file or directory
```

There's an issue in the godotenv project (https://github.com/joho/godotenv/issues/99) associated with this use-case as the error is being thrown by that library, however I've noticed in the tests for task it is explicitly intended that an error is thrown when the `.env` file does not exist. 

Once the issue is fixed in godotenv it would be good to have the intended behaviour of task reviewed -- https://github.com/go-task/task/blob/master/task_test.go#L832

chris-garrett:


andreynering:
Hi @chris-garrett,

Thanks for attacking this! ❤️

PR #387, opened 2020-10-07
feat: improve installation script documentation for shell installs

sheldonhull:
- The default didn't work well for me out of the gate.
- This is the modified version to support passing in the arguments easily as well as an example for installing to `/usr/local/bin` for using in Codespaces or equivalent development workflow.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -74,10 +74,10 @@ easily generating this script.
 
 ```bash
 # For Default Installion to ./bin with debug logging
-url -sL https://taskfile.dev/install.sh | sh -s -- -d 
+curl -sL https://taskfile.dev/install.sh | sh -s -- -d 
 
 # For Installation To /usr/local/bin with debug logging
-url -sL https://taskfile.dev/install.sh | sudo sh -s --  -b /usr/local/bin -d 
+curl -sL https://taskfile.dev/install.sh | sudo sh -s -- -d -b /usr/local/bin
 ```
 
 > This method will download the binary on the local `./bin` directory by default.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -74,10 +74,12 @@ easily generating this script.
 
 ```bash
 # For Default Installion to ./bin with debug logging
-curl -sL https://taskfile.dev/install.sh | sh -s -- -d 
+sh -c "$(curl -ssL https://taskfile.dev/install.sh)" -- -d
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -73,7 +73,11 @@ scenarios like CIs. Many thanks to [GoDownloader][godownloader] for allowing
 easily generating this script.
 
 ```bash
-curl -sL https://taskfile.dev/install.sh | sh
+# For Default Installion to ./bin with debug logging
+url -sL https://taskfile.dev/install.sh | sh -s -- -d
```
andreynering, line 77: What's this `url` command? It isn't available on macOS on ZSH. I'd personally keep `curl` it's its way more commonly available in the wild.

andreynering: This `-d` flag errors for me, on `curl` at least

andreynering (commented):
Hi @sheldonhull, thanks for opening this PR!

The suggestion to install globally at `/usr/local/bin` seems interesting, but I have a few questions.

I personally wouldn't change the default command, and would keep `curl` instead of anything else.

sheldonhull:
The url was a typo/paste fail. that's supposed to be curl.

I'll test again as while I'm also on macOS I think I ran that command inside docker probably debian or ubuntu. I'll circle back around and update this with my finding. Thanks for the feedback 🤟

docs/installation.md
sheldonhull: So I just checked in container and `curl -sL https://taskfile.dev/install.sh | sh -s -- -d` worked without any error. I thought the `--` designates the end of shell argument parsing and `-d` should be passed to the executable task then, right?

If so `-d` should work on macOS and Linux as it is the argument for debug output for your task executable right?

Regarding the install path. I'm still a bit fuzzy on where `/opt/usr` vs `/usr/local/bin` and other paths are the best fit as my background is Windows primarily. To my understanding global user tools would fit either in the `/opt` or `/usr/local/bin` , but maybe there is a better practice you can let me know of.

sheldonhull: @andreynering  try this out now. I just tested on macOS and Linux and it works fine now for me.

sheldonhull:
@andreynering just wanted to make sure you saw the update. I think it's working now.

dery345 (approved):

andreynering:
@sheldonhull Thanks!

PR #398, opened 2020-11-05
Include task name in log output

Porges:
When running tasks which have dependencies which run concurrently, it is hard to understand what is going on. Including the specific task name in the output would help with this.

```
task: go test ./... -tags=noexit
task: ./scripts/verify_boilerplate.sh
task: go fmt .
task: golangci-lint run -v
```

After:
```
task [generator:test]: go test ./... -tags=noexit
task [header-check]: ./scripts/verify_boilerplate.sh
task [generator:format-code]: go fmt .
task [generator:lint]: golangci-lint run -v
```

I structured the name to match how `output: prefixed` works.

### DIFF
--- a/task.go
+++ b/task.go
@@ -346,7 +346,7 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
 		return nil
 	case cmd.Cmd != "":
 		if e.Verbose || (!cmd.Silent && !t.Silent && !e.Taskfile.Silent && !e.Silent) {
-			e.Logger.Errf(logger.Green, "task: %s", cmd.Cmd)
+			e.Logger.Errf(logger.Green, "task [%s]: %s", t.Name(), cmd.Cmd)
 		}
 
 		if e.Dry {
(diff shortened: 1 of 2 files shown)


mradhi:
I appreciate this PR, but I think there is no need for the "task" prefix anymore the task name is really meaningful in that case, but we may think more about the layout of the displayed output.

What about:

```
* Executing task <task_1_name> [col] [timestamp]
* Executing task <task_2_name> [col] [timestamp]
etc...
```

For the commands executed for each task, I think we may need to add some flag (maybe `-v` flag to control verbosity)

Porges:
Would you merge this as a small improvement first, and then rework the whole layout later? I feel like it could take some discussion to settle onto a new output format, so it would be nice to have a small improvement in the mean time 🙂

marco-m:
I think this is a good idea indeed.
For the output format, I am fine with the format proposed in the PR (my 2 cents).

andreynering:
Hi @Porges,

Sorry for the long wait to review your PR. This project has been having more issues and PRs than I can keep up.

Thanks for your patch! I just manually merged it with a tiny change: I moved the `[...]` to after the colon `:` so we keep the consistency with the other logs which all start with `task:`.

The commit: fe917affd2b248d70c0084e8282a3190e16e4483.

PR #399, opened 2020-11-06
Include pre, post, and final run functions

zlesnr:
This allows the user to customize the run, wrapping the output of each task run
with an execution of a pre and post function.  A deferred final function is also
included to ensure any cleanup, if necessary.

This is being used as a library, where task is injected from another process.

### DIFF
--- a/task.go
+++ b/task.go
@@ -52,6 +52,10 @@ type Executor struct {
 	Output      output.Output
 	OutputStyle string
 
+	PreRun   func(*taskfile.Task, map[string]interface{})
+	PostRun  func(*taskfile.Task, map[string]interface{})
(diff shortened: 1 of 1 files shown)


sanderblue:
+1 por favor

andreynering:
Hi @zlesnr,

I'm confused by this. Can you explain better your use case?

Task as a library ™️ is something still kinda experimental, and if we allow more customization in the future, I think it'd look different than this, and it'd likely need to be more flexible.

zlesnr:
We didn't end up implementing this, so I think we can close this.  We're keen to see what changes will be implemented to get this used as a library.  We're currently using it as such, and there are some limitations currently, but those may be accounted for by those changes you mention.

Issue #383, opened 2020-10-02
Add a `-L` flag (alternative to `-l`) that prints all tasks


zx8:
Or a new flag should be introduced (e.g. `task --names`) to just output the names of available tasks, with no descriptions.

Currently, a Taskfile with no descriptions returns:

```
$ task -l
task: No tasks with description available
```

While the solution is fairly intuitive (i.e. add a description to every task), the behaviour is surprising. Not to mention it would make the completions much less hacky (for tasks that _do_ have a description):

- https://github.com/go-task/task/blob/ce4ac97269613dc895408722fea2c63cba838f48/completion/fish/task.fish#L2

andreynering:
Hi @zx8,

Perhaps we could add `-L` flag (a an alternative to `-l`) that print all tasks?

Changing the behavior of the existing flag is a backward incompatible change, so not a good idea.

zx8:
Yeah sure, that works!

themanifold:
Slightly different command that does not needs `string`:

```
for i in $(task -l | sed 1d | awk '{print $2}' | awk 'FS=":"{print $1}'); do echo $i; done
```

PR #401, opened 2020-11-13
feat (help): allow cli option to list tasks with no description

therealkevinard:
added an add'l cli option that lists all tasks, with or without description.
orig. behavior: task -l lists tasks with desc field
new behaviour: task -la or task -a will list all tasks. if task has desc, it will be included.

BREAKING CHANGES: none, that I know of.
NOTES/Concerns:
- This is wip.
- Haven't checked how it interacts with bash completion.
- The new Executor.TaskNames func does not use e.CompiledTask(taskfile.Call{Task: task.Task})

This relates to https://github.com/go-task/task/issues/383

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -150,8 +150,12 @@ func main() {
 		return
 	}
 
-	if list || listAll {
-		e.PrintTasksHelp(listAll)
+	if list {
+		e.ListTasksWithDesc()
+	}
+
+	if listAll {
+		e.ListAllTasks()
 		return
 	}
 
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -57,6 +57,7 @@ func main() {
 		helpFlag    bool
 		init        bool
 		list        bool
+		listAll     bool
 		status      bool
 		force       bool
 		watch       bool
@@ -75,6 +76,7 @@ func main() {
 	pflag.BoolVarP(&helpFlag, "help", "h", false, "shows Task usage")
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yml in the current folder")
 	pflag.BoolVarP(&list, "list", "l", false, "lists tasks with description of current Taskfile")
+	pflag.BoolVarP(&listAll, "list-all", "a", false, "list tasks with or without a description")
 	pflag.BoolVar(&status, "status", false, "exits with non-zero exit code if any of the given tasks is not up-to-date")
 	pflag.BoolVarP(&force, "force", "f", false, "forces execution even when the task is up-to-date")
 	pflag.BoolVarP(&watch, "watch", "w", false, "enables watch of the given task")
@@ -148,8 +150,8 @@ func main() {
 		return
 	}
 
-	if list {
-		e.PrintTasksHelp()
+	if list || listAll {
+		e.PrintTasksHelp(listAll)
 		return
 	}
 
--- a/help.go
+++ b/help.go
@@ -15,7 +15,7 @@ import (
 func (e *Executor) PrintTasksHelp(listAll bool) {
 	var tasks []*taskfile.Task
 	if listAll == true {
-		tasks = e.taskNames()
+		tasks = e.allTaskNames()
 	} else {
 		tasks = e.tasksWithDesc()
 	}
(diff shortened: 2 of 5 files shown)


therealkevinard:
I'm not too emotionally attached to the actual flag, but I went with `-a` (instead of `-L`), because it's more in line with the conventional flag for this sort of thing (eg: `ls -lha`, `du -ah`, and others).

kerma:
Hey @therealkevinard! Would you like to update and finish this?

therealkevinard:
Franky: oh, wow, this pr is stale! 

Yeah, I'd like to knock the dust off it. I'll catch myself up over the next day or so and finish this off. 

Are there any concerns with this note 
> The new Executor.TaskNames func does not use e.CompiledTask(taskfile.Call{Task: task.Task})

I can grok CompiledTask to see what the implications are, but maybe someone here just... Knows? :)

cmd/task/task.go
```diff
@@ -148,8 +150,8 @@ func main() {
 		return
 	}
 
-	if list {
-		e.PrintTasksHelp()
+	if list || listAll {
+		e.PrintTasksHelp(listAll)
 		return
 	}
```
andreynering, line 156: Boolean arguments are usually discouraged for readability reasons. I'd consider having a separate function for each one:

```suggestion
	if list {
		e.ListTasksWithDesc()
		return
	}
	if listAll {
		e.ListAllTasks()
		return
	}
```

In the package you could then abstract them in s private function as below:

```go
func ListTasksWithDesc() {
 	listTasks(false)
}

func ListAllTasks() {
 	listTasks(true)
}

func listTasks(all bool) {
 	// ...
}
```

help.go
```diff
@@ -26,6 +35,15 @@ func (e *Executor) PrintTasksHelp() {
 	w.Flush()
 }
 
+func (e *Executor) taskNames() (tasks []*taskfile.Task) {
```
andreynering, line 38: `taskNames` -> `allTaskNames`

task.go
```diff
@@ -64,7 +64,9 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 	for _, c := range calls {
 		if _, ok := e.Taskfile.Tasks[c.Task]; !ok {
 			// FIXME: move to the main package
-			e.PrintTasksHelp()
+			// FIXME: (ard.kevin.84@gmail.com) changed the PrintTasksHelp signature to support show all/some.
+			// 	False preserves original behavior, but should be reviewed.
+			e.PrintTasksHelp(false)
```
andreynering, line 69: The note is not needed IMO. Let's keep the default behavior which is to list documented tasks only.

```suggestion
			e. ListTasksWithDesc()
```

andreynering (commented):
Hi @therealkevinard, thanks for your PR!

I just did a few comments. Let me know of your thoughts.

cmd/task/task.go
```diff
@@ -75,6 +76,7 @@ func main() {
 	pflag.BoolVarP(&helpFlag, "help", "h", false, "shows Task usage")
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yml in the current folder")
 	pflag.BoolVarP(&list, "list", "l", false, "lists tasks with description of current Taskfile")
+	pflag.BoolVarP(&listAll, "list-all", "a", false, "list tasks with or without a description")
```
andreynering, line 79: No strong opinion on `-a` vs. `-L`. If anyone has thoughts, let me know.

cmd/task/task.go
therealkevinard: resolved in https://github.com/therealkevinard/task/commit/42702e81b35c5b4960c94ff65577f6c4314b0ccd

help.go
therealkevinard: resolved in https://github.com/therealkevinard/task/commit/42702e81b35c5b4960c94ff65577f6c4314b0ccd

task.go
therealkevinard: resolved in https://github.com/therealkevinard/task/commit/42702e81b35c5b4960c94ff65577f6c4314b0ccd

therealkevinard:
There are no merge conflicts, but this branch obviously behind. Do you think a backmerge is in order?

andreynering:
Hi @therealkevinard, thanks for contributing!

I just merged this and did small changes at 7e0346d6eb04a54e19106611851e4533fe23815c.

PR #404, opened 2020-11-16
Fix documentation link in taskfile_version

DuboisS:


andreynering:
Hi @DuboisS,

Sorry, but I think you're mistaken. The link seems correct to me.

PR #406, opened 2020-11-17
Update curl arguments in installation.md

ezhukov:
Double -s argument does not make sense according to curl manual page.
-s stands for silent, while -S (capital S) stands for showing the error.
When used in combination, curl shows an error message if it fails, but disables progress meter.

Finally, in the end of sh command there is -d, which stands for debug and contradicts -sS curl arguments.
I suggest to remove curl silencers all together, because more debug is better in CIs.
I also suggest to use --location instead of -L for clarity.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -74,11 +74,11 @@ easily generating this script.
 
 ```bash
 # For Default Installion to ./bin with debug logging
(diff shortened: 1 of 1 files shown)


andreynering:
@ezhukov Thanks!

PR #408, opened 2020-11-22
Implements running from parent directory #298

### LINKED ISSUE #289: Search for Taskfiles in parent directories
Is it possible to run Task commands from child directories without specifying a path to the Taskfile explicitly (`task -t ../.././Taskfile.yml`)?

### LINKED ISSUE #395: Why is `go get` not supported?
Hello, first let me say I really like Task and hope to contribute to it and get more adoption of it in my OSS projects (and in my work @sourcegraph where possible.)

One aspect that confuses me a bit is that the installation docs explicitly mention it cannot be installed via `go get`:

> Task requires Go Modules and doesn’t officially support installing via go get anymore.

From what I can tell, nothing about Task prevents `go get` from working and in fact `go get -u github.com/go-task/task/cmd/task` works perfectly fine today - so long as you are running a recent Go version with module support (which I imagine most are at this point.)

I totally understand the binary options being suggested by default, but it's weird to say it doesn't officially support `go get` because it makes me (and I imagine others) think that:

- Task has some very complex build process that doesn't work with `go get`
- Task has some non-idiomatic Go code
- Task requires CGO dependencies that can't be built with `go get`

It seems i haven't been the first with this thought before, as well: https://github.com/go-task/task/issues/250#issuecomment-692723563

My suggestion is to do one of the following:

1. Declare that `go get -u github.com/go-task/task/cmd/task` works - because it does and this is often the preferred installation method of many Gophers.
2. OR at least don't declare that it isn't supported, because this makes me think it doesn't work and gives me a more negative view of Task which does it a disservice. It's a really great tool and I want to see it adopted more :)

enc:
Uses the absolute path to make sure we can move up.

It's not smart yet, but it works.

so it could fix #289

Also resolves #395 since it was so simple to change.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -108,7 +108,12 @@ This installation method is community owned.
 First, make sure you have [Go][go] properly installed and setup.
 
 Task requires [Go Modules](https://github.com/golang/go/wiki/Modules) and
-doesn't officially support installing via `go get` anymore.
+can be installed got `go get` given that you have a recent version of [Go][go].
+
+```bash
+go get -u github.com/go-task/task/cmd/task
+```
+
 
 Installing in another directory:
 
--- a/taskfile/read/taskfile.go
+++ b/taskfile/read/taskfile.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/go-task/task/v3/internal/templater"
 	"github.com/go-task/task/v3/taskfile"
@@ -23,9 +24,12 @@ var (
 
 // Taskfile reads a Taskfile for a given directory
 func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
-	path := filepath.Join(dir, entrypoint)
-	if _, err := os.Stat(path); err != nil {
-		return nil, fmt.Errorf(`task: No Taskfile found on "%s". Use "task --init" to create a new one`, path)
+	// we need the full path to iterate
+	dir, _ = filepath.Abs(dir)
+
+	path, err := findPath(dir, entrypoint)
+	if err != nil {
+		return nil, err
 	}
 	t, err := readTaskfile(path)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


enc:
Ok, this will not work on windows...

I will check this later.

amoutiers:
Great feature !
Like git, docker-compose, ...

mvgijssel:
Yeah big +1 on this change!

docs/installation.md
```diff
@@ -108,7 +108,12 @@ This installation method is community owned.
 First, make sure you have [Go][go] properly installed and setup.
 
 Task requires [Go Modules](https://github.com/golang/go/wiki/Modules) and
-doesn't officially support installing via `go get` anymore.
+can be installed got `go get` given that you have a recent version of [Go][go].
+
+```bash
+go get -u github.com/go-task/task/cmd/task
```
ghostsquad, line 114: you can no longer install task with `go get` as of 1.18.

https://tip.golang.org/doc/go1.18#introduction

This document should reflect the use of `go install`

taskfile/read/taskfile.go
```diff
@@ -133,6 +137,25 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
...
+		path = filepath.Join(dir, entrypoint)
+		if _, err := os.Stat(path); err != nil {
+			if dir == "/" {
+				return "", fmt.Errorf(`task: No Taskfile found on "%s". Use "task --init" to create a new one`, anker)
+			}
+			dirFragments := strings.Split(dir, string(os.PathSeparator))
+			dir = "/" + strings.Join(dirFragments[1:len(dirFragments)-1], string(os.PathSeparator))
+
+		} else {
+			found = true
```
ghostsquad, line 153: technically, you don't even need `found` here, you can just `break`.

ghostsquad (changes requested):
This behavior can actually be _quite_ dangerous. I would recommend against implementation of this without further discussion of risk. Additionally, this change is not backwards compatible, right now an "error" is produced if you run task from a directory that does not contain a taskfile. As such, this feature should be hidden behind a feature flag in the CLI (with a default `enabled: false`).

Git CVE-2022-24765 describes the risk actually quite well:

https://github.blog/2022-04-12-git-security-vulnerability-announced/

andreynering:
Closing in favor of #920

PR #415, opened 2020-12-07
Fix duplicate task name on "Short task syntax" example

felladrin:
To make it a valid YAML file and avoid the error:

```
yaml: unmarshal errors:
  line 6: mapping key "build" already defined at line 4
```

Also, mention `desc:` key in that section, as it's an important keyword.


andreynering:
Hi @felladrin,

Sweet, thanks for contributing!

PR #416, opened 2020-12-09
chore: include taskfile vars in task scope

zlesnr:
Without this change, the variables present on a Taskfile are not available from
within the Task.  This means that in the case of a "Live" variable, the task
template is unable to access these objects as expected.

Here we merge the Taskfile vars into the vars for the given Task, allowing the
Live variables to be accessed from within the Task template scope.

### DIFF
--- a/variables.go
+++ b/variables.go
@@ -23,6 +23,8 @@ func (e *Executor) CompiledTask(call taskfile.Call) (*taskfile.Task, error) {
 		return nil, err
 	}
 
+	vars.Merge(e.Taskfile.Vars)
+
 	v, err := e.Taskfile.ParsedVersion()
 	if err != nil {
 		return nil, err


zlesnr:
Related to #121

zlesnr:
It seems. that this is breaking the tests.  I've also tried the merge the other direction, starting with the `Taskfile` vars and then merging in the specific `Task` vars, which also fails the tests.  Is there another way to approach this?  I was digging around in the code to see how I could get access to the `Taskfile` `Vars` from within the task itself, but it looks like only the environment variable are cared for.  Guidance here would be appreciated.

andreynering:
Hi @zlesnr,

It's not clear to me what exactly you're trying to fix here. Do you mind sharing an example Taskfile showing the issue?

andreynering:
Closing for lack of feedback (and I still did not understand the intention, as mentioned before).

Feel free to comment if you still think there is something to be fixed here.

Issue #373, opened 2020-09-05
Watch: Add optional integration with Watchman?


andreynering:
[Originally suggested by someone else on Reddit](https://www.reddit.com/r/golang/comments/ib4px3/task_v300_is_released/g1wulhv/)

tylermmorton:
On this front I found https://github.com/sjansen/watchman -- seems like a Go client that does exactly what was mentioned in the Reddit post. It seems largely WIP and 3 years stale but according to [`status.md`](https://github.com/sjansen/watchman/blob/master/docs/status.md) it at least has `watch_list` and `watch_project` implemented. 

Could we fork this project and and continue development of the API's we need for under `go-task/watchman`?

trulede:
As far as I can tell, watchman is an abandoned project.

PR #423, opened 2021-01-04
Make watch behavior work better

### LINKED ISSUE #368: Task Watch is unreliable and other helpers
```Task version: 3.0.0```
```OS: Linux host 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux```

First off it would be good if there was a way to force the `cmds` part of a watched task to always fire the first time.  I've been able to work around this with my `always` task as a dependency hack.

Second, I found that the task watcher to be unreliable as it will simply stop working at some point.  I've been able to work around this by using the [reflex](https://github.com/cespare/reflex) tool.  So now I can run my dev environment with `task dev:watch` and `task -w dev`.  Since I have both styles available to me I'll continue experimenting to see if I can isolate the bug.

```yaml
version: '3'

vars:
  CMD_NAME: garage
  JS_SRC: web/raw/typescript
  JS_DST: build/web/javascript
  JS_ENTRIES:
    index
    signin

  dev:watch:
    cmds:
      - reflex --decoration=none --regex='(\.[a-z]+)$' --start-service task dev
    silent: true

  dev:
    deps: [always, build:script]
    cmds:
      - go run {{.CMD_NAME}}.go
    sources:
      - "**/*.go"
      - web/static/**
      - web/templates/**
    method: checksum

  build:script:
    deps: [build:prepare]
    cmds:
      - |
        for ENTRY in {{.JS_ENTRIES}}
        do
          deno bundle "{{.JS_SRC}}/${ENTRY}.ts" "{{.JS_DST}}/${ENTRY}.bundle.js"
        done
    sources:
      - web/raw/typescript/**/*
    method: checksum

  build:prepare:
    cmds:
      - install --directory build
      - install --directory build/web/{css,javascript}
    status:
      - test -d build
      - test -d build/web/css
      - test -d build/web/javascript
    silent: true

  always:
    cmds:
      - echo "Always run" >/dev/null
    status: [ 1 = 0 ]
    silent: true
```

jwendel:
Watch wasn't working well for me. I started digging into it some and noticed that the `watcher` cli tool itself never had problems picking up changes for me.  Digging into how `task` is using watcher, I found a few things.

Top-level change explanation

1) My linter complained about the `return err` on line 44 was returning without cancel() being called.  I added this.
2) registerWatchedFiles was removing and adding files over and over to watcher.  Details below

Details on fix:

* While debugging, I added some print statements, and saw that this code kept removing and adding files to Watcher.  I didn't track down into Watcher itself, but this loop (going every second) in Taskfile seems to be the cause of making watch behavior not work well.
* Looking at the code it's unclear to me why `w.Remove(f)` is even being called.  Removed this for loop.
* On old line 149, the `if _, ok := oldWatchedFiles[f]; ok {` was 100% miss rate (at least on Windows).  Watcher uses absolute paths, and task always operates on relative paths.
* I changed that block to use `filepath.Abs(f)` to get the full path (which is the exact same code watcher uses).  I'm not a fan of binding task and watcher together like this, but it is better than what it's doing today.
* Now it will log each time a new file is being watched.  Verbose out will help debug cases if this code brakes, as the logs would be flooded.

This likely will help fix #368 (though that was reported on linux, and I tested on windows)

Potential other change:  I don't know if `registerWatchedFiles` being called every second is good for performance (especially on large projects).  I'd recommend reducing this (maybe 10 seconds?  Maybe longer?)

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -81,12 +81,12 @@ func (e *Executor) watchTasks(calls ...taskfile.Call) error {
 	}()
 
 	go func() {
-		// re-register each second because we can have new files
+		// re-register every 20 seconds because we can have new files, but this process is expensive to run
 		for {
 			if err := e.registerWatchedFiles(w, calls...); err != nil {
 				e.Logger.Errf(logger.Red, "%v", err)
 			}
-			time.Sleep(time.Second)
+			time.Sleep(time.Second * 20)
 		}
 	}()
 
--- a/watch.go
+++ b/watch.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
@@ -41,6 +42,7 @@ func (e *Executor) watchTasks(calls ...taskfile.Call) error {
 	defer w.Close()
 	w.SetMaxEvents(1)
 	if err := w.Ignore(watchIgnoredDirs...); err != nil {
+		cancel()
 		return err
 	}
 
@@ -114,12 +116,6 @@ func (e *Executor) registerWatchedFiles(w *watcher.Watcher, calls ...taskfile.Ca
 		oldWatchedFiles[f] = struct{}{}
 	}
 
-	for f := range oldWatchedFiles {
-		if err := w.Remove(f); err != nil {
-			return err
-		}
-	}
-
 	var registerTaskFiles func(taskfile.Call) error
 	registerTaskFiles = func(c taskfile.Call) error {
 		task, err := e.CompiledTask(c)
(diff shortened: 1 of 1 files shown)


jwendel:
Updated times between `registerWatchedFiles` calls.  Looking into it, filepath.Abs ends up having to do a syscall for every file passed to it.  This is far from free to call every second.

andreynering:
Hi @jwendel,

Thanks a lot for your contribution, it's really valuable! ❤️ 

The watch feature indeed needed some love. Your changes makes total sense, and I did some further tests and improvements at d830178ef810712e51f6254c76ffee55f4a5c685.

20 seconds sounded like too much for me, so I reduced it to 5 seconds. I also improved the lib cadence from 1 second to 5 seconds to make it more stable (this will reduce doubled notifications).

Hopefully some day we'll be able to do #373.

PR #428, opened 2021-01-12
Add ARM64 in supported platforms

odidev:
The following files have been modified:

- In **install-task.sh**: Added linux/arm64 architecture in **is_supported_platform()** function in order to install the binary for arm64 through the script.


andreynering:
Hi @odidev, thanks for contributing!

This file isn't supposed to be edited and it is code generated by using [godownloader](https://github.com/goreleaser/godownloader). I just re-generated it and it should now be working.

PR #430, opened 2021-01-12
Fix attribute name for the watch paragraph

drupol:
Hi,

I'm evaluating the watch (-w) option and I noticed that the documentation needs to be updated.


andreynering:
Hi @drupol,

Thanks for opening this PR, but you are mistaken here. Actually both should work, but our convention is to use lowercase keys, so the documentation is correct.

Issue #189, opened 2019-03-19
How to force task but not its dependencies


sheerun:
Hello,

I'm working on developing a task and I'd like to re-run just it even it's up to date, but I don't want its dependencies to run because they take long time to run and I already know they are correct.

Is there a way to force running just particular task and skip its dependencies other than commenting "deps" for this task in Taskfile?

andreynering:
Hi @sheerun,

Today this still isn't possible, unless you edit the Taskfile as you said.

Maybe you could achieve that by refactoring your Taskfile. For example, having two tasks, one with deps and the other without. To share the code, you could have a third one that would be called by these two using `task: foo`.

(Not sure how much sense it makes since I don't know what you're trying to do).

Regarding allowing forcing a single task, we could add another flag for that (e.g. `--force-task=foo` as an alternative to `--force`). Does that makes sense?

sheerun:
I think `--force` should force only explicitly listed tasks. For example if there are tasks A, B, C and A depends on B and B depends on C, then `task A --force` should force only task A and do B, C only if necessary, `task A B --force` should force tasks A, B and perform C if necessary. Maybe this could be on v3 wishlist :)

smyrman:
If you do `task A B C` as described above, A->B->C, B->C, and C are started in parallel so that C is potentially run three times, B two times, and A once.

smyrman:
Just stating how it currently works.

sheerun:
In my example above the dependency is just `A->B->C`, sorry if wasn't clear enough

As for your example, running C 3 times is something to be improved as well for v3, IMO..

andreynering:
@sheerun 

> I think --force should force only explicitly listed tasks.

That's a good point. But it's also a breaking change. And I'm sure there're users out there that'd expect that to work exactly how it works today: `--force` forces everything.

So ideally, we have to find a way to make both use cases possible.

> As for your example, running C 3 times is something to be improved as well for v3, IMO..

There's a jurassic issue for that already: https://github.com/go-task/task/issues/53. That discussion is old, so don't expect our opinions to be the same nowadays. But sure, we should look into fixing that.

ghostsquad:
I'm curious how you know the dependencies are up to date? Can you encode that information in the `status` of those dependencies?

andreynering:
Closing in favor of #1200.

pd93:
[...]

PR #435, opened 2021-01-25
Add "ff" (force first) flag

minitauros:
This flag forces only the first called command.

Inspired by https://github.com/go-task/task/issues/189.

My problem is similar. I have tasks A and B. Task A downloads and installs a bunch of dependencies. Task B first runs A (either as dependency or as a cmd) and uses the downloaded dependencies to generate some code. Task B only runs if code was not yet generated (it uses a `status` entry).

It can happen that I want to _regenerate_ code, but not _redownload_ the dependencies that are used to generate that code. E.g. if my API spec changes, I do not need to redownload the dependencies that generate the code; I just want to regenerate the code.

`task -f B` would **always** run A and B again and redownload all dependencies. 

With this PR you would run `task -ff B` to **only** run B again, and not A.


pd93:
@andreynering I've recently been running into this issue. My use case is for a local k8s development cluster where each service has a task to bring it up and these all depend on a task that will bring up the cluster. The task to bring up the cluster has a status check, so that it will not run if the cluster is already running. If my service is running and I do `task myservice -f`, I do not expect that my cluster will be recreated 😛 

This PR has been open a very long time, but I'm now wondering if the behaviour of `--force` is inherently incorrect. Rather than adding a `--force-first` flag, would it be more intuitive to have `--force` adopt this behaviour and have a `--force-all` flag that will will force dependant tasks too. Would this be considered a breaking change? Maybe I'm wrong, but I don't think many people will be using `--force` to intentionally force all tasks to rerun.

pd93:
@minitauros @andreynering Closing this as the requested behaviour will be available via the [gentle force experiment](https://github.com/go-task/task/issues/1200) in the next release which follows the pattern described in my previous comment.

PR #463, opened 2021-04-04
Update Go modules installation documentation for Go 1.16

JamieEdge:
Updated the documented commands for installing Task using Go modules.


andreynering:
Thanks again @JamieEdge. Those are welcome improvements.

PR #472, opened 2021-04-18
"file does not exist" improving for watchers

butuzov:
Adds additional (initial missing) context to the go generic `os.ErrNotExist` error, in addition to other errors that (possibly) can be returned by `zglob.Glob`

Context: I am working on a few small files that are sources of different generators (that generate stuff that other stuff use to generate their own stuff). And if one of the stages fails, there literally no possible way to tell which file is it - just simple - `file not exists` if all what you got.

## Before 
image

## After
image

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -2,6 +2,7 @@ package task
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @butuzov!

Thanks for the patch, it makes total sense.

I just merged it manually at 9b42ef5d46afa9d8cfa253bdd8eff6a286b6d194 with a tiny change: added a `task: ` prefix to the output which we do by convention.

PR #474, opened 2021-04-19
Allow certain commands to run even if others have failed

mccolljr:
Something I've been working around recently is the fact that `task` will stop execution as soon as one command fails,
but using `ignore_error` at the command level will bury the error and cause the overall `task` execution to
appear successful.

I've been using this workaround:

```yaml
run-all-tests:
  cmds:
    - |
       task launch-docker-for-tests;
       ( go test ./... ); # run the tests in a subshell
       TEST_EXIT=$?;
       task cleanup-docker-for-tests;
       exit $TEST_EXIT
```

It's a bit clunky, which is what inspired me to open this PR. This PR enables the following:

```yaml
run-all-tests:
  cmds:
    - task: launch-docker-for-tests
    - go test ./...
    - task: cleanup-docker-for-tests
      always: true
```

The behavior is mostly identical, with an exception: the first version will fail to run the cleanup action if the `task run-all-tests` call is terminated with a `CTRL-C`, but the second version will still run the `cleanup-docker-for-tests` if an interrupt is triggered via `CTRL-C` or other means. The error from the first failed task is returned.

### DIFF
--- a/task.go
+++ b/task.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/go-task/task/v3/internal/compiler"
 	compilerv2 "github.com/go-task/task/v3/internal/compiler/v2"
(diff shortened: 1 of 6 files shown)


task.go
```diff
@@ -296,8 +297,24 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
...
+			continue
+		}
+
+		// Give commands with always: true their own context so that cleanup can run even after an interrupt signal
+		cmdCtx := ctx
+		if cmd.Always {
+			cleanupCtx, cancelCleanupCtx := context.WithTimeout(context.Background(), time.Second*10)
+			defer cancelCleanupCtx()
+			cmdCtx = cleanupCtx
+		}
```
mccolljr, line 315: I'm not 100% sure whether this is the correct approach or even desired/necessary - completely open to changing or removing this code based on feedback

andreynering:
@mccolljr Sorry for not taking a look at this in time. I'll comment on the issue.

Issue #458, opened 2021-03-25
In case of interrupt signal, Task does not give time to a subprocess to cleanup


marco-m:
## Background

Although Task has logic to intercept a SIGINT in getSignalContext():
https://github.com/go-task/task/blob/d3cd9f17f9d16dde7c16a20713324c37d8cea47c/cmd/task/task.go#L209
the execution of a subprocess is interrupted immediately, not giving time to cleanup. For some subprocesses such as Terraform, this can have severe consequences as explained below.

The cancellable context obtained from getSignalContext() is passed everywhere:

Executor.Run()
Executor.RunTask()
Executor.runCommand()
internal/execext/exec.go:RunCommand()

and finally to

mvdan.cc/sh/v3/interp/interp.New()

## The failing use case

I found this problem invoking `terraform apply` from Task.

[Terraform](https://github.com/hashicorp/terraform) is a tool to provision cloud infrastructure as code. When it runs, it can modify state in a remote backend, for example S3. If interrupted with CTRL-C (so SIGINT), it will take around 10 seconds to cleanup in a way that is safe for the remote state to be used again.

When run under task, this grace period is skipped and Terraform is terminated too fast. This can leave the remote state in a corrupted configuration and require non-obvius manual intervention to recover.

One could argue that Terraform should have a more resilient approach, but I think that in general this is a good example of the general problem.

## Proposal 1

In my opinion Task should actually do as Terraform, which reacts differently depending on how many SIGINT signals it has received:

- on reception of the first SIGINT, enter cleanup state. For Task, this would mean simply record that a SIGINT has been received and wait as usual for the subprocess to terminate.
- on reception of a second SIGINT, assume that the user decided that the wait is over, and terminate abruptly (the current behavior).

## Proposal 2

If proposal 1 is too complicated to implement, a fallback would be to introduce a per-task timeout: when a SIGINT is received, start a timer, and terminate abruptly only on timer expiration

## How to reproduce

Taskfile:

```YAML
version: '3'

tasks:
  default:
    cmds:
      - ./ctrl_c
```

Go code to simulate a program that needs time to cleanup on reception of SIGINT:

```Go
package main

import (
	"fmt"
	"os"
	"os/signal"
	"time"
)

func main() {
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt) // Ctrl-C -> SIGINT

	fmt.Printf("Do one of the following:\n")
	fmt.Printf("- From the Task shell: type CTRL-C\n")
[...]

PR #476, opened 2021-04-20
signals: do not try to catch uncatchable signals

marco-m-pix4d:
os.Kill is SIGKILL (kill -9), cannot be intercepted.
(see golang/go#13080)

Part of #458.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,9 @@ dist/
 
 .DS_Store
 
-# intellij idea/goland
+# editors
 .idea/
+.vscode/
 
 # exuberant ctags
 tags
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @marco-m-pix4d, thanks for opening this PR!

I'll merge this since it seems an inoffensive change, and makes the code more correct according to your references.

That said, I'm still curious: did you test it and find the this improves anything? As is, I have the impression that, although this wasn't working for SIGKILL, it wasn't causing any hurt as well (was no-op).

marco-m:
hello @andreynering 

> As is, I have the impression that, although this wasn't working for SIGKILL, it wasn't causing any hurt as well (was no-op).

That is correct, it was a (confusing) no-op.

Actually the way I spotted was thanks to a Go linter, the default one you get with vscode. Once I saw the warning, I did some research.

Issue #484, opened 2021-05-08
Running with --watch does not respect task directory when registering sources to watch


lpdyck:
- Task version: v3.4.2
- Operating System: Windows

```yaml
version: '3'
dotenv: ['app/.env']

tasks:
  build-js:
    dir: app/js
    sources:
      - controllers/*.ts
      - application.ts
      - package.json
      - webpack.config.js
      - tsconfig.json
    cmds:
      - npm run webpack
```

When running the `build-js` task, Task correctly recognizes the sources for currency checks. However, when running it with the `--watch` flag, it gives an error `task: application.ts: file does not exist`. Looking at the code, the watch file globbing function does not respect the task directory (https://github.com/go-task/task/blob/master/watch.go#L131) while the internal status checker does (https://github.com/go-task/task/blob/master/internal/status/glob.go#L16).

PR #485, opened 2021-05-08
Changes --watch to respect task directory

lpdyck:
Fixes https://github.com/go-task/task/issues/484

Exports internal/status.glob() and uses that for the watch globbing as well so that watch uses the task directory.

The only problem I see is that status.glob() never returns an error, while zglob.Glob() does, so the behavior will be changed beyond the intended scope. The only solutions I see are to re-implement the code from status.glob, change status.glob to return an error and thereby changing the behavior in other cases, or leave it be and accept the changed behavior of watch.

I prefer never silently ignoring errors, so I would change the behavior in other cases (checksum and timestamp checking), but it's not my call.

### DIFF
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -53,7 +53,7 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 	if len(c.Generates) > 0 {
 		// For each specified 'generates' field, check whether the files actually exist
 		for _, g := range c.Generates {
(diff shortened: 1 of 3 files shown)


butuzov:
@andreynering check this one, I was stumbling upon this bug too. This is an important thing to have in `task`.

andreynering:
Neat patch. Thanks for contributing @lpdyck!

PR #531, opened 2021-07-23
Update community.md about Sublime Text 4 package

biozz:
Added a simple abstract about Sublime Text 4 package with the links to the source code and package control.


andreynering:
Awesome, thanks!

PR #544, opened 2021-08-07
Fix bash completions when the current argument begins with a dash

AlexanderBeyn:
When trying to complete an argument that begins with a dash, the bash completion will potentially cause grep to output error messages:

For example, doing "task -d" followed by tabs in task source tree:
```
$ task -dgrep: option requires an argument -- 'd'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

### DIFF
--- a/completion/bash/task.bash
+++ b/completion/bash/task.bash
@@ -15,7 +15,7 @@ _task_completion()
     return
   fi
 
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks @AlexanderBeyn!

PR #552, opened 2021-08-11
Allow includes to be optional

### LINKED ISSUE #519: Allow optional includes
Currently if an include doesn't exist then task fails to run e.g.
```
❯ task --list
stat vendor/lullabot/drainpipe-dev/drainpipe-dev/tasks/test.yml: no such file or directory
```

It'd be great if an include could be marked as optional:
```
includes:
  tests:
    taskfile: ./vendor/blah/blah/tests.yml
    optional: true
```
or even a global flag to mark all includes as optional.

My specific use case is that we're sharing Taskfiles amongst PHP projects as composer packages, but some are dev dependencies only so won't always be present (i.e. when we run `composer install --no-dev`)

justafish:
Fixes #519


docs/usage.md
```diff
@@ -154,6 +154,24 @@ includes:
 > This was a deliberate decision to keep use and implementation simple.
 > If you disagree, open an GitHub issue and explain your use case. =)
 
+### Optional includes
+
+Includes marked as optional will allow Task to continue execution as normal if
+the included file is missing.
```
deviantintegral, line 160: Perhaps actually write out in words the scenario below, so someone reading can understand the value?

Something like: "This is useful for removing task files providing development and testing commands from production environments".

task_test.go
```diff
@@ -755,6 +755,41 @@ func TestIncludesCallingRoot(t *testing.T) {
 	tt.Run(t)
 }
 
+func TestIncludesOptional(t *testing.T) {
```
juampynr, line 758: What does this do? I looked at https://github.com/go-task/task/blob/8f80fc4e2c283c41cb1525b88996bd07743cbf31/task_test.go#L24 but it's not clear to me.

juampynr (approved):
Looks good! Would be great to see the tests passing in the pull request but it seems that the GitHub Action has not been set up.

task_test.go
justafish: @juampynr it's a helper to test whether the contents of a file match an expected output. In the case of the one immediately below, it's checking that `testdata/includes_optional/called_dep.txt` exists and contains the text `called_dep`. This was generated by `testdata/includes_optional/Taskfile.yml` running `echo "called_dep" > called_dep.txt` as a command

justafish:
Thanks for the review @juampynr ! Looks like @andreynering enabled the tests for this PR and they're passing :tada:

andreynering:
Thanks @justafish!

Issue #185, opened 2019-03-08
String interpolation and home dir (including ~)


Ilyes512:
Task version: 2.4.0    OS: Darwin (MacOS)

This project really looks promising, and I would really like to use this instead of `Make` except I am running into some problems.

I have the fallowing example task file. For the below test I ran this first to have a test file: `$ echo 'hello world' > ~/testfile.txt`:

```yaml
version: "2"

env:
  WORLD: world!!!!!
  HW: hello_${WORLD}
  TEST_FILE: ~/testfile.txt

tasks:
  test:
    cmds:
      - echo $HOME
      - echo ~
      - echo $WORLD
      - echo $HW
      - echo $TEST_FILE
      - echo ~/testfile.txt
      - cat ~/testfile.txt
      - echo ${TEST_FILE}
      - cat ${TEST_FILE}
```

I get the fallowing output (minus the `-----> expected` appended) when I run `$ task test`:

```bash
echo $HOME
/Users/ilyes512
echo ~
/Users/ilyes512
echo $WORLD
world!!!!!
echo $HW
hello_${WORLD} -----> expected: hello_world!!!!!
echo $TEST_FILE
~/testfile.txt
echo ~/testfile.txt
/Users/ilyes512/testfile.txt
cat ~/testfile.txt
hello world
echo ${TEST_FILE}
~/testfile.txt -----> expected: /Users/ilyes512/testfile.txt
cat ${TEST_FILE}
cat: ~/testfile.txt: No such file or directory -----> expected: hello world
```

When I set an env variable like this (and I don't even use it):

```yaml
env:
  EXAMPLE: ${HOME}/foobar
```

It will error with:

```bash
"{HOME}/foobar": executable file not found in $PATH
task: Command "{HOME}/foobar" in taskvars file failed: exit status 127
```

- So should I be able to use a earlier set env variable in another env?
- Does the `~` and `${HOME}` behave correctly?

Another issue I saw about almost the same thing: #74 It also mentions a "chat". Is this still in use, and if so, it should be documented somewhere? (would expect the readme to mention it).

BTW I have tried variables as both `${FOOBAR}` as well as `$FOOBAR` (so minus the `{` and `}`).

Thanks in advance :)

andreynering:
Hi @Ilyes512,

[...]

PR #557, opened 2021-08-21
Allow included Taskfiles to use ~/* paths

### LINKED ISSUE #539: Includes should properly handle paths with ~
I am trying to write a Taskfile that includes some common tasks from a special folder in my home directory. The `~` in the path is not expanded and task throws an error that the file at the given path could not be found.

Example:

```yaml
includes:
  common: ~/.go-task/Taskfile.yml
```

PS — great tool, loving it so far!

amancevice:
This change allows users to include Taskfiles inside the home directory. 

Paths beginning with the `~` character are expanded to absolute paths using `os.UserHomeDir()`

Example:

```yaml
version: "3"

includes:
  common: ~/.task/Taskfile.yml

tasks:
  default:
    cmds:
      - task: common:task1
      - task: common:task2
      - echo 'Finished.'
```

This references #185  and resolves #539

### DIFF
--- a/taskfile/read/taskfile.go
+++ b/taskfile/read/taskfile.go
@@ -10,6 +10,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
+	"github.com/go-task/task/v3/internal/execext"
 	"github.com/go-task/task/v3/internal/templater"
 	"github.com/go-task/task/v3/taskfile"
 )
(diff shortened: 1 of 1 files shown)


andreynering (commented):
Hi @amancevice,

We already use `execext.Expand(path)` on other places to support `~`. I'd prefer to use that here as well for consistency.

amancevice:
Thanks, @andreynering — change made & tested. Thanks!

andreynering (commented):
Hi @amancevice,

Actually, there's no need for the whole `else if` you wrote. The point of `Expand` is that it takes care of everything for you. Something like this should be enough, before the `if` block:

```go
path, err := execext.Expand(includedTask.Taskfile)
if err != nil {
  return err
}

if filepath.IsAbs(path) {
  // ...
```

amancevice:
@andreynering I see, thanks for your patience — this is my first attempt at writing Go. Updated the PR to your recommendation

andreynering:
@amancevice No problem. Thanks a lot for your contribution! 🙂

PR #562, opened 2021-09-01
Fix fish completions

notnmeyer:
The existing fish completions are broken for any task names that contain a colon. The issue seems to be incorrectly parsing the output of `task -l` to replace all colons and spaces with tabs. For example,

This is how the list is currently parsed, 
```
➜ task -l | sed '1d' | grep docs:ico | awk '{ $1=""; print $0 }' | tr ': ', '\t' | string trim
docs    ico             Generate        favicon.ico     from    Logo.png
```
image

In this situation none of the `docs:*` tasks can be tab completed. The intention _seems_ to be to replace the string `': 
'` but it's not done correctly.

After these changes you can tab complete to each colon within a name like you'd expect,
```
➜ task -l | sed '1d' | grep docs:ico | awk '{ $1=""; print $0 }' | sed 's/:\ /\t/g' | string trim
docs:ico        Generate favicon.ico from Logo.png
```
image

### DIFF
--- a/completion/fish/task.fish
+++ b/completion/fish/task.fish
@@ -1,5 +1,5 @@
 function __task_get_tasks --description "Prints all available tasks with their description"
-	task -l | awk '{ $1=""; print $0 }' | sed 's/:\ /\t/g' | string trim
+	task -l | sed '1d' | awk '{ $1=""; print $0 }' | sed 's/:\ /\t/g' | string trim
 end
 
 complete -c task -d 'Runs the specified task(s). Falls back to the "default" task if no task name was specified, or lists all tasks if an unknown task name was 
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks @notnmeyer!

notnmeyer:
Haha, so I just saw https://github.com/go-task/task/pull/371. the sed command that was originally there should work fine (and does for me on Mac OS with the default sed). The `tr` solution is very clearly not correct—I doubt it was tested with "namespaced" task names.


```
➜ which sed
/usr/bin/sed        # not gnu sed

# the mac sed hilariously has no version options, 
➜ sed --version
sed: illegal option -- -
usage: sed script [-Ealnru] [-i extension] [file ...]
        sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

compare this to gnu sed,
```
➜ docker run -it --rm debian bash
...
root@060c54c64ad4:/# sed --version
sed (GNU sed) 4.7
```

So I'm not sure what that other PR is about :\

andreynering:
Thanks for the explanation @notnmeyer.

I don't use this completion, this is why I haven't noticed that bug. If you say it works by default on macOS, then it works. 🙂

PR #564, opened 2021-09-02
Upgrade Goreleaser action to v2 - To ensure mac/m1 mac/arm release

JulienBreux:
# Description

Upgrade Goreleaser action to version 2.

Because, it's impossible to use go-task with a M1 Mac.

With this version, I think that will fix the problem.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
           go-version: 1.15.x
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v1
+        uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
           args: release --rm-dist


justafish:
@JulienBreux task_darwin_amd64 works for me on an M1 (not that I'm against this PR, but just in case you're stuck!)

JulienBreux:
> @JulienBreux task_darwin_amd64 works for me on an M1 (not that I'm against this PR, but just in case you're stuck!)

Are you using Brew?

Because, I think that with Brew, it's really not possible : https://github.com/go-task/homebrew-tap/blob/master/Formula/go-task.rb#L6-L8

Normally you should have this condition: https://github.com/JulienBreux/homebrew-clamp/blob/master/Formula/clamp.rb#L11-L20

(with a normal `wget` / `mv` / `chmod` it's perfectly functional)

justafish:
@JulienBreux hi! :laughing: 
I'm not using Brew. So `OS.mac` doesn't return `true` for that formula if I'm understanding correctly?

JulienBreux:
> @JulienBreux hi! 😆
😊 

> I'm not using Brew. So `OS.mac` doesn't return `true` for that formula if I'm understanding correctly?
When you use the amazing Goreleaser project, it generate for you the good Brew Formula (AKA the definition to install on your system using Brew). Since the version 2 of Goreleaser, a new condition has appeared to distinguish M1 and Intel.

Actually, with the Goreleaser in version 1. When you install Task, you've the Intel binary, not the ~M1~ ARM binary :)

andreynering:
Hi @JulienBreux and @justafish.

I think #567 will add support for M1, but this PR will help, too.

I'll try locally in my machine and release if it works.

Also, thanks for the Homebrew formula example! That will be useful as well.

andreynering:
Actually, Go 1.17 was release the last month as well. I'll go further and upgrade to that.

PR #567, opened 2021-09-04
Build with Go 1.16

LukeCarrier:
This enables `darwin-arm64` builds to support Apple Silicon-based Macs.


andreynering:
Thanks @LukeCarrier!

andreynering:
Actually, Go 1.17 was release the last month as well. I'll go further and upgrade to that.

PR #568, opened 2021-09-07
Allow users to override task logging colors

amancevice:
As I introduce some of my coworkers to Task some of them have expressed dissatisfaction with the default coloring scheme of Task's logs. Specifically, some of them use custom mapped values so that "magenta" on their machine is much closer to "orange" or "red" than a standard setup.

This PR attempts to provide a mechanism for users to minimally customize the logging colors without too much hassle.

Introduce 4 new loggers in internal/logging/logging.go:
- Exec: color of task commands as they execute
- Fail: color of failed task logs
- Skip: color of skipped tasks (eg, status up-to-date)
- Warn: color of warning (eg, ignored errors)

Colors of these loggers are configurable via ENV vars:
- TASK_COLOR_EXEC
- TASK_COLOR_FAIL
- TASK_COLOR_SKIP
- TASK_COLOR_WARN

Example usage:

```bash
TASK_COLOR_EXEC=36 task test
# => task executes with Cyan-colored execution logs
```

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -2,6 +2,8 @@ package logger
 
 import (
 	"io"
+	"os"
+	"strconv"
 
 	"github.com/fatih/color"
 )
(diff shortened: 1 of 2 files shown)


internal/logger/logger.go
```diff
@@ -63,3 +70,11 @@ func (l *Logger) VerboseErrf(color Color, s string, args ...interface{}) {
 		l.Errf(color, s, args...)
 	}
 }
+
+func EnvColor(envKey string, defaultColor color.Attribute) color.Attribute {
+	envColor, err := strconv.Atoi(os.Getenv(envKey))
+	if err != nil {
+		return defaultColor
+	}
+	return color.Attribute(envColor)
```
andreynering, line 79: I think that switching the control flow would make more sense for me:

```suggestion
	envColor, err := strconv.Atoi(os.Getenv(envKey))
	if err == nil {
		return color.Attribute(envColor)
	}
	return defaultColor
```

andreynering (commented):
Hi @amancevice, and thanks for opening this PR!

I think that the general proposal makes sense. Having names like `Exec`, `Fail`, `Skip` and `Warn` is certainly clearer than referencing the colors directly.

But I think that we should probably migrate completely to the new functions and delete the old ones (beside `Default` maybe) instead of having a mix of new and old styles.

What do you think?

Issue #264, opened 2019-11-13
Setup bash & zsh autocomplete as part of brew install


bfarayev:
Now that we have bash & zsh autocomplete scripts in this repo, we can install them as part of installation with homebrew.

I've noticed that we have this formula already: https://github.com/go-task/homebrew-tap/blob/master/Formula/go-task.rb

Adding 2 lines to the formula, will do that for us:
```
    bash_completion.install "completion/bash/task.bash" => "task"
    zsh_completion.install "completion/zsh/_task" => "_task"
```
I was just testing this on my machine today. Happy to raise a PR for that if it's fine.

Thanks

andreynering:
Hi @bfarayev,

Yeah, it'd be nice to have that. Contributions would be welcome. 🙂 

We'd probably have to adapt our Goreleaser setup to also include these completion scripts on the `.tar.gz` files.

Also, we need to make sure that it won't break stuff for users that try install an older version (without the completion scripts in the tarball).

andreynering:
Some additional instructions on how to do it was added by @zx8 here: https://github.com/go-task/task/issues/382#issue-713577257

Contributions with this are welcome!

notnmeyer:
hey @bfarayev! i opened a PR for this, but im not sure how to go about testing it. would love to hear a little about how you tested.

bfarayev:
hey @notnmeyer! I guess the best way to test is to remove task from your system (Eg: on Mac you'd do `brew uninstall task`) first. Assuming that you're making a change into the Homebrew formula in your own fork, install that formula and see if it installs bash/zsh completion as part of it. Delete and make changes to your formula until it works and when ready, PR to the upstream.

notnmeyer:
thanks @bfarayev, i think what i probably worded my question poorly. i was looking for tips on how to install from the tap in my fork, that's the part that seems non-obvious.

bfarayev:
ah sorry for the late answer. glad that you figured it out.

I think the best ways to test the changes to the tap (Eg the PR you're doing to go-tasks's main homebrew-tap in [this PR](https://github.com/go-task/homebrew-tap/pull/2)) is to fork that repo, make your own tap and then in your local machine, remove everything related to task, "brew tap" your forked tap which will include the autocompletion scripts change and see what you get when you say something like this:

```
brew tap notnmeyer/go-task
brew install go-task ?
```

or directly use 
```
brew install notnmeyer/tap/go-task
```
(something along these lines will work). 

[...]

PR #569, opened 2021-09-07
update goreleaser config to include `completion/*` files

notnmeyer:
tested by building and inspecting the various archives,`goreleaser release --skip-publish --skip-sign --skip-validate --rm-dist`. this is half of what it should take to install shell completions via homebrew for https://github.com/go-task/task/issues/264

once the completions are packaged with the binary, we can confirm theyre installed by homebrew,  https://github.com/go-task/homebrew-tap/pull/2

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -28,6 +28,8 @@ archives:
     format_overrides:
       - goos: windows
         format: zip
+    files:
+      - completion/*
 
 release:
   draft: true


andreynering:
As said in https://github.com/go-task/homebrew-tap/pull/2 this was fixed as part of https://github.com/go-task/task/pull/592.

Thanks for contribution anyway!

Issue #534, opened 2021-07-26
Seg fault on empty command


Tiksagol:
- Task version: 3.0.0
- Operating System: Linux (Ubuntu 18.04.04)

### Example Taskfile showing the issue

```version: '3'
tasks:
  hello:
    desc: 'hi'
    cmds:
      - echo "hi"
      -
```

When running `task -l` it fails with:
```
> task -t test_task.yml -l
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x67b381]

goroutine 1 [running]:
github.com/go-task/task/v3.(*Executor).CompiledTask(0xc00009a2c0, 0xc000016b50, 0x5, 0x0, 0x0, 0x0, 0x4008000000000000)
        /home/runner/work/task/task/variables.go:78 +0x1101
github.com/go-task/task/v3.(*Executor).tasksWithDesc(0xc00009a2c0, 0xc00000e0e0, 0x0, 0x1)
        /home/runner/work/task/task/help.go:33 +0x152
github.com/go-task/task/v3.(*Executor).PrintTasksHelp(0xc00009a2c0)
        /home/runner/work/task/task/help.go:14 +0x40
main.main()
        /home/runner/work/task/task/cmd/task/task.go:152 +0xdbc
```

SwitchedToGitlab:
As of version 3.6.0 this issue has been resolved. @Tiksagol please update your ```task``` binary to the latest version and try again, and thank you for your bug report.

@andreynering this issue can be marked as CLOSED as FIXED in the latest version. 

```yaml
☢  cat -pp Taskfile.yml 
tasks:
  hello:
    desc: 'hi'
    cmds:
      - echo "hi"
      -python3/sandbox on  FEATURE/USE_LANG_IMAGE [!?] 
```

```bash
task: Could not parse taskfile version "": strconv.ParseFloat: parsing "": invalid syntax
```

andreynering:
Yep, closing as fixed.

If you are able to reproduce this on the current version, let us know and we'll reopen the issue.

Tiksagol:
Hi,

Still happens with go-task 3.7.0 when '**version**' is declared:
``` > cat Taskfile.yml 
version: '3'

tasks:
  hi:
    cmds:
      - echo
      -

> task --version
[...]

PR #573, opened 2021-09-18
Fix segmentation fault on nil slice element for issue #534

masaushi:
This is to fix the issue #534 .
Only `cmds` attribute is pointed out in the issue, but I think other attributes such as `deps` and `preconditions` may have same problems, so I fix them as well.

Please let me know if there are any problems in fixed codes.


andreynering:
Thanks @masaushi!

PR #579, opened 2021-09-27
Avoid directory not found error in evaluating dynamic variables with newly created directory #481

### LINKED ISSUE #481: dynamic vars are broken when task has also a dir: attribute
Hello @andreynering :-)

- Task version: v3.4.2
- Operating System: macOS 11.3

Consider this Taskfile:

```
version: '3'

tasks:
  default:
    dir: x
    cmds:
      - echo hello "{{.FOO}}"
    vars:
      FOO:
        sh: echo impossible
```

If the directory `x` exists, everything is fine:

```
$ mkdir x
$ task
task: [default] echo hello "impossible"
hello impossible
```

But if the directory doesn't exist, Task evaluates the dynamic variable before creating the directory and bugs:

```
$ rm -r x
$ task
task: Command "echo impossible" in taskvars file failed: could not stat: stat /Users/foo/bar/x: no such file or directory
```

and the directory is never created, so also subsequent invocations will fail.

masaushi:
This PR is to fix #481.

When a directory is specified on `dir` parameter in Taskfile, a file/directory not found error occurs when evaluating dynamic variables with the directory which does not exist.

That error shouldn't be raised because the specified directory will be created in later operations.
So I added some changes to avoid directory not found error.

### DIFF
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -47,10 +47,10 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
 
 	r, err := interp.New(
 		interp.Params("-e"),
-		interp.Dir(opts.Dir),
 		interp.Env(expand.ListEnviron(environ...)),
 		interp.OpenHandler(openHandler),
 		interp.StdIO(opts.Stdin, opts.Stdout, opts.Stderr),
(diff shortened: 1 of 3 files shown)


andreynering:
Thanks @masaushi!

PR #589, opened 2021-10-16
Show variable names on verbose mode

### LINKED ISSUE #532: Verbose Mode Missing task tag
I'm trying to create a wrapper for Task in Node.js. I'd like to report the time each task takes. However, the task variables show up as:

```
task: dynamic variable: 'if [ -f meta/main.yml ]; then yq e '.galaxy_info.role_name' meta/main.yml; else jq -r '.name' package.json | sed 's/.*\///'; fi' result: 'androidstudio'
```
They SHOULD show up as:
```
task: dynamic variable:  [common:update-ansible-galaxy-id] 'if [ -f meta/main.yml ]; then yq e '.galaxy_info.role_name' meta/main.yml; else jq -r '.name' package.json | sed 's/.*\///'; fi' result: 'androidstudio'
```

Thank you for your consideration.

jespino:
This added the variable name in the output of the verbose mode.

I duplicated some logic around the HandleDynamicVar to avoid changing the
public interface. I would prefer to add the `k` parameter there to know the
variable name, but that would break the backward compatibility.

Also, I could print it for each task that defines the variables, but because
all the tasks have access to the variables, it would show it in every single
task, without distintion between the tasks that are using it or not. So I
prefered to show this only once with the variable name, instead of once per
task with the task name.

Fixes #532

### DIFF
--- a/internal/compiler/v3/compiler_v3.go
+++ b/internal/compiler/v3/compiler_v3.go
@@ -64,10 +64,21 @@ func (c *CompilerV3) getVariables(t *taskfile.Task, call *taskfile.Call, evaluat
 			if err := tr.Err(); err != nil {
 				return err
 			}
+
+			debugLog := false
+			if _, ok := c.dynamicCache[v.Sh]; !ok {
+				debugLog = true
+			}
+
 			static, err := c.HandleDynamicVar(v, dir)
 			if err != nil {
 				return err
 			}
+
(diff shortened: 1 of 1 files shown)


jespino:
No feedback for a long time, I'm closing it, feel free to reopen it if you feel is useful for anybody.

PR #592, opened 2021-10-27
fix: goreleaser brew completions

caarlos0:
- fixed release to use a Personal Access Token: judging by history/blame, brew upload was disabled because it wasn't working, its likely because you need to use a PAT for it to work, check https://goreleaser.com/ci/actions/#token-permissions for more details
- added completion files to the archives
- installing the completions on the homebrew-tap

this should close https://github.com/go-task/homebrew-tap/pull/2 as well

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,4 +23,4 @@ jobs:
           version: latest
           args: release --rm-dist
         env:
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @caarlos0,

I think I tried a different Personal Access Token and it still failed. Anyway, let's try again; just added a brand new token.

Thanks!

PR #613, opened 2021-11-25
Fix quoting of CLI_ARGS

Peter554:
Consider a task:

```yml
test:
  cmds:
    - pytest {{.CLI_ARGS}}
```

Running `task test -- -m "not foo"` should be equivalent to running `pytest -m "not foo"`. However, with the current implementation the quoting of CLI_ARGS is lost and the task executes `python -m not foo`, which results in an error.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @Peter554, thanks for the PR!

Good catch! This was really missed by me at the time.

Issue #585, opened 2021-10-15
Nested tasks should inherit the calling task's dir


amancevice:
- Task version: 3.9.0
- Operating System: macOS 11.6 (Big Sur)

### Example Taskfile showing the issue

When a task invokes another task as part of its `cmds` block, the called task should inherit the `dir` of the calling task.

Consider the following taskfile at `/tmp/task/Taskfile.yml`:

```yaml
---
version: "3"
tasks:
  inner:
    cmds:
      - pwd
  outer:
    dir: inner
    cmds:
      - pwd
      - task: inner
```

I would expect the command `task outer` to produce:

```
/tmp/task/inner
/tmp/task/inner
```

but instead it produces:

```
/tmp/task/inner
/tmp/task
```

I _think_ this is unexpected behavior, but I'm happy to hear another point of view

sylv-io:
Hi,
I had the same issue once and created a draft PR as PoC: #620
But I'm not sure if this is a breaking change in some use cases.

kerma:
I'm afraid this is a breaking change. It's also a bit unambiguous to handle properly. Consider the example:

```yaml
version: '3'

tasks:
  inner:
    dir: other
    cmds:
      - pwd
  
  test:
    dir: first
    cmds:
      - pwd
      - task: inner
```

Would the test task dir still override inner set dir? What if tasks are defined in separate files?

It's pretty hard to implement in a way that it would not surprise the user. The workaround @amancevice suggested is more clear for the user.

amancevice:
I'm of two minds — I do agree that this could be considered a breaking change, but I also think making the change would make more intuitive sense to a user because it would behave more like how variables work in Task. What I mean by that is if I define a task that _uses_, but doesn't define a variable then it resolves up the stack up to the user's ENV.

In your example I think a typical user would understand that `inner` task runs from the `./other/` directory and won't be overridden by a calling task. On the other hand, if `inner` omitted the `dir:` entry then I think it _would_ make sense for the calling task to impose its working directory on the callee.

Expanding on your example:

```yaml
# /tmp/task/Taskfile.yml

version: '3'

tasks:
  inner-overrides:
    dir: other
    cmds:
      - pwd

  inner-inherits:
    cmds:
      - pwd

  test:
    dir: first
    cmds:
      - pwd
      - task: inner-overrides
      - task: inner-inherits
```

I would expect the output of `task test` to be:

```bash
task: [test] pwd
/tmp/task/first
task: [inner-overrides] pwd
[...]

PR #620, opened 2021-12-01
inherit dir from calling task

sylv-io:
Signed-off-by: Marcello Sylvester Bauer <sylv@sylv.io>


sylv-io:
Fix #585

sylv-io:
https://github.com/go-task/task/issues/585#issuecomment-986760375

PR #621, opened 2021-12-04
Add support for yaml extension

### LINKED ISSUE #584: parse Taskfile.yaml files
Per the [yaml.org FAQ](https://yaml.org/faq.html):
> Please use ".yaml" when possible.

The last PR I saw about this was several years ago, but I think it's worth a new discussion.

kerma:
This PR implements support for both yml and yaml file extensions. 

`--init` creates Taskfile.yaml now

Fixes #584

EDIT: 
Removed the initial implementation where the included file name needed to match the top level file. Taskfile.yml and Taskfile.yaml are used (in that order) as a fallback in cases where the file name is not specified by the user, either from command line or in `includes` block in yaml.


task.go
```diff
@@ -105,9 +105,6 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 
 // Setup setups Executor's internal state
 func (e *Executor) Setup() error {
-	if e.Entrypoint == "" {
-		e.Entrypoint = "Taskfile.yml"
-	}
```
andreynering, line 108: This empty line should be removed

taskfile/read/taskfile.go
```diff
@@ -22,11 +22,14 @@ var (
 )
 
 // Taskfile reads a Taskfile for a given directory
+// Uses Taskfile.yaml or Taskfile.yml when entrypoint is left empty
 func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
-	path := filepath.Join(dir, entrypoint)
-	if _, err := os.Stat(path); err != nil {
-		return nil, fmt.Errorf(`task: No Taskfile found on "%s". Use "task --init" to create a new one`, path)
+
```
andreynering, line 27: This empty line should be removed

taskfile/read/taskfile.go
```diff
@@ -141,3 +144,28 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
 	var t taskfile.Taskfile
 	return &t, yaml.NewDecoder(f).Decode(&t)
 }
+
+func exists(dir, name string) (string, string, error) {
```
andreynering, line 148: There's a bit of code duplication in this function.

The minimum we could do here is assign the error to a global variable like we do on other places, as the same message is returned in two different places.

Bit I think a better solution would be to split this function into at least two separate functions, one with the reading part (Join + Stat calls) and the other calling the previous with the 3 tries (`entrypoint`, `"Taskfile.yml"`, and `"Taskfile.yaml"`).

taskfile/read/taskfile.go
```diff
@@ -141,3 +144,28 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
...
+	var path string
+	if name != "" {
+		path = filepath.Join(dir, name)
+		if _, err := os.Stat(path); err != nil {
+			return path, name, fmt.Errorf(`task: No Taskfile found on "%s". Use "task --init" to create a new one`, path)
+		}
+		return path, name, nil
```
andreynering, line 155: One of the reasons I was hesitant to do this before is because it could generate some bugs on inclusion of external Taskfiles.

Here we probably have one: if the given entry point was `Foobar.yaml` we don't want for included directories to look for `dir/Foobar.yaml`, but for `dir/Taskfile.yaml`. We need to remember the extension for inclusion, but not the actual file name.

andreynering (commented):
Hi @kerma, thanks for addressing this!

This was asked many times before, so I think people care enough to be worth doing it. I tested it seems to work fine!

I made a few comments. If you have questions/ideas or need help to finish the actual code, let me know!

taskfile/read/taskfile.go
kerma: I played around with this a bit and wrote some tests and tbh I'm not sure which is a better behaviour. You are correct, if your entrypoint is foobar.yaml then includes without filename results in an error:

`open included/foobar.yaml: no such file or directory`

But to me it feels that maybe this is the expected behaviour: Your included taskfile name needs to match the parent unless explicitly specified. 

An alternative would be to add a generic fallback for included dir which looks for foobar.yaml, Taskfile.yml, Taskfile.yaml (in that order, where first is the user specified taskfile name). But to me this feels a bit more "magical" and may create more confusion or require documentation.

dirien:
Very happy to see this PR! Good work @kerma!

taskfile/read/taskfile.go
kerma: Thought about this a bit more and I think I'm now on-board with the behavior you described.

kerma:
Hey @andreynering! Would you mind taking another look at this? I've addressed your earlier comments and added couple of tests. Most cases should be covered by `TestIncludes`. 

Have a nice year end and HNY!

taskfile/read/taskfile.go
```diff
@@ -22,11 +22,20 @@ var (
...
-	path := filepath.Join(dir, entrypoint)
-	if _, err := os.Stat(path); err != nil {
-		return nil, fmt.Errorf(`task: No Taskfile found on "%s". Use "task --init" to create a new one`, path)
+	if dir == "" {
+		d, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+		dir = d
+	}
```
andreynering, line 33: I think this whole `if` block can just go. `filepath.Join(wd, "foo")` and `filepath.Join("", "foo")` should be equivalent. What do you think? Any particular reason on while you added it?

taskfile/read/taskfile.go
```diff
@@ -141,3 +148,22 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
 	var t taskfile.Taskfile
 	return &t, yaml.NewDecoder(f).Decode(&t)
 }
+
+func exists(path string) (string, error) {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return path, err
```
andreynering, line 155:
```suggestion
		return "", err
```

taskfile/read/taskfile.go
```diff
@@ -141,3 +148,22 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
...
+	}
+	if fi.Mode().IsRegular() {
+		return path, nil
+	}
+
+	for _, n := range []string{"Taskfile.yml", "Taskfile.yaml"} {
```
andreynering, line 161: This slice could optionally be moved into an non-exported global variable.

taskfile/read/taskfile.go
```diff
@@ -141,3 +148,22 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
...
+		fpath := filepath.Join(path, n)
+		if _, err := os.Stat(fpath); err == nil {
+			return fpath, nil
+		}
+	}
+
+	return path, fmt.Errorf(`task: No Taskfile found in "%s". Use "task --init" to create a new one`, path)
```
andreynering, line 168:
```suggestion
	return "", fmt.Errorf(`task: No Taskfile found in "%s". Use "task --init" to create a new one`, path)
```

testdata/includes_yaml/included_with_custom_file.txt
```diff
@@ -0,0 +1 @@
+included_with_custom_file
```
andreynering, line 1: These 3 generated `.txt` files should be on a local `.gitignore` file and not commited.

andreynering (approved):
Great job @kerma, thanks for your work! 👏 👏 👏 

Added some very minor comments. Also:

1) Would you mind adding an entry to the CHANGELOG? Should be at the top in a section `Unreleased` for now
2) If you could squash your commits into a single one with a nice message that would be a plus 🙌 

After this, this is good to go 🚀

taskfile/read/taskfile.go
kerma: `exits(path string)` needs a path and won't work with an empty string. `dir` and `entrypoint` are both "" when user doesn't specify anything as the default filename handling was removed from other places and is defined only in the read package. 

That didn't come up in tests as they all specify a dir. But the problem is seen here: https://github.com/go-task/task/runs/4665736655?check_suite_focus=true

andreynering:
Thanks @kerma!

Issue #475, opened 2021-04-20
Ability to run certain commands even if other commands have failed


mccolljr:
As usual, I am loving this tool and I use it frequently. This has been a minor pain point (there is a work-around), but I would love to see a feature added which would reduce friction around cleaning up after a command or task.

The use case I've encountered is cleaning up test resources after running tests, regardless of whether or not the tests passed.

Here is a trimmed down version of something I have in a Taskfile currently:

```yaml
test:setup:
  - echo "set up docker containers"
test:run:
  - echo "execute tests"
test:cleanup:
  - echo "remove docker containers"
test:
  - |
     (task test:setup test:run);
     EXIT_CODE=$?;
     task test:cleanup;
     exit $EXIT_CODE;
```

It is clearly possible to achieve the desired result with the current implementation, but it's not ideal, since the cleanup doesn't run if I terminate execution with `CTRL-C` or if some other interrupt mechanism kills the task.

I also have nested task output because I call `task blah:blah` from the shell within a task, which doesn't hurt anything but can be kind of annoying with longer task names.

I'm not 100% sure what form the solution should take, but I have a couple of ideas:

### 1. `always` flag for commands

Allow commands (both `task: ...` and `cmd: ...` forms) to have a property `always` that can be set to `true` to indicate that the command should run even if a previous command within the same task has failed.

Example:

```yaml
test:
  - task: test:setup
  - task: test:run
  - task: test:cleanup
    always: true
```

### 2. `cleanup` for commands

Allow commands (both `task: ...` and `cmd: ...` forms) to accept a `cleanup` property which specifies one or more commands to be run as cleanup at the end of task execution (success or fail) once that command has been successfully executed. This is sort of analogous to `defer` in go, in the sense that once the top-level command is executed, successfully, the cleanup command is guaranteed to run regardless of whether or not the containing task succeeds or fails.

Example:

```yaml
test:
  - task: test:setup
    cleanup:
      - task: test:cleanup
  - task: test: run
```

I have opened a PR (#474) to show how option 1 might be implemented, but ultimately I'd love to find a solution that the maintainers feel fits best with the tool. As long as I can avoid using sub-shells and long sections of bash to ensure cleanup is run, I will be happy :)

andreynering:
Regarding the syntax, I think `defer` could be interesting, to mimic Go's way:

```yaml
version: '3'

tasks:
  foo:
    cmds:
      - echo 'Hello'
      - defer: echo '!'
      - echo 'World'
```

[...]

PR #626, opened 2021-12-15
Deferred Commands

mccolljr:
Simple support for deferred commands.

Addresses #475

Deferred commands are commands that will run after all other commands, even if the task fails.
The semantics of deferred commands are identical to the semantics of defer statements in go:
1. Defers are added to a stack as they are encountered during task execution
2. After the last command completes, or if some command fails, the deferred tasks are executed in a last-in-first-out order

Given the following task definitions:
```yaml
do_some_setup:
  - echo 'doing setup'
  # ...
do_some_teardown:
  - echo 'doing teardown'
  # ...
do_some_stuff:
  cmds:
    - do_some_setup
    - defer: echo 'called after teardown'    # a deferred shell command
    - defer: { task: do_some_teardown, vars: {} }    # a deferred task call, indicated by the ^ prefix
    - defer: echo 'called before teardown'
    - echo 'oops we failed' && exit 1
    - defer: echo 'I will be skipped, because execution failed before we got here'
```

...and given the following call to `task`:
```sh
$ task do_some_stuff
```

...the expected output would be:
```
doing setup
oops we failed
called before teardown
doing teardown
called after teardown
```


mccolljr:
Fixes #475

taskfile/cmd.go
```diff
@@ -41,6 +42,18 @@ func (c *Cmd) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	var deferredCmd struct {
+		Defer string
+	}
+	if err := unmarshal(&deferredCmd); err == nil && deferredCmd.Defer != "" {
+		c.Defer = true
+		if strings.HasPrefix(deferredCmd.Defer, "^") {
+			c.Task = strings.TrimPrefix(deferredCmd.Defer, "^")
+		} else {
+			c.Cmd = deferredCmd.Defer
+		}
```
andreynering, line 54: This behavior is deprecated and undocumented. It was kept above for compatibility reasons (but will be eventually removed). Since here it's a new feature we should just do:

```suggestion
		c.Cmd = deferredCmd.Defer
```

taskfile/taskfile_test.go
```diff
@@ -41,6 +43,16 @@ vars:
...
+		{
+			yamlDeferredCmd,
+			&taskfile.Cmd{},
+			&taskfile.Cmd{Cmd: "echo 'test'", Defer: true},
+		},
+		{
+			yamlDeferredTask,
+			&taskfile.Cmd{},
+			&taskfile.Cmd{Task: "some_task", Defer: true},
+		},
```
andreynering, line 55: As explained in the previous comments, this second test should be removed as well.

testdata/deferred/Taskfile.yml
```diff
@@ -0,0 +1,12 @@
+version: "3"
+
+tasks:
+  task-1:
+    - echo 'task-1 ran'
+
+  task-2:
+    - defer: "^task-1"
```
andreynering, line 8: This one too:

```suggestion
  task-2:
```

andreynering (commented):
Great job @mccolljr, thanks for contributing! 🚀 

I tested it and it's working great for me. It's great that you added tests, too.

I added comments about a deprecated behavior that we shouldn't maintain anymore, and after addressing this is good to go.

mccolljr:
@andreynering just saw your review - thanks, I'll go through and apply the requested changes tomorrow and then ping you

taskfile/cmd.go
mccolljr: That's fine with me - but I think being able to call a task here is a use case I would like to support.

For example, I currently have a couple of projects with tasks such as `test:setup` and `test:teardown` which encompass some number of complex setup steps. During local development, I might keep the test environment set up for a prolonged period of time, and only tear down when I need to change config for a container. In CI, I'll call those tasks as a part of a `test:ci` task which sets up an environment for tests, runs them, and tears down the environment. Some projects have multiple test environments to test against different versions of external software, for example.

I could of course use `defer: "task test:setup"`. If `task` isn't in `PATH` on all machines, it gets a little more complicated, and the double-layered task output is kind of annoying (but not truly problematic).

Do you feel like this is something worth supporting? I'm envisioning something like:

```yaml
test:ci
    cmds:
        - task: test:setup
        - defer: { task: test:teardown }
        - task: test:run_unit
        - task: test:run_integration
```

 Perhaps if we want to be ambitious we could even support something like:

```yaml
defer: { task: test:teardown, vars: { dep_version: v1.2.3 } }
```

andreynering: @mccolljr Yeah, I think it makes sense to support that.

FYI, we already have a common struct for a call: https://github.com/go-task/task/blob/master/taskfile/call.go

I would consider creating a `Defer` struct (and file) in the `taskfile` package. It could look something like this:

```go
struct Defer {
  Cmd string
  Call Call
}
```

With a `UnmarshalYAML` function to handle the cmd vs. call marshalling.

In theory any of this could be valid:

```yaml
- defer: echo 'foo'
- defer: { cmd: echo 'bar' }
- defer: { task: baz }
- defer: { task: baz, vars: { FOO: bar } }
```

Let me know if you have thoughts or questions.

mccolljr: @andreynering for some reason I totally missed that `Call` type.

I ended up going with a slightly modified version of the above approach to keep the code changes within the executor to a minimum, but the supported syntax is basically what you described above minus the `{ cmd: "echo 'blah'" }` version.

andreynering:
Thanks for working on this @mccolljr!

I just merged and did small changes at d6c185580a758cf09c301f76e4589185861f5450.

mccolljr:
Awesome news 😀

Thanks for updating the documentation, I qasn't sure if that needed to be a part of the PR or not

PR #635, opened 2021-12-20
fix color on Windows

### LINKED ISSUE #343: v3: color output doesn't work properly on Windows
- Task version: 3.0.0-preview4
- Windows version: Windows 10 Enterprise 1909 64-bit

Taskfile.yml:

```
# https://taskfile.dev

version: '3'

tasks:
  default:
    cmds:
      - winver
```

Color output doesn't work in cmd.exe and powershell.exe:

Works well in ConEmu and Terminal:

With `version: '2'`, it works well in ConEmu and Terminal (with no color), but still fails in cmd.exe and powershell.exe:

Maybe mattn's [colorable](https://github.com/mattn/go-colorable) is a good solution?

nichady:
fixes #343 

Here's a before and after for one of my Taskfile projects

before fix:
image

after:
image

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"syscall"
 
+	colorpkg "github.com/fatih/color"
 	"github.com/spf13/pflag"
 	"mvdan.cc/sh/v3/syntax"
 
@@ -140,8 +141,8 @@ func main() {
 		Concurrency: concurrency,
 
 		Stdin:  os.Stdin,
-		Stdout: os.Stdout,
-		Stderr: os.Stderr,
+		Stdout: colorpkg.Output,
+		Stderr: colorpkg.Error,
 
 		OutputStyle: output,
 	}


viktorasm:
Having no option to turn off colors permanently (nobody wants to re-type env variable) is a big issue for windows users. Please get this merged:)

jcrben:
@ghostsquad @andreynering thoughts?

ghostsquad:
@jcrben I see no harm in this honestly. How do we test this behavior?

VictorYap-at-Symend:
@jcrben @ghostsquad 

1. Use Windows, having "go (language)" installed.
2. Use Visual Studio Code and its terminal(s) - git-bash (recommended), cmd.exe (untested), powershell (untested).
3. Checkout and install some stable branch/tag of "go-task"
    - if you tweak the "version" variable and do `task.exe --version`, that could help you be more confidence that you're using a custom build.
5. Using that freshly created `task.exe` binary to run some Taskfile for some dummy project/solution that somehow outputs colours. Note that it's mangled (see initial screenshot).
6. Cherry-pick @nichady's commit atop the currently checked-out branch/tag.
7. Rebuild the binary and install it
    - if you tweak the "version" variable and do `task.exe --version`, that could help you be more confidence that you're using a custom build.
9. Re-run the Taskfile and the result should be colourized output instead of mangled colour codes.

(optionally) Re-install "go-task" to get back to an official release, if you like.

I've been happy with the results atop 3.3.0 and some other relatively recent "latest tag" too (probably 3.14.1).

andreynering:
Sorry for taking so long to review this.

I'm not sure what is happening on Windows, but the change in this PR is is certainly not the right way to implement this, sorry. It would have side-effect and cause other bugs.

Not sure what is happening on Windows. I think the problem may be related to "reset" being used for white here (but I'm not sure, we need to debug that):

https://github.com/go-task/task/blob/421cb522d9a6874bd072491bbc466dd5300acae2/internal/logger/logger.go#L15

andreynering:
I have a secondary Windows machine to debug this, just need to find the time...

pd93:
Closing this. [See comment](https://github.com/go-task/task/issues/343#issuecomment-1609108588).

Issue #277, opened 2019-12-14
How to handle shell and shell functions


tommyalatalo:
I'm trying to set up nvm and node using go-task, but the gosh the tasks are running in is giving me some trouble. 

You can see the task I'm trying to run below, where the first command works fine, $SHELL is zsh in my case, and the scripts installs correctly. But nvm is installed as a shell function, not a binary, so the commands "which nvm" and "nvm install node" fail to find "nvm".

```
  setup:
    desc: setup common tools
    cmds:
      - "[ ! -d ~/.nvm ] && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.1/install.sh | $SHELL || echo nvm installed, skipping..."
      - which nvm
      - nvm install node
    silent: no
```

nvm also adds some things to my .zshrc when the install script runs, and it requires to set some environment variables in order to immediately use nvm, or to restart your terminal to load those vars from .zshrc:

```
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
```
Is this possible to do with go-task or am I outside the scope of the application?
It would be really great to be able to set this task up so that I can install nvm and node (npm) so that I can install some common tools that we use in one task. Would putting these installation commands for nvm and node into a script be a better approach?

andreynering:
Hi @tommyalatalo, thanks for opening this issue!

I don't know if it'd be even possible to support that.

I'll keep this issue open for now, and give this some thought and experimentation once I have some available time.

tommyalatalo:
> Hi @tommyalatalo, thanks for opening this issue!
> 
> I don't know if it'd be even possible to support that.
> 
> I'll keep this issue open for now, and give this some thought and experimentation once I have some available time.

The problem in my original question is probably that the tasks are executed in separate shells, so the environment is reset for each following task. I think the solution to this would be similar to the directive that make has, called "[ONESHELL](https://www.gnu.org/software/make/manual/html_node/One-Shell.html)". 

I can't really say anything about whether or not this is possible to support or not, just that it would be really useful :)

melivoro:
version: '3'

vars:
  GREETING: hello

tasks:
  default:
    cmds:
      - export WHAT=world
      - echo "{{.GREETING}}" ${WHAT}
[...]

PR #643, opened 2021-12-27
Re-use shell for commands of the same task

melivoro:
closes issue: https://github.com/go-task/task/issues/277


internal/execext/exec.go
```diff
@@ -16,17 +16,19 @@ import (
...
+	OneShell bool
+	Stdin    io.Reader
+	Stdout   io.Writer
+	Stderr   io.Writer
 }
 
 var (
 	// ErrNilOptions is returned when a nil options is given
 	ErrNilOptions = errors.New("execext: nil options given")
+    runner *interp.Runner
```
andreynering, line 31: Wrong indentation here. Be sure to `go fmt` the code.

andreynering (commented):
Hi @melivoro, thanks for opening this PR!

I think you have a point that doing this could be beneficial, but the proper way of doing it needs a bit of discussion and would be a bit different IMO.

1) The shell certainly shouldn't  be re-used for every command, but for commands of that task run, because of possible bugs, concurrency issues, etc. This means that the runner shouldn't be global, but declared likely in the [`RunTask`](https://github.com/go-task/task/blob/d5d198411625c638084326dd3a4d21504e9e44fb/task.go#L300) function (not necessarily directly, but maybe through a helper struct. Some refactor of the `execext` package may be needed)
2) I think this probably shouldn't be a CLI flag, but a global option in the Taskfile (like `output:` and `silent:` currently). If this work well and we think it won't cause considerable backward incompatibilities, maybe we could promote this by default in the next major (v4) version.

Sorry for the long text, just want to ensure the best behavior 🙂. If you have questions or need help with the implementation, please let me know.

internal/execext/exec.go
melivoro: Hi @andreynering. I corrected the indentation, thanks for pointing that out! I also added documentation, please have a look. The runner is not global, it acts the following way: oneshell command line option is used as default value for all tasks where oneshell parameter is not set explicitly in a taskfile. Please let me know if you see any possible issues in this implementation, thanks.

andreynering: We certainly can't keep `runner` as a global variable here. This would be reused by different tasks and this would cause many different issues: shared state (a `cd` from one task messing with another task, for example), concurrency issues since the library is not thread safe, etc. That's why scoping the runner within each task run / goroutine is needed, even if shared for commands _within the same task_. Some refactor will be needed in the `interp` package to allow that: I think that `interp.RunCommand` will have to be scoped in a struct to keep the runner as shared prop between runs.

Point 2 of my previous comment is still valid as well: I'd keep `oneshell` as a global Taskfile option only (no CLI argument, no task option). If this works well we could promote it to being a default in v4, so the option is just a temporary thing.

Sorry for the amount if info. If you think this is too much, let me know and I can continue the work once I have the time.

melivoro: 1) I can implement this behaviour without refactoring mvdan/sh project: create new internal package (interpmanager) with mutex-protected map of interps (for every task). What do you think?
2) I agree that it's reasonable to implement oneshell as a global option if you're planning to make it a default behaviour in v4.

PR #651, opened 2022-01-14
Add support for begin/end messages with grouped output

### LINKED ISSUE #647: Grouped output with GitHub Actions friendly begin/end markers
Hi, thanks for the awesome project!

I was wondering if you'd consider supporting a variant of the `group` output, but also emitting a `::group::$TASK` line before the group and `::endgroup::` after it, so that [GitHub Actions will group all the output for a single task and allow it to be collapsed](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines).

For example, if there's a Taskfile like this:

```yaml
# Taskfile.yml
output: github_actions
tasks:
  default:
    cmds:
      - "echo 'Hello, World!'"
```

I'd expect `task default` to emit the following:

```bash
::group::default
Hello, World!
::endgroup::
```

Alternatively (as a non GitHub Actions-specific solution), adding support for custom begin/end markers, e.g.,

```
# Taskfile.yml
output:
  group:
    begin: '::group::%s'
    end: '::endgroup::'
```

I'm happy to put together a PR for this, if you'd consider reviewing it.

janslow:
Fixes #647

This allows CI systems that support grouping (such as with [GitHub Actions's `::group::` command](docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines) and [Azure Devops](docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands)) to collapse all of the logs for a single task, to improve readability of log

## Example

The following Taskfile

```
# Taskfile.yml
version: 3
output:
  group:
    begin: "::group::{{ .TASK }}"
    end: "::endgroup::"
tasks:
  default:
    cmds:
      - "echo 'Hello, World!'"
```

Results in the following output
```bash
$ task
task: [default] echo 'Hello, World!'
::group::default
Hello, World!
::endgroup::
```

See [this GitHub Actions job](https://github.com/janslow/task/runs/4811059609?check_suite_focus=true) for a full example:

image
image

### DIFF
--- a/internal/output/interleaved.go
+++ b/internal/output/interleaved.go
@@ -6,6 +6,6 @@ import (
 
 type Interleaved struct{}
 
-func (Interleaved) WrapWriter(w io.Writer, _ string) io.Writer {
+func (Interleaved) WrapWriter(w io.Writer, _ string, _ Templater) io.Writer {
 	return w
 }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,4 +27,4 @@ jobs:
         run: go build -o ./bin/task -v ./cmd/task
 
       - name: Test
-        run: ./bin/task test
+        run: ./bin/task test --output=group --output-group-begin='::group::{{.TASK}}' --output-group-end='::endgroup::'
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"syscall"
 
+	outputpkg "github.com/go-task/task/v3/internal/output"
 	"github.com/spf13/pflag"
 	"mvdan.cc/sh/v3/syntax"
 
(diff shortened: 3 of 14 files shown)


internal/output/interleaved.go
```diff
@@ -6,6 +6,6 @@ import (
 
 type Interleaved struct{}
```
janslow, line 7: Note, the `Interleaved`/`Prefixed`/`Group` types are no longer used outside of the `output` package (they're now only created via `output.Style.Build()` and only consumed as the `output.Output` interface), so they could be unexported

andreynering:
Hi @janslow,

You did an awesome job here, thanks a lot!

I did some small improvements at b323531dd5f7d0043033124278a19aab4d044995, mostly moving some files around where they better fit.

Issue #390, opened 2020-10-20
Legitimate use-case of multi-level `includes`


LouisAmon:
My company is specialised in infra-as-code and cloud automation.
A typical architecture we usually end up working with is as follows:

```bash
# Client-facing web platform
app/
    # Next.js application
    frontend/
    # Python infra-as-code (Pulumi)
    infra/
# Internal back-office
admin/
    frontend/
    infra/
```

As you can see here, we have 2 Python folders (each with their `virtualenv`) and 2 Next.js folders (using `yarn`)

In order to keep it DRY (Don't Repeat Yourself), I created a `common/taskfiles/python.yml` and a `common/taskfiles/yarn.yml` in my monorepo, and then I include those in the relevant taskfiles.

For example:
```
# app/infra/Taskfile.yml
version: "3"

includes:
  python: ../../common/taskfiles/python.yml

tasks:
  build:
    # I use the included Taskfile here
    deps: [python:install]
    cmds:
      - do something
```
This way, I ensure that every Python install uses the same tooling (checking the Python version, package manager, etc.), and the same thing goes for Next.js (Node version, yarn package manager, etc.) : it's the best way I found to avoid having drift between each install.

That's all fine and dandy, but I also need to coordinate my `infra` and my `frontend` for each of my platforms (`app` and `admin`).

So I end up having a Taskfile for that:
```
# app/Taskfile.yml
versions: "3"

# I'd like to do that, but can't :(
# includes:
#   frontend: ./frontend
#   infra: ./infra

tasks:
  deploy:
    # I'd like to do something like that
    # deps: [frontend:deploy, infra:deploy]

    # Instead, I have to do something like that
    deps: [deploy:frontend, deploy:infra]

  deploy:frontend:
    dir: ./frontend
    cmds:
      - task deploy

  deploy:infra:
    dir: ./infra
    cmds:
      - task deploy
```

The namespacing and log statements end up being messier than they could be, although it does work perfectly fine that way.

The main difficulty here are variables as it becomes kind of messy to know which tasks are called as subtasks (e.g. `task: subtask` and which are called from the shell directly `task subtask`) : variables should be passed differently in each scenario (`vars` vs environment variable).

This issue is the main reason why we're considering Buck or Pants as our build system, so I'd rather post the issue here for the community as it was literally written in the docs:

> Also, for now included Taskfiles can’t include other Taskfiles. This was a deliberate decision to keep use and implementation simple. If you disagree, open an GitHub issue and explain your use case. =)

[...]

PR #656, opened 2022-01-15
Add support for multi-level includes

tylermmorton:
This PR is in response to #390 -- This adds support for multi-level includes by rewiring `read.Taskfile` to be a recursive function that merges Taskfiles in depth first order. It also uses the new `read.ReaderNode` to keep track of included Taskfile parent nodes and checks for include cycles that could cause overflow issues.

I've also added some basic tests to confirm the new functionality. Happy to continue to refine this if any interest is shown in the feature. This would absolutely be a game changer for a lot of the projects I'm involved in.


docs/usage.md
```diff
@@ -8,7 +8,7 @@ The example below allows compiling a Go app and uses [Minify][minify] to concat
 and minify multiple CSS files into a single one.
 
 ```yaml
-version: '3'
+version: "3"
```
andreynering, line 11: This was changed this the whole file, but single quotes are really preferred according to out [styleguide](https://taskfile.dev/#/styleguide).

taskfile/read/taskfile.go
```diff
@@ -170,3 +179,25 @@ func exists(path string) (string, error) {
 
 	return "", fmt.Errorf(`task: No Taskfile found in "%s". Use "task --init" to create a new one`, path)
 }
+
+func checkCircularIncludes(node *ReaderNode) error {
+	if node == nil {
+		return errors.New("failed to check for include cycle: node was nil")
```
andreynering, line 185: We prefix all errors with `task: `.

```suggestion
		return errors.New("task: failed to check for include cycle: node was nil")
```

taskfile/read/taskfile.go
```diff
@@ -170,3 +179,25 @@ func exists(path string) (string, error) {
...
+	}
+	if node.Parent == nil {
+		return errors.New("failed to check for include cycle: node.Parent was nil")
```
andreynering, line 188:
```suggestion
		return errors.New("task: failed to check for include cycle: node.Parent was nil")
```

taskfile/read/taskfile.go
```diff
@@ -170,3 +179,25 @@ func exists(path string) (string, error) {
...
+	}
+	var curNode = node
+	var basePath = filepath.Join(node.Dir, node.Entrypoint)
+	for curNode.Parent != nil {
+		curNode = curNode.Parent
+		curPath := filepath.Join(curNode.Dir, curNode.Entrypoint)
+		if curPath == basePath {
+			return fmt.Errorf("include cycle detected between %s <--> %s",
```
andreynering, line 196:
```suggestion
			return fmt.Errorf("task: include cycle detected between %s <--> %s",
```

testdata/includes_cycle/Taskfile.yml
```diff
@@ -0,0 +1,12 @@
...
+includes:
+  'one': ./one/Taskfile.yml
+
+tasks:
+  default:
+    cmds:
+      - echo "called_dep" > called_dep.txt
+  level1:
+    cmds:
+      - echo "hello level 1"
```
andreynering, line 12: We need to add a final newline to all these test files.

andreynering (commented):
Hi @tylermmorton,

Thanks again for yet another PR!

I haven't tested this one yet, but the code looks good, and I added a few comments for now. Hope I'll find the time to test it soon.

docs/usage.md
tylermmorton: Fixed!

taskfile/read/taskfile.go
```diff
@@ -68,24 +74,27 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
...
 		}
 
-		path, err = exists(path)
-		if err != nil {
-			if includedTask.Optional {
-				return nil
-			}
+		includeReaderNode := &ReaderNode{
+			Dir:        filepath.Dir(path),
+			Entrypoint: filepath.Base(path),
```
andreynering, line 82: Seems that we have the wrong value here when inclusion is by a directory (omits `Taskfile.yml`).

andreynering (commented):
Hi @tylermmorton,

The code looks great, I think this is the right direction. I really appreciate your effort.

I did some basic testing and found a small bug: when you do the inclusion like `foo: ./foo` instead of `foo: ./foo/Taskfile.yml` you get an error.

Reproduce:

- I added a new Taskfile on `docs/foo/Taskfile.yml` as:

```yaml
version: '3'

tasks:
  echo:
    desc: Includes inside includes
    cmds:
      - echo "Includes inside includes"
      - pwd
```

- Referenced it from the `docs/Taskfile.yml` like:

```yaml
includes:
  foo: ./foo
```

And the error is like: `stat /Users/andrey/Developer/andrey/task/foo: no such file or directory`. Note the dir is `task/foo` instead of `task/docs/foo`.

I think the issue is the place pointed in a review comment below. We should have a test case covering this.

The directory where the task _runs_ should also be tested, but this one seems to be working fine in my tests (once `/Taskfile.yml` was added to includes due to the previous issue).

taskfile/read/taskfile.go
tylermmorton: Fixed! I added a check for `exists()` so all default taskfile names can be used 😄

lukepighetti:
Hate to be that guy, but my experience is that big lift PRs that sit stale never land. We're looking for this feature and would love to see this merged.

andreynering (approved):

andreynering:
Thanks again @tylermmorton!

Issue #661, opened 2022-01-23
"default" task from included taskfile should be executed when no task name is provided


ankushg:
I really like the behavior that
> If you omit a task name, “default” will be assumed.

Unfortunately, it doen't seem to play well with included taskfiles.

For example, if I have a root `Taskfile.yml` with the following contents:
```yaml
version: '3'

includes:
  main: .taskfiles/main.yml
```

and `.taskfiles/main.yml`:
```yaml
version: '3'

tasks:
  default:
    - task: build
    - task: assets

  build:
    cmds:
      - go build -v -i main.go

  assets:
    cmds:
      - minify -o public/style.css src/css
```

I would like to run something like `task main`  and have it automatically run `task main:default`

Currently though, I need to explicitly run `task main:default`

tylermmorton:
Hi @ankushg! Thank you for the report. 

This does seem like the intended behavior because of the way we merge included taskfiles into the parent. I think an additional step could be added to check for the existence of `default` in an included taskfile and 'flatten' it, for lack of a better term. 

Like here https://github.com/go-task/task/blob/2373743eacee1d6d38f8f97047d241204bbb5061/taskfile/read/taskfile.go#L116

```go
if(includedTaskfile.Tasks["default"] != nil) {
	t.Tasks[namespace] = includedTaskfile.Tasks["default"]
}
```

I'm interested in how @kerma and @andreynering feel about this.

andreynering:
Hi @ankushg,

I had the impression that this was proposed before, but I couldn't find the related issue in a quick search.

This probably makes sense, but we need to take possible conflicts into account. For example: it's possible that both a `foo` and a `foo:default` tasks exists. In this case, `foo` probably should be preferred instead of `foo:default`.

ghostsquad:
thinking in terms of go, go has `func main() {}` as the "default" entrypoint for an executable file, as long as that file's package is "main". 

I like to look at upstream code/binaries/packages not as a way to "follow the crowd", but as a way to help "preload" intent with what a user may already be used to or expect.

Some other things I see is how does `Make` work? 
https://www.gnu.org/software/make/manual/html_node/Goals.html
https://stackoverflow.com/questions/17834582/run-make-in-each-subdirectory

I kind of see namespaces in `task` as subcommands, and maybe that makes sense for how they should be treated (more or less). There's currently no enforcement of this concept though (afaik), such that you can make a task at the root called `main:default` that essentially "shadows" the included `main:default` task.

[...]

PR #665, opened 2022-01-29
Add support to alias included default task

tylermmorton:
This PR addresses the issue brought up in #661. I've also updated the docs and added test coverage!

Included Taskfiles that declare a `default` task will be aliased automatically to their namespace only if there is not a conflicting task in the parent Taskfile.

```yaml
version: '3'
includes:
  docker: ./DockerTasks.yml
tasks:
  deploy:
    - task: docker # will run docker:default
```

...

```yaml
version: '3'
includes:
  docker: ./DockerTasks.yml
tasks:
  docker: # declaring `docker` will prevent the above behavior
    cmds:
      - task: docker:default # but the default task can still be used
```

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -117,6 +117,30 @@ namespace. So, you'd call `task docs:serve` to run the `serve` task from
 `documentation/Taskfile.yml` or `task docker:build` to run the `build` task
 from the `DockerTasks.yml` file.
 
+Included Taskfiles that declare a `default` task will be aliased automatically
+to their namespace only if there is not a conflicting task in the parent Taskfile.
+
+```yaml
+version: '3'
+includes:
+  docker: ./DockerTasks.yml
+tasks:
+  deploy:
+    - task: docker # will run docker:default
+```
+
+...
+
+```yaml
+version: '3'
+includes:
+  docker: ./DockerTasks.yml
+tasks:
+  docker: # declaring `docker` will prevent the above behavior
+    cmds:
+      - task: docker:default # but the default task can still be used
+```
+
 ### OS-specific Taskfiles
 
 With `version: '2'`, task automatically includes any `Taskfile_{{OS}}.yml`
--- a/taskfile/read/taskfile.go
+++ b/taskfile/read/taskfile.go
@@ -113,6 +113,11 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
 		if err = taskfile.Merge(t, includedTaskfile, namespace); err != nil {
 			return err
 		}
+
+		if t.Tasks[namespace] == nil && includedTaskfile.Tasks["default"] != nil {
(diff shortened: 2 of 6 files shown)


testdata/includes_flatten_default/.gitignore
```diff
@@ -0,0 +1 @@
+file.txt
```
andreynering, line 1: Files should end with a line break.

The same applies to the next two files where they're also missing.

taskfile/read/taskfile.go
```diff
@@ -113,6 +113,11 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
 		if err = taskfile.Merge(t, includedTaskfile, namespace); err != nil {
 			return err
 		}
+
+		if t.Tasks[namespace] == nil && includedTaskfile.Tasks["default"] != nil {
+			t.Tasks[namespace] = includedTaskfile.Tasks["default"]
+		}
```
andreynering, line 119: I think this implementation has unintended side-effects. Tasks assigned here will be duplicated on `task --list`.

I added a `default` task to our `docs/Taskfile.yml` as an example. See the duplicated `docs:default`:

```shell
$ task --list
* clean: 		Cleans temp files and folders
* cli-deps: 		Downloads CLI dependencies
* docs:default: 	Serves the documentation site locally
* docs:default: 	Serves the documentation site locally
* docs:ico: 		Generate favicon.ico from Logo.png
* docs:install: 	Installs docsify to work the on the documentation site
* docs:serve: 		Serves the documentation site locally
* gen-install-script: 	Generate install script using https://github.com/goreleaser/godownloader
* install: 		Installs Task
* lint: 		Runs golint
* mod: 			Downloads and tidy Go modules
* test: 		Runs test suite
* test-release: 	Tests release process without publishing
```

My proposal is that we implement this near the code below, by looking for `foo:default` if a `foo` task is not available:

https://github.com/go-task/task/blob/2373743eacee1d6d38f8f97047d241204bbb5061/variables.go#L25-L28

@tylermmorton What do you think? Do you have a different opinion?

docs/usage.md
```diff
@@ -117,6 +117,30 @@ namespace. So, you'd call `task docs:serve` to run the `serve` task from
...
+Included Taskfiles that declare a `default` task will be aliased automatically
+to their namespace only if there is not a conflicting task in the parent Taskfile.
+
+```yaml
+version: '3'
+includes:
+  docker: ./DockerTasks.yml
+tasks:
+  deploy:
+    - task: docker # will run docker:default
```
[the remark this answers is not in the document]
ghostsquad: inline comments are usually harder to follow, and some IDEs have trouble finding them.
I would also suggest putting comments _above_ the line you are referring to.

andreynering:
Closes in favor of #815

Issue #344, opened 2020-06-12
Add task pre and post operations (before, after)


fairbairn:
Be great if there was a way to add post processing similar in concept to preconditions but a bit more flexible.

The goal is to be able to have pre- and post- operations surround a task sequence.
For example, the pre- would be the setup (much like a test runner), and the post- would be the cleanup.

It's conceivable it can be done with sequential tasks under cmds: , but it gets messy because of the added bash code to track exit codes, shove in variables, add ignore_errors (which you may or may not forget to add or remove at times), and test those to abort follow on tasks but not the cleanup task.

It would be cleaner to have a pre- and post- sequence that could surround the cmd, instead of writing nuance around cmd sequences or adding additional tasks.

Thinking something like...

```
task:
  example:
    before:
      - docker run -d --name server -p 8080:80 nginx
    cmds:
      - curl http://localhost:8080 | awk '{print}/error/{exit 1}'
      - echo "I may or may not get here"
    after:
      - docker rm server -f
```

To extend it further, you could conceivably have different flags that could determine which operations in after should/should not run depending on the eventual exit result (it could even be a number, or simply a success/failure) ...

```
after:
   - echo "Fires only if the cmds sequence failed"
   if_errors: true
   - echo "Fires if no error"
   if_errors: false
   - echo "Fires regardless"
   - echo "Send message to slack if we got a 25, as we know where that came from"
   if_error: 25
   
```

andreynering:
Related issues: https://github.com/go-task/task/issues/141 and maybe https://github.com/go-task/task/issues/275.

stephencheng:
@fairbairn you may want to try: https://github.com/upcmd/up

It simply use composition model for you to assemble pre/post tasks, it is flexible to chain through, loop, do if/else logic etc.

marco-m:
I am not sure I understand the need for the `before` stanza: to me, it is the same as putting its contents as the first line (or lines) in `cmds`, also the expected behavior in case of error (exit with error) is the same.

On the other hand I understand the concept for `after`, it is equivalent to, say, Python `finally`: execute no matter what happened in `cmds`. While the proposal `on_error` from #141 is only in case of error.

My 2 cents is that `after` should be simple (no `if_errors`) and that it should be added after `on_error` is added.

ghostsquad:
I like how ConcourseCI does this. It feels very natural and quite powerful:

https://concourse-ci.org/modifier-and-hook-steps.html

markmsmith:
[...]

Issue #204, opened 2019-05-16
Add a "setup:" setting to allow running setup scripts before any command?


jooola:
I'm looking for a tool that can run `source venv/bin/activate` and run the commands in that same shell. It is not cleanly doable using make, and so I tried Task.

It appears we can't do this using Task. Is there a way to have a workaround or add an option to run all commands in the same shell ?

andreynering:
Hi @jooola,

Can you explain more why you think this'd be useful? Perhaps an example...

jooola:
When developing a Python project within a Virtual environment with all its dependencies, we always need to activate this virtual environment in order to run project tasks like testing, linting, building.

To set up the environment by hand, we run the following commands (where `requirements.txt` lists all the dependencies):
```sh
python3 -m venv venv
source venv/bin/activate
pip -q install --upgrade -r requirements.txt
```

Since the `source` command is altering the shell to activate it, we need to stay in it to execute the next commands. So I suggested to allow to run all the commands within the same shell.

The option should also all to run all the dependencies in the same shell.

Sorry for my previous message, I clearly wasn't describing my problem enough.

smyrman:
You can do this today as a multi-line command, but you need to somehow duplicate the setup of venv to each command, e.g. as a template variable.

jooola:
It would be something like:
```yaml
vars:
  VENV: |-
    test -d venv || python3 -m venv venv
    source venv/bin/activate


tasks:
  setup:
    cmds:
      - |
        {{.VENV}}
        pip -q install --upgrade -r requirements.txt

```

smyrman:
👍 

Remember to also add a `set -e` on of multi-line commands to ensure they exit on error.

```yaml
vars:
  VENV: |-
    test -d venv || python3 -m venv venv
    source venv/bin/activate


tasks:
  setup:
    cmds:
    - |
      set -e
      {{.VENV}}
      pip -q install --upgrade -r requirements.txt
```

smyrman:
> But dependencies inheritance cannot work in this case, that was the point of chaining dependencies. 

No it can't; dependencies, or any other cmd, can never change the environment in task, which is what would be needed for that to work.

At an early stage of task, there was a `"set"` command, that could set the environment within a task, but it was deprecated.

andreynering:
As @smyrman said, multi-line strings should solve this use case. The downside is that you have to concat the setup stage in each needed command.

Maybe we could add a `setup` flag to task/global to setup stuff before all commands?

[...]

Issue #141, opened 2018-10-22
New feature: action on error


edillmann:
Hi,

I'm using go-task to automate simple data integration workflows, and i'am dreaming on a way to execute a task when a task return's an error,

Eric

marco-m:
Hello @andreynering, would you consider this feature for v3 ? I find it quite useful. Do you have any suggested workaround to obtain the same effect with the current `task` ?

Conceptually, I think we are talking about something like `on-error`:

```
tasks:
  foo:
    cmds:
      - cmd1
      - ...
      - cmdN
    on-error:
      - cmdA
      - ...
      - cmdK
```

Am I on the same page @edillmann ?

andreynering:
Hi @marco-m,

Not for v3.0 because this version is late enough (we've been working on it for more than a year).

But for v3.1 or later, yes.

marco-m:
Hello @andreynering, I am happy that v3.3.0 added #327, thanks!
Now, I would like to bring back to your attention this one :-)

ghostsquad:
I like how ConcourseCI does this. It feels very natural and quite powerful:

https://concourse-ci.org/modifier-and-hook-steps.html

andreynering:
IDEA: We could easily add this by adding a new special variable on [`defer:`](https://taskfile.dev/usage/#doing-task-cleanup-with-defer) commands.

We need to think about the best UX, though. It could be `IS_ERROR` (bool), `EXIT_CODE` (int) or maybe just `ERROR` (string).

tennox:
I'd say `EXIT_CODE` is most ~~intuitive~~ well-known :sweat_smile: 

```
build:
  cmds:
    - foo.sh
    - defer: if [[ {{EXIT_CODE}} -ne 0 ]]; notify-send "Build error"
```

It's a bit of a mouthful though, and I'd much prefer something like:
```
build:
  cmds:
    - foo.sh
  on_error:
    - notify-send "Build error"
```

trulede:
Defer with EXIT_CODE has been available for some time now, and addresses this issue.

PR #667, opened 2022-01-29
Add task hooks

tylermmorton:
This PR is in response to a series of issues: #141, #204 and #344. This code adds "hooks" as a feature, with the initial implementation containing 5 hooks. I've added unit tests to cover all of the new behavior, and additional documentation containing examples of each hook!

- `before_all` commands are called before a task starts execution
- `after_all` commands are called after a task completes, success or failure
- `on_success` commands are called when a task completes successfully
- `on_failure` commands are called when a task fails due to an exit code other than 0
- `on_skipped` commands are called when a task is skipped due to status, precondition or checksum

Here is an example of the usage:
```yml
version: '3'

tasks:
  notify:
    vars:
      STATUS: "NONE"
    cmds:
      - curl -d "status={{.STATUS}}" -X POST http://webhook.example.com
      
  test:
    cmds:
      - go test
    hooks:
      on_success:
        - task: notify
          vars:
            STATUS: "PASS" 
      on_failure:
        - task: notify
          vars:
            STATUS: "FAIL"
```

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -653,8 +653,10 @@ There are a series of hooks that can be used to execute commands in response to 
 - `before_all` commands are called before a task starts execution
 - `after_all` commands are called after a task completes, success or failure
 - `on_success` commands are called when a task completes successfully
-- `on_failure` commands are called when a task fails an exit code other than `0`
+- `on_failure` commands are called when a task completes with an exit code other than `0`
 - `on_skipped` commands are called when a task is skipped due to status, precondition or checksum
+
+Hooks are simply event listeners and do not have the ability to control the flow of a task. If a command within a hook fails to execute, the failure will simply be pushed to the verbose logger and the task will continue its execution.
 
 Examples:
 ```yml
@@ -719,6 +721,7 @@ tasks:
 
   skippy2:
     status:
+      - touch status.txt
       - test -f status.txt
     cmds:
       - echo "i won't get here"
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -643,9 +643,89 @@ tasks:
   cleanup: rm -rf tmpdir/
 ```
 
-> NOTE: Due to the nature of how the
+> `NOTE` Due to the nature of how the
 [Go's own `defer` work](https://go.dev/tour/flowcontrol/13), the deferred
 commands are executed in the reverse order if you schedule multiple of them.
+
+## Hooks
+There are a series of hooks that can be used to execute commands in response to an event happening during a task's execution. The available hooks are as follows:
+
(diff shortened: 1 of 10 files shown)


tylermmorton:
I'm running into issues in VS Code where the Taskfile YAML Schema is complaining about the new `hooks` field. Any ideas on how to update this?
image

andreynering:
Hi @tylermmorton, thanks for yet another PR.

I see future in this PR, but I understand we need to discuss what we really want to do here before proceeding.

The idea of wrapping setting like this into `hooks` is probably interesting, making it consistent with each other and easy to add more in the future.

The `on_failure` and `on_skipped` hooks are likely very useful.

The `before_all` and `on_success` feels not needed to me initially, since one could just add more commands to the beginning or end of a task.

In a similar manner, the `after_all` duplicates functionality we recently released with the `defer` keyword, so I'd be inclined to remove this one as well.

Regarding #204 you linked, the idea of a possible `setup:` keyword is that it would be _global_, so it would kinda run before every declared task. We could consider turning that into a kind of hook, but that would means we would have two kind of hooks, global and local (task-specific) ones.

---

This is the kind of feature where the input from user is very important. So if whoever is reading this would like to add thoughts, they're more than welcome!

tylermmorton:
Hey @andreynering, thanks the feedback! :)

I'd like to make the argument to keep `after_all` as a hook _because_ of the way that `defer` works. There's some nuance there that isn't completely apparent to users not familiar with Go. To me, users that are already using other hooks to run commands are naturally going to reach for `after_all` instead of `defer`. 

I agree that `before_all` and `on_success` are mostly redundant hooks and I only added them to give the feature some symmetry. If there's an `on_failure` hook it seems natural to me to have an `on_success` hook as well. Users defining tasks using hooks might appreciate the semantics that even these redundant hooks provide.

I will say to implement and maintain a hook is much easier than other features, as a hook in the current implementation does not control the flow of the or branch any of the task logic. Every new hook definition is roughly 12 lines of code, which could be reduced, especially all of the boilerplate within executor's `hooks.go`. The call to a hook is just a one-liner!

-- 

I just read deeper into #204 and I guess I did not get far enough into the discussion to realize this PR does not cover that 😅  My bad. I do really like the idea of a global `setup` hook that can modify the environment tasks are run in. I'm actually trying to solve this exact problem on one of my projects and the idea of using `source` and a setup script to build the shell environment seemed like a great solution. There's some value in being able to do that within the Taskfile to help keep all of my team's IaC centralized.

I can probably get some code in for this next time I'm able to pull up my IDE. I will take any notes I have to that issue thread instead of this PR.

Issue #591, opened 2021-10-20
Dynamic variables are being evaluated before dependencies are run


srohr-luther:
- Task version: 3.9.0, 3.9.0
- Operating System: EndeavourOS, Windows Server 2019 Datacenter

While designing our build pipeline based on task I stumbled over the following issue. I would expect that any dependencies are executed before the task itself and that dynamic variables can be created from artifacts of those dependent tasks. At least from a user perspective it would make sense I think.

I added an example with preconditions which works as expected and I would expect dynamic variables behaving the same way.

### Example Taskfile showing the issue

```yaml
version: "3"

tasks:
  greeting:
    cmds:
      - echo "Hello World!" > greeting.txt
  default:
    deps: [greeting]
    cmds:
      - echo "$GREETING"
    env:
      GREETING:
        sh: head -n 1 greeting.txt
```
### Expected output

```txt
task: [greeting] echo "Hello World!" > greeting.txt
task: [default] echo "$GREETING"
Hello World!
```

### Actual output

```txt
head: cannot open 'greeting.txt' for reading: No such file or directory
task: Command "head -n 1 greeting.txt" failed: exit status 1
```

### Example Taskfile with preconditions

```yaml
version: "3"

tasks:
  greeting:
    cmds:
      - echo "Hello World!" > greeting.txt
  default:
    deps: [greeting]
    cmds:
      - cat greeting.txt
    preconditions:
      - test -f greeting.txt
```

andreynering:
Hi @srohr-luther, thanks for opening this issue!

This work as is because variables can be used in any part of a task, including the dependencies, so they're evaluated first.

But your use case makes sense. I'll keep this issue open for future consideration, but we likely won't change that soon.

trulede:
Closing stale issue.

If this issue still impacts your work, please collect an example/logs from the most recent version of Task, and reopen this issue.

Issue #582, opened 2021-10-12
Add posibility to include by dotenv


antonio-antuan:
Hi!
I want to include taskfiles by dotenv var. For now I see that it is not possible because dotenv read when files already included...

.env:
```
TASKFILE_PATH=./child
```

Taskfile.yml:
```
version: '3'

dotenv:
  - .env

includes:
  child: 
    taskfile: $TASKFILE_PATH/Taskfile2.yml
    optional: true
    dir: $TASKFILE_PATH
```

Taskfile2.yml:
```
version: '3'

tasks:
  default:
    cmds:
      - echo child
```

Also, it is not possible for now to include with env variable:
Taskfile.yml:
```
version: '3'

env:
  TASKFILE_PATH: Taskfile2.yml

includes:
  child: 
    taskfile: $TASKFILE_PATH
    optional: true
```

trulede:
This really is not possible.

And alternative would be to use a an envar or var at the CLI, and the solution being developed for #2443 (PR #2445). Its a remarkably confusing problem, and possibly has no solution. The processing is: Read (imports done here), compile (vars, envars resolved here) and then execute.

We can track the issue under 2443.

Issue #535, opened 2021-07-28
Vars are not interpolated before deps in a task


weeping-somnambulist:
- Task version: 3.6.0
- Operating System: OpenSUSE Tumbleweed 

I believe this is a bug as it is not expected behavior from a user's point of view. Variables that use the shell function, e.g. ```sh: foo```, run *before* the dependencies in a task. This is, in my opinion, not the expected behavior for the user and is undesirable for most  use cases, as it prevents information from being passed from the depended task to the dependent task. 

I am trying to pass information from a dependency to a task. I need the dependency to generate a container image ID then use that image ID for subsequent tasks. It should be noted that not all tasks in this Taskfile create a container image, so this is not something that I would prefer to do globally.  

If there is a better way to solve this problem please let me know.

### Example Taskfile showing the issue

```yaml
version: '3'

tasks:

  generate-bar:
    cmds:
      - echo foo > bar
    generates:
      - bar

  default:
    deps: [generate-bar]
    vars:
      BAZ:
        sh: cat bar 
    cmds:
      - echo {{.BAZ}}
```

### Output

```bash
> task                        
cat: bar: No such file or directory
task: Command "cat bar" failed: exit status 1
```

andreynering:
Hi @SwitchedToGitlab,

This is a side-effect of how caching of variables work, and it's not trivial to fix, specially without introducing breaking changes.

I'll keep this open for consideration if we have another vars refactor in the future.

SwitchedToGitlab:
So the issue that I'm having is that I effectively cannot pass information from one task to another using any method. Is there any way to pass information from one task to another? If there is I don't see it, and to me that's a critical flaw in design if I may say so. 

My specific use case is that I am building a container image with Task, and I need the container's image ID to run operations on it, such as mounting it's file system and working with binaries installed on the container. As it stands I have no way to instantiate the container and then pass the ID to a task. The only way that I have to do this is to instantiate a container as a global var using SH, but this is less than ideal as it creates a container for every operation whether it's needed or not. 


Thoughts? I'm of the opinion that the functionality detailed in this issue is critical if Task is going to replace Make.

amancevice:
Are you defining the var on the called task? If so i think you want to take that out.

I find the order of preference to interpolation confusing. It feels like defining a var at the task level ought to behave as a default, but I think actually will override anything you try to pass through.

Eg

```yaml
tasks:
  caller:
    cmds:
      - task: callee
        vars:
          MY_VAR: fizz

[...]

PR #668, opened 2022-02-03
Add setup hook

tylermmorton:
This PR is in response to issues described in #204. I've modified `RunCommand` to accept a slice of strings to execute in the same shell. Each string in the slice is run through `syntax.Parser` and the resulting statements are appended to a single `syntax.File` as if it were a small script. 

This ultimately results in all commands under `setup` being run before every single command, the advantage here is that they are executed in the same shell.
```yml
version: '3'

setup:
  - source .env
  - export DIR=$(pwd)

tasks:
  test:
    cmds:
      - echo $DIR
```

This is a rough draft for this feature. Personally I feel like there could be some performance implications with running setup commands before _every single_ command in a task. I'm open to suggestions/feedback from the community!

It might make more sense to have all commands under a task run in the same shell by using the same strategy applied here, ie appending all commands into one script and passing them all at once to the shell.

### DIFF
--- a/internal/compiler/v2/compiler_v2.go
+++ b/internal/compiler/v2/compiler_v2.go
@@ -104,12 +104,12 @@ func (c *CompilerV2) HandleDynamicVar(v taskfile.Var, _ string) (string, error) 
 
 	var stdout bytes.Buffer
 	opts := &execext.RunCommandOptions{
-		Command: v.Sh,
-		Stdout:  &stdout,
-		Stderr:  c.Logger.Stderr,
+		Commands: []string{v.Sh},
+		Stdout:   &stdout,
+		Stderr:   c.Logger.Stderr,
 	}
 	if err := execext.RunCommand(context.Background(), opts); err != nil {
-		return "", fmt.Errorf(`task: Command "%s" failed: %s`, opts.Command, err)
+		return "", fmt.Errorf(`task: Command "%s" failed: %s`, v.Sh, err)
 	}
 
 	// Trim a single trailing newline from the result to make most command
--- a/internal/compiler/v3/compiler_v3.go
+++ b/internal/compiler/v3/compiler_v3.go
@@ -130,13 +130,13 @@ func (c *CompilerV3) HandleDynamicVar(v taskfile.Var, dir string) (string, error
 
 	var stdout bytes.Buffer
 	opts := &execext.RunCommandOptions{
-		Command: v.Sh,
-		Dir:     dir,
-		Stdout:  &stdout,
-		Stderr:  c.Logger.Stderr,
+		Commands: []string{v.Sh},
+		Dir:      dir,
+		Stdout:   &stdout,
+		Stderr:   c.Logger.Stderr,
 	}
 	if err := execext.RunCommand(context.Background(), opts); err != nil {
-		return "", fmt.Errorf(`task: Command "%s" failed: %s`, opts.Command, err)
+		return "", fmt.Errorf(`task: Command "%s" failed: %s`, v.Sh, err)
 	}
 
 	// Trim a single trailing newline from the result to make most command
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -16,12 +16,12 @@ import (
 
 // RunCommandOptions is the options for the RunCommand func
 type RunCommandOptions struct {
-	Command string
-	Dir     string
-	Env     []string
-	Stdin   io.Reader
-	Stdout  io.Writer
-	Stderr  io.Writer
+	Commands []string
+	Dir      string
+	Env      []string
+	Stdin    io.Reader
+	Stdout   io.Writer
+	Stderr   io.Writer
 }
 
 var (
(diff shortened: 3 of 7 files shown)


andreynering:
Hi @tylermmorton,

> Personally I feel like there could be some performance implications with running setup commands before every single command in a task. I'm open to suggestions/feedback from the community!

Yeah. There's a refactor I have in mind that would benefit both this `setup` feature and another thing some users have been requesting: that commands in a single task would share state:

```yaml
version: '3'

cmds:
  default:
    - export FOO=bar
    - echo $FOO # has access to previous declared variables
```

For this to be possible, I think we need to refactor the [execext](https://github.com/go-task/task/blob/master/internal/execext/exec.go) package so the `mvdan.cc/sh/v3/interp.Runner` struct is stored in another struct in our app. This way we would be able to do successive calls to `runner.Run` so the state is kept between them.

For `setup` specifically I think we would be able to run the setup once and cache the result. For the task, we could just clone/copy this struct/state before running the task commands. We may need to contribute to mvdan/sh to have a clone function.

My proposal is to do this refactor alone on its own PR as a POC, and once in master we should work on `setup` and this state behavior on their own PRs.

ghostsquad:
I don't agree with this being an improvement. I think this would result in more confusion, especially given the precedent that no other tool combines cmd lists into a single shell. Not Make, not GitHub actions, not Docker, etc.

ghostsquad:
@tylermmorton what I would like to understand in greater clarity is the use case that makes combining the list solves. As you may know, you can literally just write a small multiline script in yaml.

https://yaml-multiline.info/

tylermmorton:
@ghostsquad You bring up a good point that no other tool runs command lists in a single shell. I'm for keeping the precedent consistent across tools. There is also the problem that `cmds` and `deps` would then have diverging functionality if this change were to go in.

The use case where you can source an `.env` or `.bashrc`-like file containing variable and function definitions to be used by all tasks was my original inspiration for this PR.

As a specific example, in a monorepo I'm working in, the goal is to have a script that can resolve relative paths and export them as a series of environment variables. Now that I'm thinking about it, perhaps this could be done with the [`dotenv` option](https://taskfile.dev/#/usage?id=env-files)... I'll have to try

I'm willing to close this and move on. Perhaps a better action item coming out of this is to document solutions for use cases mentioned in #204

ghostsquad:
@tylermmorton ok, sounds good. Ya, try the `dotenv` functionality described here: https://taskfile.dev/#/usage?id=env-files

And if you run into a situation that isn't working, please open a new issue here.

tylermmorton:
@ghostsquad So I just tried the `dotenv` feature and realized why it wasn't working for me at first. Here's a use case:

build.env
```
REPO_ROOT=$(git rev-parse --show-toplevel)
```

Taskfile.yml
```yml
version: 3

dotenv:
  - ./build.env

tasks:
  test:
    cmds:
      - echo $REPO_ROOT
```
In this case, `test` echos: "$(git rev-parse --show-toplevel)"

Obviously .env files are not shell scripts so it makes sense why this won't resolve to anything besides the literal string. But there is a workaround that isn't 100% apparent when reading the usage docs under Environment Variables.

env.taskfile.yml
```yml
version: 3

env:
  REPO_ROOT:
    sh: git rev-parse --show-toplevel
```

Taskfile.yml
```yml
version: 3

includes:
  env: ./env.taskfile.yml

tasks:
  test:
    cmds:
      - echo $REPO_ROOT
```

This works as you would expect and feels like a good strategy.  However, relying on the environment variables here to declare additional variables does not work:

env.taskfile.yml
```yml
version: 3

env:
  REPO_ROOT:
    sh: git rev-parse --show-toplevel
  PROJECT1_ROOT:
    sh: $REPO_ROOT/project1
```

Taskfile.yml
```yml
version: 3

includes:
  env: ./env.taskfile.yml

tasks:
  test:
    cmds:
      - echo $PROJECT1_ROOT
```

Output:
```console
stat /project1: no such file or directory
task: Command "$REPO_ROOT/project1" failed: exit status 127
```

I have not yet dug into the task source to figure out whats going on here. Curious to know your thoughts.

ghostsquad:
I'm not totally sure what's going on in V3 here. My plan in v4 is to lazily evaluate variables. Essentially waiting until they are _used_ before discovering (and caching) the value. This would allow `task` to produce the correct order without the user having to think about it. [jsonnet](https://jsonnet.org/articles/design.html) does this. It is an incredibly powerful feature I hope I can bring to `task`.

tylermmorton:
Lazy loading does sound like the solution to a lot of problems. (#535, #582, #591) Do we have an issue created where we can discuss the feature and cross reference issues? I'm down to help design/implement/test it if needed.

ghostsquad:
I'm working on this in v4. I don't have anything yet that's ready for collaborative work yet.

ghostsquad:
> Do we have an issue or discussion created where we can discuss the feature and cross reference issues?

The discussion is #694 - but unfortunately, when you mention a discussion from an issue (or vice versa), it only makes a link in the comment, not a reference in the discussion (like it works in issues).

PR #677, opened 2022-02-23
Add vars to included taskfiles

dballweg:
When including task files allow for specifying variables that should be global variables in the included files.

I was looking for a way to do this so that I can have a template Taskfile repo and include those in my projects.
The included file will have a set of tasks that are our standard tasks for the type of project (terraform, docker image, etc).  These tasks will all need to know which environment they are in.  Included taskfiles already account for directory, but there isn't a way to pass variables to the included taskfile.

The taskfile in the root of the project repo will have this for the includes:
```yaml
includes:
  <account-profile>-tf:
    taskfile: Taskfiles/Taskfile-tf.yml
    dir: terraform/<account-profile>
    vars: { AWS_PROFILE: <account-profile> }
```


ghostsquad:
Do the variables in the parent Taskfile not get plumbed into included files?

taskfile/read/taskfile.go
```diff
@@ -109,6 +110,12 @@ func Taskfile(dir string, entrypoint string) (*taskfile.Taskfile, error) {
 			}
 
 			for _, task := range includedTaskfile.Tasks {
+				if includedTask.Vars != nil {
+					if task.Vars == nil {
+						task.Vars = &taskfile.Vars{}
+					}
+					task.Vars.Merge(includedTask.Vars)
```
ghostsquad, line 117: I think the parent task file vars should take precedence over the included vars. Also, it might be useful to include the parent vars in the included.

```go
included.Vars.Merge(task.Vars)
task.Vars.Merge(included.Vars)
```

This behavior should be extracted to a separate function and have a couple tests written for it as well. This will also document what we expect to happen, and allow us to write some scenarios to see if we can find edge cases that might behave strangely.

ghostsquad (changes requested):

taskfile/read/taskfile.go
dballweg: I was thinking that the expected behavior should be that if there are vars in the included taskfile then they should be used unless overridden by the vars specified to the include.  I don't think that root taskfile vars should be applied because if you wanted them to override the included files vars you would put them in the vars map on the include statement.
I will go ahead and work up some tests. I wanted to get this posted to see if there is any opposition in the idea of doing this.

dballweg: I added a test to the root tests to show how it works and how I expect it to work.

dballweg: I can help with other tests if you let me know the requirements that you are looking for and where you want them to go

ghostsquad: Oh, you know, I think I missed that this _adds_ vars to the `include`. All we need is for the included file to have access to the parent vars, right?

dballweg: If you look at the test I added it shows off the functionality that I'm after. I want to set the same variable to different values for each include

dballweg: @ghostsquad  Is there anything else that is needed to get this considered?  I would like to start using this if possible.

pascal-hofmann: > Oh, you know, I think I missed that this _adds_ vars to the `include`. All we need is for the included file to have access to the parent vars, right?

I think the way the PR implements it is much more useful. This way you can properly re-use tasks with a different set of variables.

It would be great to have this!

andreynering:
Hi @dballweg, and thanks for opening this PR!

I'll take a deeper look once I find the time for it.

Regarding the precedence discussion above, this is indeed tricky. For one side I see that people expect the ones declared in the inclusion to be the most significant.

But for the other side, for a matter of consistency, it would make sense to require the included Taskfile to have something like this:

```yaml
MY_VAR: '{{default "foobar" .MY_VAR}}'
```

See: https://go-task.github.io/slim-sprig/defaults.html

This would be consistent with how calls to tasks and tasks work: the task also has the control of what is definitive, and need to use `default` if it wants to be overridable.

This is not a final decision, but I'm inclined to keep the consistency as mentioned. Open for opinions...

dballweg:
> Hi @dballweg, and thanks for opening this PR!
> 
> I'll take a deeper look once I find the time for it.
> 
> Regarding the precedence discussion above, this is indeed tricky. For one side I see that people expect the ones declared in the inclusion to be the most significant.
> 
> But for the other side, for a matter of consistency, it would make sense to require the included Taskfile to have something like this:
> 
> ```yaml
> MY_VAR: '{{default "foobar" .MY_VAR}}'
> ```
> 
> See: https://go-task.github.io/slim-sprig/defaults.html
> 
> This would be consistent with how calls to tasks and tasks work: the task also has the control of what is definitive, and need to use `default` if it wants to be overridable.
> 
> This is not a final decision, but I'm inclined to keep the consistency as mentioned. Open for opinions...

The thing I am after is being able to create a "Terraform" template taskfile.  I want to include this template multiple times with different variables for each.  

currently we have a bigger file that has plan-env1 apply-env1 plan-env2 apply-env2 plan-env3 apply-env3 and so on.  I want to be able to include the template file and do `task env1:plan` `task env2:plan` etc. 

This way no matter how many tasks are in the template I don't have to triplicate them in the root taskfile.  I just include the template taskfile with the env var passed down to it. then every task in the template is automatically triplicated/quadruplicated/etc however many times I include the template taskfile.

ghostsquad:
I can see allowing a `vars` field on includes, in order to be explicit about default values that the included Taskfile will see.

dballweg:
> I can see allowing a `vars` field on includes, in order to be explicit about default values that the included Taskfile will see.

That is the spirit of this PR. I added that field on the includes and made it so that the vars in that field overwrite the ones in the included taskfile.  I was just thinking if you wanted we could rename that to "overrides" or something to explicit that these vars override the vars in the included taskfile.

ghostsquad:
> > I can see allowing a `vars` field on includes, in order to be explicit about default values that the included Taskfile will see.
> 
> That is the spirit of this PR. I added that field on the includes and made it so that the vars in that field overwrite the ones in the included taskfile. I was just thinking if you wanted we could rename that to "overrides" or something to be explicit that these vars override the vars in the included taskfile.

I don't think it would be an override. What it should do is act like VAR passing in any other way. Allowing you to declare your intention to be absolute about the value, or declare it as a default with an override, e.g. `{{ .Ham | Default "Bone"}}`.

taskfile/read/taskfile.go
ghostsquad: The way `includes` currently reads, because that it's a `namespace` (tasks in an includes are automatically prefixed), the `included` Taskfile is "more specific" than the parent task file. To quote the documentation:

> When doing interpolation of variables, Task will look for the below. They are listed below in order of importance (e.g. most important first):

    1. Variables declared in the task definition
    2. Variables given while calling a task from another (See [Calling another task](https://taskfile.dev/#/usage?id=calling-another-task) above)
    3. Global variables (those declared in the vars: option in the Taskfile)
    4. Environment variables
    
An exercise in appending this would be:

```diff
    1. Variables declared in the task definition
    2. Variables given while calling a task from another (See [Calling another task](https://taskfile.dev/#/usage?id=calling-another-task) above)
+   3. Global Vars within the _included_ Taskfiles
    4. Global variables (those declared in the vars: option in the Taskfile)
    5. Environment variables
```

The more I think about this, the current mental model of Task is such of passing vars from high to low. The highest point is the parent Taskfile.

If I'm reading `merge` correctly, then it really is an override.

ghostsquad:
> Hi @dballweg, and thanks for opening this PR!
> 
> I'll take a deeper look once I find the time for it.
> 
> Regarding the precedence discussion above, this is indeed tricky. For one side I see that people expect the ones declared in the inclusion to be the most significant.
> 
> But for the other side, for a matter of consistency, it would make sense to require the included Taskfile to have something like this:
> 
> ```yaml
> MY_VAR: '{{default "foobar" .MY_VAR}}'
> ```
> 
> See: https://go-task.github.io/slim-sprig/defaults.html
> 
> This would be consistent with how calls to tasks and tasks work: the task also has the control of what is definitive, and need to use `default` if it wants to be overridable.
> 
> This is not a final decision, but I'm inclined to keep the consistency as mentioned. Open for opinions...

I am in strong agreement with this. Includes taskfiles are _more_ specific than the parent. Values flow from least specific to most specific. 

@dballweg to make sure I understand your use case, essentially you want something like:

> `Taskfile.terraform.yml`
```yaml
vars: 
   EXTRA_VARS: '{{.EXTRA_VARS | default ""}' 
   WORKSPACE: '{{.WORKSPACE | default "test" }}'
   
tasks:
   workspace:select:
      cmds:
         - terraform workspace select {{.WORKSPACE}}
         
   plan:
      deps: ["workspace:select"]
      cmds:
         - terraform plan {{.EXTRA_VARS}}
```

> `Taskfile.yml`
```yaml
vars: 
   EXTRA_VARS: '-compact-warnings'
   
includes:
   tf:dev: ./Taskfile.terraform.yml
   tf:staging:
     taskfile: ./Taskfile.terraform.yml
     vars:
       WORKSPACE: "staging"
```

such that the following command:

```
task tf:staging:workspace:select
```

runs

```
terraform workspace select staging
```

Is this accurate?

andreynering:
Hi again @dballweg, and thanks again for working on this feature! 😄 

Also thanks @ghostsquad for the insigths.

I just merged with further improvements + documentations + CHANGELOG at de098434673a34f18085987cb4765e7472620ab9.

I advice you (and anyone else interested) to test this in master before I ship (in a week maybe?) to see if it is working as expected.

I went with the approach of having the Taskfile have more priority over the inclusion, requiring the `MY_VAR: '{{.MY_VAR | default "my-default-value"}}'` trick as mentioned above.

PR #679, opened 2022-02-25
completion: zsh: overhaul and sync to current flags

philpennock:
* List all current option flags
* Provide descriptions for every flag
* Pass the `task -l` descriptions as descriptions for the task completions
  + The prior logic had 4 invocations of sed and 1 of awk, and only kept the task name
  + Do all filtering in zsh without forking (except for `task` itself)
* When `--taskfile` is used, complete tasks from _that_ file
  + And otherwise, enable completions if only the `.dist` variant files are present
* Ensure mutually exclusive options preclude each other
  + the `+ '(groupname)'` clause defines this
* Fix `--dir` to take directories, not files

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -1,25 +1,60 @@
 #compdef task
 
-# Listing commands from Taskfile.yml
-function __list() {
-    local -a scripts
+local context state state_descr line
+typeset -A opt_args
 
-    if [ -f Taskfile.yml ] || [ -f Taskfile.yaml ]; then
-        scripts=($(task -l | sed '1d' | sed 's/^\* //' | awk '{ print $1 }' | sed 's/:$//' | sed 's/:/\\:/g'))
-        _describe 'script' scripts
+# Listing commands from Taskfile.yml
+function __task_list() {
+    local -a scripts cmd
+    local -i enabled=0
+    local taskfile item task desc
+
+    cmd=(task)
+    taskfile="${(v)opt_args[(i)-t|--taskfile]}"
+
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @philpennock, thanks for opening this PR!

> Has there been any consideration to adopting use of spf13/cobra? That has native completion support.

I didn't know it supports completion. Given that's the case, it could be interesting, indeed, something we could consider using.

In the meantime this PR is still useful, so still merging. 🙂

philpennock:
1. cobra carries a lot of dependencies; there is muesli/coral which is a fork with far fewer dependencies (I haven't used it yet, just know it exists)
2. cobra targets bash completion and zsh completion is normally via the bashcompinit compatibility layer, IIRC (but I haven't looked recently, so my knowledge might be stale).  What I normally find is I'm frustrated because of the limit hooks to plug things in to supply data fields and am hamstrung by the completion model.

My stance is that using cobra is a good way to get minimally acceptable completion for free, but not good completion.

ghostsquad:
> 1. cobra carries a lot of dependencies; there is muesli/coral which is a fork with far fewer dependencies (I haven't used it yet, just know it exists)
> 2. cobra targets bash completion and zsh completion is normally via the bashcompinit compatibility layer, IIRC (but I haven't looked recently, so my knowledge might be stale).  What I normally find is I'm frustrated because of the limit hooks to plug things in to supply data fields and am hamstrung by the completion model.
> 
> My stance is that using cobra is a good way to get minimally acceptable completion for free, but not good completion.

Good insights. Thanks!

PR #682, opened 2022-03-06
fix: image file path

tokizuoh:
Thank you for the best product :)
The file path was incorrect and the image was not displayed, so I fixed it.

| Before | After |
| --- | --- |
| スクリーンショット 2022-03-06 18 38 34 |  スクリーンショット 2022-03-06 18 39 03 |
| https://github.com/go-task/task/tree/master/docs | https://github.com/tokizuoh/task/tree/master/docs |

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 # Task
 
 <div align="center">
-  <img id="logo" src="/Logo.png" height="250px" width="250px" />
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @tokizuoh, thanks for opening this PR.

This proposal has came before, but actually it is as is by design.

What happens is that the documentation is supposed to be viewed on the site: https://taskfile.dev

Applying this change would break the image on the site.

PR #691, opened 2022-03-21
Add capability to print only task names

ardnew:
#310 was closed recently in favor of the much better shell-completion refactor of #671.

However, #310 included useful functionality that has been lost in the process – specifically the capability to list only task names (without descriptions and without `*` decorations), which, like @shilangyu pointed out, is useful for more than just shell completions; IDE integration, CI/CD tooling, and SCM/container tooling to name a few.

 - This PR reintroduces that capability without any changes to the shell-completion scripts. 
 - The capability was modified to support the new `--list-all, -a` flag such that: 
 	- `task --list --silent` (`task -ls`) will print only task names that have a description
 	- `task --list-all --silent` (`task -as`) will print all task names
 - The review comments from the previous PR have also been addressed.

Note that if this is merged, consider updating the shell-completion scripts to eliminate the `awk` dependency. The scripts can use this functionality to obtain the desired format.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -158,10 +158,12 @@ func main() {
 
 		OutputStyle: output,
 	}
-	if list && silent {
-		e.PrintTaskNames()
+
+	if (list || listAll) && silent {
+		e.ListTaskNames(listAll)
 		return
 	}
+
 	if err := e.Setup(); err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


andreynering:
Hi @ardnew,

Tests were failing, but I managed to fix them at 0f4621fb02ad8c6fdc6ed12e5dc04ee38b63bfac.

Thanks for a lot contributing!

PR #718, opened 2022-05-07
Fix broken completion when no taskfile is found

budimanjojo:
When you are in a directory where there is no `Taskfile.yaml`, the completion will go crazy and keep adding new line on every `Tab` keypress.
```
 task task: No Taskfile found in "/mnt". Use "task --init" to create a new one
tatask task: No Taskfile found in "/mnt". Use "task --init" to create a new one
tatask task: No Taskfile found in "/mnt". Use "task --init" to create a new one
tatask task: No Taskfile found in "/mnt". Use "task --init" to create a new one
tatask task: No Taskfile found in "/mnt". Use "task --init" to create a new one
tatask -l
```
This will silence the error when there's nothing to complete. Of course it will still show completion when there is a `Taskfile.yaml` file.


andreynering:
Thanks @budimanjojo!

PR #725, opened 2022-05-11
Clean up SVG logo markup

sisp:
I've done some cleanup of the SVG logo markup. The paths were a bit chaotic and redundant, there were some unneeded attributes, and the viewbox height was imprecise. I've also beautified the formatting.


andreynering:
Thanks @sisp!

I exported the original from [Canva](https://www.canva.com/), which explains why it was like that.

PR #738, opened 2022-05-22
Make progname easier to configure in task.{bash,fish}

carlsmedstad:
As task is a very generic name that conflicts with for example
taskwarrior, some packagers might choose to change it. See my package in
the AUR: https://aur.archlinux.org/packages/go-task.

This change makes it easier to configure the name for downstream
packagers.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 indent_style = tab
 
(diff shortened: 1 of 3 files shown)


ghostsquad (approved):

carlsmedstad:
Made similar changes for fish completion.

andreynering:
Thanks @carlsmedstad!

PR #755, opened 2022-06-02
Add --exit-code (-x) flag to enable carrying error codes from task cmds

BrunoDelor:
Hello,

I recently found the need to get the return code that was returned from a failed command from a task's cmds. After investigating quickly it appeared to me that the task tool would always return with status code 1 when an error occurred during a task.

This PR proposes an additional command flag (--carry), false by default, that will carry the error code, if any, over to the calling shell.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -68,6 +68,7 @@ func main() {
 		silent      bool
 		dry         bool
 		summary     bool
+		carryErr    bool
 		parallel    bool
 		concurrency int
 		dir         string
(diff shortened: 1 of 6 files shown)


ghostsquad:
Please add some tests to assert on this functionality.

andreynering:
Hi @BrunoDelor, thanks for your PR!

Given this would be a pretty small breaking change, perhaps we could even get rid of the flag and always return the right exit code? Do you guys have an opinion on this?

Another question: does other tools like Make does this as well?

mgbowman:
Always returning the right exit code seems logical to me.

Looks like `make` only uses 0, 1 or 2

``` 
$ man make

EXIT STATUS
       GNU make exits with a status of zero if all makefiles were successfully parsed and no targets that were built failed.
       A status of one will be returned if the -q flag was used and make determines that a target needs to be rebuilt.
       A status of two will be returned if any errors were encountered.
```

ghostsquad:
I think maybe this flag could be renamed too. I've seen it described as exit code pass thru.

Heroku uses `--exit-code` as the flag and describes it as allowing the exit code of the sub process to "pass thru" the main/parent process.

https://github.com/heroku/heroku-run

I'm trying to recall specifically where else I've seen this.

BrunoDelor:
This sounds good to me, I'll push a change for -x --exit-code if this is agreed upon

andreynering:
@BrunoDelor Thanks for contributing! ❤️

PR #479, opened 2021-04-28
Fix signal handling for subprocess

marco-m-pix4d:
hello @andreynering, the fix is quite simple :-D, what is complicated and worthwhile to spend some time to understand is the first commit a496a75, that introduces extensive tests to reproduce the behavior.

I realized that to fix this problem, it is not possible to tweak the timeout of mvdan.cc/sh, we must be more drastic and don't pass to it any cancellable context. The problem that I observed is the following (reproduced in the tests):

- As explained in  #458, when hitting CTRL-C, the terminal driver sends a SIGINT to the foreground process group, which is composed of Task and of all the subprocesses that Task has spawned via mvdan.cc/sh.
- Task receives the signal and forwards it to mvdan.cc/sh, which _immediately_ sends it to the subprocess.
- The result is that the subprocess receives always 2 signals: one from the process group, and one from Task.
- Depending on how the program of the subprocess deals with interrupts, this can be unnoticeable or cause immediate termination, which is the case of Terraform.
- Terraform on reception of the first signals enters a cleanup phase, but if it receives a second signal, it will terminate immediately the cleanup, leaving a broken state.
- Due to the signal amplification explained above, Terraform always receives two signals when driven from Task...

I suggest to checkout a496a75, run
```
task install && go test -run '^TestSignalSentToProcessGroup' -count=1 -v=1
```
and get acquainted with the (expected) failures.

Then, checkout the tip of the branch and run the same tests, to validate that everything is fixed.

There are some things that are still missing: testing what happens on Windows and deciding how far propagate in the Task functions the fact that the context should not even passed. I mean: `func RunCommand(ctx context.Context, opts *RunCommandOptions)` is now lying, since the context will not be passed to anything.

As a reference, here are the failing tests on Linux:

```
task install && go test -run '^TestSignalSentToProcessGroup' -count=1 -v=1
task: [install] go install -v -ldflags="-w -s -X main.version=a496a75b05" ./cmd/task
github.com/go-task/task/v3/internal/execext
=== RUN   TestSignalSentToProcessGroup
=== RUN   TestSignalSentToProcessGroup/child_simulates_terraform:_receives_1_sigint_and_does_cleanup
    unix_test.go:174: 
        wanted but not found:
        [sleepit: cleanup done
         task: Failed to run task "default": exit status 3
        ]
    unix_test.go:179: 
        unwanted but found:
        [sleepit: got signal=interrupt count=2
         sleepit: cleanup canceled
         task: Failed to run task "default": exit status 4
        ]
[...]

PR #769, opened 2022-06-12
Fix behavior of interrupt (SIGINT, SIGTERM) signals

### LINKED ISSUE #728: Ctx/Signal Handling still not totally correct
#479 was recently merged (associated issue: #458), but I realized that you don't want to give `mvdan/sh` an arbitrary `context.Background()`, what we need to do is give it a new context, and forward the cancellation from the parent _IF_ that cancellation isn't coming from a signal.

I use github.com/oklog/run and it's associated SignalHandler in the CLI's I make, and I think this can be improved with the following:

https://github.com/go-task/task/blob/367c0b38a6787bebf6ee5af864be1574ad40b24a/cmd/task/task.go#L205-L220

more or less, this can be replaced with

```go
var group run.Group
group.Add(run.SignalHandler(ctx, syscall.SIGINT, syscall.SIGTERM))
group.Add(func() error {
	if status {
		return e.Status(ctx, calls...)
	}

	return e.Run(ctx, calls...)
}, func(err error) {
	cancel()
})

if err := group.Run(); err != nil {
	e.Logger.Errf(logger.Red, "%v", err)
	os.Exit(1)
}
```

And then this would be updated:

https://github.com/go-task/task/blob/367c0b38a6787bebf6ee5af864be1574ad40b24a/internal/execext/exec.go#L59-L67

with

```go
subCtx := context.Background()
subCtx, cancel := context.WithCancel(subCtx)

var group run.Group
group.Add(func() error {
	return r.Run(subCtx, p)
}, func(err error) {
	if cancel != nil {
		cancel()
	}
})

group.Add(func() error {
	select {
	case <-ctx.Done():
		// If we receive a SIGINT, let signal propagation handle the cancellation
		// This is to avoid the problem presented in: https://github.com/go-task/task/issues/458
		if !errors.Is(ctx.Err(), run.SignalError{}) {
			cancel()
		}
	case <-subCtx.Done():
		cancel()
	}
	return nil
}, func(err error) {
	cancel()
})
return group.Run()
```

mentions to get some eyes to sanity check this thinking:
@andreynering @aliculPix4D @marco-m-pix4d

andreynering:
Task will now give time for the processes running to do cleanup work

Ref #458
Ref #479
Fixes #728

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ dist/
 # exuberant ctags
 tags
 
-/bin
+/bin/*
+!/bin/.keep
 /testdata/vars/v1
 /tmp
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fix behavior of interrupt (SIGINT, SIGTERM) signals. Task will now give time
+  for the processes running to do cleanup work
+  ([#458](https://github.com/go-task/task/issues/458), [#479](https://github.com/go-task/task/pull/479), [#728](https://github.com/go-task/task/issues/728)).
 - Add new `--exit-code` (`-x`) flag that will pass-through the exit form the
   command being ran
   ([#755](https://github.com/go-task/task/pull/755)).
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -23,6 +23,8 @@ tasks:
 
   install:
     desc: Installs Task
+    sources:
+      - './**/*.go'
     cmds:
       - go install -v -ldflags="-w -s -X main.version={{.GIT_COMMIT}}" ./cmd/task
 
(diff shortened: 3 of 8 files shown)


andreynering:
I took some time to understand the problem, but thanks to the `sleepit` helper @marco-m-pix4d implemented I could do some manual tests figure it out what was the problem.

We had code to intercept the signals and it triggered a cancel of the context, which means the process was actually receiving two signals: one because of the `cancel()` and the other because the OS was still sending the signals to the sub-processes anyway.

https://github.com/go-task/task/blob/c9a582fbcc81aa86fe9e5b9b38fb8de10c342583/cmd/task/task.go#L252-L262

Changed it so it will now only intercept the signal but don't cancel the context, so it's actually sent only once to the sub-processes as expected.

As an addition, I added a force kill behavior if SIGINT is sent for the third time, so users at least have way to force Task to quit if any sub-process happen to be stuck.

Also changed the timeout to send KILL to processes in case of context cancellation from 2 to 15 seconds, which give them more time to do the necessary cleanup.

I'll merge it to master. Please give it a test if possible and let me know if it's working correctly. I'll wait a couple days to release it to be more certain that it is stable.

/cc @marco-m-pix4d @aliculPix4D @ghostsquad

marco-m-pix4d:
> We had code to intercept the signals and it triggered a cancel of the context, which means the process was actually receiving two signals: one because of the cancel() and the other because the OS was still sending the signals to the sub-processes anyway.

This is what I remember I had discovered in my work, yes.

> Changed it so it will now only intercept the signal but don't cancel the context, so it's actually sent only once to the sub-processes as expected.

This looks correct to me. I will try to test it today. Thanks for the effort!

andreynering:
@marco-m-pix4d Great! I plan to make a release tonight.

We still have that intermittently failing test as before. Hopefully we'll find a solution for that soon.

PR #726, opened 2022-05-11
Add monochromatic SVG logo

sisp:
I've added a monochromatic version of the SVG log as discussed in https://github.com/go-task/task/discussions/720#discussioncomment-2730626.

PR #772, opened 2022-06-13
Simplify vectorial logo paths

ssbarnea:
Previous version of the logo had complex shapes which caused rendering
issues in some cases (a thin horizontal line appearing in the middle of the image).

This change ensures only 3 simple shapes are used and sorts the visual artifacts. Older version had 3 different shapes of same color for each of the left and sides of the cube. The top side of the cube was already ok, as a single path.

A side effect is also a small decrease in file size. Apparently we do not
need to update the raster logos as they do not present the bug.


andreynering:
Hi @ssbarnea, thanks for opening this PR!

This had been done before by @sisp on #725, and a monochromatic version added at #726.

I rolled it back by mistake when I migrated the website to a new documentation system, but just restore that work again.

Let me know if you think there's any additional improvement to be done here.

PR #775, opened 2022-06-13
Correct install instructions to use standard user bin

ssbarnea:
Update user-level install script to advertise the recommended location
for user binaries, which is `~/.local/bin`.

Most operating systems already has this inside their default PATH.

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -108,8 +108,8 @@ scenarios like CI. Many thanks to [GoDownloader][godownloader] for enabling the
 easy generation of this script.
 
 ```bash
-# For Default Installation to ./bin with debug logging
-sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d
+# For user-level installation to ~/.local/bin with debug logging
+sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b ~/.local/bin
 
 # For Installation To /usr/local/bin for userwide access with debug logging
 # May require sudo sh


ghostsquad:
Can you provide the source material that you found?

andreynering:
Hi @ssbarnea,

I like the intention, but I'm not so sure that having a `~/.local/bin` directory created and added by default to `$PATH` is that common.

That script is intended to work on macOS and Windows as well, something to keep in mind. (For Windows you need a Bash-like terminal, of course). I checked in my macOS and a `~/.local/bin` dir neither exist nor it's added to the `$PATH`.

So my proposal here is to keep that first section which install to `./bin/task` and add a second one with the `~/.local/bin` example mentioning that it will work on most Linux distros.

The existing section mentioning `/usr/local/bin` can be probably kept as is as the third section.

Opinions?

---

There's probably some room to refactor the text/markdown formatting as well here (have sections on different code blocks?). But you can let that to me if you prefer.

ssbarnea:
Some findings:
* Ubuntu 22.04 will recognize both ~/bin and ~/.local/bin once they are created (re-reun .profile)
* Fedora 36 will do the same, remark that `~/.local/bin` takes precedence on Fedora but ~/bin on Ubuntu
* MacOS 12.4 does not recognize any of those but to my surprise seems to include `~/.dotnet/tools` inside PATH but I am not sure if that this true on all versions as mine is managed, so IT might have injected it as I do not see Apple fancy dotnet stuff.
* Windows: TBH and likely irrelevant as we have no curl/bash AFAIK
* FreeBSD has only `~/bin` in PATH, by default.

Based on that I would suggest default value to be first already existing folder found on disk out of the two or `~/bin` as fallback.

andreynering:
Thought that the better to do in the short term was updating the documentation mentioning the possibilities.

We could in theory update the script to follow @ssbarnea's suggestion for the default directory, but there are some problems:

- It would be backward incompatible (some CI script might be relying on the existing `./bin` default)
- [GoReleaser](https://github.com/goreleaser/godownloader), the tool we used to generate that script, is unfortunately not maintained anymore

ghostsquad:
I'm most familiar with, and have most commonly seen `~/bin` used. However, I find what might be able to put this to rest

>  It is recommended that, apart from autosave and lock files, programs should refrain from creating non dot files or directories in a home directory without user consent.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s08.html#idm236092748016

Just some light searching I see these results:

`~/bin`
- https://zwbetz.com/how-to-add-a-binary-to-your-path-on-macos-linux-windows/
- https://mycyberuniverse.com/create-personal-bin-directory-run-scripts-without-specifying-full-path.html

The link below mentions this is default in RHEL 6+, Debian, and Ubuntu.

https://unix.stackexchange.com/a/215717

`~/.bin or ~/.local/bin`
- https://forums.linuxmint.com/viewtopic.php?t=313460

This article mentions both `~/bin` and `~/.local`

https://www.quora.com/Why-do-some-applications-use-HOME-local-bin-and-not-usr-local-bin-on-Linux-and-similar-systems?share=1

ssbarnea:
Few months ago I randomly encountered a project that was about online line installers for most common software but I failed to find it again. I think they also had a template for creating `curl|bash` installers. Somehow I have the impression they already clarified this.

For sure, current installing into `./bin` is not a good idea, it means that it might create bin folder in random directories, as CWD is rarely $HOME when someone tried to run a command.

The tricky bit is that when running a script as root, you do not want to do the same, instead you want to install the tool in a global location. There is almost no reason for someone to install stuff in `/root/`, regardless which subfolder, mostly because nobody should use root user as a normal user.

I would not be shy about creating a standard folder if it does not exist, especially as on several important operating systems, those user directories are added to the PATH only when they exist, during login scripts. You might not even have to touch profile dotfiles to add it as they are often already there.

I can understand that you do not want to spend a lot of time to get the installer right, especially with so many combinations. Let's try to find a good example to follow, one that is mature, maintained, so we do not have to the same work.

PR #783, opened 2022-06-21
Fix GHA grouping examples

ssbarnea:
Our example was not using the correct syntax required for grouping
command output on GHA.

Documentation at https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines


andreynering:
Thanks @ssbarnea!

PR #790, opened 2022-06-27
docs: scoop extra bucket is not required anymore

Trim21:
it's in main bucket now

https://github.com/ScoopInstaller/Main/blob/master/bucket/task.json


andreynering:
Thanks @Trim21!

PR #791, opened 2022-06-27
Correct `method: timestamp` example block in usage docs

MPLew-is:
The text above this block currently references a `timestamp` method for checking if a task should be run, then the example uses `checksum` instead.


andreynering:
@MPLew-is Good catch! Thanks for noticing and opening a PR for it!

PR #802, opened 2022-07-08
Update api_reference.md

nokome:
Using `cmd` (singular) with a single string is valid syntax also.


andreynering:
Thanks @nokome!

PR #803, opened 2022-07-09
Fix Zsh completion for tasks without description

carlsmedstad:
Use --list-all instead of --list in order include tasks without
description in the auto-completion.

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -27,7 +27,7 @@ function __task_list() {
     (( enabled )) || return 0
 
     scripts=()
-    for item in "${(@)${(f)$("${cmd[@]}" --list)}[2,-1]#\* }"; do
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks @carlsmedstad!

ghostsquad:
This was so needed. In fact, I'd like to open another pull request.. I feel like `--list` should include all tasks regardless, and you would just add another option to _filter_ to include only those with descriptions. E.g.

```shell
# old command
task --list

# new
task --list --only-described

# and deprecate this
task --list-all
```

ghostsquad:
I suppose it's important to add a description to tasks that a user may want to run...

Nvm, that way `--list` becomes the easiest thing to get the tasks that are most relevant.

I'll open an issue with this once I have some more time to think on it.

PR #809, opened 2022-07-14
Create default taskfile by constructing go object

yeahdongcn:
Signed-off-by: R0CKSTAR <yeahdongcn@gmail.com>

Currently, I use `task` with static `Taskfile.yaml` to perform CI in `Drone`. While the `Taskfile.yaml` becomes larger and larger, I would like to use a go program to generate that file without worrying about the lint and incorrect syntax.

But then I found the definition of Taskfile under `/taskfile/taskfile.go` is a kind of structure used internally, so I decided to add new definition in init.go (I don't know if this is the best place) and use it to generate the default `Taskfile.yaml`.

`Taskfile` structure is now a minimal one that only targets to support `task --init`, we can extend all possible configurations later.

### DIFF
--- a/init.go
+++ b/init.go
@@ -5,21 +5,30 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+
+	"gopkg.in/yaml.v3"
 )
 
-const defaultTaskfile = `# https://taskfile.dev
+const defaultTaskfileHeader = `# https://taskfile.dev
 
-version: '3'
-
-vars:
-  GREETING: Hello, World!
-
-tasks:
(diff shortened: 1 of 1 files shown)


andreynering:
Hi @yeahdongcn!

I don't see any benefit of having a duplicated Go schema in the code base, neither in generating the default Taskfile using it.

You can still use Go code to generate Taskfiles if you want, you just have to declare the schema yourself. You can use [the documentation](https://taskfile.dev/api/#schema) as a reference.

Let me know if you have thoughts.

PR #815, opened 2022-07-20
feat: run default task in included file when task is omitted

pd93:
I was experimenting with refactoring my (now very large) Taskfile into smaller, more manageable files and I came across #661. This functionality would really help us to make our tasks a bit simpler.

This PR adds the functionality mentioned by @tylermmorton and @andreynering in that issue. Namely:

- When a Taskfile with a `default` task inside is included, the `default` task is available by calling `task <namespace>` as opposed to `task <namespace>:default`.
- If the parent Taskfile contains a task with the same name as the namespace (i.e. the task is "shadowed"), the task in the _parent_ is run, and _not_ the default task in the included file.
  - This ensures that we maintain backward compatibility with existing Taskfiles.
- A couple of unit tests for ensuring the functionality works going forwards.

Note: Not sure about the "shadowed" terminology. Open to better suggestions.

An example below:

```yaml
# Taskfile.yml
version: '3'

includes:
  included:
    taskfile: Taskfile2.yml
```

```yaml
# Taskfile2.yml
version: '3'

tasks:
  default:
    cmds:
      - echo "included task"
```

`task included` will output: `included task`.
Currently you are required to run `task included:default`.

If we amend the parent taskfile like so:

```yaml
# Taskfile.yml
version: '3'

includes:
  included:
    taskfile: Taskfile2.yml

tasks:
  included:
    cmds:
      - echo "shadowed task"
```

`task included` will now output: `shadowed task` instead.

### DIFF
--- a/taskfile/read/taskfile.go
+++ b/taskfile/read/taskfile.go
@@ -153,6 +153,13 @@ func Taskfile(readerNode *ReaderNode) (*taskfile.Taskfile, error) {
 		if err = taskfile.Merge(t, includedTaskfile, &includedTask, namespace); err != nil {
 			return err
 		}
+
+		if includedTaskfile.Tasks["default"] != nil && t.Tasks[namespace] == nil {
+			defaultTaskName := fmt.Sprintf("%s:default", namespace)
+			t.Tasks[defaultTaskName].Aliases = append(t.Tasks[defaultTaskName].Aliases, namespace)
+			t.Tasks[defaultTaskName].Aliases = append(t.Tasks[defaultTaskName].Aliases, includedTask.Aliases...)
+		}
+
 		return nil
 	})
 	if err != nil {
--- a/task_test.go
+++ b/task_test.go
@@ -1107,6 +1107,30 @@ func TestInternalTask(t *testing.T) {
 	}
 }
 
+func TestIncludesShadowedDefault(t *testing.T) {
+	tt := fileContentTest{
+		Dir:       "testdata/includes_shadowed_default",
+		Target:    "included",
+		TrimSpace: true,
+		Files: map[string]string{
+			"file.txt": "shadowed",
+		},
+	}
+	tt.Run(t)
+}
+
+func TestIncludesUnshadowedDefault(t *testing.T) {
+	tt := fileContentTest{
+		Dir:       "testdata/includes_unshadowed_default",
+		Target:    "included",
+		TrimSpace: true,
+		Files: map[string]string{
+			"file.txt": "included",
+		},
+	}
+	tt.Run(t)
+}
+
 func TestSupportedFileNames(t *testing.T) {
 	fileNames := []string{
 		"Taskfile.yml",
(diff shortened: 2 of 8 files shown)


pd93:
I've just seen #665. I didn't realise someone had already made an attempt at this. However, since the comments there don't seem to have been addressed, I've made an attempt to address them here instead.

> I think this implementation has unintended side-effects. Tasks assigned here will be duplicated on task --list.

@andreynering I tried out your suggestion in `variables.go`, but couldn't get it to work. This duplication (as I'm sure you know) is happening because we're merging the included task and then copying it to a new map key as well. The names are the same due to them sharing a memory address and the loop at the end of the `read.Taskfile` function is setting the name for _both_ map keys. Interestingly, because looping over maps is non-deterministic, the name is sometimes `docs:default` and sometimes `docs` depending on which map entry is looped over last.

I can see two fixes to this depending on the desired behaviour:

1. Copy the value so that they no longer share a pointer. This results in both `docs` and `docs:default` being written to the `--list` output.

```go
if includedTaskfile.Tasks["default"] != nil && t.Tasks[namespace] == nil {
	t.Tasks[namespace] = &taskfile.Task{}
	*t.Tasks[namespace] = *(includedTaskfile.Tasks["default"])
}
```

2. Delete the original map entry. This will result in _only_ `docs` being written to the `--list` output.

```go
if includedTaskfile.Tasks["default"] != nil && t.Tasks[namespace] == nil {
	t.Tasks[namespace] = includedTaskfile.Tasks["default"]
	delete(t.Tasks, namespace+":default")
}
```

IMO, the first is preferrable, since both `docs` and `docs:default` are still callable (and should remain so for backward compatibility). The second approach also breaks many tests 😞

taskfile/read/taskfile.go
```diff
@@ -152,6 +152,12 @@ func Taskfile(readerNode *ReaderNode) (*taskfile.Taskfile, error) {
 		if err = taskfile.Merge(t, includedTaskfile, includedTask.Internal, namespace); err != nil {
 			return err
 		}
+
+		if includedTaskfile.Tasks["default"] != nil && t.Tasks[namespace] == nil {
+			t.Tasks[namespace] = &taskfile.Task{}
+			*t.Tasks[namespace] = *(includedTaskfile.Tasks["default"])
+		}
```
andreynering, line 159: Hey @pd93!

Now that we have a whole `alias` infra in the code, I think we should change how we plan to do this.

Instead of copying the whole task here, we should add an aliases to the task pointing to the alternative names. This way `task --list` will also behave as expected, not duplicating tasks and documenting the right aliases as well.

```yaml
version: '3'

includes:
  docs:
    aliases: [d]
    taskfile: ./docs
    dir: ./docs
```

```yaml
# docs/Taskfile.yml

version: '3'

tasks:
  default:
    desc: Something
    cmds:
      - ...
```

`task -l` should print something like this:

```
$ task -l
* docs:default:         Something                              (aliases: docs, d, d:default)
```

pd93: This is a great idea. I've pushed a change to use aliases instead

andreynering:
Awesome. Thanks @pd93!

PR #817, opened 2022-07-21
Resolve contradiction.

white-gecko:
> Vars declared in the included Taskfile have preference over the included ones!

is a contradiction. It should be

> Vars declared in the included Taskfile have preference over the variables in the including Taskfile!


ghostsquad:
This is not backwards compatible, and some folks use includes differently, and expect this behavior.

PR #818, opened 2022-07-22
Internal tasks

pd93:
## Problem

Building on the fantastic work introduced by #656, I have been refactoring some of my projects with a `taskfiles` subdirectory. This helps me manage my extemely long files much more effectively and also allows me to take full advantage of the namespacing features (`a:b:c`).

However, I've noticed that in my more complex projects, the `--list-all` output is now _extremely_ long, as some of my "utility" taskfiles are being included in multiple other taskfiles. I've provided a simplified example below:

```yaml
# Taskfile.yml
version: '3'

includes:
  install: ./taskfiles/Install.yml
  gen: ./taskfiles/Gen.yml
  # etc...
```

```yaml
# ./taskfiles/Install.yml
version: '3'

tasks:
  
  proto-tools:
    # ...
  mock-tools:
    # ...
  # etc...
```

```yaml
# ./taskfiles/Gen.yml
version: '3'

includes:
  install: ./Install.yml

tasks:
  
  protos:
    deps:
      - install:proto-tools
  mocks:
    deps:
      - install:mock-tools
  # etc...
```

```ini
# --list-all output
* install:proto-tools
* install:mock-tools
* gen:protos
* gen:mocks
* gen:install:proto-tools
* gen:install:mock-tools
# etc...
```

I'd like to be able to call `install:*` from the command line, which is fine. However, I _really_ don't need to be able to call `gen:install:*` directly and I definitely don't want it to appear in my nicely documented list of tasks when running `--list-all`. Imagine this scenario, but with an much more complex setup, with several layers of includes and the list of tasks can very quickly become 100s of lines long and lag out my terminal autocompletion 😞 

## Solution

This PR adds a boolean `internal` parameter to tasks which stops them from being callable on the command-line. It also stops the task from being output using `--list|--list-all`.

In addition to this, it solves the problem illustrated above by adding the same parameter to `includes`. This will mark all the included tasks as `internal`, thus removing them from the task list. Internal tasks can still be called by other tasks in the usual way.

My example would now looks something like this:

```yaml
# ./taskfiles/Gen.yml
version: '3'

includes:
  install:
    taskfile: ./Install.yml
    internal: true

tasks:
  
  protos:
    deps:
      - install:proto-tools
  mocks:
    deps:
      - install:mock-tools
  # etc...
```

```ini
# --list-all output
* install:proto-tools
* install:mock-tools
* gen:protos
* gen:mocks
# etc...
```

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -194,6 +194,21 @@ tasks:
       - echo "This command can still be successfully executed if ./tests/Taskfile.yml does not exist"
 ```
 
+### Internal includes
+
+Includes marked as internal will set all the tasks of the included file to be
+internal as well (See the [Internal Tasks](#internal-tasks) section below). This is useful when including utility tasks that are not
+intended to be used directly by the user.
+
+```yaml
+version: '3'
+
+includes:
+  tests:
+    taskfile: ./taskfiles/Utils.yml
+    internal: true
+```
+
 ### Vars of included Taskfiles
 
 You can also specify variables when including a Taskfile. This may be useful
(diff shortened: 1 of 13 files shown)


andreynering:
Thanks @pd93!

PR #831, opened 2022-08-08
Attempt to fix Task not following symlinks

ilewin:
Please check if this could be a good solution for the described bug.


internal/status/glob.go
```diff
@@ -26,13 +25,13 @@ func globs(dir string, globs []string) ([]string, error) {
...
 	g = filepathext.SmartJoin(dir, g)
-
 	g, err := execext.Expand(g)
 	if err != nil {
 		return nil, err
 	}
 
-	fs, err := zglob.Glob(g)
+	//fs, err := zglob.Glob(g) //Looks like this does not evaluate **/* glob correctly
+	fs, err := filepath.Glob(g)
```
andreynering, line 34: `filepath.Glob` does not support double star (`foo/**/*.js`). We should keep using `zglob.Glob`

andreynering (commented):
Hi @ilewin,

Given there's an issue open on https://github.com/mvdan/sh/issues/843, I think we should not try to fix it when running shell commands and wait for a fix upstream.

Let's code the fix for the source calculation only.

---

Regarding the PR itself, please write a meaningful commit message and be careful with unnecessary changes to the codebase like changing new lines, commented code, etc (specially on files you didn't change). The diff should be on relevant code only.

internal/status/glob.go
ilewin: Ok, Thanks I will dig more.
Zglob does not follow symlinks so I will try to figure out how to have both ** and symlinks
https://github.com/mattn/go-zglob/issues/3

andreynering:
Thanks @ilewin!

PR #832, opened 2022-08-10
Modify goreleaser defaults for reproducible builds

jfhovinne:
This should make it easier to generate the same builds locally (tested with --snapshot).
https://goreleaser.com/customization/build/#reproducible-builds


andreynering:
Thanks @jfhovinne!

PR #835, opened 2022-08-11
Improved bash completion

MarioSchwalbe:
Some incremental update. Point 1 below was most annoying during daily use.

Features:
- Handles the -t and -d options correctly, i.e., completes with tasks in the target directory.
- Calls the correct binary for completions, if bound multiple times, by using `$1` or `COMP_WORDS`.

### DIFF
--- a/completion/bash/task.bash
+++ b/completion/bash/task.bash
@@ -1,48 +1,48 @@
-# vim: set tabstop=4 shiftwidth=4 expandtab:
+# vim: set tabstop=2 shiftwidth=2 expandtab:
 
 _GO_TASK_COMPLETION_LIST_OPTION='--list-all'
 
 function _task()
 {
-    local cur prev words cword
-    _init_completion -n : || return
+  local cur prev words cword
+  _init_completion -n : || return
 
-    # Handle special arguments of options.
-    case "$prev" in
-        -d|--dir)
-            _filedir -d
-            return $?
-        ;;
-        -t|--taskfile)
-            _filedir yaml
-            _filedir yml
-            return $?
-        ;;
-        -o|--output)
-            COMPREPLY=( $( compgen -W "interleaved group prefixed" -- $cur ) )
-            return 0
-        ;;
-    esac
+  # Handle special arguments of options.
+  case "$prev" in
+    -d|--dir)
+      _filedir -d
+      return $?
+    ;;
+    -t|--taskfile)
+      _filedir yaml || return $?
+      _filedir yml
+      return $?
+    ;;
+    -o|--output)
+      COMPREPLY=( $( compgen -W "interleaved group prefixed" -- $cur ) )
+      return 0
+    ;;
+  esac
 
-    # Handle normal options.
-    case "$cur" in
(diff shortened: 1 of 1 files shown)


completion/bash/task.bash
```diff
@@ -1,26 +1,48 @@
-#!/bin/bash
+# vim: set tabstop=4 shiftwidth=4 expandtab:
 
-GO_TASK_PROGNAME=task
```
andreynering, line 3: AFAIK this variable was here to allow someone to have completion even if the Task binary is named differently.

Due to a conflict with taskwarrior, some installation methods may call it `go-task` for example.

Don't you think it's worth keeping it?

completion/bash/task.bash
```diff
@@ -1,26 +1,48 @@
-#!/bin/bash
+# vim: set tabstop=4 shiftwidth=4 expandtab:
```
andreynering, line 1: Isn't `#!/bin/bash` really needed here?

What does this vim syntax actually does here?

andreynering (commented):
Thanks for your contribution!

I wasn't the one that wrote these completion scripts (and my knowledge about completion is very limited), so excuse if my comments looks dumb.

completion/bash/task.bash
MarioSchwalbe: The shebang line `#!/bin/bash` informs the (UNIX-like) system with which interpreter to run a script provided it is made executable. However, those bash completion script fragments are designed to be included (sourced) into an already running interpreter. The bash completion framework does so automatically on demand. There's no point in running them on their own. Hence, the shebang is useless.

The other line configures editors like vim (and compatible) how to format (indent) this source code. Other bash completions contain that as well.

completion/bash/task.bash
MarioSchwalbe: Within completion functions you better use `$1` or the array variable `$COMP_WORDS` to get the actual command the completion is running for. There's no need for a global variable. This is what the code does.

The line
```sh
complete -F _task task
````

attaches the completion function to a program named `task` (the last word).

It may make some sense to allow customization here. However, the previous approach was also useless, because:
1. The variable `GO_TASK_PROGNAME` was set unconditionally. How can any user change it between the beginning and the end of the script?
2. The bash completion framework loads completions on demand matching the command name `task` with the file name `task.bash`. This doesn't work if the binary is called differently, but the script still has the old name. It always requires more changes which those creating packages can provide easily (rename the file and change the last line with a regex).

I remove it because it wasn't really working and I didn't expect that use case.

completion/bash/task.bash
andreynering: @MarioSchwalbe OK, last question: can we keep 2 space indentation as before? That's our convention for non-Go code as you can see on the `.editorconfig` file.

MarioSchwalbe: The bash completion standard is actually 4 ;-), but I don't really care. I'm going to push a new commit.

andreynering:
Thanks @MarioSchwalbe!

PR #844, opened 2022-08-23
Use --silent to get the list of tasks (bash completion)

MarioSchwalbe:
This is a small update of PR <https://github.com/go-task/task/pull/835>.

I may have missed the `--silent` option which is much better suited to get the list of tasks than digging in output intended for human readers with regular expressions.


andreynering:
Thanks @MarioSchwalbe!

PR #859, opened 2022-09-05
Document how to reproduce release binaries

jfhovinne:
Since published release executables can now be locally reproduced and verified (see #832, tested with latest release v3.15.0), I would mention it in the docs.


andreynering:
Thanks @jfhovinne!

PR #861, opened 2022-09-07
Make zsh completion list configurable

cdaguerre:
This makes the zsh completion list configurable through an env variable exposed before sourcing the completion script without changing current behavior.
The var name is taken from the [bash completion script](https://github.com/go-task/task/blob/master/completion/bash/task.bash#L3) which seems to provide similar functionality.

Add:
```sh
export GO_TASK_COMPLETION_LIST_OPTION="--list"
```
To limit autocompletion suggestions to only tasks with descriptions.

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -3,6 +3,8 @@
 local context state state_descr line
 typeset -A opt_args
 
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks @cdaguerre!

PR #863, opened 2022-09-08
Fix handling of CLI_ARGS

MarioSchwalbe:
Improvements:

1. Do not attempt to propose completions if the current position is behind `--` as the arguments would go to task's `CLI_ARGS` variable.
2. Only use the command-line before `--` when proposing task names to prevent the list option from going to task's `CLI_ARGS` variable.


andreynering:
Thanks again @MarioSchwalbe!

Issue #813, opened 2022-07-17
Make watch interval configurable (currently fixed at 5 seconds)


gregorriegler:
I have `gradle -t test` (this is watching for file-changes and then reruns the tests) and  `task test --watch` open at the same time and task is significantly slower. It reacts to filechanges about a second later, while gradles watcher is instant. Do you see an easy way to make it faster?

ghostsquad:
Can I ask you to provide a minimal reproduction for Gradle and Task so that benchmarks can be made and used for comparison?

gregorriegler:
[watch-slow.zip](https://github.com/go-task/task/files/9131344/watch-slow.zip)
So i made this zip.
In the left corner we have Taskfile
```
version: '3'

tasks:
  test:
    cmds:
      - echo Taskfile
    sources:
      - test.txt
```

fighting against the oldschooler on the right corner: Makefile with inotifywait
```
.PHONY: test
test: 
	echo "Makefile";

.PHONY: watch
watch: 
	while true; do \
		make test; \
		inotifywait -qr -e modify -e create -e delete -e move test.txt; \
	done
```

run both, and change test.txt to trigger the task.

andreynering:
Hi everybody,

I believe that having an attribute on the Taskfile (and possible a flag, too) to have a different polling interval would do the trick here.

We could also consider switching to a library that uses the OS notification API, but I wanted to avoid it initially because it may behave differently based on the user's OS.

gregorriegler:
I use watchexec and just for right now.
So I'm not waiting for this.
But I think it could be beneficial for you to have it built in. Easier setup and scripting for users.

> We could also consider switching to a library that uses the OS notification API, but I wanted to avoid it initially because it may behave differently based on the user's OS.

Do you say 'may' because we're not sure? In that case I recommend trying 
For the thing I'm doing I'm not sure if an interval would be sufficient. But maybe it's worth trying, too

ilewin:
I'd like to give it a try.

andreynering:
This was implemented by @ilewin at #865.

I plan to make a new release still today.

PR #865, opened 2022-09-08
Configurable watch interval #813

ilewin:
Attempt to make watch interval configurable from Taskfile global settings and from CLI. If both are used (CLI and Taskfile setting), then Taskfile's setting has a higher priority.


cmd/task/task.go
```diff
@@ -206,6 +208,10 @@ func main() {
 		e.InterceptInterruptSignals()
 	}
 
+	if e.Taskfile.Interval == "" {
+		e.Taskfile.Interval = strings.TrimSpace(interval)
+	}
```
andreynering, line 213: For consistency with other configurations, we should have an attribute on the `Executor` struct instead of setting in the Taskfile here. See `Silent` for example.

This allow more flexibility for those who use Task as a library.

The priority order is: CLI flag > Taskfile > Default (5s).

testdata/watcher_interval/src/a
```diff
@@ -0,0 +1 @@
+test
```
andreynering, line 1: I feel that this file shouldn't be commited? We usually have a `.gitignore` inside each directory inside `testdata`.

watch.go
```diff
@@ -36,6 +37,11 @@ func (e *Executor) watchTasks(calls ...taskfile.Call) error {
 		}()
 	}
 
+	wInt, err := e.Taskfile.ParsedWatchInterval()
+	if err != nil || wInt == 0 {
+		wInt = watchInterval
```
andreynering, line 42: The `watchInterval` const in this file should be renamed to `defaultWatchInterval`.

I would then rename the `wInt` variable to `watchInterval`.

andreynering (commented):
Sound good, thanks for working on this!

Added a few comments about small adjustment before this is ready to be merged.

andreynering:
Thanks @ilewin!

PR #866, opened 2022-09-08
fix: ignore empty dotfile file names

### LINKED ISSUE #858: Getting "is a directory" for Task v3.15.0
- Task version: v3.15.0
- Operating System: macOS 12.2, Ubuntu

```yaml
version: '3'

dotenv:
  - "{{.DOTENV}}"
```

Hey!

We've started getting a weird error message after upgrading to v3.15.0 (v3.14.0 works fine).

```
read /path/to/repository: is a directory
```

It seems the issue is with loading `dotenv` which we use like this:

```yaml
dotenv:
  - "{{.DOTENV}}"
```

I'm not sure if that's a documented feature and if it was working by accident or is an actual bug. We're using this as a way to conditionally load a set of environment variables.

pd93:
This PR fixes #858 - A bug (introduced in f45dd11) that causes an "is a directory" error when a taskfile contains a `dotfiles` key with unresolved variable.

The bug exists because the new `e.setCurrentDir()` function is setting `e.Dir` to the taskfile's root directory where it was previously `""`. When this variable was empty and the given dotfile was also empty, `os.Stat` would have returned an `os.ErrNotExist` and the loop would have `continued`. However, now that `e.Dir` is set to the taskfile directory, this `os.Stat()` check passes (as the directory exists) and the `godotenv.Read()` function errors causing the program to fail.

This fix simply checks if each value in the list of `dotfiles` is empty (not set) and `continues` if it is. This allows users to continue using the behaviour described in #858 where they could conditionally set environment variables via the `dotfiles` key.

### DIFF
--- a/taskfile/read/dotenv.go
+++ b/taskfile/read/dotenv.go
@@ -27,6 +27,9 @@ func Dotenv(c compiler.Compiler, tf *taskfile.Taskfile, dir string) (*taskfile.V
 
 	for _, dotEnvPath := range tf.Dotenv {
 		dotEnvPath = tr.Replace(dotEnvPath)
+		if dotEnvPath == "" {
+			continue
+		}
 		dotEnvPath = filepathext.SmartJoin(dir, dotEnvPath)
 
 		if _, err := os.Stat(dotEnvPath); os.IsNotExist(err) {


andreynering:
Thanks @pd93!

PR #874, opened 2022-09-29
Colorize --list, --list-all and --summary output

pd93:
@andreynering @orenmazor This seemed like low-hanging fruit, so I've done a quick first draft (see the [associated branch](https://github.com/go-task/task/tree/845-colorize-output) and screenshots:

`task --list-all`:

image

`task foo --summary`:

image

Let me know what you think of the colours - I'm happy to change them. I went with: 
- `green` for tasks as this matches the green when running a task
- `cyan` for descriptions as it complements the green (and the Task logo)
- `yellow` for commands

Could we do something more interesting with namespaces? Each layer of the namespace be a different colour? This might be too noisy... Or maybe just a different colour for the colons?

---

Sidenote: I noticed that `logger.Outf` automatically adds a `\n` to the input string. I'm not a fan of this behaviour as it gives less control to the caller. The new `logger.FOutf` function does not do this to mimic the behaviour of the standard `fmt.Printf` functions and this allows me to use multiple colours on the same line.

I've left the existing `logger.Outf` functionality as-is for now, but I think we should consider changing this in the future.

### DIFF
--- a/help.go
+++ b/help.go
@@ -42,9 +42,12 @@ func (e *Executor) printTasks(listAll bool) {
 	e.Logger.Outf(logger.Default, "task: Available tasks for this project:")
 
 	// Format in tab-separated columns with a tab stop of 8.
-	w := tabwriter.NewWriter(e.Stdout, 0, 8, 0, '\t', 0)
+	w := tabwriter.NewWriter(e.Stdout, 0, 8, 6, ' ', 0)
 	for _, task := range tasks {
-		fmt.Fprintf(w, "* %s: \t%s\n", task.Task, task.Desc)
+		e.Logger.FOutf(w, logger.Yellow, "* ")
+		e.Logger.FOutf(w, logger.Green, task.Task)
+		e.Logger.FOutf(w, logger.Default, ": \t%s", task.Desc)
+		fmt.Fprint(w, "\n")
 	}
 	w.Flush()
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add colored output to `--list`, `--list-all` and `--summary` flags ([#845](https://github.com/go-task/task/pull/845), [#874](https://github.com/go-task/task/pull/874)).
 - Fix unexpected behavior where `label:` was being shown instead of the task
   name on `--list`
   ([#603](https://github.com/go-task/task/issues/603), [#877](https://github.com/go-task/task/pull/877)).
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -53,6 +53,11 @@ type Logger struct {
 
 // Outf prints stuff to STDOUT.
 func (l *Logger) Outf(color Color, s string, args ...interface{}) {
+	l.FOutf(l.Stdout, color, s+"\n", args...)
+}
+
+// FOutf prints stuff to the given writer.
+func (l *Logger) FOutf(w io.Writer, color Color, s string, args ...interface{}) {
 	if len(args) == 0 {
 		s, args = "%s", []interface{}{s}
 	}
@@ -60,7 +65,7 @@ func (l *Logger) Outf(color Color, s string, args ...interface{}) {
 		color = Default
 	}
 	print := color()
-	print(l.Stdout, s+"\n", args...)
+	print(w, s, args...)
 }
 
 // VerboseOutf prints stuff to STDOUT if verbose mode is enabled.
--- a/internal/summary/summary.go
+++ b/internal/summary/summary.go
@@ -56,7 +56,8 @@ func printTaskSummary(l *logger.Logger, t *taskfile.Task) {
 }
 
 func printTaskName(l *logger.Logger, t *taskfile.Task) {
-	l.Outf(logger.Default, "task: %s", t.Name())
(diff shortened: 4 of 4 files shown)


andreynering:
Hi @pd93, thanks for taking this one!

I liked the color choices in overall, but I think longer texts like descriptions are better as white. I also tried putting `*` in yellow and liked the result. What do you think?

Screen Shot 2022-09-29 at 21 49 36

help.go
```diff
@@ -42,9 +42,13 @@ func (e *Executor) printTasks(listAll bool) {
...
 	// Format in tab-separated columns with a tab stop of 8.
-	w := tabwriter.NewWriter(e.Stdout, 0, 8, 0, '\t', 0)
+	w := tabwriter.NewWriter(e.Stdout, 0, 8, 6, ' ', 0)
 	for _, task := range tasks {
-		fmt.Fprintf(w, "* %s: \t%s\n", task.Name(), task.Desc)
+		fmt.Fprintf(w, "* ")
+		e.Logger.FOutf(w, logger.Green, task.Name())
+		fmt.Fprintf(w, ": \t")
+		e.Logger.FOutf(w, logger.Cyan, task.Desc)
+		fmt.Fprintf(w, "\n")
```
andreynering, line 51: For consistency you could use `e.Logger.FOutf` for all calls, but using `logger.Default` for the white texts.

pd93: I did give this a go, but it resulted in this weird spacing issue and I couldn't figure out what was going on (see red mark in screenshot below).

```diff
diff --git a/help.go b/help.go
index 3457683..4914347 100644
--- a/help.go
+++ b/help.go
@@ -44,11 +44,11 @@ func (e *Executor) printTasks(listAll bool) {
        // Format in tab-separated columns with a tab stop of 8.
        w := tabwriter.NewWriter(e.Stdout, 0, 8, 6, ' ', 0)
        for _, task := range tasks {
-               fmt.Fprintf(w, "* ")
+               e.Logger.FOutf(w, logger.Default, "* ")
                e.Logger.FOutf(w, logger.Green, task.Name())
-               fmt.Fprintf(w, ": \t")
+               e.Logger.FOutf(w, logger.Default, ": \t")
                e.Logger.FOutf(w, logger.Cyan, task.Desc)
-               fmt.Fprintf(w, "\n")
+               e.Logger.FOutf(w, logger.Default, "\n")
        }
        w.Flush()
 }
```

image

andreynering: I suppose it's fine to keep as is then...

pd93:
> I liked the color choices in overall, but I think longer texts like descriptions are better as white. I also tried putting * in yellow and liked the result. What do you think?

Happy to keep the text descriptions white, that makes sense. I assume we should change the description in the `--summary` white too? Updated in [2df6908](https://github.com/go-task/task/pull/874/commits/2df6908d0f3c54a3a7bc2f615fbfb1a8c21f7266)

andreynering:
> I assume we should change the description in the --summary to white too?

IMO, yes.

help.go
pd93: I agree it would be nicer to be consistent. I'll take another look at this tomorrow and see if I can figure out what's going on.

pd93: I had a bit of time to look at this today and worked out what's going on. This first thing to note is that `text/tabwriter` doesn't care about escape sequences. This means that when you add color escape sequences, you can get some strange behaviour.

To make debugging easier, I replaced the tab writer with a regular `bytes.buffer` and printed the output. This is what we were getting (I've reformatted this slightly so its easier to tell what's going on):

```
"       \x1b[33m* \x1b[0m\x1b[32m <task_1> \x1b[0m\x1b[0m: \t <task 1 desc>"
"\x1b[0m\x1b[33m* \x1b[0m\x1b[32m <task_2> \x1b[0m\x1b[0m: \t <task 2 desc>"
...
"\x1b[0m\x1b[33m* \x1b[0m\x1b[32m <task_n> \x1b[0m\x1b[0m: \t <task n desc>"
"\x1b[0m"
```

As you can see, because each color print function _wraps_ the text sent to the writer, when you send a `\n` to it, the color reset escape sequence will be printed on a new line. This means that the tabwriter package will see the next line (and all subsequent lines) as longer than the first line and this leads to oddities with the columns.

As there is no way to change this behaviour in `text/tabwriter` (I've tried using `tabwriter.StripEscape` to no effect). I propose that we consistently use `e.Logger.FOutf` for all output except the final newline which will use `fmt.Fprint`. This will result in the following:

```
"\x1b[33m* \x1b[0m\x1b[32m <task_1>           \x1b[0m\x1b[0m: \t <task_1_desc> \x1b[0m"
"\x1b[33m* \x1b[0m\x1b[32m <task_2>           \x1b[0m\x1b[0m: \t <task_2_desc> \x1b[0m"
...
"\x1b[33m* \x1b[0m\x1b[32m <task_n>           \x1b[0m\x1b[0m: \t <task_n_desc> \x1b[0m"
```

Because the escape sequences are consistent on every line, `text/tabwriter` will now deal with it correctly.

andreynering (approved):
@pd93 This needs conflicts to be resolved, a CHANGELOG entry, and maybe squash commits into a single one.

After that, I authorize you to merge this PR yourself if you want. 🙂

PR #877, opened 2022-09-30
Fix task names for `--list` and `--list-all`

### LINKED ISSUE #603: Do not show label instead of task name in some places like `--list`
- Task version: 3.9.0
- Operating System: Ubuntu

```yaml
---
version: '3'

vars:
  REPOSITORY_SUBTYPE: config
  REPOSITORY_TYPE: npm

tasks:
  one:
    cmds:
      - task: two
      - pip3 install ansible

  two:
    interactive: true
    label: three
    desc: Hey
    cmds:
      - poetry env info --path
      - poetry --help
      - source "$(poetry env info --path)/bin/activate"


```

When assigning a label to a task, the `task --list` is updated with the task with the new label. However, when you attempt to run the task by running `task [label_name]`, task fails to recognize the task.

Also, normally `task` alphabetizes the list shown in `task --list`. However, labeled items are not sorted alphabetically.

I propose that we:

1. Allow the user to run tasks by specifying their label name (e.g. `task label_name`)
2. Alphabetize the `task --list` with the label names included

biozz:
Closes #603.

### DIFF
--- a/help.go
+++ b/help.go
@@ -44,7 +44,7 @@ func (e *Executor) printTasks(listAll bool) {
 	// Format in tab-separated columns with a tab stop of 8.
 	w := tabwriter.NewWriter(e.Stdout, 0, 8, 0, '\t', 0)
 	for _, task := range tasks {
-		fmt.Fprintf(w, "* %s: \t%s\n", task.Name(), task.Desc)
+		fmt.Fprintf(w, "* %s: \t%s\n", task.Task, task.Desc)
 	}
 	w.Flush()
 }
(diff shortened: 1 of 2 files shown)


pd93 (approved):
I think this is a good change. The linked issue was asking to allow calling tasks by their labels, but I think this is a job for aliases (in a future feature) instead.

Worth noting that this will conflict with #874, but I don't want to discard this contribution, so I feel we should merge this first and I can rebase the other branch later.

andreynering:
Thanks @biozz!

PR #878, opened 2022-10-01
feat: optional task filtering with `task -l=foo`

notnmeyer:
adds optional filtering for the `--list` argument to only display tasks that match a supplied regex.

to demonstrate,
```
➜ task -l=test
* test: 	Runs test suite
* test-release: Tests release process without publishing
* test:signals: Runs test suite with signals tests included
```

if you omit the argument to -l/--list, you get the existing behavior,
```
➜ task -l | wc -l
      17
```

for some context around why this is useful for me, at [Optic](https://useoptic.com) we're using Task to manage our internal monorepo. we have quite a few different services and tools and lean heavily on included Taskfiles. we wind up with lots of tasks like, `foo:build`, `foo:deploy`, `bar:build`, etc. long story short, you could say we've got a few tasks,

```
➜ task -l | wc -l
     107
```

shell completions make it somewhat manageable for us to discover and find the tasks we're after, but frequently we wind up running `task -l | grep foo` to view a subset of the output. piping to grep isn't really a problem, but I like not having to do it :)

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -57,7 +57,7 @@ func main() {
 		versionFlag bool
 		helpFlag    bool
 		init        bool
-		list        bool
+		list        string
 		listAll     bool
 		status      bool
 		force       bool
@@ -79,7 +79,7 @@ func main() {
 	pflag.BoolVar(&versionFlag, "version", false, "show Task version")
 	pflag.BoolVarP(&helpFlag, "help", "h", false, "shows Task usage")
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yaml in the current folder")
-	pflag.BoolVarP(&list, "list", "l", false, "lists tasks with description of current Taskfile")
+	pflag.StringVarP(&list, "list", "l", "", "lists tasks with description of current Taskfile. optionally accepts an regex argument to filter tasks")
 	pflag.BoolVarP(&listAll, "list-all", "a", false, "lists tasks with or without a description")
 	pflag.BoolVar(&status, "status", false, "exits with non-zero exit code if any of the given tasks is not up-to-date")
 	pflag.BoolVarP(&force, "force", "f", false, "forces execution even when the task is up-to-date")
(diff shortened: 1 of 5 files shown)


pd93:
I feel like this change breaks the UNIX philosophy (i.e. `Make each program do one thing well`). Is it necessary to duplicate this functionality in Task when tools like `grep` or `Select-String` already do the job perfectly well (arguably better than Task would)?

I think a more interesting approach would be one similar to how the Docker CLI works. For example, filtering based on internal task properties rather than just searching for the task name:

- `task --list --filter="name=foo"`
- `task --list --filter="label=foo" --filter="desc=bar"`
- etc...

Separating the `--filter` from the `--list` flag also allows this to work with `--list-all` and allows us to specify multiple filters at once.
This is far more flexible, achieves what you want and its doing something that tools like `grep` cannot.

Having said all that, this is @andreynering's call.

---

Otherwise the PR looks ok. Ideally the filters should work when the `--silent` flag is enabled. Currently if you run `task --list="test" --silent`, you still get a list of all the tasks rather than just the ones containing `test`.

notnmeyer:
thanks for the feedback @pd93! i considered something like `--filter` (but admittedly only for filtering names), but didn't want to have to think through through how it would interact with other arguments, and whether its just ignored in situations where it doesn't make sense or produces a warning, etc.

my logic here is that `--list` _already_ shows a subset of the available tasks in a project—its filtering based on the existence of a description. so i'm not sure i agree that adding an additional filter for a task name fundamentally alters what `--list` is or breaks the unix philosophy.

not filtering `--list-all` was intentional. the option itself implies that it should be a comprehensive list. it was born out of a need to work around `--list`s inherent filtering.

>  Ideally the filters should work when the --silent flag is enabled. Currently if you run task --list="test" --silent, you still get a list of all the tasks rather than just the ones containing test

fair, hadn't tested this. i'll fix if there's interest in merging this.

pd93:
> I know that #818 implements a way to hide tasks from `--list-all`

Just to add that this PR actually does more than just hide the task from the `--list-all` output. It actually stops you from being able to call it from the command line at all. The task not displaying in the `--list`/`--list-all` output is just a side effect of this.

This doesn't change anything you said, I just thought I'd mention it 😛

notnmeyer:
@pd93 added support for filtering `--list` when `--silent` is also specified 👍

andreynering:
Hi @notnmeyer and @pd93,

I think that having an additional `--filter` flag can be useful given if could be reused between different flags. For now it would be just `--list` and `--list-all`, but soon we will likely have `--json` and `--debug` for example. So a separated flag would be the more future-proof route.

The conflicts you're seeing is because of #890, so there will be some rework needed here (sorry).

notnmeyer:
if `--filter` is the preferred way, we can probably scrap this PR. i think it makes more sense to just implement `--filter` directly rather than use this.

andreynering:
OK.

Thanks anyway for your willingness to contribute @notnmeyer.

Issue #675, opened 2022-02-20
Feature: target aliases and matrix generation


ghostsquad:
This is a continuation of #268 and #340 - It appears the conversation abruptly ended without conclusion.

## Describe in detail what feature do you want to see in Task.

Given a task file such as:

```yaml
version: 3

tasks:
  install:
    cmds:
      - yarn install
```

I'd like to be able to do something like `task i`. Since `i` can with 100% confidence resolve to `install` since there are no other tasks that start with `i`, it could be automatically aliased without the need to force the user to define aliases.

Even though in this situation it feels fluid and expected, the following scenario may end up with suboptimal alias names:

```yaml
version: 3

tasks:
  get:serviceham:
    cmds:
      - ./serviceham.sh
  get:serviceaccount:
    cmds:
      - kubectl -n hello get serviceaccount {{ .CLI_ARGS }}
  get:service:
     cmds:
       - kubectl -n hello get service {{ .CLI_ARGS }}
```

aliases generated

```
task get:service
task get:servicea
```

In this particular case, since `service` is a strict substring of `serviceaccount`, This may surprising, especially in the case when new tasks are added to the file that the user was not aware of, and they were used to using this alias functionality. E.G. if `task get:service` was what the user was used to running that resolved in  `get:serviceaccount`, and later `get:service` was added, all of a sudden the alias `get:service` is no longer and alias, but a target. I feel this is dangerous.

One way to resolve this is to give the footgun to the user, such that it's their responsibility to create aliases that could not be ambiguous:

```yaml
tasks:
  get:serviceham:
    aliases:
    - gsh
    cmds:
      - ./serviceham.sh
  get:serviceaccount:
    aliases:
    - gsa
    cmds:
      - kubectl -n hello get serviceaccount {{ .CLI_ARGS }}
  get:service:
    aliases:
    - gsvc
     cmds:
       - kubectl -n hello get service {{ .CLI_ARGS }}
```

In addition to supporting manually managed aliases, it maybe be interesting to programmatically generate tasks & aliases similar to https://github.com/ahmetb/kubectl-aliases?ts=4#syntax-explanation

```yaml
tasks:
  kubectl:
    aliases:
      - k
    cmds:
      - kubectl {{ .MATRIX.command }} {{ .MATRIX.resource }} {{ .MATRIX.outputformat }} {{ .MATRIX.namespace }}  {{ .CLI_ARGS }}
    matrix:
       command:
          g: "get"
       resource:
          po: "pod"
[...]

Issue #340, opened 2020-06-08
Feature: aliases


kovetskiy:
It would be great to have a sort of aliases for tasks. For example, I have a task `atlas:install` and I'd like to have an alias for `i`.

It could be just a task and it will work well:
```
tasks:
   i:
      cmds: [task: atlas:install]
```
It would be great to have a separated section for such tasks called `aliases` and it would look like
```
aliases:
    i: atlas:install
    p: frontend:bundle
```

@andreynering what do you think about this feature? I could implement it and send PR, if the idea looks good for you.

andreynering:
Hi @kovetskiy,

I'm closing this as a duplicate of https://github.com/go-task/task/issues/268.

I see you proposed a different syntax, though. It'd be nice to comment about it on https://github.com/go-task/task/issues/268 so we could discuss about which one would be a better fit.

Issue #273, opened 2019-12-08
Add option to 'flatten' included taskfiles


tommyalatalo:
It would be great to be able to set an option to avoid having to call the tasks using the `task includename:task` syntax when including a taskfile.

So this example assumes you have files like:
./Taskfile.yml
./common/Taskfile.yml

The use case for this is when creating a main Taskfile for a repository and importing some common tasks, lets say "list", "clean" and "version", right now I need to then call `task common:clean` etc to run them.

Instead I would like to be able to do something like this in my main taskfile:

```
includes:
  common: ./common
    flatten:true
    override: false
```

Where the 'flatten' property would define that the tasks from 'common' should be available as if they were present in the main taskfile, so "task clean" would actually call "task common:clean". This would probably have to throw a conflict error if the "clean" task exists in both the main taskfile and the 'common' one, meaning that anything from an included taskfile that would be flattened out needs to be unique. Possibly an "override" property could be added instead, to let the included task take precedence over any task with the same name in the main taskfile.

This kind of functionality would allow you to add taskfiles as a submodule to a repository, create a main taskfile for that repo where you include the submodule taskfiles, and still be able to run common tasks with a short syntax without having to point out the sub-taskfile. Or in other words include other taskfiles as transparent "libraries".

andreynering:
Hi @tommyalatalo, thanks for opening this issue!

I'm still not convinced to add this feature. I think this brings too little benefit to be worth it: mostly allowing shorter task names (i.e. `build` instead of `common:build`).

I like how namespaced tasks allow you to immediately know where a task if defined (you just have to follow the import).

Am I missing something on other ways this would be useful?

tommyalatalo:
> Hi @tommyalatalo, thanks for opening this issue!
> 
> I'm still not convinced to add this feature. I think this brings too little benefit to be worth it: mostly allowing shorter task names (i.e. `build` instead of `common:build`).
> 
> I like how namespaced tasks allow you to immediately know where a task if defined (you just have to follow the import).
> 
> Am I missing something on other ways this would be useful?

I think there is a valid use case which you're missing.
If I want to do multiple includes from a main taskfile, and those includes can vary, lets say like this:
```
Taskfile.yml
- build
- test

Taskfile.go.yml
- build
- test

Taskfile.another.yml
- build
- test
```

[...]

PR #879, opened 2022-10-02
Aliases

pd93:
Task `aliases` have been discussed a few times in #268, #340, #675. These discussions have ranged in scope and detail, but no design has yet to be established. This PR attempts to implement the simplest possible version of `aliases` as a proof of concept.

Aliases are alternative names for tasks. They can be used to make it easier and quicker to run tasks with long or hard-to-type names. You can use them on the command line, when calling sub-tasks in your Taskfile and when including tasks with aliases from another Taskfile.

There are 3 parts to this PR:

1. Task aliases
1. Namespace aliases
1. `--list` / `--list-all` / `--summary`

## Task Aliases

Task aliases allow users to set a list of alternative strings that can be used to call a task. In the example below, we have 2 tasks: `foo` and `bar`. We can call either of these tasks in the normal way with `task foo` or `task bar` respectively, but thanks to aliases, we are now _also_ able to call `task f` and `task b` with the same behaviour.

Both tasks also have `x` aliased. If we were to run `task x`, Task will return an error as aliases are not allowed to collide.

Note that `foo` calls `bar` and that it is able to do this by using `bar`'s alias: `b`.

```yaml
# Taskfile.yml
version: "3"

tasks:
  foo:
    aliases:
      - f
      - x
    cmds:
      - echo "foo"
      - task: b

  bar:
    aliases:
      - b
      - x
    cmds:
      - echo "bar"
```

All of the above also works with included taskfiles. In the below example we can call `task f` and `task included:b` from the command-line with the same behaviour as the previous example.

```yaml
# Taskfile.yml
version: "3"

includes:
  included:
    taskfile: Taskfile2.yml

tasks:
  foo:
    aliases:
      - f
    cmds:
      - echo "foo"
      - task: included:b
```

```yaml
# Taskfile2.yml
version: "3"

tasks:
  bar:
    aliases:
      - b
    cmds:
      - echo "bar"
```

## Namespace Aliases

Namespace aliases are a way to create alternative names for the namespace of an included taskfile. In the example below, we have adjusted the taskfile so that the included taskfile: `included` is now aliased to `inc`. This means that we can now call `task inc:b` and it will behave the same as `task included:b`:

```yaml
# Taskfile.yml
version: "3"

includes:
  included:
    taskfile: Taskfile2.yml
    aliases:
      - inc

tasks:
  foo:
    aliases:
      - f
    cmds:
      - echo "foo"
      - task: inc:b
```

```yaml
# Taskfile2.yml
version: "3"

tasks:
  bar:
    aliases:
      - b
    cmds:
      - echo "bar"
```

When a task with aliases is merged into a file with a namespace alias, these aliases form a matrix and they are all merged into the task. This means that all of the following are callable:

- `task foo`
- `task f`
- `task included:bar`
- `task included:b`
- `task inc:bar`
- `task inc:b`

## `--list` / `--list-all` / `--summary`

These flags all print information about one or many tasks. Now that we have aliases, these need to be updated. Below are some examples of the new output for each flag when run against the taskfiles in the Namespace Aliases section above:

`task --list` / `task --list-all`:

```txt
* foo|f:                                prints foo
* included:bar|included:b|inc:b:        prints bar
```

Aliases are displayed after the main task name and separated by `|` characters. I'm slightly concerned about how long some of these lines might get with the matrix merging, so we may need to find a better way of displaying this.

`task foo --summary`:

```txt
task: foo (f)

prints foo

aliases:
 - f

commands:
 - echo "foo"
 - Task: inc:b
```

The output above show _two options_ for how we could display aliases in the summary text. Either by having a `|` separated list in brackets next to the task name (similar to the `--list`/`--list-all` flags), or having a dedicated section with a bullet list like the commands - I think I prefer the latter.

### Other noteworthy changes

- There is a longstanding `FIXME` in the codebase about deep copying the tasks when merging an included taskfile. This caused some issues when implementing this change, so I have also added a couple of new utility functions in `taskfile/copy.go` called `deepCopySlice` and `deepCopyMap`. These leverage generics to create a deep copy of slices and maps respectively. These utils are then used in 3 new `DeepCopy` methods on the `Task`, `Vars` and `IncludedTaskfile` structs to create a full deep copy of a task when merging. Hopefully this will help to reduce bugs in the future.
- I'm leveraging the `golang.org/x/exp/slices` package to do generic "slice contains element" lookups. This also means I was able to remove the non-generic version of this function in `taskfile/slice.go`. This package is expected to become a part of the standard library in the next few Go releases.
- Updated `usage.md` and `api_reference.md` docs.
  - The API reference doc was actually missing a couple of other keys (like `label`), so I added these too.
- Updated `changelog.md` doc.


docs/docs/changelog.md
```diff
@@ -5,6 +5,10 @@ sidebar_position: 6
 
 # Changelog
 
+## Unreleased
+
+- Add ability to set `aliases` for tasks and namespaces (#879).
```
andreynering, line 10: - This may be confusing, but the right flow at the moment is to write changes to `/CHANGELOG.md` only, and on release I call `task docs:changelog` which will actually update this file according to this one.
- Unfortunately linking issues/PR with just `#num` won't always work (on the website, for example). We need to make a proper link as `([#num](url))` as the others below.
- Mention https://github.com/go-task/task/issues/268 and https://github.com/go-task/task/issues/340 as well?

docs/docs/usage.md
```diff
@@ -230,6 +230,20 @@ includes:
...
+When including a taskfile, you can give the namespace a list of `aliases`. This works in the same way as [task aliases](#task-aliases) and can be used together to create shorter and easier-to-type commands.
+
+```yaml
+version: '3'
+
+includes:
+  generate:
+    taskfile: ./taskfiles/Generate.yml
+    aliases:
+      - gen
```
andreynering, line 244: What about using one line arrays for `aliases` mentions on documentation? This would be consistent with `deps` and is shorter.

```suggestion
    aliases: [gen]
```

docs/docs/usage.md
```diff
@@ -936,6 +950,27 @@ If the task does not have a summary or a description, a warning is printed.
 
 Please note: *showing the summary will not execute the command*.
 
+## Task aliases
+
+Aliases are alternative names for tasks. They can be used to make it easier and quicker to run tasks with long or hard-to-type names. You can use them on the command line, when [calling sub-tasks](#calling-another-task) in your Taskfile and when [including tasks](#including-other-taskfiles) with aliases from another Taskfile. They can also be used together with [namespace aliases](#namespace-aliases).
```
andreynering, line 955: As a convention, we try to wrap paragraphs in 80 chars when possible.

help.go
```diff
@@ -44,7 +44,8 @@ func (e *Executor) printTasks(listAll bool) {
 	// Format in tab-separated columns with a tab stop of 8.
 	w := tabwriter.NewWriter(e.Stdout, 0, 8, 0, '\t', 0)
 	for _, task := range tasks {
-		fmt.Fprintf(w, "* %s: \t%s\n", task.Name(), task.Desc)
+		taskNames := append([]string{task.Name()}, task.Aliases...)
+		fmt.Fprintf(w, "* %s: \t%s\n", strings.Join(taskNames, "|"), task.Desc)
```
andreynering, line 48: I believe this may possible break the auto-completion scripts that use `--list` to parse the list of available tasks.

Because of this, I think we may need to move the somewhere after. Not sure about the format, but maybe something like this?:

```bash
* foo:                prints foo (aliases: f)
* included:bar        prints bar (aliases: included:b, inc:b)
```

internal/summary/summary.go
```diff
@@ -56,10 +58,25 @@ func printTaskSummary(l *logger.Logger, t *taskfile.Task) {
...
 
+func printTaskAliases(l *logger.Logger, t *taskfile.Task) {
+	if len(t.Aliases) == 0 {
+		return
+	}
+	l.Outf(logger.Default, "")
+	l.Outf(logger.Default, "aliases:")
+	for _, alias := range t.Aliases {
+		l.Outf(logger.Default, " - %s", alias)
+	}
```
andreynering, line 77: I liked the format choice this this 👍

task.go
```diff
@@ -62,17 +63,18 @@ type Executor struct {
...
+	// Check if given tasks exist
+	for _, call := range calls {
+		task, err := e.GetTask(call)
+		if err != nil {
 			e.ListTasksWithDesc()
-			return &taskNotFoundError{taskName: c.Task}
+			return err
 		}
-		if t.Internal {
+		// If the task is internal
```
andreynering, line 74: I don't think this comment is needed. The code is self-descriptive.

```suggestion
```

task.go
```diff
@@ -330,3 +332,35 @@ func (e *Executor) startExecution(ctx context.Context, t *taskfile.Task, execute
...
 }
+
+// GetTask will return the task with the name matching the given call from the taskfile.
+// If no task is found, it will search for tasks with a matching alias.
+// If multiple tasks contain the same alias or no matches are found an error is returned.
+func (e *Executor) GetTask(call taskfile.Call) (*taskfile.Task, error) {
+	// Search for a matching task
+	matchingTask, ok := e.Taskfile.Tasks[call.Task]
+	// If didn't find one, search for a task with a matching alias
+	if !ok {
```
andreynering, line 343: Early returns are preferred. Check for `if ok { return }` and move the current block to after it.

```go
if ok {
  return matchingTask, nil
}

// Rest of the code
```

andreynering (commented):
Awesome work! 👏 

I didn't test it yet, but the code is almost perfect. I did a couple of comments only.

Thanks a lot for your help 🙌

postlund:
I like this a lot 👍 One thing I feel is missing though is aliases across namespaces. In my particular use case (a mono repo), I for instance have various tools under the `tools` directory and each tools have set of generic tasks, e.g.  `build`, `run`, `clean`, etc. to make it look like a build system. So running for instance the "foobar" tool, I would call `tools:foobar:run` and potentially pass some arguments. If "foobar" is a tool that I run often, then typing that would be tedious. So I would much rather have just `foobar` on top-level map to that task. Pseudo-something like this:

```yaml
version: "3"

includes:
  tools: ./tools

tasks:
  foobar:
    alias: tools:foobar:run
```

Could something like this fit within the alias concept somehow?

internal/summary/summary.go
pd93: I've updated the PR and removed the alternative syntax in favour of this. I've also set the color of aliases to Cyan since the colored output is now merged. Wdyt?

docs/docs/changelog.md
pd93: Whoops, yeah that makes sense. I missed that there are two copies of the file. Should be fixed now

pd93:
@postlund "flattening" included files is something I considered, but I decided against it for a few reasons: 

- Firstly, if this were to be a feature, I think it fits a `flatten` keyword in `includes` better than `aliases`.
- Secondly, there has been some discussion regarding this in #273, and the general consensus was against it.
- Lastly, I wanted the first version of aliases to be as simple as possible. Adding flattening would increase the complexity quite a bit as we'd have to decide how to deal with conflicting task names.

I would suggest adding to the conversation in #273 if you're interested in moving this idea forwards.

help.go
pd93: Fine with this. Do we want to print them in a new column? or just inline next to the description. A column might make this info easier to parse.

Also, same question as `--summary` flag below - Should we colour the aliases?

ssbarnea (approved):
That is cool, I do often like to have 1-2 letter aliases, like with git.

andreynering:
Awesome work @pd93!

help.go
andreynering: I end up changing that myself, taking your suggestion of printing in a separate column, and colored.

Issue #867, opened 2022-09-17
Show "Did you mean ...?" message when user mistypes a task name


andreynering:
This is a proposal to suggest the most similar task(s) name(s) when a given task does not exist.

```bash
$ task docs:stat
task: Task "docs:stat" does not exist. Did you mean "docs:start"?
```

This sounds like a good package to use for this: https://github.com/sajari/fuzzy.

PR #880, opened 2022-10-02
Suggest the most similar task name when a given task does not exist

maxpushka:
This PR closes issue #867.

If the fuzzy model fails to find the most similar task name, it will just say "task does not exist" and the suggestion message won't be shown

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -13,11 +13,20 @@ var (
 )
 
 type taskNotFoundError struct {
-	taskName string
+	taskName   string
+	didYouMean string
 }
 
 func (err *taskNotFoundError) Error() string {
-	return fmt.Sprintf(`task: Task %q not found`, err.taskName)
+	if err.didYouMean != "" {
+		return fmt.Sprintf(
+			`task: Task %q does not exist. Did you mean %q?`,
+			err.taskName,
+			err.didYouMean,
+		)
+	}
+
+	return fmt.Sprintf(`task: Task %q does not exist`, err.taskName)
 }
 
 type taskInternalError struct {
--- a/help.go
+++ b/help.go
@@ -78,7 +78,7 @@ func (e *Executor) tasksWithDesc() (tasks []*taskfile.Task) {
 	return
 }
 
-// PrintTaskNames prints only the task names in a Taskfile.
+// ListTaskNames prints only the task names in a Taskfile.
 // Only tasks with a non-empty description are printed if allTasks is false.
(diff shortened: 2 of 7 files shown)


errors.go
```diff
@@ -13,11 +14,22 @@ var (
 )
 
 type taskNotFoundError struct {
-	taskName string
+	taskName   string
+	fuzzyModel *fuzzy.Model
```
andreynering, line 18: I think it would make more sense to assign the computed value of `fuzzyModel.SpellCheck(...)` here instead of a more complex struct (errors are meant to be dumb). What do you think?

```suggestion
	didYouMean string
```

help.go
```diff
@@ -101,6 +101,10 @@ func (e *Executor) ListTaskNames(allTasks bool) {
 	// sort and print all task names
 	sort.Strings(s)
 	for _, t := range s {
-		fmt.Fprintln(w, t)
+		_, err := fmt.Fprintln(w, t)
+		if err != nil {
+			log.Fatal(err)
+			return
+		}
```
andreynering, line 108: I recommend rolling back any changes on this file as it's not needed for this patch and will conflict with other PRs where we're editing it, like #874.

setup.go
```diff
@@ -82,6 +84,21 @@ func (e *Executor) readTaskfile() error {
...
+		model.SetDepth(5)
+
+		var words []string
+		for taskName := range e.Taskfile.Tasks {
+			words = append(words, taskName)
+		}
+
+		model.Train(words)
+		e.fuzzyModel = model
+	}
```
andreynering, line 100: Extract this to its own function like the other below in this file. Could be named `setupFuzzyModel()`, for example.

Also, this function should be called from `Setup()` and not from `readTaskfile()`.

andreynering (commented):
Hi @maxpushka, thanks for working on this PR!

I made a few comments, but once addressed this is good to be merged.

Can you also squash all your commits and keep a single one? Thanks!

maxpushka:
Hi @andreynering, thank you for the suggestions! A simpler error struct is the thing I should have gone on with from the start

pd93 (approved):
This PR looks good to me now. This is a really nice addition to Task!

One quick thing to think about. How do we imagine this working with #879? Should we also consider alias names when fuzzy matching?

andreynering:
Thanks @maxpushka!

Merged with small changes, including training the model with the aliases as well, not that they are available on master.

PR #477, opened 2021-04-21
Evaluate sources also if status is up-to-date

mrngsht:
Hi, 

This PR fixes #411.

(This is my first PR for OSS! Sorry if I made a mistake:bow:)

PR #881, opened 2022-10-03
Clearify that sources/generate and status can be used together

### LINKED ISSUE #788: Docs are not correct sources and status can co-exist
Docs indicate that status overrides/precedes sources. But #477 says otherwise.

https://taskfile.dev/changelog/#v360---2021-07-10

danquah:
Resolves #788 by added a couple of paragraphs and a code-example that clearifies that `sources` / `generates` and `status` can be used in the same task.

I'm not sure whether the code-example should have been even more generic, or the (slightly simplified) real-life example borrowed from https://github.com/go-task/task/pull/477#issuecomment-867808366 is ok?

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -465,6 +465,11 @@ tasks:
     method: timestamp
 ```
 
+In situations where you need more flexibility the `status` keyword can be used.
+You can even combine the two. See the documentation for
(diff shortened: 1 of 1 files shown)


andreynering:
Looks good to me. Thanks @danquah!

PR #882, opened 2022-10-03
Add a contributions guide

pd93:
As per a discussion on Discord, this PR adds a contribution guide to the Taskfile docs.

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -1,0 +1,122 @@
+---
+slug: /contributing/
+sidebar_position: 7
+---
+
+# Contributing
+
+Contributions to Task are very welcome, but we ask that you read this document
+before submitting a PR.
+
+## Before you start
+
+- **Check existing work** - Is there an existing PR? Are there issues discussing
+the feature/change you want to make? Please make sure you consider/address these
+discussions in your work.
+- **Backwards compatibility** - Will your change break existing Taskfiles? It is
+much more likely that your change will merged if it backwards compatible. Is
+there an approach you can take that maintains this compatibility? If not,
+consider opening an issue first so that API changes can be discussed before you
+invest you time into a PR.
+
+## 1. Setup
+
+- **Go** - Task is written in [Go]. We always support the latest two major Go
+  versions, so make sure your version is recent enough.
+- **Node.js** - [Node.js] is used to host Task's documentation server and is
+  required if you want to run this server locally.
+- **Yarn** - [Yarn] is the Node.js package manager used by Task.
+
+## 2. Making changes
+
+- **Code style** - Try to maintain the existing code style where possible and
+  ensure that code is formatted by `gofmt`. We use `golangci-lint` in our CI to
+  enforce a consistent style and best-practise. There's a `lint` command in
+  the Taskfile to run this locally.
+- **Documentation** - Ensure that you add/update any relevant documentation. See
+  the [updating documentation](#updating-documentation) section below.
+- **Tests** - Ensure that you add/update any relevant tests and that all tests
+  are passing before submitting the PR. See the [writing tests](#writing-tests)
+  section below.
+
+### Running your changes
+
+To run Task with working changes, you can use `go run ./cmd/task`. To run a
+development build of task against a test Taskfile in `testdata`, you can use `go
+run ./cmd/task --dir ./testdata/<my_test_dir> <task_name>`.
+
+### Updating documentation
+
+Task uses [Docusaurus] to host a documentation server. This can be setup and run
+locally by using `task docs:setup` and `task docs:start` respectively (requires
+`nodejs` & `yarn`). All content is written in Markdown and is located in the
+`docs/docs` directory. All Markdown documents should have an 80 character line
+wrap limit.
+
+When making a change, consider whether a change to the [Usage Guide](./usage.md)
+is necessary. This document contains descriptions and examples of how to use
+Task features. If you're adding a new feature, try to find an appropriate place
+to add a new section. If you're updating an existing feature, ensure that the
+documentation and any examples are up-to-date. Ensure that any examples follow
+the [Taskfile Styleguide](./styleguide.md).
+
+If you added a new command/flag, ensure that you add it to the [API
+Reference](./api_reference.md).
+
+### Writing tests
+
+Most of Task's test are held in the `task_test.go` file in the project root and
+this is where you'll most likely want to add new ones too. Most of these tests
+also have a subdirectory in the `testdata` directory where any Taskfiles/data
+required to run the tests are stored.
+
+When making a changes, consider whether new tests are required. These tests
+should ensure that the functionality you are adding will continue to work in the
+future. Existing tests may also need updating if you have changed Task's
+behaviour.
+
+## 3. Committing your code
+
+Try to write meaningful commit messages and avoid having too many commits on
+the PR. Most PRs should likely have a single commit (although for bigger PRs it
+may be reasonable to split it in a few). Git squash and rebase is your friend!
+
+## 4. Submitting a PR
+
+- **Describe your changes** - Ensure that you provide a comprehensive
+  description of your changes.
+- **Issue/PR links** - Link any previous work such as related issues or PRs.
+  Please describe how your changes differ to/extend this work.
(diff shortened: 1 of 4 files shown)


pd93:
@andreynering Please feel free to add/remove anything you like from this PR. This is just intended to be a starting point for a discussion.

Some things to consider:
- @danquah mentioned branch naming / commit messages on Discord, but I've not seen any formal requirements on this. I've been (roughly) following [Conventional Commits](https://www.conventionalcommits.org/) and I've noticed others doing this too. I don't think branch naming matters too much as long as they relate to the change and don't break any obvious profanity rules.
- I don't know if there is any formal code style that we are trying to conform to, or whether just saying "try to maintain the existing style" is enough. Should we add a code styleguide at some point? (I personally quite like [gofumpt](https://github.com/mvdan/gofumpt), though it does differ from the existing style in some places.
- There are probably a lot more FAQs we could add.

danquah:
First thoughts from my initial quick read
* Are there any requirements on which version of node should be installed? With regards to Yarn I assume "latest stable" should work fine?
* On the codestyle - As a contributor it helps me a lot that it explicitly says that there is no special code-style per say, but that I should be aware of following the style in whatever file I'm working on. It's much better than not mentioning a code-style just because the project has not landed on anything formal
* Apologies if this is something that has already been discussed - but would it make sense to have some targets in the projects Task-file for eg. running a local build against the Taskfile in testdata. It's out of scope for this PR, but could be looked at afterwards / in parallel
* ❤️ the "Updating documentation" section
* I'm pretty sure the "Writing tests" section will be expanded going forward as there are probably all sorts of tips and tricks that are good to know when testing a task-runner. Eg. how good test-data looks. But, I think the section is pretty good as is.
* Would it be too meta to add paragraph that invites the reader to contribute to making the contribution guide even better should they run in to any problems?

Branching, I 100% agree on branch-naming not being all that important, in particular for forked repos as the name will only show up that one time in the merge-commit.

Commits: The pragmatic way of going about it is to just ask the contributor to keep the commit messages to the point of the change that is being made. Maybe point to a couple of good examples, and then remind the contributor not to do too many  many commits (or do you normally squash them)?

I like Conventional Commits - we should probably be careful to make it clear whether this is a suggestion (as in, if you're in doubt, go look here, they have some good thoughts on the matter), or a requirement. If we're getting closer to the latter, some standardisation of the types/scopes is probably required.

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+Contributions to Task are very welcome, but we ask that you read this document before submitting a PR.
+
+## Before you start
+
+- **Check existing work** - Is there an existing PR? Are there issues discussing the feature/change you want to make? Please make sure you consider/address these discussions in your work.
+- **Backwards compatibility** - Will your change break existing Taskfiles? It is much more likely that your change will merged if it backwards compatible. Is there an approach you can take that maintains this compatibility? If not, consider opening an issue first so that API changes can be discussed before you invest you time into a PR.
+
+## 1. Setup
+
+- **Golang** - Task is written in [Go]. You will need to [install version 1.18+].
```
andreynering, line 17:
```suggestion
- **Go** - Task is written in [Go]. We always support the latest two major Go versions, so make sure your version is recent enough.
```

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+- **Node.js** - [Node.js] is used to host Task's documentation server and is required if you want to run this server locally.
+- **Yarn** - [Yarn] is the Node.js package manager used by Task.
+
+## 2. Making changes
+
+- **Code style** - Try to maintain the existing code style where possible and ensure that code is formatted by `gofmt`.
```
andreynering, line 23: We also use `golangci-lint`. There's a `lint` command on the Taskfile to run it, and CI will fail if it found any offenses as well.

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+- **Documentation** - Ensure that you add/update any relevant documentation. See the [updating documentation](#updating-documentation) section below.
+- **Tests** - Ensure that you add/update any relevant tests and that all tests are passing before submitting the PR. See the [writing tests](#writing-tests) section below.
+
+### Running my changes
+
+To run Task with working changes, you can use `go run ./cmd/task`. To run a development build of task against a test Taskfile in `testdata`, you can use `go run ./cmd/task --dir ./testdata/<my_test_dir> <task_name>`.
```
andreynering, line 29: `cd`ing into a `testdata/` subdirectory also works 🙂

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+
+Task uses [Docusaurus] to host a documentation server. This can be setup and run locally by using `task docs:setup` and `task docs:start` respectively (requires `nodejs` & `yarn`). All content is written in Markdown and is located in the `docs/docs` directory.
+
+When making a change, consider whether a change to the [Usage Guide](./usage.md) is necessary. This document contains descriptions and examples of how to use Task features. If you're adding a new feature, try to find an appropriate place to add a new section. If you're updating an existing feature, ensure that the documentation and any examples are up-to-date. Ensure that any examples follow the [Taskfile Styleguide](./styleguide.md).
+
+If you added a new command/flag, ensure that you add it to the [API Reference](./api_reference.md).
+
+### Writing tests
+
+All of Task's test are held in the `task_test.go` file in the project root. Any new tests should also be added here. Most of these tests also have a subdirectory in the `testdata` directory where any Taskfiles/data required to run the tests are stored.
```
andreynering, line 41: We have a couple of other `_test.go` files, even if most tests are on the mentioned one. Also, creating new test files is not prohibited.

```suggestion
Most of Task's test are held in the `task_test.go` file in the project root. New tests should mostly likely be added here. Most of these tests also have a subdirectory in the `testdata` directory where any Taskfiles/data required to run the tests are stored.
```

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+
+When making a changes, consider whether new tests are required. These tests should ensure that the functionality you are adding will continue to work in the future. Existing tests may also need updating if you have changed Task's behaviour.
```
andreynering, line 43: Let's add an extra paragraph after this one:

> Try to write meaningful commit messages. Also, avoid having too many commits on the PR: most PRs should likely have a single commit (although for bigger PRs it may be reasonable to split it in a few). Git squash and rebase is your friend!

docs/docs/contributing.md
```diff
@@ -0,0 +1,73 @@
...
+
+- **Code style** - Try to maintain the existing code style where possible and ensure that code is formatted by `gofmt`.
+- **Documentation** - Ensure that you add/update any relevant documentation. See the [updating documentation](#updating-documentation) section below.
+- **Tests** - Ensure that you add/update any relevant tests and that all tests are passing before submitting the PR. See the [writing tests](#writing-tests) section below.
+
+### Running my changes
+
+To run Task with working changes, you can use `go run ./cmd/task`. To run a development build of task against a test Taskfile in `testdata`, you can use `go run ./cmd/task --dir ./testdata/<my_test_dir> <task_name>`.
+
+### Updating documentation
```
andreynering, line 31: A convention that needs to be mentioned here is trying to wrap most lines in 80 chars, a common convention for markdown files.

Also, can you do that for this file as well? Thanks 😄

andreynering (commented):
Looks like a great start @pd93, thanks for working on this!

I made a couple of comments about things to improve, but this does not need to be perfect on the first interaction. We can improve this document with time. 🙂

docs/docs/contributing.md
pd93: I'm not quite sure how to phrase this. `cd`ing into another directory would require the contributor  to `go install` a development build or run something like `go run ../../cmd/task ...` which seems a little confusing for a guide. Do you have a suggestion?

docs/docs/contributing.md
pd93: I've added an extra section as commit messages felt quite distinct from the other sections.

docs/docs/contributing.md
andreynering: Yeah, I think we can keep it as you wrote for now. No changes required.

PR #884, opened 2022-10-05
feat: add a new --json to better supports interactivity in scripts 

### LINKED ISSUE #764: Add flag to output JSON info to be used by editor extensions, etc
When integrating with other tools (like vscode extension, or when trying to script exporting tasks to github actions) we really want to avoid having to parse the free-form output, especially as this will likely change and break integration, even between minor versions.

If we would output as JSON, we can make the parsing far more reliable and less likely to break between versions, especially as if we add extra data, it would not render the original format obsolete.

Things to be included in JSON output:
- all available tasks as keys inside a mapping
- each task dependencies
- task description and summary
- task status (built or not)

davidalpert:
resolves #764

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -57,8 +57,7 @@ func main() {
 		versionFlag bool
 		helpFlag    bool
 		init        bool
-		list        bool
-		listAll     bool
+		listOptions task.ListOptions
 		status      bool
 		force       bool
 		watch       bool
@@ -78,8 +77,8 @@ func main() {
 	pflag.BoolVar(&versionFlag, "version", false, "show Task version")
 	pflag.BoolVarP(&helpFlag, "help", "h", false, "shows Task usage")
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yaml in the current folder")
-	pflag.BoolVarP(&list, "list", "l", false, "lists tasks with description of current Taskfile")
-	pflag.BoolVarP(&listAll, "list-all", "a", false, "lists tasks with or without a description")
+	pflag.BoolVarP(&listOptions.ListWithDescriptionsOnly, "list", "l", false, "lists tasks with description of current Taskfile")
+	pflag.BoolVarP(&listOptions.ListAll, "list-all", "a", false, "lists tasks with or without a description")
 	pflag.BoolVar(&status, "status", false, "exits with non-zero exit code if any of the given tasks is not up-to-date")
 	pflag.BoolVarP(&force, "force", "f", false, "forces execution even when the task is up-to-date")
 	pflag.BoolVarP(&watch, "watch", "w", false, "enables watch of the given task")
@@ -159,8 +158,12 @@ func main() {
 		OutputStyle: output,
 	}
 
-	if (list || listAll) && silent {
-		e.ListTaskNames(listAll)
+	if err := listOptions.Validate(); err != nil {
+		log.Fatal(err)
+	}
+
+	if (listOptions.ShowList()) && silent {
+		e.ListTaskNames(listOptions)
 		return
 	}
 
@@ -173,13 +176,8 @@ func main() {
 		return
 	}
 
-	if list {
-		e.ListTasksWithDesc()
-		return
-	}
-
-	if listAll {
-		e.ListAllTasks()
+	if listOptions.ShowList() {
+		e.ListTasks(listOptions)
 		return
 	}
 
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -79,6 +79,7 @@ func main() {
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yaml in the current folder")
 	pflag.BoolVarP(&listOptions.ListWithDescriptionsOnly, "list", "l", false, "lists tasks with description of current Taskfile")
 	pflag.BoolVarP(&listOptions.ListAll, "list-all", "a", false, "lists tasks with or without a description")
+	pflag.BoolVarP(&listOptions.AsJson, "json", "j", false, "lists tasks as JSON")
 	pflag.BoolVar(&status, "status", false, "exits with non-zero exit code if any of the given tasks is not up-to-date")
 	pflag.BoolVarP(&force, "force", "f", false, "forces execution even when the task is up-to-date")
 	pflag.BoolVarP(&watch, "watch", "w", false, "enables watch of the given task")
--- a/help.go
+++ b/help.go
@@ -1,6 +1,7 @@
 package task
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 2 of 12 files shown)


ssbarnea:
Can you please also update the docs to give an example of output file?

If you want I can easily write a JSON Schema specification once I have the example, I happened that I authored quite a few in the recent years.

davidalpert:
@ssbarnea let's first validate that this output is what you expect, then I am happy to update the docs.

## `--list-all --silent --json`
Given
```
go run ./cmd/task/task.go -t ./testdata/env/Taskfile.yml --list-all --silent --json | jq .
```
I get 
```
[
  "default",
  "global",
  "local"
]
```
## `--list-all --json`

Given
```
go run ./cmd/task/task.go -t ./testdata/env/Taskfile.yml --list-all --json | jq .
```
I get
```
[
  {
    "task": "default",
    "cmds": [
      {
        "Cmd": "",
        "Silent": false,
        "Task": "local",
        "Vars": null,
        "IgnoreError": false,
        "Defer": false
      },
      {
        "Cmd": "",
        "Silent": false,
        "Task": "global",
        "Vars": null,
        "IgnoreError": false,
        "Defer": false
      }
    ],
    "deps": null,
    "label": "",
    "desc": "",
    "summary": "",
    "sources": null,
    "generates": null,
    "status": null,
    "preconditions": null,
    "dir": "",
    "vars": null,
    "env": null,
    "silent": false,
    "interactive": false,
    "internal": false,
    "method": "",
    "prefix": "",
    "ignore_error": false,
    "run": "",
    "include_vars": null,
    "included_taskfile_vars": null,
    "included_taskfile": null
  },
  {
    "task": "global",
    "cmds": [
      {
        "Cmd": "echo \"FOO='$FOO' BAR='$BAR' BAZ='$BAZ'\" > global.txt",
        "Silent": false,
        "Task": "",
        "Vars": null,
        "IgnoreError": false,
        "Defer": false
      }
    ],
    "deps": null,
    "label": "",
    "desc": "",
    "summary": "",
    "sources": null,
    "generates": null,
    "status": null,
    "preconditions": null,
    "dir": "",
    "vars": null,
    "env": {
      "keys": [
        "BAR"
      ],
      "mapping": {
        "BAR": {
          "static": "overriden",
          "live": null,
          "sh": "",
          "dir": ""
        }
      }
    },
    "silent": false,
    "interactive": false,
    "internal": false,
    "method": "",
    "prefix": "",
    "ignore_error": false,
    "run": "",
    "include_vars": null,
    "included_taskfile_vars": null,
    "included_taskfile": null
  },
  {
    "task": "local",
    "cmds": [
      {
        "Cmd": "echo \"GOOS='$GOOS' GOARCH='$GOARCH' CGO_ENABLED='$CGO_ENABLED'\" > local.txt",
        "Silent": false,
        "Task": "",
        "Vars": null,
        "IgnoreError": false,
        "Defer": false
      }
    ],
    "deps": null,
    "label": "",
    "desc": "",
    "summary": "",
    "sources": null,
    "generates": null,
    "status": null,
    "preconditions": null,
    "dir": "",
    "vars": {
      "keys": [
        "AMD64"
      ],
      "mapping": {
        "AMD64": {
          "static": "amd64",
          "live": null,
          "sh": "",
          "dir": ""
        }
      }
    },
    "env": {
      "keys": [
        "GOOS",
        "GOARCH",
        "CGO_ENABLED"
      ],
      "mapping": {
        "CGO_ENABLED": {
          "static": "",
          "live": null,
          "sh": "echo '0'",
          "dir": ""
        },
        "GOARCH": {
          "static": "{{.AMD64}}",
          "live": null,
          "sh": "",
          "dir": ""
        },
        "GOOS": {
          "static": "linux",
          "live": null,
          "sh": "",
          "dir": ""
        }
      }
    },
    "silent": false,
    "interactive": false,
    "internal": false,
    "method": "",
    "prefix": "",
    "ignore_error": false,
    "run": "",
    "include_vars": null,
    "included_taskfile_vars": null,
    "included_taskfile": null
  }
]

```

Notes:
- this PR only supports `--json` output on the `--list` and `--list-all` flags (with and without `--silent`)
- this PR does not include task status, as the `--list` and `--list-all` flags do not include task status
- including task status seems to require a larger refactoring or alternative design optimized for the use cases you are trying to support (or a generalized version of them)
- in order to support JSON output more widely (i.e. for all subcommands) I would recommend introducing an abstraction over output (e.g. in tools like `kubectl` I have seen a global(ish) `--output` format flag) something like [`go-printers`](https://github.com/davidalpert/go-printers) which I put together adapting a few ideas I've seen around the community.

davidalpert:
I can add struct tags to `Cmd`; I must have missed that one.

JSON key formatting is easy to switch up if `camelCase` is preferred; I tend to default to `snake_case` but I am also indifferent here.

davidalpert:
updated the sample JSON output

andreynering:
Hi @davidalpert, thanks for opening this PR.

I think this will need to be more advanced, though. Instead of just dumping tasks in JSON, I would like to have a separated package to handle this, with handcrafted structs with just the things that are relevant to code editors. It'll need to have variables in its computed state, for example.

Also, we'll probably need to ability to output only for a specific task as well, because always outputting the whole thing may be a performance issue on big Taskfiles, specially when dynamic variables are in place.

Let me know if you would still like to continue working on this. I'm open to give you the direction needed. If you don't want or won't have the time for that, just let us know so someone else will be free to work on it without having a conflict with your work.

ssbarnea:
Bit of feedback unrelated to proposed implementation: 

- Keep JSON schema inside task project itself (separated means more maintenance work and it will almost always lag behind)
- Don't try to get it perfect from start, in fact I would suggest an [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) approach, like only outputting the rendered tasks and their names for start. Set additionalProperties=true to allow later extension without breaking the older schema.
- Avoid over-engineering by trying to build schema dynamically. It might give you the impression that it would be easier but in long run it would be quite costly. Doing it by hand is fine as long you have positive/negative tests to validate it.

Over the last 3 years, I wrote 10+ JSON schemas for various projects, including big ones like Ansible. I made many mistakes, but I think I did learn from them (these bullet points being part of conclusions).

davidalpert:
Hi @andreynering 

Thank you for the feedback. I initially picked #764 during Hactoberfest as a way into this code base but I am happy to continue the effort to refine this PR (or rewrite it from the ground up) so that it can align with your expectations. Being new to the project I wasn't sure that this was the right way to approach this request, so I appreciate the guidance.

Specifically, you mentioned:
- a separate package with handcrafted structs - do you have a package name or a list of what code editors need? shall we take @ssbarnea's suggestion and start with an MVP and then iteratively add more detail?  Sorin originally requested this set of details:
    - all available tasks as keys inside a mapping
    - each task's dependencies
    - task description and summary
    - task status (built or not)
- ability to output for a specific task - can you jot down some notes about the CLI experience you have in mind, in terms of the various combinations of flags and arguments you envision?

Please jot down whatever level of detail you have in mind in terms of what you would like to see this PR turn into. As I mentioned, this was an initial spike into the code for me, so I am happy to let this one go and start over in whatever direction you prefer.

davidalpert:
in fact, looking at the code changes in October I would like to close this PR, take the lessons learned and the discussion, and start fresh with your guidance as detailed as you would like to give it.

andreynering:
Closing in favor of #936.

PR #886, opened 2022-10-06
global variables, as well as task level variable can now be exported, (closes #181)

### LINKED ISSUE #181: Is it possible to send all the variables to the environment?
I want to run something like Helmsman which takes environment variables and substitutes them into a template. I want to be able to use the variables defined in `Taskvars.yml` in the environment but I don't have to have to explicitly include each one like:
```
tasks:
  test:
    env:
      ENV1: {{.ENV1}}
      ENV2: {{.ENV2}}
    cmds:
      - helmsman
```
is there a way to do something like:

```
tasks:
  test:
    env:
      setAllVariablesToEnvironment: true
    cmds:
      - helmsman
```

nxtcoder17:
I have added the ability to export task vars into Env for the subsequent commands. I needed this feature, while building kubernetes operators with make, as it expects values from Env

It works like this
screenshot-14316

- export global variables and task variables to all tasks, by setting `exportVars: true` globally
- export only task variables, by setting `exportVars: true` on individual task

### DIFF
--- a/task.go
+++ b/task.go
@@ -65,6 +65,25 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 	// check if given tasks exist
 	for _, c := range calls {
 		t, ok := e.Taskfile.Tasks[c.Task]
+
+		if e.Taskfile.ExportVars && e.Taskfile.Vars != nil {
+			for k, v := range e.Taskfile.Vars.Mapping {
+				if t.Env == nil {
+					t.Env = &taskfile.Vars{}
+				}
+				t.Env.Set(k, v)
+			}
+		}
+
+		if (e.Taskfile.ExportVars || t.ExportVars) && t.Vars != nil {
+			for k, v := range t.Vars.Mapping {
+				if t.Env == nil {
+					t.Env = &taskfile.Vars{}
+				}
+				t.Env.Set(k, v)
+			}
+		}
+
 		if !ok {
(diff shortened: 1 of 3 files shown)


pd93:
@nxtcoder17 Thanks for the PR. What's the advantage of this over just using the `env` key instead of `vars`? See my example below:

```yaml
version: "3"

env:
  v1:
    sh: echo "Hello"
  v2: "world"

tasks:
  sample:
    desc: "hi"
    silent: true
    cmds:
      - echo "$v1, $v2"
```

nxtcoder17:
resolves #181

andreynering:
Hi @nxtcoder17,

Thanks for taking your time to work into this PR, but I don't this is the right direction here.

For example, in a (far in the future) next major version, I want to separate `vars` and `env` even more, while today `env`s are accessible as `var`s  as well, which is confusing. In this way this PR goes to the opposite direction on what we plan to do.

I recommend users to keep `vars` for interpolation and `env` for things that needs to be exposed to external commands. If you really need both, you can do something like this:

```yaml
vars:
  FOO: foo
  BAR: bar

env:
  FOO: '{{.FOO}}'
  BAR: '{{.BAR}}'
```

nxtcoder17:
ok, i personally needed this feature, and also saw an open issue, so i implemented a pull request. Any way, you guys have been doing great work, Taskfile has been a very integral part of my workflow, kudos to the team

PR #890, opened 2022-10-10
refactor: implement task list filtering

### LINKED ISSUE #806: On `--list`, show tasks in the root Taskfile first
Currently, when running `task -l` the available tasks are listed alphabetically. 
While this works for simple taskfiles, in cases with several includes and many compound tasks in the root file, I'd hoped that those tasks defined in the root taskfile are listed first (visually separated by an empty line).

Example:
Given two taskfiles to be included:
```yaml
# taskfile.lib.yaml
---
version: 3
tasks:
  some_task: 
    desc: greet someone
    cmds: ['echo "hello"']
---
# taskfile.tools.yaml
---
version: 3
tasks:
  some_other_task: 
    desc: point somewhere
    cmds: ['echo "there"']
---
```

The files are included in the root taskfile:

```yaml
# Taskfile.yaml
version: 3
include:
  lib:
    taskfile: taskfile.lib.yaml
  tools:
    taskfile: taskfile.tools.yaml
tasks:
  main_task:
    desc: do the thing.
    cmds:
      - task: lib:some_task
      - task: tools:some_other_task
```

The output would currently look like this:
```yaml
* lib:some_task                    greet someone
* main_task                        do the thing.
* tools:some_other_task            point somewhere
```

I'm suggesting it should group the tasks like this instead:

```yaml
* main_task                        do the thing.  # several tasks in the root file can be listed alphabetically again

* lib:some_task                    greet someone
* tools:some_other_task            point somewhere
```

Alternatively, add an option to `include` to not list the included tasks, even if they have a `desc` attribute:

```yaml
include:
  lib:
    taskfile: taskfile.lib.yaml
    hide: true
  tools:
    taskfile: taskfile.tools.yaml
    hide: true
```

pd93:
This PR refactors the code used to fetch the list of tasks when outputting them to the CLI. This is used for the `--list` and `--list-all` flags as well as when any errors are displayed to the user.

## Why?

This code has a lot of duplication in it and isn't very extensible. i.e. it requires a lot of custom code and bloat to add new features.

## Shiny things

I've added a `GetTaskList` function to the executor (this nicely matches the new `GetTask` function in #879). This function will return a sorted list of tasks from the task map and filter any tasks according to the list of given filters. These filters can be any function that matches the new `FilterFunc` type. 

I have implemented the two filter functions that are currently required in the codebase: 

- `FilterNoDesc` - Removes tasks with no description
- `FilterInternal` - Removes tasks that are marked as internal

To add a new filter, all you'd need to do is create a function that takes a single task, decides if it should be filtered or not and return true/false accordingly. This function should then be wrapped with the `Filter` helper function which does all the actual filtering for you.

Note: The `Filter` function uses an in-place, zero-allocation method for filtering. This is much more efficient than creating a new slice for each filter and using `append` many times. See [this StackOverflow post](https://stackoverflow.com/a/20551116/1599633) for details.

## Sorting

This PR also happens to fix #806. The duplication of the task list sorting was what led me down this path in the first place. The sort function has been replaced with something that will put tasks without a `:` first.

It's worth noting that #806 requested that the tasks in the root Taskfile get sorted at the top of the list. This will not necessarily happen as if you create a task called `foo:bar` in your root Taskfile, it will see it as a namespace. However, I think this behaviour is fine (good, even).

## Other

Worth noting that the error handling for there being no tasks has moved from `help.go` into the `cmd/task.go` file. This is because functions aren't comparable and therefore `ListTasks` can't check if the `FilterNoDesc` function was given or not and decide which error to show.

This has the knock-on effect that the `task: No tasks with description available. Try --list-all to list all tasks` error message will no longer display when an error occurs and there are no tasks with descriptions. I personally don't mind this as it felt like a side-effect anyway, but I wanted to gather thoughts from others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Tasks in the root Taskfile will now be displayed first in `--list`/`--list-all`
+  output ([#806](https://github.com/go-task/task/pull/806), [#890](https://github.com/go-task/task/pull/890)).
 - It's now possible to call a `default` task in an included Taskfile by using
   just the namespace. For example: `docs:default` is now automatically
   aliased to `docs`
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -177,12 +177,16 @@ func main() {
 	}
 
 	if list {
-		e.ListTasksWithDesc()
+		if ok := e.ListTasks(task.FilterOutInternal(), task.FilterOutNoDesc()); !ok {
+			e.Logger.Outf(logger.Yellow, "task: No tasks with description available. Try --list-all to list all tasks")
+		}
 		return
 	}
 
 	if listAll {
-		e.ListAllTasks()
+		if ok := e.ListTasks(task.FilterOutInternal()); !ok {
+			e.Logger.Outf(logger.Yellow, "task: No tasks available")
+		}
 		return
 	}
 
--- a/help.go
+++ b/help.go
@@ -10,34 +10,15 @@ import (
 	"text/tabwriter"
 
 	"github.com/go-task/task/v3/internal/logger"
-	"github.com/go-task/task/v3/taskfile"
 )
 
-// ListTasksWithDesc reports tasks that have a description spec.
-func (e *Executor) ListTasksWithDesc() {
-	e.printTasks(false)
-}
-
-// ListAllTasks reports all tasks, with or without a description spec.
-func (e *Executor) ListAllTasks() {
-	e.printTasks(true)
-}
-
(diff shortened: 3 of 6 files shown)


andreynering (approved):
Looks good! 👏 

I'd consider renaming `FilterInternal` and `FilterNoDesc`, though, because the current name looks like we're filtering _only_ with desc or _only_ internal instead of filtering _out_ them.

Could be something like `FilterOutInternal` or `RejectInternal` (ideas, no strong opinion).

Once addressed, you can merge this PR yourself 🙂. A changelog entry mentioned the sorting of `:` first is welcome as well.

PR #893, opened 2022-10-11
Update prefixed output to use the label, not name (if set)

jonathanio:
Although the `--list` and `task: ...` output from `task` use the `label:` argument for the task itself (if present), the prefixed output from each of the `cmds` uses the original name of the task rather than the label. As such, when running multiple tasks (especially in parallel) it can be challenging to know which version of the task the output belongs to.

This change allows the prefixed output to use the task's `label:` argument instead if it's been provided.


jonathanio:
I didn't see the `prefix:` argument, so closing.

Issue #842, opened 2022-08-22
Install script is not preperly detecting the ARM v7 architecture


varac:
Currently, Taskfile doesn't install on RPi 4 (armv7):

```
pi@pi:~ $ sudo sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d
go-task/task crit platform linux/armv7 is not supported.  Make sure this script is up-to-date and file request at https://github.com/go-task/task/issues/new
```

Please provide armv7 binaries.

andreynering:
Hi @varac,

Does the binary work in your machine if you download `task_linux_arm.tar.gz` from the [releases page](https://github.com/go-task/task/releases)?

I'm trying to understand if the problem is the binary or just the download script.

We use v6 for ARM binaries which is the default on Goreleaser.

varac:
I just tried the latest `task_linux_arm.tar.gz ` on a `Raspberry Pi 4 Model B Rev 1.1` (armv7l) and it works.
So it must be an issue with the installer script.

PR #895, opened 2022-10-13
fix: add armv5 & armv7 builds

Eun:
# Description
This pull request adds builds for armv5 and armv7 architectures.

# Notes 
Should also fix https://github.com/go-task/task/issues/842

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,7 +11,9 @@ build:
     - arm
     - arm64
   goarm:
+    - 5
     - 6
+    - 7
   ignore:
     - goos: darwin
       goarch: 386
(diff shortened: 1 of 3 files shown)


andreynering:
Hi @Eun!

Some concerns:

1) Changing the existing file name from `arm.tar.gz` to `armv6.tar.gz` is not a good idea because it will break automation around downloading these binaries, not only our own automation but also scripts from users.
2) I believe that supporting ARMv6 should be enough because in theory v7 devices should be able to run v6 binaries. Am I wrong? Never used ARM apart from Mac M1.
3) I also believe that v5 devices are rare to be worth supporting?

Eun:
@andreynering 
Thanks for the feedback, all your comments make total sense.
I reverted the changes and unified the logic for armv5, armv6, armv7 in the install script.

So this should only fix #842

andreynering:
Thanks @Eun!

PR #897, opened 2022-10-14
Improved fish shell completions

mymmrac:
- Added check for non-zero exit code of `task --list-all` that happens if there is no `Taskfile` present or etc.
- Added comments
- Added fish to `.editorconfig`

### DIFF
--- a/completion/fish/task.fish
+++ b/completion/fish/task.fish
@@ -1,15 +1,23 @@
 set GO_TASK_PROGNAME task
 
 function __task_get_tasks --description "Prints all available tasks with their description"
-	set -l output ($GO_TASK_PROGNAME --list-all | sed '1d; s/\* \(.*\):\s*\(.*\)/\1\t\2/' | string split0)
+  # Read the list of tasks (and potential errors)
+  $GO_TASK_PROGNAME --list-all 2>&1 | read -lz rawOutput
+
+  # Return on non-zero exit code (for cases when there is no Taskfile found or etc.)
+  if test $status -ne 0
+    return
+  end
+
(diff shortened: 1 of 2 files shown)


completion/fish/task.fish
```diff
@@ -27,3 +35,4 @@ complete -c $GO_TASK_PROGNAME -s t -l taskfile  -d 'choose which Taskfile to run
 complete -c $GO_TASK_PROGNAME -s v -l verbose   -d 'enables verbose mode'
 complete -c $GO_TASK_PROGNAME      -l version   -d 'show Task version'
 complete -c $GO_TASK_PROGNAME -s w -l watch     -d 'enables watch of the given task'
+
```
andreynering, line 38: This extra empty line here is not needed. (Should have a single empty line, we now have two).

andreynering (approved):
Hi @mymmrac, thanks for this PR!

I see you changed indentation from spaces to tabs. Can you use 2 space indention? That's what we've been using for all non-Go files.

I'd also appreciate if you added `fish` to the file extension list here:

https://github.com/go-task/task/blob/6335878317390ad5535d73d9898e91028d75b538/.editorconfig#L11

Squashing all changes and keeping a single commit is preferred.

completion/fish/task.fish
mymmrac: Hmm, previously there were no new line at the end, so I added it, right now it looks like this:
image

mymmrac: Removed that to match all other files

andreynering:
Perfect, thanks @mymmrac!

Issue #389, opened 2020-10-15
Allow `dotenv` per task


zx8:
Currently, `dotenv` is a global configuration.

It would be great if we could specify a different `dotenv` configuration depending on the task we're invoking in addition to the current global option, similar to how `vars` and `env` work.

shinebayar-g:
Yeah this would work nice since `env` can be declared at task level. Currently `env` and `dotenv` configurations are inconsistent because of this.

andreynering:
Hi @zx8,

I'm not ure this is a common enough use case. But I'll keep this issue open for a while to collect more opinons.

andreynering:
@shinebayar-g That's a valid suggestion, but I suggest you to open another issue about it.

Making feature request in comments of an other issue makes it way more likely that this will just be forgotten.

zx8:
@andreynering I don't think @shinebayar-g is saying anything different to the OP? They seem to just be reenforcing what I'm suggesting in this feature request.

andreynering:
@zx8 You're right, I totally misread that comment. I was responding to many issues/comments earlier today and definitely made some confusion.

Sorry.

rugwirobaker:
I need this 2.

SwitchedToGitlab:
+1 on this.

Use case: 

I use Task (exceptional tool btw) to build OCI containers for use with Podman (or Docker). Most images are child images of a parent image. When building a child image, the parent sometimes needs to be re-build for the latest updates. Each image has it's own dotenv file, with a dozen or more variables. As it is now the only way I can re-build a parent image from the child image is to include all the variables in the Taskfile, which is inconvenient and makes the Taskfile much longer and more difficult to read.

Having the dotenv file read at the task level would allow including the parent image for rebuild much easier and simpler, and in my opinion more predictable for the user. 

I hope that makes sense. 

Thanks!!!!

Hades32:
In case it helps, this is my current workaround:
```yaml
  my-task:
    cmds:
    -  npm install
    - |
      export $(grep -v '^\s*#' ./{{.stage}}/env.ini | xargs)
      command --that --needs the-env-vars
```

ghostsquad:
I'll plan on doing this in v4 which is my current priority.

ToshY:
> I'll plan on doing this in v4 which is my current priority.

@ghostsquad could you give an ETA on v4?

ghostsquad:
> > I'll plan on doing this in v4 which is my current priority.
> 
[...]

PR #904, opened 2022-10-21
feat: add task-level dotenv support

pd93:
This PR adds support for the task-level `dotenv` key - A [much requested feature](https://github.com/go-task/task/issues/389).

The key works much the same as it does at a global level:

```yaml
version: '3'

tasks:
  default:
    dotenv: ['.env']
    cmds:
      - echo "$FOO"
```

It also supports filenames as variables (like the global key):

```yaml
version: '3'

tasks:
  default:
    vars:
      ENV: .env
    dotenv: ['{{.ENV}}']
    cmds:
      - echo "$FOO"
```

- Global `env` and `dotfile` variables will be overridden by task-level `dotfile` variables.
- Task-level `dotfile` variables will be overridden by task-level `env` variables.


pd93:
This is a note-to-self more than anything. I've just spotted `ErrIncludedTaskfilesCantHaveDotenvs`. I need to look into why this restriction exists and whether we need to add something for task-level `dotenv`

https://github.com/go-task/task/blob/a160baae9df97a7e8077c2fff151a3a4089ef2ea/taskfile/read/taskfile.go#L124-L126

pd93:
> I've just spotted `ErrIncludedTaskfilesCantHaveDotenvs`. I need to look into why this restriction exists and whether we need to add something for task-level `dotenv`

@andreynering I've added 552802cf5310e0b2319b0553af821ceb792132bc to address this, but I'm still unsure as to what the original decision was to omit dotfiles from included Taskfiles. Is this something that can be revisited now?

> any ETA on when this will come into a release ?

@nxtcoder17 No ETA. It's still a WIP. It will be released once its ready and another maintainer has time to review it.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add task-level `dotenv` support ([#389](https://github.com/go-task/task/issues/389)).
```
andreynering, line 5: You can mention this own PR (#904) here as well.

docs/static/schema.json
```diff
@@ -134,7 +141,7 @@
           "method": {
             "description": "Defines which method is used to check the task is up-to-date. `timestamp` will compare the timestamp of the sources and generates files. `checksum` will check the checksum (You probably want to ignore the .task folder in your .gitignore file). `none` skips any validation and always run the task.",
             "type": "string",
-            "enum": ["none", "checksum", "timestamp"],
+            "enum": [ "none", "checksum", "timestamp" ],
```
andreynering, line 144: I mentioned this on another PR IIRC, but I'd revert this style change as it's not common IMO

```suggestion
            "enum": ["none", "checksum", "timestamp"],
```

taskfile/read/taskfile.go
```diff
@@ -121,7 +121,15 @@ func Taskfile(readerNode *ReaderNode) (*taskfile.Taskfile, error) {
...
-		if v >= 3.0 && len(includedTaskfile.Dotenv) > 0 {
+		tasksWithDotenv := false
+		for _, v := range includedTaskfile.Tasks {
+			if len(v.Dotenv) > 0 {
+				tasksWithDotenv = true
+				break
+			}
+		}
+
+		if v >= 3.0 && (len(includedTaskfile.Dotenv) > 0 || tasksWithDotenv) {
```
andreynering, line 132: Hey @pd93,

Regarding your question on this...

I believe that the reason I added this validation at the time was because, if we allowed dotenv on included Taskfiles, this would cause an unintended side-effect how things work today on the code. Basically the env in the included Taskfile would be _merged_ into the main Taskfile. Given that:

1) I believe you can revert your changes in this validation here and keep the way it works on master
2) In a future PR, we could allow dotenv on included Taskfiles, but that would require a refactor so these env would be _scoped_ in the included Taskfile and not merged into the main Taskfile. In short, dotenv would only affect tasks on its own Taskfile and not all tasks. At least that's how I would expect it to work, but this is open to discussion, and we may want to open an issue about this.

andreynering (approved):
Hi @pd93, thanks for working on this!

I tried to answer your question in a comment below, and I don't think we can undo the changes you made on the validation.

Other than that, if you don't have further questions, you're free to merge this PR to master yourself. 🙂

docs/static/schema.json
pd93: Whoops. I'm not doing this on purpose, I swear 🤣

taskfile/read/taskfile.go
pd93: I suspect I'm misunderstanding slightly, but these sound contradictory to me?

> I don't think we can undo the changes you made on the validation

> you can revert your changes in this validation here and keep the way it works on master

Could you please clarify for me. Would you like to keep the changes in 4614c77 or revert them?

andreynering: Hey @pd93,

Sorry for the confusion.

I meant keep the same behavior as master. Taskfiles with _global_ dotenvs would be problematic (until we refactor to _scope_ them to be valid for tasks in that specific Taskfile only). Tasks with local dotenvs are fine, because they are already scoped by design.

I just reverted your last commit and merged to master. Thanks! 🙂

PR #919, opened 2022-10-31
feat: include aliases in --list --silent output

pd93:
This PR adds aliases to the `--list`/`--list-all` flags when combined with `--silent`.

I wasn't 100% sure about this one, but I tried to tab-complete an aliased task today and it didn't work. Do we think that users should be able to tab complete aliased tasks or is this not needed because tab-completion will fill in the long name anyway? Thoughts appreciated.

image


andreynering:
Thanks @pd93!

PR #924, opened 2022-11-04
fix: log the checked files if verbose

aminya:
This adds a logging step to the function that checks if the inputs/generates are up to date. After hours of debugging, It helped me to finally detect that the globbing library that `task` uses [fails to check some patterns](https://github.com/mattn/go-zglob/issues/38#issuecomment-1302839605).

The following checks the files inside the `dev` folder
```
"{dev/**/*,something}"
```

But this doesn't!
```
"{something,dev/**/*}"
```

Or for example, `**` doesn't check all the files of the current project.

Performance-wise, this is not a good solution because it calls the glob outside the ` checker.IsUpToDate` function. I did this because that was the only way I could access the logger without changing the checker API. If you are OK with the API change, we can remove the duplicate glob calls.

### DIFF
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -34,7 +34,7 @@ func (c *Checksum) IsUpToDate() (bool, error) {
 	data, _ := os.ReadFile(checksumFile)
 	oldMd5 := strings.TrimSpace(string(data))
 
-	sources, err := globs(c.TaskDir, c.Sources)
+	sources, err := Globs(c.TaskDir, c.Sources)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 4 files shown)


andreynering:
Hi @aminya, thanks for creating this PR.

I gave this some thought and decided to close because:

1) `sources`/`generates` can be _a lot_ a files (if `**/*.go` in a big repo for example) and can potentially generate a lot of noise in the logs, which could annoy users debugging other stuff
2) We plan to add another feature to allow this kind of debugging. We'll probably have a `--debug` flag some day to output useful data about the Taskfiles, like computed variables, sources, generates, and more.

andreynering:
I just opened #931 to track this idea.

aminya:
Thanks. Could you please make an issue for the Glob errors too? The glob package that you use is not validated

PR #926, opened 2022-11-04
BASH: Also complete with task aliases

MarioSchwalbe:
A minor update for the new alias feature as of task 3.17.0:
- Also propose task aliases next to task names.

Note: A better approach might be to include aliases in the output of `task --list --silent` on extra lines.

### DIFF
--- a/completion/bash/task.bash
+++ b/completion/bash/task.bash
@@ -11,7 +11,7 @@ function _task()
   local i
   for i in "${!words[@]}"; do
     if [ "${words[$i]}" == "--" ]; then
-      # Do not complete words following `--` passed to CLI_ARGS.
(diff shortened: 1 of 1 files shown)


pd93:
> A better approach might be to include aliases in the output of task --list --silent on extra lines.

See https://github.com/go-task/task/pull/919

andreynering:
Hi @MarioSchwalbe,

With #919 merged to master, this is PR still useful?

MarioSchwalbe:
> Hi @MarioSchwalbe,
> With #919 merged to master, this is PR still useful?

No. Since `task --list --silent` now includes task aliases we don't have to extract them the hard way anymore. Thanks.

PR #929, opened 2022-11-10
feat: set and shopts directives

pd93:
## Summary

This PR adds 2 new directives:

- `set` - `[]string` - Equivalent of the POSIX `set` builtin
- `shopts` - `[]string` - Equivalent of the Bash `shopt` builtin

Both of these directives are available at global, task and command level. See example Taskfile below:

```yaml
version: '3'

# Global-level - Set for all commands in all tasks
set: [e] # Shorthand flags work for set too
shopts: [globstar]

tasks:
  default:
    # Task-level - Set for all command in this task
    set: [pipefail]
    shopts: [globstar]
    cmds:
      # Command-level - Set for just this command
      - cmd: shopt
        shopts: [globstar]
      - cmd: set -o
        set: [allexport]
```

## Considerations

- [`mvdan/sh`](https://github.com/mvdan/sh) allows us to set POSIX shell options via `interp.Params()`. However, from what I can tell, there is no equivalent for Bash options. This means that this PR is currently prepending the `shopt` builtin before each command where `shopts` are specified. I'm not a huge fan of this and have been looking at contributing a change to `mvdan/sh` to add a new `interp.Shopts()` function or similar. I'd love some input from @mvdan on whether this is a good approach.
- Originally, I implemented both directives as a `map[string]bool` so that you could enable/disable opts at different levels. However, I came to the conclusion that this level of complexity wasn't really necessary as all the [currently supported shell options](#currently-supported-shell-options) are disabled by default anyway. Implementing as a `[]string` still allows us to enable at each level and I think this covers most use cases. It also keeps the Taskfile looking clean as we can inline the list with `[]`.
- I did also consider merging the two directives as none of the option names conflict. However, this seems confusing and would require us to maintain a list of both inside Task so that we know how to pass them into the interpreter. I think its easier just to keep them separate.
- The names `set` and `shopts` are definitely up for debate. I wasn't sure about the plurality for example...

### [POSIX (`set`)](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)

If you have [`gosh`](https://github.com/mvdan/sh#gosh) installed, you can find these by running `gosh -c "set -o"`.

| Option Name | Short Name | Default |
| ----------- | ---------- | ------- |
| `allexport` | `a`        | off     |
| `errexit`   | `e`        | off     |
| `noexec`    | `n`        | off     |
| `noglob`    | `f`        | off     |
| `nounset`   | `u`        | off     |
| `xtrace`    | `x`        | off     |
| `pipefail`  |            | off     |

### [Bash (`shopt`)](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html)

If you have [`gosh`](https://github.com/mvdan/sh#gosh) installed, you can find these by running `gosh -c "shopt"`.

| Option Name      | Default |
| ---------------- | ------- |
| `expand_aliases` | off     |
| `globstar`       | off     |
| `nullglob`       | off     |

### Don't have `gosh`?

If you don't have `gosh` installed, you can use the following Taskfile to get these lists:

```yaml
version: '3'

tasks:
  default:
    cmds:
      - set -o
      - shopt
```


internal/execext/exec.go
```diff
@@ -17,12 +18,14 @@ import (
...
-	Command string
-	Dir     string
-	Env     []string
-	Stdin   io.Reader
-	Stdout  io.Writer
-	Stderr  io.Writer
+	Command   string
+	Dir       string
+	Env       []string
+	POSIXOpts []string
```
andreynering, line 24:
```suggestion
	PosixOpts []string
```

internal/execext/exec.go
```diff
@@ -36,6 +39,31 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
...
+
+	var s []string
+
+	// Manually adding bash opts before each command as there is no interp.Shopts
+	if len(opts.BashOpts) > 0 {
+		s = append(s, fmt.Sprintf("shopt -s %s", strings.Join(opts.BashOpts, " ")))
+	}
+
+	if len(s) > 0 {
+		opts.Command = strings.Join(append(s, opts.Command), " && ")
```
andreynering, line 64: Adding `interp.BashOpts` would be better, but if we need to do this ourselves, I would do it differently.

We can call `r.Run` multiple times, so it's just a matter of calling every `shopt` command before the user command. This would be safer than changing `opts.Command`.

andreynering (commented):
First look.

I liked the direction here, but adding `interp.BashOpts` to `mvdan/sh` first would be great IMO.

pd93:
@andreynering I've fixed the issues you raised, but haven't had the time to look into the codebase over at mvdan/sh yet. Instead, I've opened an issue (https://github.com/mvdan/sh/issues/962) over there instead. Hopefully I (or someone else) will have the time to work on this soon. In the meantime, if you're happy to merge this as-is, I can open another issue to track the improvements once (if) the functionality is added upstream.

andreynering (approved):
Yes, let's get this merged and we can improve once https://github.com/mvdan/sh/issues/962 gets implemented.

I've renamed `shopts` to `shopt` for consistency with the proper builtin. Also added documentation and README entry.

Thanks again!

marco-m:
I love
```
tasks:
  default:
    # Task-level - Set for all command in this task
    set: [pipefail]
```

thanks!!!

pd93:
> Yes, let's get this merged and we can improve once [mvdan/sh#962](https://github.com/mvdan/sh/issues/962) gets implemented.
> 
> I've renamed `shopts` to `shopt` for consistency with the proper builtin. Also added documentation and README entry.

@andreynering Thanks for tidying this up. I forgot to add the docs/schema etc. when marking as ready. I've opened #984 to track the improvements as discussed.

docs/docs/usage.md
```diff
@@ -1420,6 +1420,31 @@ tasks:
...
 
+## `set` and `shopt`
+
+It's possible to specify options to the
+[`set`](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)
+and [`shopt`](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html)
+builtins. This can be added at global, task or command level.
+
+```yaml
+version: '2'
```
marco-m-pix4d, line 1431: @pd93 I think that this is an error and should be instead: 
```
version: '3'
```
?

pd93: @marco-m-pix4d good spot! yes, this should be `version: '3'`. I'll get this updated.

PR #936, opened 2022-11-17
feat: add --json flag to be used by editor extensions

### LINKED ISSUE #764: Add flag to output JSON info to be used by editor extensions, etc

davidalpert:
a new `--json` flag works only with the `--list` and `--list-all` flags to format those task lists as JSON output.

- can't use `--json` by itself:
```
go run ./cmd/task --taskfile ./testdata/output_group/Taskfile.yml hello --json | jq .
task: --json only applies to --list or --list-all
exit status 1
```
- `--json --list` ignores tasks that are missing descriptions:
```
go run ./cmd/task --taskfile ./testdata/output_group/Taskfile.yml hello --json --list | jq .
{
  "tasks": []
}
```
- `--json --list-all` includes all tasks
```
go run ./cmd/task --taskfile ./testdata/output_group/Taskfile.yml hello --json --list-all | jq .
{
  "tasks": [
    {
      "name": "bye",
      "desc": "",
      "summary": "",
      "up_to_date": false
    },
    {
      "name": "hello",
      "desc": "",
      "summary": "",
      "up_to_date": false
    }
  ]
}
```

this PR includes a temporary WIP (work in progress) commit with additional detail printed to the console around the three failing tests; this commit is intended to help illustrate what is happening when these tests expect that we print nothing to the console so that we can review and confirm that we want to fix the code to maintain that expectation or perhaps update the expectation to provide more detailed output to clients and users.

may resolve #764

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -184,8 +184,10 @@ func main() {
 	}
 
 	if listOptions.ShouldListTasks() {
-		_ = e.ListTasks(listOptions)
-		return // if we are listing tasks, we don't expect a task to run
+		if foundTasks, err := e.ListTasks(listOptions); !foundTasks || err != nil {
+			os.Exit(1)
+		}
+		return
 	}
 
 	var (
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -59,6 +59,7 @@ func main() {
 		init        bool
 		list        bool
 		listAll     bool
+		listJson    bool
 		status      bool
 		force       bool
 		watch       bool
@@ -81,6 +82,7 @@ func main() {
 	pflag.BoolVarP(&init, "init", "i", false, "creates a new Taskfile.yaml in the current folder")
 	pflag.BoolVarP(&list, "list", "l", false, "lists tasks with description of current Taskfile")
 	pflag.BoolVarP(&listAll, "list-all", "a", false, "lists tasks with or without a description")
+	pflag.BoolVarP(&listJson, "json", "j", false, "formats task list as json")
 	pflag.BoolVar(&status, "status", false, "exits with non-zero exit code if any of the given tasks is not up-to-date")
 	pflag.BoolVarP(&force, "force", "f", false, "forces execution even when the task is up-to-date")
 	pflag.BoolVarP(&watch, "watch", "w", false, "enables watch of the given task")
@@ -162,7 +164,12 @@ func main() {
 		OutputStyle: output,
 	}
 
-	if (list || listAll) && silent {
+	var listOptions = task.NewListOptions(list, listAll, listJson)
+	if err := listOptions.Validate(); err != nil {
+		log.Fatal(err)
+	}
+
+	if (listOptions.ShouldListTasks()) && silent {
 		e.ListTaskNames(listAll)
 		return
 	}
@@ -176,18 +183,9 @@ func main() {
 		return
 	}
 
-	if list {
-		if ok := e.ListTasks(task.FilterOutInternal(), task.FilterOutNoDesc()); !ok {
-			e.Logger.Outf(logger.Yellow, "task: No tasks with description available. Try --list-all to list all tasks")
-		}
-		return
-	}
-
-	if listAll {
-		if ok := e.ListTasks(task.FilterOutInternal()); !ok {
-			e.Logger.Outf(logger.Yellow, "task: No tasks available")
-		}
-		return
+	if listOptions.ShouldListTasks() {
+		_ = e.ListTasks(listOptions)
+		return // if we are listing tasks, we don't expect a task to run
 	}
 
 	var (
--- a/help.go
+++ b/help.go
@@ -1,8 +1,10 @@
 package task
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/go-task/task/v3/taskfile"
 	"io"
 	"log"
 	"os"
@@ -65,9 +67,14 @@ func (o ListOptions) Filters() []FilterFunc {
 func (e *Executor) ListTasks(o ListOptions) (bool, error) {
 	tasks := e.GetTaskList(o.Filters()...)
 	if o.FormatTaskListAsJSON {
+		output, err := e.ToEditorOutput(tasks)
+		if err != nil {
+			return false, err
+		}
+
 		encoder := json.NewEncoder(e.Stdout)
 		encoder.SetIndent("", "  ")
-		if err := encoder.Encode(editors.ToOutput(tasks)); err != nil {
+		if err := encoder.Encode(output); err != nil {
 			return false, err
 		}
 
@@ -131,4 +138,23 @@ func (e *Executor) ListTaskNames(allTasks bool) {
 	for _, t := range s {
 		fmt.Fprintln(w, t)
 	}
+}
+
+func (e *Executor) ToEditorOutput(tasks []*taskfile.Task) (*editors.Output, error) {
+	o := &editors.Output{
+		Tasks: make([]editors.Task, len(tasks)),
+	}
+	for i, t := range tasks {
+		upToDate, err := e.isTaskUpToDate(context.Background(), t)
+		if err != nil {
+			return nil, err
+		}
+		o.Tasks[i] = editors.Task{
+			Name:     t.Name(),
+			Desc:     t.Desc,
+			Summary:  t.Summary,
+			UpToDate: upToDate,
+		}
+	}
+	return o, nil
 }
--- a/help.go
+++ b/help.go
@@ -1,6 +1,7 @@
 package task
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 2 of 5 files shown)


help.go
```diff
@@ -1,7 +1,9 @@
 package task
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/go-task/task/v3/internal/editors"
```
andreynering, line 6: This should be moved to the second import group, together with `"github.com/go-task/task/v3/internal/logger"`.

help.go
```diff
@@ -12,12 +14,65 @@ import (
...
+
+// ShouldListTasks returns true if one of the options to list tasks has been set to true
+func (o ListOptions) ShouldListTasks() bool {
+	return o.ListOnlyTasksWithDescriptions || o.ListAllTasks
+}
+
+// Validate validates that the collection of list-related options are in a valid configuration
+func (o ListOptions) Validate() interface{} {
+	if o.ListOnlyTasksWithDescriptions && o.ListAllTasks {
+		return fmt.Errorf("cannot use --list and --list-all at the same time")
```
andreynering, line 41: We prepend `task: ` to errors that are from Task itself.

This a common Go pattern to know where an error came from. And it's useful here because Tasks prints outputs from other CLIs as well.

```suggestion
		return fmt.Errorf("task: cannot use --list and --list-all at the same time")
```

help.go
```diff
@@ -12,12 +14,65 @@ import (
...
+	}
+	if o.FormatTaskListAsJSON && !(o.ShouldListTasks()) {
+		return fmt.Errorf("--json only applies to --list or --list-all")
```
andreynering, line 44:
```suggestion
		return fmt.Errorf("task: --json only applies to --list or --list-all")
```

help.go
davidalpert: fixed in a rebased commit

help.go
davidalpert: fixed in a rebased commit

help.go
davidalpert: fixed in a rebased commit

cmd/task/task.go
```diff
@@ -176,18 +183,9 @@ func main() {
...
-	}
-
-	if listAll {
-		if ok := e.ListTasks(task.FilterOutInternal()); !ok {
-			e.Logger.Outf(logger.Yellow, "task: No tasks available")
-		}
-		return
+	if listOptions.ShouldListTasks() {
+		_ = e.ListTasks(listOptions)
+		return // if we are listing tasks, we don't expect a task to run
```
andreynering, line 188: Needs to exit with error if printing failed.

```suggestion
		if ok := e.ListTasks(listOptions); !ok {
			os.Exit(1)
		}
		return
```

help.go
```diff
@@ -9,15 +10,69 @@ import (
...
+	}
+}
+
+// ShouldListTasks returns true if one of the options to list tasks has been set to true
+func (o ListOptions) ShouldListTasks() bool {
+	return o.ListOnlyTasksWithDescriptions || o.ListAllTasks
+}
+
+// Validate validates that the collection of list-related options are in a valid configuration
+func (o ListOptions) Validate() interface{} {
```
andreynering, line 39:
```suggestion
func (o ListOptions) Validate() error {
```

help.go
```diff
@@ -9,15 +10,69 @@ import (
...
+	if o.ListOnlyTasksWithDescriptions && o.ListAllTasks {
+		return fmt.Errorf("task: cannot use --list and --list-all at the same time")
+	}
+	if o.FormatTaskListAsJSON && !(o.ShouldListTasks()) {
```
andreynering, line 43:
```suggestion
	if o.FormatTaskListAsJSON && !o.ShouldListTasks() {
```

help.go
```diff
@@ -9,15 +10,69 @@ import (
...
 // ListTasks prints a list of tasks.
 // Tasks that match the given filters will be excluded from the list.
-// The function returns a boolean indicating whether or not tasks were found.
-func (e *Executor) ListTasks(filters ...FilterFunc) bool {
-	tasks := e.GetTaskList(filters...)
+// The function returns a boolean indicating whether tasks were found.
+func (e *Executor) ListTasks(o ListOptions) bool {
+	tasks := e.GetTaskList(o.Filters()...)
+	if o.FormatTaskListAsJSON {
+		_ = json.NewEncoder(os.Stdout).Encode(editors.ToOutput(tasks))
```
andreynering, line 67: 1. Use `executor.Stdout` instead of global ones (may be overriden by tests, usage as a library, etc)
2. Print with indentation
3. Handle error (need to change function signature and callers)

```suggestion
		encoder := json.NewEncoder(e.Stdout)
		encoder.SetIndent("", "  ")
		if err := encoder.Encode(editors.ToOutput(tasks)); err != nil {
			return false, err
		}
```

internal/editors/output.go
```diff
@@ -0,0 +1,38 @@
...
+	Summary string `json:"summary"`
+
+	// "up-to-date" vs. "out-of-date"? Alternatively could be a "UpToDate bool"
+	//Status string `json:"status"`
+	UpToDate bool `json:"up_to_date"`
+
+	//// These could be added on the future. Don't need to be on the MVP
+	//Env map[string]string `json:"env"`
+	//Vars map[string]string `json:"vars"`
+}
```
andreynering, line 23:
```suggestion
type Task struct {
	Name     string `json:"name"`
	Desc     string `json:"desc"`
	Summary  string `json:"summary"`
	UpToDate bool   `json:"up_to_date"`
}
```

internal/editors/output.go
```diff
@@ -0,0 +1,38 @@
...
+func ToOutput(tasks []*taskfile.Task) *Output {
+	o := &Output{
+		Tasks: make([]Task, len(tasks)),
+	}
+	for i, t := range tasks {
+		o.Tasks[i] = Task{
+			Name:     t.Name(),
+			Desc:     t.Desc,
+			Summary:  t.Summary,
+			UpToDate: false, // TODO
```
andreynering, line 34: This fields needs to be either implemented or removed (if you decide to implement in another PR).

task.go
```diff
@@ -72,13 +72,11 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 	for _, call := range calls {
 		task, err := e.GetTask(call)
 		if err != nil {
-			e.ListTasks(FilterOutInternal(), FilterOutNoDesc())
-			return err
+			return fmt.Errorf("task: executor.Run(): executor.GetTask(): %v", err)
```
andreynering, line 75: No need to wrap the error here. The extra context is not useful to the user and it's not making the test break.

```suggestion
			return err
```

What is making some tests break is the printing of tasks on `e.ListTasks`. With your change, it's not being print automatically anymore when the user tries to run a task that does not exist. That needs to be added back. Maybe in another place? I don't know, need to figure it out where.

task.go
```diff
@@ -72,13 +72,11 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
...
 		}
 
 		if task.Internal {
-			e.ListTasks(FilterOutInternal(), FilterOutNoDesc())
-			return &taskInternalError{taskName: call.Task}
+			return fmt.Errorf("task: executor.Run(): %v", &taskInternalError{taskName: call.Task})
```
andreynering, line 79: Same here

```suggestion
			return err
```

andreynering (commented):
Hi @davidalpert,

Good progress! I added a few things to be addressed before this can be merged.

task.go
davidalpert: I put back the original `&taskInternalError{taskName: call.Task}` because there is a unit test expecting an error there and `err` is nil, but it has the effect you were looking for.

davidalpert:
I will look into restoring the task list when a task is not found.

in the meantime, let me know if you see anything else which needs addressing.

once it's set I will rebase to squash it all together into a single `feat:` commit

task.go
davidalpert: I tried putting this
```
_, _ = e.ListTasks(ListOptions{ListOnlyTasksWithDescriptions: true})
```
in place of restoring
```
e.ListTasks(FilterOutInternal(), FilterOutNoDesc())
```
but now two more tests are breaking, expecting nothing where they are now getting
```
task: No tasks with description available. Try --list-all to list all tasks\n
```

I think that the old `ListTasks(...)` method didn't print anything if it just filtered out all the tasks while the new one picked up that "try --list-all" method from somewhere else.

probably I tried reducing duplication and tripped over two places in the code that were almost the same but not quite.

what do you think?

andreynering:
Thanks for your contribution @davidalpert!

I fixed the test issues on 957bff4b89584fd06e93dee298d4e66c5c32adf3.

max-sixty:
Quick message to say very excited to see this! 

Would love to see a VSCode integration — `task` is much better than the native VSCode tasks (as well as working outside of VSCode...), and would be awesome if we could move towards `task` being the standard way of running things. And I could imagine this being the basis for even more — e.g. creating CI / GitHub action specs from task definitions.

PR #938, opened 2022-11-18
JetBrains JSON Schema Validator Error Workaround

### LINKED ISSUE #847: Taskfile schema fits multiple variants error
- Task version: 3.12.1
- Operating System: macos x64

Using the current schema complains that multiple variants fit the same file.

This was caused by #757 because using the schema from before that commit works fine.

```yaml
# https://taskfile.dev

version: '3'

# Just here for convenience
includes:
  devops:
    taskfile: "./devops/Taskfile.dist.yaml"

tasks:
  default:
    cmds:
      - stuff

```

rrrix:
`schema.json` Workaround for these two JetBrains issues:

- [IDEA-236928 json schema: False positive "Schema validation: Validates to more than one variant" with "oneOf"](https://youtrack.jetbrains.com/issue/IDEA-236928)
- [IDEA-265710 JSON schema validation fails if anyOf is nested in oneOf](https://youtrack.jetbrains.com/issue/IDEA-265710)

There is no material effect on the actual schema validation, as the `oneOf`/`allOf` evaluates only one list entry.

`allOf([1])` is `1`, and `oneOf([1])` is also `1`.

Fixes #847

### DIFF
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -260,7 +260,7 @@
       }
     }
   },
-  "oneOf": [
+  "allOf": [
     {
       "type": "object",
       "properties": {


andreynering:
Thanks @rrrix!

PR #957, opened 2022-12-17
FAQ doc

pd93:
As discussed - First draft of a new page on the documentation site containing frequently asked questions. This will allow us to direct users asking these common questions to carefully worded, well-explained answers. Hopefully this will save us some time when answering questions and give us a unified stance on these topics.

@andreynering I've left the "Why Task" section as WIP for now. I feel like it might be better for you to answer this one as its your project. As a user of Task, my personal reasons for using Task over the alternatives would be:

- Simple/Standard YAML configuration is much easier to use than a custom syntax (Make) or a full language (Mage).
- Written in Go (fast, x-platform, single binary install)

If you can think of any other questions to add, please feel free to modify the PR, or add a comment and I can give it a go :)

Also, WDYT about the menu ordering? I wasn't sure where it should go:

image

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -1,0 +1,54 @@
+---
+slug: /faq/
+sidebar_position: 5
+---
+
+# FAQ
+
+This page contains a list of frequently asked questions about Task.
+
+- [Why won't my task update my shell environment?](#why-wont-my-task-update-my-shell-environment)
+- ['x' builtin command doesn't work on Windows](#x-builtin-command-doesnt-work-on-windows)
+
+## Why won't my task update my shell environment?
+
+This is a limitation of how shells work. Task runs as a subprocess of your
+current shell, so it can't change the environment of the shell that started it.
+This limitation is shared by other task runners and build tools too.
+
+A common way to work around this is to create a task that will generate output
+that can be parsed by your shell. For example, to set an environment variable on
+your shell you can write a task like this:
+
+```yaml
+my-shell-env:
+  cmds:
+    - echo "export FOO=foo"
+    - echo "export BAR=bar"
+```
+
+Now run `eval $(task my-shell-env)` and the variables `$FOO` and `$BAR` will be
+available in your shell.
+
+## 'x' builtin command doesn't work on Windows
+
+The default shell on Windows (`cmd` and `powershell`) do not have commands like
+`rm` and `cp` available as builtins. This means that these commands won't work.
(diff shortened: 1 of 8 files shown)


docs/docs/faq.md
```diff
@@ -0,0 +1,63 @@
...
+    - echo "export FOO=foo"
+    - echo "export BAR=bar"
+```
+
+Now run `eval $(task my-shell-env)` and the variables `$FOO` and `$BAR` will be
+available in your shell.
+
+## 'x' builtin command doesn't work on Windows
+
+The default shell on Windows (`cmd` and `powershell`) do not have commands like
```
andreynering, line 44: Just a FYI, it seems to PowerShell _do_ allow some commands like `cp` but they're actually aliases to other commands with a different interface, so far from being 100% compatible.

- https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/copy-item?view=powershell-7.3
- https://ilovepowershell.com/2011/11/03/list-of-top-powershell-alias/

docs/docs/faq.md
```diff
@@ -0,0 +1,63 @@
...
+`rm` and `cp` available as builtins. This means that these commands won't work.
+If you want to make your Taskfile fully cross-platform, you'll need to work
+around this limitation using one of the following methods:
+
+- Use the `{{.OS}}` variable to run an OS-specific script.
```
andreynering, line 49: You mean the `{{OS}}` function, right? It returns in the Go format: `windows`, `linux`, `darwin`.

https://github.com/go-task/task/blob/0307ca8ac6eaaa79b89602bce8176db1625a54c1/internal/templater/funcs.go#L19

docs/docs/faq.md
```diff
@@ -0,0 +1,63 @@
...
+- Use something like `{{if eq .OS "Windows_NT"}}powershell {{end}}<my_cmd>` to
```
andreynering, line 50:
```suggestion
- Use something like `{{if eq OS "windows"}}powershell {{end}}<my_cmd>` to
```

andreynering (approved):
Nice work, thanks for giving it a start!

Added a few commands, but otherwise feel free to merge it _if_ you're done for now.

EDIT: I don't have a strong opinion about the order, but I think I'd put it right _after_ "API Reference".

docs/docs/faq.md
pd93: Ah yeah, my bad. For some reason, I had it in my head that this was just an environment variable called `OS`, but that's coming from Windows, not Task.

PR #958, opened 2022-12-17
feat: add local-prefixes flag to goimports linter

pd93:
I noticed that you added `goimports` as a linter. This change extends that by forcing users to separate internal imports (ones starting with `github.com/go-task/task`) from other non standard lib imports. This is a pattern followed elsewhere in the project, but not currently enforced.

VSCode users can add this to their settings to do this formatting automatically:

```jsonc
{
    "gopls": {
        "formatting.local": "github.com/go-task/task",
    }
    ...
}
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,3 +6,7 @@
 linters:
   enable:
     - goimports
+
+linters-settings:
+  goimports:
+    local-prefixes: github.com/go-task/task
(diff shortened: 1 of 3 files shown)


andreynering:
I thought this was the default behavior by just invoking `goimports`, but looks that I was wrong.

Good catch!

PR #959, opened 2022-12-17
feat: improve unmarshal error handling and use v3 yaml interface ever…

pd93:
Small changes to the Unmarshal functions to make them all use the newer v3 YAML interface. The flow of these functions has changed in a couple of places (e.g. `task.go`) since we can now tell the type of the `Node` ahead of unmarshalling. We can rely on error handling instead of having to check the state of the unmarshalled object.

### DIFF
--- a/taskfile/cmd.go
+++ b/taskfile/cmd.go
@@ -1,5 +1,11 @@
 package taskfile
 
+import (
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


andreynering:
Very welcome refactor. Thanks! 👏

PR #964, opened 2022-12-22
Close Taskfile after reading it

### LINKED ISSUE #963: Taskfile.yml remains open during task execution
- Task version: latest/`master` branch
- Operating System: Windows 10

Hello! I've recently gotten this error on GitHub Desktop when switching branches after having modified the Taskfile, while some tasks were running:

...which led me to discover that after reading `Taskfile.yaml`, Task does not `Close()` the file, which I believe is what's causing the error above.

The current workaround is to close all running tasks (which forcefully closes open files) and **then** switch branches.

The relevant code can be found here: https://github.com/go-task/task/blob/master/taskfile/read/taskfile.go#L195,L205

Is this behavior expected or is it in fact an oversight during development? I've written a PR that should fix this: #964

HeCorr:
Properly close `Taskfile.yml` after reading it.

This should fix issues preventing modifications to the Taskfile while tasks are still running, like switching git branches for example.

Fixes #963

### DIFF
--- a/taskfile/read/taskfile.go
+++ b/taskfile/read/taskfile.go
@@ -197,6 +197,7 @@ func readTaskfile(file string) (*taskfile.Taskfile, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer f.Close()
 	var t taskfile.Taskfile
 	if err := yaml.NewDecoder(f).Decode(&t); err != nil {
 		return nil, fmt.Errorf("task: Failed to parse %s:\n%w", filepathext.TryAbsToRel(file), err)


pd93 (approved):
Thanks for the PR. I don't think this is intentional. We should definitely be closing the file once it's read into memory. Keeping the file open when running the actual tasks (even if its `O_RDONLY`) isn't necessary.

andreynering:
Thanks @HeCorr!

And @pd93, feel free to merge PRs to master when additional discussion isn't needed. 🙂

PR #965, opened 2022-12-23
fix: incorrect schema type for interval

### LINKED ISSUE #962: JSONSchema `interval` incorrect
image

In https://taskfile.dev/schema.json interval uses `"$ref": "#/definitions/3/run"` which seems invalid.

pd93:
Fixes #962

### DIFF
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -370,7 +370,8 @@
         },
         "interval": {
           "description": "Sets a different watch interval when using `--watch`, the default being 5 seconds. This string should be a valid Go duration: https://pkg.go.dev/time#ParseDuration.",
-          "$ref": "#/definitions/3/run"
+          "type": "string",
+          "pattern": "^[0-9]+(?:m|s|ms)$"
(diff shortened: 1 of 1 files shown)


docs/static/schema.json
```diff
@@ -370,7 +370,7 @@
         },
         "interval": {
           "description": "Sets a different watch interval when using `--watch`, the default being 5 seconds. This string should be a valid Go duration: https://pkg.go.dev/time#ParseDuration.",
-          "$ref": "#/definitions/3/run"
+          "type": "string"
```
pd93, line 373: @andreynering We could also do something like this so that it validates the syntax of the string. Do we need to support anything other than `ms` and `s`?

```suggestion
          "type": "string"
          "pattern": "^[0-9]+m?s$"
```

andreynering: Although minute (`m`) is unlikely to be used, I'd add it just to be safe.

andreynering (approved):

PR #967, opened 2022-12-23
feat: add action for when an issue is closed

pd93:
Small addition to the GitHub actions to tidy up `needs triage` labels if someone closes an issue before it is triaged.

Screenshot from my test repo below:

image

No change if issue has already had the `needs triage` label removed:

image

### DIFF
--- a/.github/workflows/issue-closed.yml
+++ b/.github/workflows/issue-closed.yml
@@ -1,11 +1,11 @@
-name: issue needs triage
+name: issue closed
 
 on:
   issues:
-    types: [opened]
+    types: [closed]
 
 jobs:
-  issue-needs-triage:
+  issue-closed:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/github-script@v6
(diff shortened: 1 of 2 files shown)


.github/workflows/issue-closed.yml
```diff
@@ -0,0 +1,28 @@
...
+  issues:
+    types: [closed]
+
+jobs:
+  issue-closed:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
```
andreynering, line 13: Hey @pd93,

We have the @task-bot account which is already used for some automation like the website build & publish process at the [gh-pages](https://github.com/go-task/task/tree/gh-pages) branch.

I think we can use it for these scripts as well. Its tokens is set to `GH_PAT`. We need to apply it to the other GA workflow files as well.

Doc: https://github.com/actions/github-script#using-a-separate-github-token

```suggestion
          github-token: ${{secrets.GH_PAT}}
          script: |
```

andreynering (approved):

PR #970, opened 2022-12-29
fix: watch interval

### LINKED ISSUE #969: 'interval' directive in Taskfile.yaml has no effect
- Task version: 3.19.0 (or manual build from commit [c7d9efe](https://github.com/go-task/task/commit/c7d9efebf9e6ef6d94f3efeae12f7d8d6e53e65b))
- Operating System: Linux

Given this `Taskfile.yaml`:
````
# https://taskfile.dev

version: '3'

interval: 500ms

tasks:
  default:
    sources:
    - test_watch.txt
    cmds:
      - ksuid > test_watch.txt
````

Running:
````
task -w -I 500ms
````
executes the task every 500ms, as expected. However, running:
````
task -w
````
runs it every *5 seconds*, ignoring the `interval: 500ms` directive.

I think that this section:
````
if e.Interval != "" {
	watchIntervalString = e.Interval
} else if e.Taskfile.Interval != "" {
	watchIntervalString = e.Taskfile.Interval
}
````
in [watch.go](https://github.com/go-task/task/blob/e4158dc5e465a28bfeaa927ce9adb1cf352ea4aa/watch.go) will always match on the first "if" statement since `e.Interval` will always be set to **5s** by default if not specified on the commandline due to:
````
pflag.StringVarP(&interval, "interval", "I", "5s", "interval to watch for changes")
````
in [task.go](https://github.com/go-task/task/blob/e4158dc5e465a28bfeaa927ce9adb1cf352ea4aa/cmd/task/task.go).

### PR DESCRIPTION
**Pull request by pd93**
Fixes #969 by setting the default value of the pflag to 0 instead of "5s".

Additionally, this PR changes both the `Taskfile` struct and the `interval` flag variable to use `time.Duration` directly rather than a string. This still allows the user to pass values using "5s" or "500ms" etc, but defers the passing of the value to the yaml and pflag libraries respectively. This slightly simplifies the `watch.go` file.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"runtime/debug"
 	"strings"
+	"time"
 
 	"github.com/spf13/pflag"
 	"mvdan.cc/sh/v3/syntax"
@@ -74,7 +75,7 @@ func main() {
 		entrypoint  string
 		output      taskfile.Output
 		color       bool
-		interval    string
+		interval    time.Duration
 	)
 
 	pflag.BoolVar(&versionFlag, "version", false, "show Task version")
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cmd/task/task.go [hunk 1]
```diff
@@ -99,7 +100,7 @@ func main() {
 	pflag.StringVar(&output.Group.End, "output-group-end", "", "message template to print after a task's grouped output")
 	pflag.BoolVarP(&color, "color", "c", true, "colored output. Enabled by default. Set flag to false or use NO_COLOR=1 to disable")
 	pflag.IntVarP(&concurrency, "concurrency", "C", 0, "limit number tasks to run concurrently")
-	pflag.StringVarP(&interval, "interval", "I", "5s", "interval to watch for changes")
+	pflag.DurationVarP(&interval, "interval", "I", 0, "interval to watch for changes")
```
andreynering, line 103: I didn't know pflag was able to parse durations. Good catch!

andreynering (approved):

PR #971, opened 2022-12-30
fix: add missing nil check

### LINKED ISSUE #961: Task throwing runtime error, 'invalid memory address'
I'm doing a fresh go development environment and installed task with 'go install' and 'choco install'.  In both cases I get the following.

```
PS C:\Users\xxx\workspace\semaphore> task --version
Task version: v3.19.0 (h1:jS1s2fHoSgYySw+kHEtbSxS6emE620unBsDDcNqdRbc=)
PS C:\Users\xxx\workspace\semaphore> task deps
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x21 pc=0x932dda]

goroutine 1 [running]:
github.com/go-task/task/v3/taskfile.Merge(0xc000138000, 0xc000138240, 0x0, {0x0, 0x0, 0x0})
        github.com/go-task/task/v3@v3.19.0/taskfile/merge.go:48 +0x41a
github.com/go-task/task/v3/taskfile/read.Taskfile(0xc0000f4d20)
        github.com/go-task/task/v3@v3.19.0/taskfile/read/taskfile.go:178 +0x2a5
github.com/go-task/task/v3.(*Executor).readTaskfile(0xc0000fc5a0)
        github.com/go-task/task/v3@v3.19.0/setup.go:83 +0x85
github.com/go-task/task/v3.(*Executor).Setup(0xc0000fc5a0)
        github.com/go-task/task/v3@v3.19.0/setup.go:29 +0x49
main.main()
        github.com/go-task/task/v3@v3.19.0/cmd/task/task.go:170 +0xb6a
```

OS Info...

Edition	Windows 10 Pro
Version	22H2
Installed on	‎12/‎5/‎2022
OS build	19045.2364
Experience	Windows Feature Experience Pack 120.2212.4190.0

pd93:
Fixes #961.

This problem occurs when a user uses `version: '2'` in their Taskfile in combination with a `Taskfile_{{OS}}.yml` file. In the case of #961, I'm guessing there was only a `Taskfile_windows.yml` file which is why the issue doesn't repro on Linux.

This happens because the `Merge` function is called with `includedTaskfile` set to `nil`:

https://github.com/go-task/task/blob/5be1f6bb2b612eeeae21a1c5bea6e7cd0e0f852e/taskfile/read/taskfile.go#L171-L182

This value is then not being nil checked in the `Merge` function here:

https://github.com/go-task/task/blob/5be1f6bb2b612eeeae21a1c5bea6e7cd0e0f852e/taskfile/merge.go#L48

The fix simply adds a nil check to this line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Small bug fix: closing "Taskfile.yml" once we're done reading it
   ([#963](https://github.com/go-task/task/issues/963), [#964](https://github.com/go-task/task/pull/964) by @HeCorr).
+- Fixes a bug in v2 that caused a panic when using a `Taskfile_{{OS}}.yml` file
+  ([#961](https://github.com/go-task/task/issues/961), [#971](https://github.com/go-task/task/pull/971) by @pd93).
 - Add `--json` flag (alias `-j`) with the intent to improve support for code
   editors and add room to other possible integrations. This is basic for now,
   but we plan to add more info in the near future
--- a/taskfile/merge.go
+++ b/taskfile/merge.go
@@ -45,7 +45,7 @@ func Merge(t1, t2 *Taskfile, includedTaskfile *IncludedTaskfile, namespaces ...s
 
 		// Set the task to internal if EITHER the included task or the included
 		// taskfile are marked as internal
-		task.Internal = task.Internal || includedTaskfile.Internal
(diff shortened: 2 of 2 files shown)


andreynering:
Thanks!

Issue #681, opened 2022-03-06
Out-of-date dependency force execution of an up-to-date task.


ConX:
## TL;DR

Force the execution of all dependent tasks when a dependency (defined by `deps`) is out-of-date.  

_Note: A more detailed discussion can be found in [here](https://github.com/go-task/task/discussions/669).  To assess the popularity of this feature please upvote (👍) or downvote it (👎 )_ 

## Current behavior

Task currently only evaluates the `source/generates/status` fields of a task in order to determine if the task's commands need to run or to be skipped. As a result, when dependencies are out-of-date, their execution does not affect their dependent task and subsequently the execution of the target task (i.e., the task whose name was passed as the command line argument). As a result, a target task might be skipped despite having some of its dependencies executed due to being out-of-date.

### Overview
Make the execution of any of the dependencies trigger the execution of all their dependent tasks. Especially for project management and DevOps uses of go-task, where not all tasks operate on files, this behavior will greatly simplify the definition of tasks and help minimize the need for forced runs while still preventing unnecessary work.

This change, as noted by @ghostsquad [here](https://github.com/go-task/task/discussions/669#discussioncomment-2259239), will also make go-task behave closer to Makefiles (GNU make).

### Details

A potential algorithm of how to resolve whether the commands of a given task need to be executed or skipped can be the following:

1. Evaluate the dependencies of a target task. If any of the dependencies was out-of-date mark the target task as out-of-date.
2. When returning to evaluate the up-to-date/out-of-date status of a task, if it hasn't already been marked as out-of-date by its dependencies, evaluate its sources/generates/status as normal to determine if it should be skipped or not.

For example, for a task dependency tree as shown below:
![image](https://user-images.githubusercontent.com/338764/156909936-36666232-2330-4414-8f5a-3bbb30de82ac.png)

If only tasks dep1.2.1 and dep2.2 are out-of-date, this would also force tasks dep1.2, dep1, dep2, and of course the target to be marked as out-of-date and run.

trulede:
Closing stale issue.

PR #973, opened 2022-12-31
feat: Out-of-date dependency to force execution of an up-to-date task

ConX:
This is meant to be a transitioning implementation for #681 before it becomes the default behavior on a later version. When reviewing this PR, please take into consideration that I don't have much of an experience writing in Go.

The optional command line flag for this feature "--propagate-status," without a shortcut.

### DIFF
--- a/task.go
+++ b/task.go
@@ -44,6 +44,8 @@ type Executor struct {
 	Summary     bool
 	Parallel    bool
 	Color       bool
+	PropStatus  bool
+	OutOfDate   bool
 	Concurrency int
 	Interval    string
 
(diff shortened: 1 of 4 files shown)


task.go
```diff
@@ -192,6 +194,9 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 				return &TaskRunError{t.Task, err}
 			}
 		}
+		if e.PropStatus && !e.OutOfDate {
+			e.OutOfDate = true
```
andreynering, line 198: Hi @ConX.

Unfortunately, implementing this will actually be more complicated IMO.

FYI, changing a variable here without a mutex or similar is not ideal because tasks can be run in parallel.

That said, I think the implementation would actually be different. For example, it's possible to run multiple tasks like `task foo bar baz`, even in parallel with the `-p` flag, so this would require code to check specific dependencies of each task before running each of them, instead of having a global one.

I'm not sure _how_ to proceed exactly because I haven't dived in, but I wanted to let you know my initial thought about this proposal.

ConX: Hi @andreynering,

Thanks for checking the PR.

You are correct on both points; I didn't think of the parallel aspect of task, which is definitely one of its strengths. I will think about it myself, but please let me know if you have an approach in mind, and I could give it a try if you don't have time.

PR #977, opened 2023-01-02
fix: avoid reruns when the timestamp method is used

### LINKED ISSUE #976: Timestamp method for checking the state does not work
I noticed that the `timestamp` method doesn't seem to work. It reruns the tasks all the time even though nothing has changed. 

This happens when the task generates some artifacts in the first run and doesn't re-generate them in the subsequent runs. This causes the following run to return `false`.
https://github.com/go-task/task/blob/63c50d13eec142694ff5a074c3f9cda6c53d8be3/internal/status/timestamp.go#L41

To fix this issue, `task` should include the last run time in the `generates`. This way, the time of the generates will be newer than the sources, and so there is no need for rerunning. 

This solution
- would also fix the issue with an empty `generates` array. 
- is consistent with the `checksum` method, and these two should work the same in most cases. 
- is way faster than the checksum method because the source times are checked lazily and getting the metadata of a file is faster.

aminya:
Fixes #976 

> I noticed that the `timestamp` method doesn't seem to work. It reruns the tasks all the time even though nothing has changed. 
> 
> This happens when the task generates some artifacts in the first run and doesn't re-generate them in the subsequent runs. This causes the following run to return `false`.
> https://github.com/go-task/task/blob/63c50d13eec142694ff5a074c3f9cda6c53d8be3/internal/status/timestamp.go#L41
> 
> To fix this issue, `task` should include the last run time in the `generates`. This way, the time of the generates will be newer than the sources, and so there is no need for rerunning. 
> 
> 
> This solution
> - would also fix the issue with an empty `generates` array. 
> - is consistent with the `checksum` method, and these two should work the same in most cases. 
> - is way faster than the checksum method because the source times are checked lazily and getting the metadata of a file is faster.

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -595,9 +595,9 @@ The method `none` skips any validation and always run the task.
 
 :::info
 
-For the `checksum` (default) method to work, it is only necessary to
-inform the source files, but if you want to use the `timestamp` method, you
-also need to inform the generated files with `generates`.
+For the `checksum` (default) or `timestamp` method to work, it is only necessary to
+inform the source files.
+When the `timestamp` method is used, the last time of the running the task is considered as a generate.
 
 :::
 
--- a/internal/status/checksum.go
+++ b/internal/status/checksum.go
@@ -109,12 +109,12 @@ func (*Checksum) Kind() string {
 }
 
 func (c *Checksum) checksumFilePath() string {
-	return filepath.Join(c.TempDir, "checksum", c.normalizeFilename(c.Task))
+	return filepath.Join(c.TempDir, "checksum", NormalizeFilename(c.Task))
 }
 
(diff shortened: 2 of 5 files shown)


max-sixty:
Commenting as an observer — would this miss a source file change _during_ the previous test run?

aminya:
> Commenting as an observer — would this miss a source file change _during_ the previous test run?

The time snapshot is taken before running the test. To make sure, I can move `time.now()` to the top of the function

andreynering:
Hi @aminya, thanks for your contribution! 🚀 

I thought about changing the implementation to save the timestamp as the content of the file (a unix timestamp), but I think using the metadata is fine as well.

aminya:
You're welcome! 

I am going to switch my default to the timestamps after this release due to the performance boost it gives.

Issue #978, opened 2023-01-03
[Proposal] Inline platform/arch key for commands


leaanthony:
Firstly, thanks for such an awesome project!

I've been trying to write a simple taskfile that does slightly different things depending on platform/arch. This has led me to try the 2 current methods for doing this:

  - Go Template method
  - Multiple Taskfiles

The first is hard to read and I had issues with it outside of simple echos (easy to make invalid YAML). For multiple platforms, this would get messier:

```
- echo '{{if eq OS "windows"}}windows-command{{else}}unix-command{{end}}'
```

The second option can lead to many Taskfiles on disk making it harder to manage. 

As an example scenario, I want to run a `go build` on linux, darwin & windows but need to do slightly different things on each:

- On Windows, generate a .syso file
- Run go build
- On Darwin, create an app structure
- On Darwin, copy the app to the app structure
- On windows, copy the correct 3rd party dll for the architecture
- On linux, copy a `.desktop` file to the directory

For the templating option, `Taskfile.yaml` would be very messy and I'm not even sure it's achievable in any elegant way.
The multi-taskfile option would be better IMHO, but now there's potentially 4 task files and 1 `main.go` file which seems overkill, not to mention code duplication. I mean it's possible but....

When I break down the problem, I simply want to be able to run certain commands on certain platforms *within a task*. This led me to think about an alternative approach:

For a `cmd`, allow the following syntax:  `- [platform/arch:] command`

Example:
```
version: '3'

tasks:
  build:
    cmds:
      - darwin: mkdir -p x/y/z
      - windows: winres -I app.manifest -o app.syso
      - go build
      - windows/386: copy ..\32bit\app.dll .
      - windows/amd64: copy ..\64bit\app.dll .
      - linux: cp ../app.desktop .
```

I've created a PoC and it's a reasonably simple change and works well. I'm not well-versed in Task and the nuances this may introduce so I thought I'd put it out there and see what people think. If there's agreement that it's a good addition, I'll create a PR so I can get feedback on the best way to implement it.

Cheers.

andreynering:
Hi @leaanthony!

I agree that we can improve this use case. Needs a bit of discussion on how, though.

I'm curious to see your PoC work. If you could push it to a branch in your fork, or create a PR (whatever you prefer), so I can take a look, that would be helpful!

FYI, we have an open (and overdue) issue about having an `if` attribute able to control when a task/cmd will run: #608. That's a more generic solution, but would help in this use case.

[...]

PR #980, opened 2023-01-05
Add support for 'platforms' in both task and command

leaanthony:
This is a draft PR based on the discussions in #978. It allows limiting a task or a command to a particulate OS/Arch combination. It also validates any given OS/Arch against a list of valid values.

Example Taskfile that is valid with this PR:

```yaml
version: '3'

tasks:
  build-windows:
    platforms: [windows]
    cmds:
      - echo 'building on windows'

  build-darwin:
    platforms: [darwin]
    cmds:
      - echo 'building on darwin'

  build-mixed:
    cmds:
      - cmd: echo 'building on windows/arm64'
        platforms: [windows/arm64]
      - cmd: echo 'building on linux/amd64'
        platforms: [linux/amd64]
      - cmd: echo 'building on darwin'
        platforms: [darwin]
```

Todo:

- Add test for `testdata/platforms `
- Update documentation

### DIFF
--- a/taskfile/task.go
+++ b/taskfile/task.go
@@ -36,6 +36,7 @@ type Task struct {
 	IncludeVars          *Vars
 	IncludedTaskfileVars *Vars
 	IncludedTaskfile     *IncludedTaskfile
+	Platforms            []*Platform
 }
 
 func (t *Task) Name() string {
@@ -90,6 +91,7 @@ func (t *Task) UnmarshalYAML(node *yaml.Node) error {
 			Prefix        string
 			IgnoreError   bool `yaml:"ignore_error"`
 			Run           string
+			Platforms     []*Platform
 		}
 		if err := node.Decode(&task); err != nil {
 			return err
@@ -115,6 +117,7 @@ func (t *Task) UnmarshalYAML(node *yaml.Node) error {
 		t.Prefix = task.Prefix
 		t.IgnoreError = task.IgnoreError
 		t.Run = task.Run
+		t.Platforms = task.Platforms
 		return nil
 	}
 
@@ -150,6 +153,7 @@ func (t *Task) DeepCopy() *Task {
 		IncludeVars:          t.IncludeVars.DeepCopy(),
 		IncludedTaskfileVars: t.IncludedTaskfileVars.DeepCopy(),
 		IncludedTaskfile:     t.IncludedTaskfile.DeepCopy(),
+		Platforms:            deepCopySlice(t.Platforms),
 	}
 	return c
 }
--- a/docs/docs/api_reference.md
+++ b/docs/docs/api_reference.md
@@ -139,6 +139,7 @@ includes:
 | `prefix` | `string` | | Defines a string to prefix the output of tasks running in parallel. Only used when the output mode is `prefixed`. |
 | `ignore_error` | `bool` | `false` | Continue execution if errors happen while executing commands. |
 | `run` | `string` | The one declared globally in the Taskfile or `always` | Specifies whether the task should run again or not if called more than once. Available options: `always`, `once` and `when_changed`. |
+| `platforms` | `[]string` | All platforms | Specifies which platforms the task should be run on. |
 
 :::info
 
@@ -189,6 +190,7 @@ tasks:
 | `vars` | [`map[string]Variable`](#variable) | | Optional additional variables to be passed to the referenced task. Only relevant when setting `task` instead of `cmd`. |
 | `ignore_error` | `bool` | `false` | Continue execution if errors happen while executing the command. |
 | `defer` | `string` | | Alternative to `cmd`, but schedules the command to be executed at the end of this task instead of immediately. This cannot be used together with `cmd`. |
+| `platforms` | `[]string` | All platforms | Specifies which platforms the command should be run on. |
 
 :::info
 
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -439,6 +439,73 @@ tasks:
       - echo {{.TEXT}}
 ```
 
+## Platform specific tasks and commands
+
+If you want to restrict the running of tasks to explicit platforms, this can be achieved
+using the `platforms` key. Tasks can be restricted to a specific OS, architecture or a
+combination of both.
+
+The `build-windows` task below will run only on Windows, and on any architecture:
+
+```yaml
+version: '3'
+
+tasks:
+  build-windows:
+    platforms: [windows]
+    cmds:
+      - echo 'Running command on windows'
+```
+
+This can be restricted to a specific architecture as follows:
+
+```yaml
+version: '3'
+
+tasks:
+  build-windows-amd64:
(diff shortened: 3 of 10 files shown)


taskfile/task.go
```diff
@@ -90,6 +119,7 @@ func (t *Task) UnmarshalYAML(node *yaml.Node) error {
 			Prefix        string
 			IgnoreError   bool `yaml:"ignore_error"`
 			Run           string
+			On            []string
```
pd93, line 122: You could implement this as `[]Platform` and then use the [`Unmarshaler` interface](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshaler) to process it rather than calling `processPlatforms` manually. This might be a bit cleaner.

I would personally also move the `Platform` struct and its `UnmarshalYAML` function to their own `.go` file in the `taskfile` package. Similar to what has been done for `precondition` or `vars`.

pd93 (commented):
I really like this proposal and the direction this is going in. Just a couple of small comments on the draft PR :)

taskfile/task.go
```diff
@@ -115,12 +145,66 @@ func (t *Task) UnmarshalYAML(node *yaml.Node) error {
...
+// processPlatforms iterates the given platform strings,
+// validates the values and returns a slice of platform data.
+func processPlatforms(on []string) ([]*Platform, error) {
+
+	var platforms []*Platform
+
+	for _, onValue := range on {
+		splitValues := strings.Split(onValue, "/")
+		if len(splitValues) > 2 {
+			return nil, fmt.Errorf("invalid OS/Arch provided: %s", onValue)
```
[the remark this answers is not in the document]
leaanthony: The errors from this method get bubbled up and formatted as you proposed in the calling method. The reason for doing this is because one caller has the node information and one doesn't so this way would keep the errors consistent. I'm happy to change it though.

taskfile/task.go
leaanthony: Good suggestions. I was on the verge of making another package anyway so that's fine. I haven't done a yaml unmarshaller before but I'm sure it's easy enough...

leaanthony:
@pd93 - addressed comments. Feels a lot cleaner now as you suggested 👍

pd93:
> @pd93 - addressed comments. Feels a lot cleaner now as you suggested 👍 EDIT: I did try listing but the linter breaks with what seems like a timezone issue.

@leaanthony The linting issue is just the imports not being sorted. You can fix this by running `goimports -w .`. This should result in the changes below:

```diff
diff --git a/taskfile/platforms.go b/taskfile/platforms.go
index c748b15..ec0f7f2 100644
--- a/taskfile/platforms.go
+++ b/taskfile/platforms.go
@@ -2,9 +2,10 @@ package taskfile
 
 import (
        "fmt"
-       "gopkg.in/yaml.v3"
        "runtime"
        "strings"
+
+       "gopkg.in/yaml.v3"
 )
 
 // Platform represents GOOS and GOARCH values
diff --git a/taskfile/task.go b/taskfile/task.go
index 12dc76b..c5a9a55 100644
--- a/taskfile/task.go
+++ b/taskfile/task.go
@@ -2,6 +2,7 @@ package taskfile
 
 import (
        "fmt"
+
        "gopkg.in/yaml.v3"
 )
```

Not sure what is causing you to not be able to run the linter. I haven't seen that error before.

taskfile/task.go
```diff
@@ -36,6 +35,19 @@ type Task struct {
...
+func ShouldRunOnCurrentPlatform(platforms []*Platform) bool {
+	if len(platforms) == 0 {
+		return true
+	}
+	for _, platform := range platforms {
+		if platform.MatchesCurrentPlatform() {
+			return true
+		}
+	}
+	return false
```
pd93, line 50: This function feels a little lonely now that all the other platform related changes have moved. Maybe it would be better in [task.go](https://github.com/go-task/task/blob/master/task.go) alongside the Executor which is where its called from. The `taskfile` package is generally for code that reads/parses the Taskfile` whereas this is called during the execution of a command.

leaanthony: Good call. Pushed change 👍

pd93:
The code layout looks good to me now. Just checked out the branch to try it out and I have a couple of observations:

## 1.

If we have a Taskfile like:

```yaml
version: '3'

tasks:
  default:
    platforms: [linux/foo]
    #                 ^ invalid arch
    cmds:
      - echo 'building...'
```

The error message isn't quite right:

```
task: Failed to parse testdata/platforms/Taskfile.yml:
task: arch 'linux' unsupported
exit status 1
```

This should probably read `task: Arch 'foo' is not supported`. Note the capitalisation too - we've not been entirely consistent, but most of the errors in the repo are Capitalised. (Note to @andreynering, we should probably standardise this).

## 2.

Do we want to support something like this?

```yaml
  default:
    platforms: [amd64]
    cmds:
      - echo 'building...'
```

Or are we happy with only supporting `OS` and `OS/ARCH`. It would be nice to add this for situations where you want to build using a different command depending on arch, but don't care about the OS. If we do add support for `ARCH`, we'd need to update the error handling to say its "not a valid OS or arch" when only 1 valid is given.

## Finally

Once we're happy with the above, we could do with adding the following too:

- Small section in the [usage docs](https://github.com/go-task/task/blob/master/docs/docs/usage.md) to describe the feature.
- Update the [schema](https://github.com/go-task/task/blob/master/docs/static/schema.json) and [API Reference](https://github.com/go-task/task/blob/master/docs/docs/api_reference.md#command) (preferably with the same description in each). The items here are sorted in the same order as they appear in the Go structs.
- And finally, squash all the commits :)

leaanthony:
Thanks for all the quick and informative feedback. I've pushed an update with the following changes:

- A single platform value (without a slash) can now be either OS or Arch
- Invalid single value (`platforms: [foo]`) produces the following error: `task: Invalid OS/Arch value provided (foo)`
- Trailing slash value (`platforms: [linux/]`) produces the following error: `task: Blank Arch value provided`
- Invalid Arch value (`platforms: [linux/foo]`) produces the following error: `task: Invalid Arch value provided (foo)`
- Empty value (`platforms: []`) runs on all platforms
- Updated test task file

I'll update the docs next 👍

leaanthony:
Docs + Schema updated. I'm not 💯 on the schema updates so feedback appreciated 🙏

andreynering:
Hi @leaanthony!

Thanks for your work, and quick response after code review as well!

I just merged this to master and push a commit with small improvements: 2efb3533ec5e069c0782629edcb499242874a7c7.

leaanthony:
Brilliant! Thanks @andreynering and @pd93 for your fast and great feedback. This was a pleasure.

dkarter:
Hey all, can you please review my PR to add `platforms` to the JSON schema when `for` is used?

https://github.com/go-task/task/pull/1915

I believe this was accidentally missed in this one - it still works, but my editor is providing false positive lint errors.

Thanks!

PR #982, opened 2023-01-12
Optimize task filtering

pd93:
This PR makes some optimizations to the task filtering. It's currently extremely slow on complex configs. Particularly in setups with many files that include one another.

- Use goroutines for each compilation
- Filter before compiling a task
- Simplify filtering API
- Handle fast compile errors

---

Edit: Some context, this takes one of my more complex Taskfiles from around 4 seconds to compile the tasks, to around 0.5s!


andreynering:
Thanks @pd93!

PR #1004, opened 2023-02-08
feat: use semver package for taskfile schema version

pd93:
Simple change to switch back to using the [Masterminds/semver](https://github.com/Masterminds/semver) package for storing/comparing the Taskfile schema version.

Also see [this related PR](https://github.com/Masterminds/semver/pull/196) which may make some of our comparisons a bit more readable.

### DIFF
--- a/taskfile/taskfile.go
+++ b/taskfile/taskfile.go
@@ -2,15 +2,20 @@ package taskfile
 
 import (
 	"fmt"
-	"strconv"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"gopkg.in/yaml.v3"
+)
+
+var (
+	V3 = semver.MustParse("3")
+	V2 = semver.MustParse("2")
 )
 
 // Taskfile represents a Taskfile.yml
 type Taskfile struct {
-	Version    string
+	Version    *semver.Version
 	Expansions int
 	Output     Output
 	Method     string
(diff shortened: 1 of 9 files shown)


taskfile/taskfile.go
```diff
@@ -31,7 +36,7 @@ func (tf *Taskfile) UnmarshalYAML(node *yaml.Node) error {
 
 	case yaml.MappingNode:
 		var taskfile struct {
-			Version    string
+			Version    *semver.Version
```
andreynering, line 39: Make sure this doesn't add any side effects. Otherwise it could be safer to keep as `string` and parse afterwards.

(It's fine to keep `*semver.Version` on the `Taskfile` struct in this case, I'm more concerned about `UnmarshalYAML`).

andreynering (approved):

taskfile/taskfile.go
pd93: Mastermind/semver supports the `UnmarshalYAML` interface. See [this PR](https://github.com/Masterminds/semver/pull/45). Unless I'm misunderstanding the problem?

pd93: @andreynering I did a bit of research. Since the version string is a scalar value (string), [the YAML parser will check if the the type implements the `encoding.TextUnmarshaler` interface](https://github.com/go-yaml/yaml/blob/f6f7691b1fdeb513f56608cd2c32c51f8194bf51/decode.go#L591-L609), which [`semver.Version` does](https://github.com/Masterminds/semver/blob/8e6ddaa702c4ef24d5a34119b993eafba1810a82/version.go#L458-L467). This means the value is parsed correctly and should be safe to use directly.

andreynering: @pd93 Alright, then 👍. Feel free to merge once ready.

PR #1010, opened 2023-02-13
chore: update to go 1.20

pd93:
Updates the repo from 1.18/1.19 to 1.19/1.20.

Also:

- Updates the [golang.org/x/exp](https://pkg.go.dev/golang.org/x/exp) package (which we use for `slices`) to the latest commit.
- Updates `golangci-lint` to 1.51.1 in order to solve [this issue](https://github.com/golangci/golangci-lint/issues/3107).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,11 +14,11 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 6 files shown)


andreynering (approved):
Thanks @pd93!

I wanted to do this this weekend, but lacked the time to proceed. 👍

Issue #990, opened 2023-01-24
Please consider to add new special variable TASK_VERSION


skaluzka:
I'm using few various task executables in very mixed and inconsistent
environment. Sometimes I miss an equivalent of `MAKE_VERSION` or `BASH_VERSION`
variable.

In bash, there is a `BASH_VERSION` variable available to determine the bash
interpreter version:

    $ bash -c 'echo $BASH_VERSION'
    5.1.16(1)-release
    $ 

In GNU Make, there is also similar variable, called `MAKE_VERSION` (you can
see more details about it in official GNU Make documentation
https://www.gnu.org/software/make/manual/html_node/Quick-Reference.html#Quick-Reference).

Would it be possible to extend the list of task's special variables available
on the templating system? Could you please consider to add a new one called,
for example, a `TASK_VERSION` ??

This could be useful for some debug investigations while running the same
jobs with various task executables.

andreynering:
Hi @skaluzka,

This is a good idea, and should be easy to implement.

If anyone is interested to contribute, these most special vars are defined here:

https://github.com/go-task/task/blob/ecc65a218e6be0c17b2669b20cc0e169ec93fa3f/internal/compiler/v3/compiler_v3.go#L183-L186

We'll to add it to this section of the documentation as well:

https://taskfile.dev/api/#special-variables

PR #1014, opened 2023-02-20
TASK_VERSION special var

ja1code:
As suggested [here](https://github.com/go-task/task/issues/990). This PR adds the `TASK_VERSION` special variable. It is retrieved from the value passed on the `version` build parameter.

It can be used as follows:
```yml
version: '3'

tasks:
  default:
    cmds:
      - echo "{{.TASK_VERSION}}"
```

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"runtime/debug"
 	"strings"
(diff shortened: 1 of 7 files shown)


andreynering:
Hi @ja1code,

I just merged this with small adjustment at ec2110e58f1837b5a06257ba9a7929ed9e87a930.

Thanks for contributing!

Issue #664, opened 2022-01-26
Add `output: group_error_only` output mode


jaedle:
Hey :wave:

I'm still a huge fan of task and I am using it quite heavily, so thanks for your effort!  :+1: :clap: 

If my motivation or details about this proposal seem to be unclear, please feel free to ask. If we agree on the details and edge-cases, it would a pleasure to implement this myself :sunglasses: 

## Feature description

I would like to propose a new feature in order to make my console output less verbose in case of re-occurring tasks:

Many of my [taskfiles](https://github.com/jaedle/golang-tplink-hs100/blob/ec1da8a581ec5279e853de9a4b65d6523e962989/Taskfile.yml#L29) have a lot of tasks involved for some kind of workflow. 
The output of single tasks is (mostly) not relevant for me if nothing errors, but is very relevant if it does (i.e. a failing test).

In order to keep my console clean and focus on potentially important messages I would like to have the chance to drop the stdout/stderr of command, if it does not error. If it errors the whole output should be shown.

### Current 

```sh
> task format lint test build
output from format...
output from linter...
output from test...
output from build..
```

### Proposal for happy execution (output swallowed)

All output is swallowed as not relevant because of no errors.

```sh
> task format lint test build
(no putput)
```

### Proposal for execution with errors (output shown for failing task)

Let's assume that the tests are broken in this example.

```sh
> task format lint test build
output from test...
task: Failed to run task "test": exit status 1
```

## Proposal

I would like to add a configuration parameter such as `show_output` which could be `always` (default), `on_error` or `never`.
This should not break the current behavior as the default would be `always`.

### Example Taskfile + executions

```yml
version: '3'

silent: true

tasks:

  shows_output:
    show_output: 'always'
    cmds:
      - echo 'is always shown'

  swallows_output:
    show_output: 'on_error'
    cmds:
      - echo 'a'
      - echo 'lot of'
      - echo 'output'

  shows_output_on_error:
    show_output: 'on_error'
    cmds:
      - echo 'shows error output'
      - false

  swallows_output_even_on_error:
    show_output: 'never'
    cmds:
      - echo 'is never shown'
      - false
```

```
> task shows_output
'is always shown'

> task swallows_output
(no output)

> task shows_output_on_error
shows error output

> task swallows_output_even_on_error
(no output)
```

I would also prefer to have a global parameter to configure the default behavior in regard of the output like `silent`. 

```yml
version: '3'

silent: true
show_output: 'on_error'

tasks:
...
```

[...]

PR #1022, opened 2023-02-25
Feature 664 group error only

jaedle:
Implementation idea for #664.

> Describe your changes 

## Usage

```yaml
version: '3'

silent: true

output:
  group:
    error_only: true

tasks:
  passing: echo 'passing-output'
  failing: echo 'failing-output' && exit 1
```

```sh
> task passing
(no output)
> task failing
failing-output
task: Failed to run task "failing": exit status 1
```

## Implementation
Enrich `Group` output by an `error_only` flag which should swallow the output (stdout + stderr) if any external command did not have a non zero exit code. By default this flag is disabled.

## backwards comapatibility

Changes should not interfere with task files created in previous versions as the default behavior is not changed.

## What's already done
- [x] high level tests (`task_test.go`)
- [x] tests for output writer
- [x] code style (fmt & lint)
- [x] documentation
- [x] schema definition

I would love to have some feedback if this goes into the right direction. If so, I would love to continue working on it.

### DIFF
--- a/internal/output/group.go
+++ b/internal/output/group.go
@@ -7,6 +7,7 @@ import (
 
 type Group struct {
 	Begin, End string
+	ErrorOnly  bool
 }
 
 func (g Group) WrapWriter(stdOut, _ io.Writer, _ string, tmpl Templater) (io.Writer, io.Writer, CloseFunc) {
@@ -17,7 +18,13 @@ func (g Group) WrapWriter(stdOut, _ io.Writer, _ string, tmpl Templater) (io.Wri
 	if g.End != "" {
 		gw.end = tmpl.Replace(g.End) + "\n"
 	}
-	return gw, gw, func() error { return gw.close() }
+	return gw, gw, func(err error) error {
+		if g.ErrorOnly && err == nil {
+			return nil
+		}
+
+		return gw.close()
+	}
 }
 
 type groupWriter struct {
--- a/internal/output/output.go
+++ b/internal/output/output.go
@@ -18,7 +18,7 @@ type Output interface {
 	WrapWriter(stdOut, stdErr io.Writer, prefix string, tmpl Templater) (io.Writer, io.Writer, CloseFunc)
 }
 
-type CloseFunc func() error
+type CloseFunc func(err error) error
 
 // Build the Output for the requested taskfile.Output.
 func BuildFor(o *taskfile.Output) (Output, error) {
@@ -30,8 +30,9 @@ func BuildFor(o *taskfile.Output) (Output, error) {
 		return Interleaved{}, nil
 	case "group":
 		return Group{
-			Begin: o.Group.Begin,
-			End:   o.Group.End,
+			Begin:     o.Group.Begin,
+			End:       o.Group.End,
+			ErrorOnly: o.Group.ErrorOnly,
 		}, nil
 	case "prefixed":
 		if err := checkOutputGroupUnset(o); err != nil {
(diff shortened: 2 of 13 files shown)


internal/output/output.go
```diff
@@ -33,6 +33,12 @@ func BuildFor(o *taskfile.Output) (Output, error) {
 			Begin: o.Group.Begin,
 			End:   o.Group.End,
 		}, nil
+	case "group_error_only":
```
andreynering, line 36: After some reflection, I think that we should probably follow the existing pattern for `begin/end` and allow an additional option on `group`, instead of having a separate option. Would be like this:

```yaml
output:
  group:
    begin: '::group::{{.TASK}}'
    end: '::endgroup::'
    error_only: true
```

`begin` and `end` would be optional, of course, if the users wants only `error_only`.

What do you think?

internal/output/group.go
```diff
@@ -17,7 +18,17 @@ func (g Group) WrapWriter(stdOut, _ io.Writer, _ string, tmpl Templater) (io.Wri
 	if g.End != "" {
 		gw.end = tmpl.Replace(g.End) + "\n"
 	}
-	return gw, gw, func() error { return gw.close() }
+	return gw, gw, func(errored bool) error {
+		if g.shouldSwallowOutput(errored) {
```
andreynering, line 22: 1. I have the impression that passing an error would better follow Go's conventions (no strong opinion, though)
2. In theory, instead of clearing the buffer we could just... do nothing

```go
	return gw, gw, func(err error) error {
		if g.ErrorOnly && err == nil {
			return nil
		}

		return gw.close()
```

task.go
```diff
@@ -282,8 +282,9 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
 			return fmt.Errorf("task: failed to get variables: %w", err)
 		}
 		stdOut, stdErr, close := outputWrapper.WrapWriter(e.Stdout, e.Stderr, t.Prefix, outputTemplater)
+		var errored bool = false
 		defer func() {
-			if err := close(); err != nil {
+			if err := close(errored); err != nil {
```
andreynering, line 287: To avoid having to have an outer variable, I believe it's probably fine to move this block out of the `defer` block and after `execext.RunCommand`.

task.go
```diff
@@ -298,6 +299,10 @@ func (e *Executor) runCommand(ctx context.Context, t *taskfile.Task, call taskfi
 			Stdout:    stdOut,
 			Stderr:    stdErr,
 		})
+		if execext.IsExitError(err) {
+			errored = true
+		}
```
andreynering, line 304:
```suggestion
		if err = close(err); err != nil {
			e.Logger.Errf(logger.Red, "task: unable to close writer: %v", err)
		}
```

andreynering (commented):
Hi @jaedle!

Thanks for working on this!

This is in the right direction, but I commented a few changes before this can be merged.

Other than the documentation, something else missing the updating the [`schema.json`](https://github.com/go-task/task/blob/master/docs/static/schema.json) to reflect this change. If you don't want to do that yourself, let us know and we'll do once merged.

jaedle:
Hi @andreynering,

thanks for your feedback. I will try to catch up on your review comments on the weekend.

internal/output/output.go
jaedle: done

task.go
jaedle: defer block moved after command execution, done.

task.go
jaedle: resolved by moving defer statement after command execution

jaedle:
Hey @andreynering,

as you have suggested a different approach (add a flag instead to output `group` instead of a new output `group_error_only`), it was way easier to start the implementation from scratch. I am sorry if this creates more effort on your side, but this was way cleaner.

I have updated the issue to reflect the new implementation idea.

I tried to implement all your review comments. If there is still anything missing on this pr, feel free leave a note.

internal/output/group.go
jaedle: done

andreynering:
Thanks for your contribution @jaedle, it's very appreciated!

PR #1029, opened 2023-03-02
Add `--global` (`-g`) flag

andreynering:
This is an idea that has been in my mind for some time. Feedback and/or suggestions are welcome.

In short, this adds the ability to run a Taskfile from your home directory (`$HOME/Taskfile.yaml`), meaning that you can now have a Taskfile that you can run from anywhere with a simple `task -g ...`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Add a brand new `--global` (`-g`) flag that will run a Taskfile from your
+  `$HOME` directory. This is useful to have automation that you can run from
+  anywhere in your system!
+  ([Documentation](https://taskfile.dev/usage/#running-a-global-taskfile), [#1029](https://github.com/go-task/task/pull/1029) by @andreynering).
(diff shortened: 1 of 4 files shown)


pd93 (approved):
I don't see a downside to this. Its simple and I'm sure lots of people will find this useful. 

Worth noting that you can get this behaviour today by putting a Taskfile in `$HOME` and running `task` (without `-g`) if there are no other Taskfiles between `$PWD` and `$HOME`. However being able to call it from inside a project containing a Taskfile is super useful.

Its a little difficult to make tests for this since it relies on `os.UserHomeDir()`, but its also quite simple code - so maybe we don't need one? We could mock out the call to `os.UserHomeDir` though if you think its worth it.

andreynering:
Thanks for your comments @pd93.

The fact is that we don't have any test for `cmd/task/task.go` at the moment. Perhaps someday we could try extract some behavior out of the `main` package and write a few simple tests.

And yes, `-g` is useful because you can run even when the current directory has another Taskfile.

PR #1031, opened 2023-03-02
Fix output "prefixed" option in schema.json

pzloty:
Fix the output option to match implementation and documentation.


pd93:
Whoops! Thanks for fixing this. I blame copilot 😅

PR #1035, opened 2023-03-03
for task up to date check both status and sources

### LINKED ISSUE #840: .task/checksum is not created after running a task whose status checked returned non-zero
- Task version: v3.12.1 and v3.14.1
- Operating System: Ubuntu 18

```yaml
version: '3'
tasks:
  default:
    cmds:
      - "touch ./out.txt"
      - "echo running"
    status:
      - test -f ./out.txt
    sources:
      - ./src/**
```

### Steps to reproduce

```
❯ tree . -a
.
├── src
│   └── foo.txt
└── Taskfile.yaml

❯ task -vvv
task: "default" started
task: status command test -f ./out.txt exited non-zero: exit status 1
task: [default] touch ./out.txt
task: [default] echo running
running
task: "default" finished

❯ tree . -a # NOTE .task/checksum/ is missing
.
├── out.txt
├── src
│   └── foo.txt
└── Taskfile.yaml

❯ task -vvv # NOTE default task ran again erroneously because .task/checksum/ doesn't exist
task: "default" started
task: status command test -f ./out.txt exited zero
task: [default] touch ./out.txt
task: [default] echo running
running
task: "default" finished

❯ tree . -a # This time .task was created successfully
.
├── out.txt
├── src
│   └── foo.txt
├── .task
│   └── checksum
│       └── default
└── Taskfile.yaml

❯ task -vvv # This time it correctly skips it because .task/checksum exists
task: "default" started
task: status command test -f ./out.txt exited zero
task: Task "default" is up to date
```
I suspect the issue has to do with the fact that the first time status command exited 1. If I remove the status command or replace it with "true" the .task/checksum is created the first time.

harelwa:
>
>

Closes #840 

- **Disclaimer:** I'm not a `go` programmer. so if there is a more "idiomatic go" way of doing what i did, do tell.
- **Test:** As `TestStatusChecksum` checks exactly what I think should be checked in this case, I've added a `status` to the tested `Taskfile`. Without my changes to `status.go` this test will fail when running on the modified `Taskfile`

### DIFF
--- a/status.go
+++ b/status.go
@@ -29,10 +29,6 @@ func (e *Executor) Status(ctx context.Context, calls ...taskfile.Call) error {
 }
 
 func (e *Executor) isTaskUpToDate(ctx context.Context, t *taskfile.Task) (bool, error) {
-	if len(t.Status) == 0 && len(t.Sources) == 0 {
-		return false, nil
-	}
-
 	isUpToDateStatus := true
 	isUpToDateChecker := true
 
--- a/status.go
+++ b/status.go
@@ -33,13 +33,16 @@ func (e *Executor) isTaskUpToDate(ctx context.Context, t *taskfile.Task) (bool, 
 		return false, nil
 	}
 
+	isUpToDateStatus := true
+	isUpToDateChecker := true
+
 	if len(t.Status) > 0 {
 		isUpToDate, err := e.isTaskUpToDateStatus(ctx, t)
 		if err != nil {
 			return false, err
 		}
 		if !isUpToDate {
-			return false, nil
+			isUpToDateStatus = false
 		}
 	}
 
@@ -53,11 +56,13 @@ func (e *Executor) isTaskUpToDate(ctx context.Context, t *taskfile.Task) (bool, 
 			return false, err
 		}
 		if !isUpToDate {
-			return false, nil
+			isUpToDateChecker = false
 		}
 	}
 
-	return true, nil
+	isUpToDate := isUpToDateStatus && isUpToDateChecker
+
+	return isUpToDate, nil
 }
 
 func (e *Executor) statusOnError(t *taskfile.Task) error {
--- a/task_test.go
+++ b/task_test.go
@@ -447,36 +447,43 @@ func TestGenerates(t *testing.T) {
 func TestStatusChecksum(t *testing.T) {
 	const dir = "testdata/checksum"
 
-	files := []string{
-		"generated.txt",
-		".task/checksum/build",
+	tests := []struct {
+		files []string
+		task  string
+	}{
+		{[]string{"generated.txt",".task/checksum/build"}, "build"},
+		{[]string{"generated.txt",".task/checksum/build-with-status"}, "build-with-status"},
 	}
 
-	for _, f := range files {
-		_ = os.Remove(filepathext.SmartJoin(dir, f))
+	for _, test := range tests {
+		t.Run(test.task, func(t *testing.T) {
+			for _, f := range test.files {
+				_ = os.Remove(filepathext.SmartJoin(dir, f))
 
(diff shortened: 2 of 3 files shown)


status.go
```diff
@@ -33,13 +33,16 @@ func (e *Executor) isTaskUpToDate(ctx context.Context, t *taskfile.Task) (bool,
 		return false, nil
 	}
```
ghostsquad, line 34: while we are here, there's really no point in this if statement. It's redundant, and only clutters this function.

status.go
```diff
@@ -53,11 +56,13 @@ func (e *Executor) isTaskUpToDate(ctx context.Context, t *taskfile.Task) (bool,
 			return false, err
 		}
 		if !isUpToDate {
-			return false, nil
+			isUpToDateChecker = false
```
ghostsquad, line 59: does `checker.IsUpToDate()` have a side effect of writing the checksum file?

ghostsquad (changes requested):
I think this still needs a test

harelwa:
> ~I think this still needs a test~ maybe a _new_ test can be written to catch this case? We want to ensure both cases are valid, so please don't modify existing tests.

it can be the same test running on two difference `Taskfile`s - the prev one + the one here

status.go
harelwa: indeed

harelwa:
> ~I think this still needs a test~ maybe a _new_ test can be written to catch this case? We want to ensure both cases are valid, so please don't modify existing tests.

\+ if you have noticed something the existing test does not cover for this case, please share

harelwa:
> > ~I think this still needs a test~ maybe a _new_ test can be written to catch this case? We want to ensure both cases are valid, so please don't modify existing tests.
> 
> it can be the same test running on two difference `Taskfile`s - the prev one + the one here

i do think such an addition is somewhat redundant though, don't you ?

status.go
harelwa: https://github.com/go-task/task/blob/d4ed7c3cfc2408bf0978b9d693ff226e3b9c0707/internal/status/checksum.go#L49

ghostsquad:
The test is verifying the behaviour of a single piece of functionality in isolation. This change affects the combination of features. Thus, technically we want a permutation of tests.
- status - OK
- status - failed
- source - OK
- source - failed
- status - OK & source OK
- ... etc.

harelwa:
> The test is verifying the behaviour of a single piece of functionality in isolation. This change affects the combination of features. Thus, technically we want a permutation of tests.
> 
> * status - OK
> * status - failed
> * source - OK
> * source - failed
> * status - OK & source OK
> * ... etc.

I'd not re-test `status, as `TestStatus` already covers some of these [subtests](https://go.dev/blog/subtests).

I'd rather re-use/ride `TestStatusChecksum` as it tests the desired functionality correctness.

Following your remark, I'd also rather not "combine" tests like I did.

[thus] I have added subtests to `TestStatusChecksum` and added a separate subtest accordingly, leaving the previous one unchanged.

ghostsquad (approved):

harelwa:
thank you @ghostsquad for the kind and fast review. and approval.

Are we good to merge ?

pd93:
@andreynering @ghostsquad @abrahammurciano This hasn't been properly tested. It seems to have broken checksums for any tasks that don't declare `status` or `sources`.

On `d4ed7c3`

```sh
❯ task --dir ./testdata/env
task: [local] echo "GOOS='$GOOS' GOARCH='$GOARCH' CGO_ENABLED='$CGO_ENABLED'" > local.txt
task: [global] echo "FOO='$FOO' BAR='$BAR' BAZ='$BAZ'" > global.txt
```

On `master` (`6e346de`)

```sh
❯ task --dir ./testdata/env
task: Task "default" is up to date
```

See fix here: #1037

Issue #1038, opened 2023-03-07
Environment variables from shell takes precedence over task file definition


jaedle:
Hey :wave: 

I guess I have found a bug yesterday on task: It looks like previously defined environment variable are not overwritten by the definition within the task file.

## Taskfile

```yaml
version: '3'

silent: true

tasks:
  default:
    env:
      KEY: 'other'
    cmds:
      - echo "$KEY"
```

## Expected

I would expect the environment variable of KEY to be overwritten whatever the outside environment looks like.

```sh
$ task
other
$ KEY=some task
other
```

## Actual

The environment variable from the outside shell takes precedence over the variable defined within the taskfile. 

```sh
$ task
other
$ KEY=some task
other
```

### Os + Taskfile-version

```sh
$ cat /etc/lsb-release 
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=21.1
DISTRIB_CODENAME=vera
DISTRIB_DESCRIPTION="Linux Mint 21.1 Vera"
$ task --version
Task version: v3.21.0 (h1:pVGAGXxJ9Pk5mvjqv/CsdAD4ZIzNMjF+vrXMueM/WFk=)
```

jaedle:
Any feedback on this?

matthchr:
I also just ran into this and was also surprised

andreynering:
Unfortunately, this is a breaking change. This means we'd have to wait for the next major version release to do this by default, and that can take quite some time to happen.

In the meantime, we can consider having an opt-in setting for the behavior, so those who want to anticipate that can just set a flag.

zbindenren:
In my opinion this is a bug. If I set a variable in the command I expect it to run with this configuration. Otherwise the stuff I see in the Taskfile is not what is happening when running the task. This is really confusing for the user.

If you are not willing to change the behavior because of backward compatibility concerns I would prefer to set the opt-in option in the Taskfile, otherwise when I forget it, commands are not working.

jaedle:
Thanks for your feedback!

I kindly agree with @zbindenren considering it a bug. 

I also agree that this may be a breaking chance in respect of current behavior even though it was never documented. 

Nevertheless I would consider documenting (and also adding tests) the current behavior as useful.
When creating the PR changing the behavior did not break any test.

I would argue that my pull request to be not relevant in respect of the direction implementing it. I would close it.

Shall I put some effort into testing and adding documentation for the current behavior?

andreynering:
[...]

PR #1053, opened 2023-03-12
Environment definitions in task files should have precedence over predefined environment variables

jaedle:
Hey :wave: 

I have tried to add a test + a fix for the (potential) bug #1038.

> Describe your changes 

Environment variable definitions within a task file take precedence over environment variables defined in shell.

## backwards compatibility

Changes **may** have problems about backwards compatibility:
Despite no test are failing nor any documentation guaranteed that behavior, it may have an impact.
I would like to discuss that.

## Tasks
- [x] high level test (`task_test.go`)
- [x] code style (fmt & lint)
- [x] document change

## Testing

In order to test that behavior, it was mandatory to compile / run the task binary itself. 
As an explicit and repeatable control over the outside environment configuration is mandatory, I think following that path is the only path that leads to real feedback.


andreynering:
See my comment: https://github.com/go-task/task/issues/1038#issuecomment-1558355071.

andreynering:
Closing in favor of #1633

PR #1056, opened 2023-03-13
feat: add task location data to json output

pd93:
This PR adds location data to the `--json` output flag. There is a new location string in the JSON root which describes the filepath of the root Taskfile (entrypoint) and there is a new location object on each task which describes filepath of the Taskfile containing the task and the line/column that the task is defined on.

Since Taskfiles can include other Taskfiles, the location of a task can differ from location of the root taskfile.

Example:

```jsonc
{
  "tasks": [
    {
      "name": "clean",
      "desc": "Cleans temp files and folders",
      "summary": "",
      "up_to_date": false,
      "location": {
        "line": 54,
        "column": 3,
        "taskfile": "/path/to/Taskfile.yml"
      }
    },
    // ...
  ],
  "location": "/path/to/Taskfile.yml"
}
```

### DIFF
--- a/docs/docs/api_reference.md
+++ b/docs/docs/api_reference.md
@@ -67,7 +67,8 @@ When using the `--json` flag in combination with either the `--list` or `--list-
       }
     },
     // ...
-  ]
+  ],
+  "location": "/path/to/Taskfile.yml"
 }
 ```
 
--- a/help.go
+++ b/help.go
@@ -146,9 +146,10 @@ func (e *Executor) ListTaskNames(allTasks bool) {
 	}
 }
 
-func (e *Executor) ToEditorOutput(tasks []*taskfile.Task) (*editors.Output, error) {
-	o := &editors.Output{
-		Tasks: make([]editors.Task, len(tasks)),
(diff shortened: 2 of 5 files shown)


taskfile/tasks.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+// Tasks represents a group of tasks
+type Tasks map[string]*Task
+
+func (t *Tasks) UnmarshalYAML(node *yaml.Node) error {
+	switch node.Kind {
+
```
andreynering, line 14:
```suggestion
```

taskfile/tasks.go
```diff
@@ -0,0 +1,47 @@
...
+	case yaml.MappingNode:
+		tasks := map[string]*Task{}
+		if err := node.Decode(tasks); err != nil {
+			return err
+		}
+
+		for name := range tasks {
+
```
andreynering, line 22:
```suggestion
```

andreynering (approved):

Issue #1066, opened 2023-03-17
add environment variable-based configuration


droopy4096:
When using task in larger setups it would be more convenient to pass some of the configuration options via the environment variable so as not to update every line containing `task` invocation to add or remove options. Consider scenario of having several scripts with contents like:

```shell
#!/bin/sh

task target1
#....

task target2
```

to make them run in `dry` mode one has to alter all the shell scripts to include `--dry` option. However with environment variable configuration one can simply:

```shell
export TASK_DRY=yes
```

and run scripts with all `task` invocations picking up desired option automatically. 

Of course `--dry` is only a sample, control of concurrency between workstation and CI node would be another good example. As well as increasing verbosity while debugging etc.

jwater7:
I love this, exactly what I was looking for (specify the taskfile/entrypoint via an env var) - how do we get this in main so we can start using it?

droopy4096:
someone will have to review and approve https://github.com/go-task/task/pull/1067

vmaerten:
Hello!
We’ve already added several configuration options to `.taskrc`, and we don’t plan to support additional ones via environment variables.
If something is missing from `.taskrc`, feel free to open an issue or submit a PR.

trulede:
For a lot of use cases, especially automation/CI, the preference is to use environment variables over configuration files. For these use cases, configuration files can be quite unpractical. More often than not, if a configuration file _is_ used, that is only as a basis for operation - it is expected to use the environment to further refine that operation.

If you take a look at git-sync, there is a good example of how to do this. Nearly _every_ CLI option is mirrored by a environment variable:
https://github.com/kubernetes/git-sync?tab=readme-ov-file#manual

Now, to work around these kinds of issues it not difficult, but the question remains, why can't Task be so easy to operate?

vmaerten:
I understand!
@pd93 / @andreynering should we revisit and add env variables for each CLI / taskrc option ?

jwater7:
For me, the most important piece of this to get is an environment variable that defines the location of the task file.  If you need to change the task file that task is reading dynamically from the environment, there is currently no way of doing it.

For example, if there is a shared build setup:
~~~
.build-base:
  script:
    - … bunch of stuff
    - task build

build-ci:
  expands: .build-base
  env:
    TASK_TASKFILE: ~/conf-ci/Taskfile.yml

build-bvt:
  expands: .build-base
  env:
    TASK_TASKFILE: /opt/Taskfile.dist.yml
~~~

[...]

PR #1067, opened 2023-03-17
add environment variable-based configuration

droopy4096:
this is a sample implementation of environment variable based configuration outlined in #1066


trulede:
The code that does this has moved to internal/flags/flags.go. Perhaps something like this following generated code to use generics and simply calculate the envar name automatically: `list-all` -> `TASK_LIST_ALL`. Then it becomes pretty easy to operate as the envars are easy to "guess" from whatever the cli arguments are.

<details>

<summary>Anyway, the code:</summary>
```
func GetTaskEnv[T any](name string, defaultValue T) T {
    if name == "" {
        return defaultValue
    }    envName := "TASK_" + strings.ToUpper(strings.ReplaceAll(name, "-", "_"))
    value, ok := os.LookupEnv(envName)
    if !ok {
        return defaultValue
    }

    switch v := any(defaultValue).(type) {
    case bool:
        boolVal, err := strconv.ParseBool(value)
        if err != nil {
            return defaultValue
        }
        return any(boolVal).(T)
    case int:
        intVal, err := strconv.Atoi(value)
        if err != nil {
            return defaultValue
        }
        return any(intVal).(T)
    case string:
        return any(value).(T)
    default:
        return defaultValue
    }
}

func FlagVar[T any](p *T, name, short string, defaultValue T, usage string) {
    *p = GetTaskEnv(name, defaultValue)

    switch v := any(defaultValue).(type) {
    case bool:
        pflag.CommandLine.BoolVarP((*bool)(any(p).(*bool)), name, short, *p, usage)
    case int:
        pflag.CommandLine.IntVarP((*int)(any(p).(*int)), name, short, *p, usage)
    case string:
        pflag.CommandLine.StringVarP((*string)(any(p).(*string)), name, short, *p, usage)
    }
}

func main() {
    var ListAll bool
    FlagVar(&ListAll, "list-all", "a", false, "Lists tasks with or without a description")
}
```

</details>

vmaerten:
Hello!
Thanks for your PR. I am closing it. See my comment [here](https://github.com/go-task/task/issues/1066#issuecomment-3651366204)

PR #1072, opened 2023-03-21
fix: deep copying pointers inside slices

pd93:
This is another piece of work that has fallen out of my changes in the `dag` branch. AFAIK This doesn't cause any issues currently. However, it is causing issues on my `dag` branch and I see no reason to delay merging the fix while I work on other things.

Deep copying currently isn't working as intended when a slice contains pointers. This PR fixes this by adding a `DeepCopier` interface which allows you to specify how a type should be deep copied. If a slice type implements this interface then the `DeepCopy()` function will now be called on each element. This gives us the ability to recurse into our structures and deep copy everything. Many of our types already satisfy this interface. However, I have added a method to the remaining ones that didn't.

I did consider using a 3rd party library to do all of this for us. However, they all use reflection to do it which is much slower. This custom implementation needs to be maintained, but by using an interface, this should be quite manageable.

Sidenote: I split up `Dep` and `Cmd` too as they have both grown and I think they deserve their own file now.


andreynering:
Thanks @pd93!

PR #1073, opened 2023-03-21
fix: broken link to scoop bucket

pd93:
The Scoop project was moved from extras into the main bucket quite some time ago, so our link is broken. This PR fixes it.

Related:
- https://github.com/ScoopInstaller/Main/pull/1985
- https://github.com/ScoopInstaller/Extras/pull/8015


andreynering:
Thanks @pd93!

PR #1079, opened 2023-03-22
Update schema: support several `version` values

EmilySeville7cfg:
- don't use `anyOf` for `version`
- add "2" for version

>
>

### DIFF
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -348,17 +348,9 @@
       "type": "object",
       "properties": {
         "version": {
-          "description": "Specify the Taskfile format that this file conforms to.",
-          "anyOf": [
-            {
-              "type": "number",
-              "enum": [3]
-            },
-            {
-              "type": "string",
-              "enum": ["3"]
-            }
-          ]
+          "description": "A version for the current Taskfile",
+          "type": "string",
+          "enum": ["2", "3"]
         },
         "output": {
           "description": "Defines how the STDOUT and STDERR are printed when running tasks in parallel. The interleaved output prints lines in real time (default). The group output will print the entire output of a command once, after it finishes, so you won't have live feedback for commands that take a long time to run. The prefix output will prefix every line printed by a command with [task-name] as the prefix, but you can customize the prefix for a command with the prefix: attribute.",


docs/static/schema.json
```diff
@@ -348,17 +348,9 @@
...
-              "enum": [3]
-            },
-            {
-              "type": "string",
-              "enum": ["3"]
-            }
-          ]
+          "description": "A version for the current Taskfile",
+          "type": "string",
+          "enum": ["2", "3"]
```
andreynering, line 353: Can you please keep the support for `number` as before?

EmilySeville7cfg: Is there something wrong with my change? Yes, I can keep as before, but it doesn't explain why.

pd93:
@andreynering are you sure that we want to do this? Much of the schema has changed since version 2. Merging this change would imply to our users that they can use functionality like `internal` or `platforms` (or any number of other v3-only features) in v2 since the schema would consider it as valid.

I think the correct way to do this is:

```yaml
{
  "definitions": {
    "2": { ... }
    "3": { ... }
  },
  "oneOf": [
    {},  # v2 schema
    {}   # v3 schema
  ]
}
```

However, this is quite a bit of work!

This leads me to question how many of our users are still using v2 and whether it might soon be time to stop supporting it. It would certainly help to clean up the code a bit - especially when we might need to support a v4 in the future too. Are there any particular reasons why we would still have users on v2 that cannot easily upgrade to v3? Is there something we could do to make that upgrade easier for users before discontinuing support?

@EmilySeville7cfg It would be helpful to know why you want to add v2 support to the schema. Is there anything blocking you from upgrading to v3?

andreynering:
@pd93 You have a good point. When I first saw this I was in doubt if allowing v2 was a good idea or not.

It's certainly not worth it defining the whole v2 schema there given only a few people still use it.

Given v2 is a subset of v3, though, we could in theory just allow "v2" as this PR does to at least avoid an error to be shown for these few users.

Balancing the pros and cons, I think allowing only v3 is likely better, specially given v4 will happen at some point.

docs/static/schema.json
andreynering: `version: 3` is valid. Quotes like in `version: '3'` are not required.

EmilySeville7cfg:
@pd93, I want to do as u suggested, via `oneOf`. I can support schema myself, I just want it to reflect the documentation and not to lie users that they can't use v2 even it's deprecated.

PR #1086, opened 2023-03-26
feat: create release command

pd93:
Hey @andreynering, I was messing around with the changelog yesterday and I came up with a way that we can automate the release process a bit more. I'm curious on your thoughts as this would mostly effect you.

- New command in `./cmd/release`
- Can be run with `task release -- major|minor|patch|x.x.x`
- Automatically set the version/date in the both changelogs
- Automatically convert github autolinks (#issue/@user) to FQDNs in the website changelog
  - This means we can keep the short autolinks in the main `CHANGELOG.md` which is a bit easier to write/manage

I've also reformatted both copies of the changelog slightly.

- 80 character limit
- Removed the hard-to-write FQDNs from `CHANGELOG.md` as they now get converted

Its worth noting that PRs will be linked with `/issues` in the URL as there is no way to differentiate between issues/PRs using autolinks. However, as PRs are just a special type of issue in GitHub, the link will still work - it just redirects.


andreynering:
Looks good. Thanks!

PR #1091, opened 2023-03-27
docs: optimize document style

misitebao:
Optimize the style of links in tables in documents so that Crowdin can parse them correctly.


andreynering:
Thanks!

PR #1096, opened 2023-03-28
feat: release tool improvements

pd93:
I noticed that the npm version bump was added to the Taskfile. I figured that we could just add this to the new release tool so now `task release -- major|minor|patch|x.x.x` will bump the changelog _and_ yarn files.


andreynering:
Thanks!

PR #1097, opened 2023-03-29
Add Taskfile.yml as a default watcher file and reload content

### LINKED ISSUE #1095: Watch of Taskfile.yml (reload file)
I like to watch the Taskfile.yml itself. It works that it react on changes but the file won't be reload with the new content of that file.

Maybe Taskfile.yml could be watched by default?

```
version: '3'

interval: 5s

tasks:
  default:
    deps: [first, second]
    sources:
      - Taskfile.yml

  first:
    cmds:
      - echo "First"

  second:
    cmds:
      - echo "Second"
```

misery:
This fixes #1095

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -160,89 +160,95 @@ func main() {
 		}
 	}
 
-	e := task.Executor{
-		Force:       force,
-		Watch:       watch,
-		Verbose:     verbose,
-		Silent:      silent,
-		Dir:         dir,
-		Dry:         dry,
-		Entrypoint:  entrypoint,
-		Summary:     summary,
-		Parallel:    parallel,
-		Color:       color,
-		Concurrency: concurrency,
-		Interval:    interval,
+	for {
+		e := task.Executor{
+			Force:       force,
+			Watch:       watch,
+			Verbose:     verbose,
+			Silent:      silent,
+			Dir:         dir,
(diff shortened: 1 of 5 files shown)


misery:
Any help would be nice. :-)

misery:
I switch to fsnotify, too. This is not necessary for the reload of Taskfile.yml. This can be another pull request later if requested. :-)

andreynering:
Hi @misery,

I think that reloading the Taskfile itself could cause some bugs, in special related to concurrency, so I prefer not to do that, specially by default.

With regard to changing to `fsnotify`, that's something we could consider, but I don't have a strong opinion. Do you mind sharing if you're having issues with the current library to consider changing that? Also, that's needs to be a separate PR.

misery:
Well, I that is sad. I really like the idea of Taskfile. I planed to contribute some more changes to improve it to a "tiny task runner" like a self-hosted github-action on own servers.
What do you think if that will be an extra option that needs to be enabled to reload the Taskfile? Concurrency will be a problem already as you don't have an ordered event right now.

andreynering:
@misery I'm not willing to allow reload of the `Taskfile.yml` for now. I think that's probably an uncommon use case. Nothing that we couldn't reconsider in the future, though...

That said, this don't need to stop you from contributing with other features, if you feel like it.

PR #1099, opened 2023-03-30
refactor: enable gofumpt linter and fix all issues

pd93:
Pretty simple change. Just enabled `gofmt` in the linter settings to enforce code formatting. Also forces rewrites of `interface{}` to the new `any` syntax.

Any opinions on `gofumpt`? I use it on my personal projects, but not bothered either way if we use it here.


andreynering (approved):
Looks good. I'm open to the idea of using `gofumpt` if you think that's better. I already follow most of its rules.

pd93:
@andreynering I quite like `gofumpt`. I've added it to the linter and fixed the issues it found.

I've also added a `task lint:fix` to the Taskfile and updated the contribution guide to reflect the changes.

andreynering:
Thanks!

PR #1100, opened 2023-03-31
refactor: use require for testing errors

pd93:
Uses [`github.com/stretchr/testify/require`](https://github.com/stretchr/testify/require) to test errors instead of [`github.com/stretchr/testify/assert`](https://github.com/stretchr/testify/assert). This applies to both `NoError` and `Error` functions.

This means that execution of the test will stop if we get an into an unexpected state and it will stop further cascading error assertions from obscuring the root cause.

The `assert` package is still used everywhere else as this means that we can see when multiple values are incorrect if the test is in a "good" state.

### DIFF
--- a/internal/output/output_test.go
+++ b/internal/output/output_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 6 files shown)


andreynering (approved):
This is definitely an improvement.

There's a couple of conflicts, but once fixed feel free to merge.

PR #1104, opened 2023-04-02
docs: fix usage.md

misitebao:


docs/docs/usage.md
```diff
@@ -595,7 +595,7 @@ The above syntax is also supported in `deps`.
 
 NOTE: If you want to call a task declared in the root Taskfile from within an
 [included Taskfile](#including-other-taskfiles), add a leading `:` like this:
-`task: :task-name`.
+`task::task-name`.
```
andreynering, line 598: This was actually correct before.

misitebao: So this is expected behavior?

pd93: Maybe this example would be clearer with syntax highlighting:

```yaml
tasks:
  my-task:
    cmds:
      - task: :task-in-parent-file
```

misitebao: Ok, I see, thank you.

PR #1105, opened 2023-04-02
feat: implement task sorting with `--sort` flag

### LINKED ISSUE #946: Allow for changing the order of tasks when listing tasks
Right now tasks are always listed alphabetically.
For large taskfiles such ordering does not always provide the most logical way of displaying/grouping tasks.

Please implement the ability to list task in the order in which they are defined via new a Taskfile.yaml toplevel setting (e.g. called "order", or some other name to that effect)

---
Discussion: What should be the order of tasks when including other Taskfiles?
Proposed implementation:

When definition-order is used:
(1) the order of included taskfiles should matter
(2) the order of tasks coming from included taskfiles should be the same as when listing those included taskfiles directly

So, for example, given 3 taskfiles:

```
Taskfile.yaml  # the main one, importing the other two
---
order: definition    # we're changing from the default alphanumeric ordering
includes:
  b: ./TaskfileB.yaml
  a: ./TaskfileA.yaml

tasks:
  
  foobar:
     (...)
  bar:
     (...)
```

```
TaskfileB.yaml:
---
order: alphanumeric # the default ordering

tasks:
  task2:
    (...)
  task1:
    (...)
```

```
TaskfileA.yaml
---
# no ordering specified, so use the default
tasks:
  task2:
    (...)
  task1:
    (...)
```

The result of listing tasks in Taskfile.yaml:
```
b:task2   # List "b" tasks first - respect the order of "includes".  List task2 first - respect the "order: definition" inside TaskfileB)
b:task1
a:task1  # within "a", we have alphanumeric ordering (as implicitly defined by defaults in TaskfileA.yaml)
a:task2
foobar  # finally, tasks defined in the main file
bar
```

Optionally, additionally it should also be possible to define the relative placement of tasks define in the main Taskfile.yaml - whether they end up after include tasks, or before them.

pd93:
Fixes #946

This PR implements custom sorting for tasks using a new `--sort` flag. This will work with any of the commands that output a list of tasks. i.e.

- `task --list`
- `task --list --silent`
- `task --list --json`
- `task --list-all`

The `--sort` flag will accept a string that specifies the sort order. The following values are supported:

- `default` - The same as `alphanumeric`, but root tasks (with no namespace) will be always appear first.
  - This is the current sorting behavior 
- `none` - No sorting. Tasks will appear in the order they are defined in the taskfile
  - If multiple taskfiles are used, they will be traversed in lexicographical order of the taskfile path.
- `alphanumeric` - Sorts tasks alphabetically by name ignoring the files they are defined in.

The implementation for this was quite complex so here is a bullet point summary of the changes:

- The contents of `copy.go` has been moved to a new package in `internal/deepcopy` so that it can be more easily reused
- We have two other new packages:
  - `internal/sort` - Contains the task sorting interface and implementations
  - `internal/orderedmap` - Contains a generic ordered map implementation
- Refactored `taskfile.Tasks` to use the orderedmap implementation
- Refactored `taskfile.Vars` to use the orderedmap implementation
  - This replaces the custom ordered map that was previously used
  - I had to do some interesting things here because most uses of `taskfile.Vars` use a pointer. When this pointer is `nil`, the method on the underlying `OrderedMap` cannot be called (because of how embedding works in Go). I've simply added some wrapper methods on `taskfile.Vars` to handle these cases which do `nil` checks before calling the underlying method. This is not an issue with `taskfile.Tasks` because they aren't implemented with a pointer.

---

Side note... this would be so much nicer if we had https://github.com/golang/go/discussions/56413 in Go 😛

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-task/task/v3"
 	"github.com/go-task/task/v3/args"
 	"github.com/go-task/task/v3/internal/logger"
+	"github.com/go-task/task/v3/internal/sort"
 	ver "github.com/go-task/task/v3/internal/version"
 	"github.com/go-task/task/v3/taskfile"
 )
@@ -57,6 +58,7 @@ func main() {
 		list        bool
 		listAll     bool
 		listJson    bool
+		taskSort    string
 		status      bool
 		force       bool
 		watch       bool
@@ -81,6 +83,7 @@ func main() {
 	pflag.BoolVarP(&list, "list", "l", false, "Lists tasks with description of current Taskfile.")
 	pflag.BoolVarP(&listAll, "list-all", "a", false, "Lists tasks with or without a description.")
 	pflag.BoolVarP(&listJson, "json", "j", false, "Formats task list as JSON.")
+	pflag.StringVar(&taskSort, "sort", "", "Changes the order of the tasks when listed.")
 	pflag.BoolVar(&status, "status", false, "Exits with non-zero exit code if any of the given tasks is not up-to-date.")
 	pflag.BoolVarP(&force, "force", "f", false, "Forces execution even when the task is up-to-date.")
 	pflag.BoolVarP(&watch, "watch", "w", false, "Enables watch of the given task.")
@@ -160,6 +163,14 @@ func main() {
 		}
 	}
 
+	var taskSorter sort.TaskSorter
+	switch taskSort {
+	case "none":
+		taskSorter = &sort.Noop{}
+	case "alphanumeric":
+		taskSorter = &sort.AlphaNumeric{}
+	}
+
 	e := task.Executor{
 		Force:       force,
 		Watch:       watch,
@@ -179,6 +190,7 @@ func main() {
 		Stderr: os.Stderr,
 
 		OutputStyle: output,
+		TaskSorter:  taskSorter,
 	}
 
 	listOptions := task.NewListOptions(list, listAll, listJson)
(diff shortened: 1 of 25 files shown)


internal/orderedmap/orderedmap.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+func (om *OrderedMap[K, V]) UnmarshalYAML(node *yaml.Node) error {
+	switch node.Kind {
+	// Even numbers contain the keys
+	// Odd numbers contain the values
+	case yaml.MappingNode:
+		for i := 0; i < len(node.Content); i += 2 {
+
```
andreynering, line 143:
```suggestion
```

cmd/task/task.go
```diff
@@ -81,6 +83,7 @@ func main() {
 	pflag.BoolVarP(&list, "list", "l", false, "Lists tasks with description of current Taskfile.")
 	pflag.BoolVarP(&listAll, "list-all", "a", false, "Lists tasks with or without a description.")
 	pflag.BoolVarP(&listJson, "json", "j", false, "Formats task list as JSON.")
+	pflag.StringVar(&taskSort, "sort", "", "Changes the order of the tasks when listed.")
```
andreynering, line 86: We need to update the API documentation page to include this flag.

andreynering (approved):
Nice work! 👏 

This flag needs to be mentioned on the API page. Other than that, this is ready to be merged.

cmd/task/task.go
pd93: Unless I'm missing something, this is already updated?

andreynering: Oh, I missed this change when reviewing yesterday.

Ready to merge, then. 👍

Issue #546, opened 2021-08-08
Make silent flag also hide "is up to date" logs


ProfessorManhattan:
For every task, I list the dependencies like this:

```
my-task:
  deps:
    - :npm:prettier
    - :software:jq
    - :software:yq
```

With the new `run: once` option it is working absolutely fantastic (thank you Task developers). However, since I include the deps on each task, I get a lot of spam from "is up to date" and the information is not that useful. Is there a way of silencing the tasks without silencing any of the other logs?

ghostsquad:
this could be a UI feature of V4. Making the UI less static/append-only could really improve things.

Though, @ProfessorManhattan are you interested in seeing/parsing the result? or just the spam of watching things run? Because I was also thinking of a "reporting" framework, so you could get the output of a task run in a machine readable format, like JSON.

![image](https://user-images.githubusercontent.com/903488/165890484-679c7901-9858-4ce7-a8bf-60bb4ab381c9.png)

ssbarnea:
To be honest I am not sure if we need a option to hide it. Instead we can only display it when verbosity level is increased via `-v`. For normal execution, I do not see much value in it, so if we ensure we display it only when we have extra verbosity, we sort the problem.

danquah:
Would it make sense to use the existing [silent mode](https://taskfile.dev/usage/#silent-mode) flag?

I ran in to this when I needed a task to only output the value i explicitly echo'ed as I needed another process to use the output. And setting `silent: true` worked great - except for any tasks that used a status.

So, for my need it would be perfect if `silent: true` also silenced status checks.

andreynering:
@danquah I think that's a good idea, indeed. I would be happy to accept a PR for that.

ghostsquad:
A better option would be use of stderr for operator only messaging, and stdout for machine messaging

danquah:
@ghostsquad as far as I can tell we [already log to stderr](https://github.com/go-task/task/blob/master/task.go#L180) in this case. I guess I could have solved my particular issue by just directing stderr away. I do really like to have the option of turning the verbosity down via the `silent` flag - and now that it's already there, it feels like something the Status feature should respect.

ghostsquad:
[...]

PR #1107, opened 2023-04-04
fix: status-checks should respect task and global silent-mode

danquah:
As discussed in #546 this pull-request updates the condition surrounding the "up to date" message emitted from the status-check to now also respect the task-level silence flag. It also allows the user to switch off silence via the Verbose flag.

I view this more as a fix than a new feature so I have not added any documentation. I'll be happy to do so if it is deemd nessecary.

### DIFF
--- a/task.go
+++ b/task.go
@@ -141,7 +141,7 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 			}
 
 			if upToDate && preCondMet {
-				if !e.Silent {
+				if e.Verbose || (!t.Silent && !e.Taskfile.Silent && !e.Silent) {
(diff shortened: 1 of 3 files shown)


task.go
```diff
@@ -141,7 +141,7 @@ func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 			}
 
 			if upToDate && preCondMet {
-				if !e.Silent {
+				if e.Verbose || (!t.Silent && !e.Taskfile.Silent && !e.Silent) {
```
andreynering, line 144: I see this was already being respected for the `--silent`, but not for the Taskfile or task level attribute.

One more reason to proceed in name of consistency. 🙁

andreynering:
Thanks for contributing @danquah!

PR #1110, opened 2023-04-06
refactor: logger singleton

pd93:
In some Go programs it is useful to pass loggers around using pointers as it allows us to create distinct loggers with different contexts and levels for different things. However, I don't think Task needs this? The logging is pretty basic and passing the logging object around can be annoying at times.

This PR refactors the logger so that instead of passing a pointer around, we simply have a global singleton logger that can be called from anywhere. The `logger` package has been renamed to `log` to make calling functions like `log.Out` more natural. Additionally, I have removed the `\n` from the logging functions to make their behavior more consistent with the standard library's `log`/`fmt` package. This gives the caller more control over the output's whitespace.

The global logger's config can be changed by calling the functions in the `log` package. This is useful for tests for example, where we can use `log.SetStdout()` to use a buffer as the writer etc.


titpetric:
Sharing a global variable for the logger is a testing defect, as you disable correct use of t.Parallel. Logging is a responsibility of itself, which could be wrapped behind an interface and enable better testing. Logger interfaces aren't anything new, the reasons are usually to replace some old logger, and to better enable single-responsibility testing, t.Parallel in particular, have test specific loggers,...

internal/compiler/v2/compiler_v2.go
```diff
@@ -106,7 +105,7 @@ func (c *CompilerV2) HandleDynamicVar(v taskfile.Var, _ string) (string, error)
 	opts := &execext.RunCommandOptions{
 		Command: v.Sh,
 		Stdout:  &stdout,
-		Stderr:  c.Logger.Stderr,
+		Stderr:  os.Stdout,
```
titpetric, line 108:
```suggestion
		Stderr:  os.Stderr,
```

internal/compiler/v2/compiler_v2.go
```diff
@@ -117,7 +116,7 @@ func (c *CompilerV2) HandleDynamicVar(v taskfile.Var, _ string) (string, error)
 	result := strings.TrimSuffix(stdout.String(), "\n")
 
 	c.dynamicCache[v.Sh] = result
-	c.Logger.VerboseErrf(logger.Magenta, `task: dynamic variable: '%s' result: '%s'`, v.Sh, result)
+	log.VerboseErrf(log.Magenta, "task: dynamic variable: %q result: %q\n", v.Sh, result)
```
titpetric, line 119: Logger should handle trailing whitespace

pd93: format loggers shouldn't handle trailing whitespace - In the same way that `fmt.Printf` does not. This gives the caller more control of the output. We should be consistent with the standard library.

pd93:
> Sharing a global variable for the logger is a testing defect, as you disable correct use of t.Parallel. 

@titpetric Thanks for the review, you make a good point. Though, I would point out that this is exactly how the standard `log` package works - and we're already using this in Task.

I guess this is a bit of a trade-off. I don't think we'll ever have a need to use `t.Parallel`. Our tests are very quick to run as it is. However, I can definitely see the pitfalls of using a global anything.

I might look into some other options. For example, there would be less need to pass loggers around if we improved how error handling works. (i.e. making the caller responsible for logging an issue rather than the called functions themselves)

titpetric:
From what I see the code is structured readonably well. I just imported the taskfile subpackage, and while tests may not need t.parallel, it's a canary in the mine situation. I could reasonably hit concurrency issues over the global state as I am doing work in a http.Handler, which runs in it's own goroutines. Loggers are usually thread safe, fortunately.

internal/compiler/v2/compiler_v2.go
titpetric: Stdlib log does not require trailing \n, fmt does; structured logging also does not require \n (in fact, it is escaped to prevent log forgery, prefer %q for params in stdlib).

PR #1112, opened 2023-04-07
docs: optimize document format

misitebao:
- [x] Add markdown formatting configuration
- [x] Format all documents using prettier


docs/docs/changelog.md
pd93, line 1: This file is generated. We need to tell prettier to ignore it. We can do this by creating a `.prettierignore` file and adding `docs/docs/changelog.md` to it. (Note: Annoyingly, this file must be in the project root for VSCode to detect it). Maybe we need to consider moving the Prettier config here too since we have MD files outside the docs folder too?

pd93 (commented):
A couple of minor comments, but otherwise I'm pretty happy. Obviously @andreynering will have the final say.

Also, as I said in one of my comments, we might want to look into using Prettier outside of the `docs` directory too if we're enforcing this styling on all markdown documents.

docs/docs/changelog.md
misitebao: Maybe we don't need two identical `Changelog` files, we can clear the Changelog file in the root directory of the repository and add a link to the `Changelog` file in the `docs` directory. What do you think?

pd93: The reason that we have the second changelog is because Docusaurus doesn't understand issue/PR [autolinks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls) (e.g. `#123` or `@pd93`), so we have a script that will convert these autolinks to FQDNs so that they are clickable on the website too.

Automating this conversion has the advantage that we don't have to manually write the cumbersome links every time we add to the changelog (e.g. `([#123](https://github.com/go-task/task/issues/123) by [@pd93](https://github.com/pd93))` instead of just writing `(#123 by @pd93)`. The latter is much more readable in the `CHANGELOG.md`.

Also, having the `CHANGELOG.md` in the project root makes it much more visible in the GitHub repo.

Since the `docs/docs/changelog.md` is a generated file, I don't thing we need to care about readability/line length since people shouldn't be editing/viewing it anyway.

misitebao: I know this problem, but the automatic identification link will only take effect in `ReleaseNote`, the automatic link in [https://github.com/go-task/task/blob/master/CHANGELOG.md](https://github.com/go-task/task/blob/master/CHANGELOG.md) will not work . 

So we only need to maintain a `changelog` in the `docs/docs` directory, and use normal links when writing `changelog` (e.g. `([#123](https://github.com/go-task/task/issues/123) by [@pd93](https://github.com/pd93)`). Then when releasing a new version, modify `ReleaseNote`, change `([#123](https://github.com/go-task/task/issues/123) by [@pd93](https://github.com/pd93)` to `(#123 by @pd93)`.

misitebao:
Hi, @andreynering @pd93 

This PR needs to be resolved first, otherwise it will affect Crowdin translation and automatic sync PR.

docs/docs/changelog.md
pd93: Personally, I don't ever want to have to write the long-form version of this link. It's a pain to write out and very error prone. This is the reason I wrote the release tool in the first place. @andreynering and I write most of the changelog entries, but I think that it's also easier for others to follow the format when its simple and most first-time contributors will see the `CHANGELOG.md` in the project root, but not even know that the one in `docs/docs` exists.

Another point is that updating the website changelog in each PR would update the website immediately when that PR is merged. I don't think we really want to update the website's changelog until a release is cut.

I see no issue with having 2 files as long as one is generated from the other. It's fairly simple to add the latter to a `.prettierignore`. We can maybe add a comment to the top that says its a generated file and that it shouldn't be edited too. Happy to be overridden on this, but my vote would be to keep it as it is.

misitebao: I added `.prettierignore` and it worked.

pd93: Unfortunately, most people are going to be opening the project from the Git root.

See related issue/PRs in the Prettier repo:

- https://github.com/prettier/prettier/issues/4081
- https://github.com/prettier/prettier/pull/6203

I particular, the final comment of the PR from the author: 

> I've long since worked around the issue myself by adding all rules to a root .prettierignore and only running prettier from the root. I'd recommend others do the same.

I would prefer to have the `.prettierignore` file where you have put it too. However, this won't work for the majority of users. The only other choice would be to put it in the project root _and_ the docs folder 🫤

andreynering: I'd put this file in the root directory personally. Most people probably open the repository from the root.

I guess adding to both folders wouldn't hurt, though.

misitebao: I checked these contents carefully, and I understand what you said, so is it acceptable to have `.prettierignore` files in the project root directory and subdirectories at the same time?

misitebao: I think both ways are fine, it's mostly up to you guys, my intention is just to fix this quickly to get Crowdin working as quickly as possible. 😆

pd93: My argument for moving the config is that I now have Prettier running on all `.md` files in the project (including the ones in the root), but the custom config we made is only applied to the files in the `docs` folder. We may as well be consistent everywhere and apply the same config to everything.

misitebao: I resubmitted, can you see if this works?

pd93 (approved):
LGTM. Just needs a final 👍 from @andreynering.

Thanks for all the effort you put into this. It's definitely a lot nicer to work with the docs now.

andreynering:
Thanks @misitebao!

PR #1113, opened 2023-04-08
Fix fish completions for tasks with aliases

patricksjackson:
The existing fish completions script is broken when tasks include aliases. This change adds an additional sed replacement to match against commands with aliases.

Here's an example recording with the existing completions, then sourcing and using this fixed version: https://asciinema.org/a/jVlyeadd643J6FC3o5V6927hM


andreynering:
Thanks @patricksjackson!

PR #1114, opened 2023-04-09
feat: custom error codes

pd93:
This PR adds custom error codes to Task. Currently, we only ever output an exit code of 1 when something goes wrong. This makes it difficult programs that want to have different behaviors depending on the reason that Task failed. Specifically, I want this so that the VSCode extension can handle parsing errors differently to taskfile not found errors. However, there are a lot of other places where this could be useful too.

As well as adding the error codes, I have also split the errors out of the root `task` package into their own (non-internal) `errors` package. This is useful for a couple of reasons. These errors can now be accessed from anywhere in the codebase without having to import the `task` package (which causes nasty circular imports) and it also makes it easier for anyone using Task as a library to check for specific errors (This is also why all the types are now exported).

The way that I have implemented the custom error codes is to make a new interface called `TaskError` which extends the standard `error` interface by adding a `Code() int` method. I have then split most of the code in the `main` function into a separate `run` function which returns a normal `error`. Now all that `main` does is check if the error returned by `run` implements `TaskError` and if it does, it will exit with the code returned by `Code()`. If the error does not implement `TaskError` then it will exit with the generic code 1.

Also worth noting that this is compatible with #755. If the `-x`/`--exit-code` flag is specified, then the exit code will overwritten by the exit code of the command being run.


andreynering (approved):
Nice idea and implementation. Thanks!

PR #1118, opened 2023-04-12
feat: use mockery/testify for mocking instead of mockgen/gomock

pd93:
We're already using [Testify](https://github.com/stretchr/testify) for assertions. Why not also use it for mocking and remove the additional dependency on [GoMock](https://github.com/golang/mock)?

Testify/mockery are more actively maintained, more popular and support all the same features that GoMock/mockgen do.


andreynering:
Thanks!

PR #1131, opened 2023-04-20
Add support for single-cmd tasks (fixes #1130)

### LINKED ISSUE #1130: Allow single command/cmd under task
I can't imagine this hasn't been discussed before but maybe I suck at using search. Anyway:

Personally, I try to keep my Taskfiles really succinct. As soon as a command/script becomes a little involved, I'll quickly fork out to a dedicated script file (sacrificing Go sh support). That means that my tasks are usually one-liners.

Now, when only specifying the command itself, I can use the shorthand task declaration format, which is really intuitive:

```yaml
tasks:
  foo: scripts/foo.sh
```

However, when the task require more configuration—e.g., `deps`, `desc`, `vars`—, that forces me to specify the same command as a single-item array, adding a bit of overhead:

```yaml
foo:
  deps: [bar, baz]
  cmds:
    - scripts/foo.sh
```

I suppose I could collapse the second array too, but it will quickly become a kludge:

```yaml
foo:
  deps: [bar, baz]
  cmds: ['[[ -f file ]] && scripts/foo.sh {{.CLI_ARGS}}']
```

So I was wondering if it wouldn't make sense to also accept a string instead of an array of strings—either as the existing `cmds` or under a new `cmd` key at the task level:

```yaml
foo:
  deps: [bar, baz]
  cmd: scripts/foo.sh
```

This shouldn't be confused with `cmd` _inside_ items in the `cmds` array. However, it serves the same purpose, and therefore, I think it makes sense to allow it one level higher as well.

While I'm personally also not a huge fan of these situations where configuration items are mutually exclusive, I think this particular feature would really allow us to write cleaner and less verbose Taskfiles.

Looking forward to hearing your thoughts. Thanks in advance.

timdp:
As discussed under #1130, I believe the [documented behavior](https://github.com/go-task/task/blob/c2523796c0de468c25b19f49ca257fe43ea470e2/docs/docs/api_reference.md?plain=1#L243-L244) of allowing a single `cmd` under a task is not implemented. This PR adds it—or at least my interpretation of it.


andreynering:
Thanks @timdp!

PR #1134, opened 2023-04-21
Fix for swallowing unkown errors - fixes #1133

### LINKED ISSUE #1133: Latest refactor of task/cmd/task/task.go is swallowing errors
> Thanks for your bug report!
>
> Before submitting this issue, please make sure the same problem was
> not already reported by someone else.
>
> Please describe the bug you're facing. Consider pasting example
> Taskfiles showing how to reproduce the problem.

- Task version: 3.24.0
- Operating System: MacOS Monterey 12.6

Here's the code:

https://github.com/go-task/task/blob/c2523796c0de468c25b19f49ca257fe43ea470e2/cmd/task/task.go#L46-L54

just need a `log.Fatal(err)` before line 52

I'll send a PR

clintmod:
>
>

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -49,6 +49,7 @@ func main() {
 			log.Print(err.Error())
 			os.Exit(err.Code())
 		}
+		log.Print(err.Error())
 		os.Exit(errors.CodeUnknown)
 	}
 	os.Exit(errors.CodeOk)


cmd/task/task.go
```diff
@@ -49,6 +49,7 @@ func main() {
 			log.Print(err.Error())
 			os.Exit(err.Code())
 		}
+		log.Fatal(err)
```
pd93, line 52: This should be `log.Print(err.Error())` IMO.

`log.Fatal()` will call `log.Print()` followed by `os.Exit(1)` - which is technically what we want. However, we should stick to exiting with the `errors.CodeUnknown` const that we have defined for clarity.

clintmod: @pd93 fixed

pd93 (approved):
LGTM 🚀

Issue #701, opened 2022-04-14
Linting of commands


orenmazor:
So this is not fully fleshed out in my head yet.

The gist is as such: my team currently uses makefiles + shell scripts to do things. this is obviously messy and awful. 

I've done a few PoCs to migrate those things into taskfile and it's a beautiful world.

but here's the rub: we use shellcheck in all of our CI/vim/local machines/etc to verify shell scripts for various code smells.

I'd love to figure out a way to do this in taskfile, but I haven't used this tool enough yet to figure out how to do it.

I can think of a few hacky ways to do it. for example: `task --shell -t mytaskfile.yaml mytask` and it dumps that task as a mini-shell script. I can then pipe that to shellcheck.

or even better, build shellcheck into task: `task --lint -t mytaskfile.yaml mytask` and it would run each command through shellcheck.

the goal isn't to verify anything or do any complex stuff. the goal is to catch weird little typos or bad commands or other basic tripwire features.

andreynering:
Hi @orenmazor,

You seem to work for 1Password? I use that for both work and personal use and love the tool. 🙂 

---

Interesting idea, indeed. I'd say it's a bit unconventional for Task to have official integrations with external tools, although integrations with [watchman](https://facebook.github.io/watchman/) and Docker were discussed in the past. Go stuff usually try to be self-contained when possible.

For this reason, it'd make more sense to try to do that with [mvdan/sh](https://github.com/mvdan/sh) given we already use it for actually running stuff. The problem is, it only contains a formatter, and not a linter AFAIK.

To allow shellcheck use it'd make more sense to have a flag to output the bash so it can be manually redirected to shellcheck (or something else), as you suggested. This is probably relatively simple, and pull requests would be welcome!

orenmazor:
hey Andrey, sure do. we're testing this tool out in our ops teams :)

and cool, I agree with you. a pure bash output would be the handiest as far as integrations go. I'll kick this idea around and maybe punt a PR at you folks :)

perobertson:
Having taskfile output the contents of the bash script would work great. That way each tool can work on their own without coupling them together. In my example here, we could replace the `echo "echo 'hello'"` with the task command that would output the shell script.

```
shellcheck <(echo "echo 'hello'")

In /proc/self/fd/11 line 1:
echo 'hello'
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
```

ghostsquad:
[...]

PR #1135, opened 2023-04-22
Add new flag 'shell' which outputs shell commands from a task

the-kaustubh:
The PR adds new flag `--shell` to the task command which works similar to dry run flag. It outputs raw shell command corresponding to the task without executing it.

Commit 22310215 also adds the same functionality to the pre-condition checks.

This PR addresses #701 Linting of commands

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


q0rban:
For the purposes of passing to another linter, such as `shellcheck`, would it need to do some preprocessing of template variables?

trulede:
It might be interesting to combine this with #2114 as a trace implementation. For instance:

`task default --tracer=audit `

where the "audit" prints resolved task variables, envars (send to `sh`) and the `sh` commands.

trulede:
I think here a neater approach would be to modify the existing `--dry` flag to allow optional values, such as `--dry=cmds`.

See #2395

trulede:
Actually, this is working via the dry command.

Issue #680, opened 2022-03-02
Override `silent` option when calling other tasks


HeCorr:
This does not as expected:

```yaml
tasks:
  notsilent:
    cmds:
      - echo "hey"

  maybesilent:
    cmds:
      # maybe some other tasks here, which I may not want to be silent
      - task: notsilent
        silent: true
```

![image](https://user-images.githubusercontent.com/75134774/156277328-6113f128-1244-490b-9a95-056392407ef7.png)

For comparison, this is what I expected to see:

![image](https://user-images.githubusercontent.com/75134774/156278046-f58bc79a-d02f-46c2-b8c7-326203018896.png)

It would be really good if it worked, for the cases where one might want the called task to **not** be silent when executed directly, but be silent when executed from another task.

I know there's the `-s` flag, but it wouldn't work for the case where one might want the other cmds to **not** be silent.

mknapik:
I had a similar use case. I want to run have a task for pull docker containers with optional `ignore_error` (e.g. pulling required docker images vs pulling CI build cache).

I tried to make it work with vars
```yaml
  docker-pull:
    run: when_changed
      - cmd: docker pull {{.DOCKER_IMAGE}}
        ignore_error: '{{.IGNORE_ERROR | default false}}'
```
The solution above was silently failing. It seems `ignore_error` expects boolean and fails silently when string (e.g. `"false"`) is passed.

When using `ignore_error` on the task level:
```yaml
  docker-pull:
    run: when_changed
    cmds:
      - docker pull {{.DOCKER_IMAGE}}
    ignore_error: '{{.IGNORE_ERROR | default false}}'
    silent: '{{.SILENT | default false}}'
```
I get an type error:
```
yaml: unmarshal errors:
  line 275: cannot unmarshal !!str `{{.IGNO...` into bool
```

mknapik:
https://github.com/go-task/task/issues/363 similar functionality, but for `ignore_error`.

4x0v7:
I have a similar use case, but using `includes:`

When the calling taskfile has `silent: false` at Task level:

```yaml
version: '3'
silent: false
```

If an included taskfile has:

```yaml
version: '3'
silent: true
```

Then output is displayed, where I expect the included files `silent` to take precedence.

Swapping them over; calling task is `silent: true`, called task is `silent: false`, then there is no output.

I'm not sure how the `silent` option should propagate with `includes:` tasks

danquah:
I'm considering looking at this after #546 

[...]

PR #1142, opened 2023-04-27
Silenced Calls

danquah:
This PR addresses #680 by fixing a behaviour and introducing a small new feature.

Task now support silencing calls to other calls. The `Cmd` struct already had a Silence field, but it did not actually silence the "task: " output that is produced whenever a task is invoked.

The fix has been implemented by expanding the [`Call`](https://github.com/danquah/task/blob/21a8bb025ba733616cb3abbf84b35a2a82522d74/taskfile/call.go) struct to now also pass a ´Silent` field.

Dependencies are also a type of `Call` and to stay consistent the `Dep` struct has also been expanded with a `Silent` field.


danquah:
I've updated the documentation and API to reflect that you can now specify a `silence` attribute on dependencies.

andreynering:
Thanks @danquah!

PR #1144, opened 2023-04-27
Fix duplicate error message

### LINKED ISSUE #1141: double error output
- Task version: 3.24.0
- Operating System: Mac OS

```shell
$ task test3
* test1:        Test 1
* test2:        Test 2
task: Task "test3" does not exist
task: Task "test3" does not exist
```

```yaml
version: '3'
tasks:
  test1:
    desc: Test 1
    cmds:
      - echo "test 1"
  test2:
    desc: Test 2
    cmds:
      - echo "test 2"
```

pd93:
Fixes #1141

image

I think this PR is a nice, clean solution to the problem. Unfortunately, as you can see, it also has the side-effect of removing the red coloured text from the error message since the executor/logger are out of scope in `main` 🙁 I'm not sure what the best solution is for this. We weren't always consistent with showing errors in coloured text anyway.

- We could potentially return the executor/logger back to the main function and use the logger that way. However, the logger isn't actually setup until we call `e.Setup()` and errors can occur before that which would result in nil pointer panics.
- We could just not use the logger at all and print a raw ANSI code to get the color, but we'd need to build in cases for the `--color` flag and `NO_COLOR` envvar.

@andreynering wdyt?

Also, since we needed access to the flags in the main function, I've moved them to a package-level struct variable. It means we can access them outside the `run()` function and also has the nice side-effect of namespacing the variables. i.e. `versionFlag` -> `flag.version`

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -43,6 +43,32 @@ tasks:
 Options:
 `
 
+var flags struct {
+	version     bool
+	help        bool
+	init        bool
+	list        bool
+	listAll     bool
+	listJson    bool
+	taskSort    string
+	status      bool
+	force       bool
+	watch       bool
+	verbose     bool
+	silent      bool
+	dry         bool
+	summary     bool
+	exitCode    bool
+	parallel    bool
+	concurrency int
+	dir         string
+	entrypoint  string
+	output      taskfile.Output
+	color       bool
+	interval    time.Duration
+	global      bool
+}
+
 func main() {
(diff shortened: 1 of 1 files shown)


andreynering:
Hey @pd93, thanks for the fix.

Take a look at the commit. I just instantiated the `logger.Logger` struct manually and it was enough to make it work and still respect settings like `--color=false`, `NO_COLOR=1`, etc.

Issue #140, opened 2018-10-16
Support declaring non-string (bool, maps, slices, etc) variables


philippgille:
When writing the `Taskfile.yml` many people probably write it like any other yaml, which means values like `false` is a boolean, as the yaml spec says: http://yaml.org/type/bool.html

But it turns out to be used as string and when using the template language it's actually `"false"`. This is bad for `{{if .SOME_VAR}}` for example, which doesn't work, so you have to write `{{if eq "true" .SOME_VAR}}` for example, or even a much longer expression given the valid forms of a bool in the spec (e.g. "no", "NO", "No", n", ...).

Note: I found the same issue in a different Go repository: https://github.com/concourse/concourse/issues/294
Unfortunately they only link to a Pivotal tracker and the link is 404, so I don't know if and how they dealt with this issue.

If you explicitly don't want to adhere to the yaml spec (and of course there might be good reasons for it), this behaviour should at least be mentioned in the documentation. Currently the "Variables" section doesn't do this: https://taskfile.org/#/usage?id=variables

Example Taskfile, created with `task --init` and then customized:

```yml
# github.com/go-task/task

version: '2'

vars:
  GREETING: Hello, World!
  IS_LIB: false

tasks:
  default:
    cmds:
      - echo "{{.GREETING}}"
      - echo "{{.IS_LIB}}"
      - echo "{{if .IS_LIB}} error {{else}} ok {{end}}"
      - echo "{{if eq "true" .IS_LIB}} error {{else}} ok {{end}}"
    silent: true
```

The output is:
```
task: No argument given, trying default task
Hello, World!
false
 error 
 ok 
```

task Version: Commit `309cfb14995ec511546726ab607d53c55c803d0f`

OS: Ubuntu 14.04

smyrman:
I have met this limitaruon as well. With some effort, I believe not only bolean, but also array and nuneric values could be supported as static values.

Unfortunantly maps would be of the table
 because the syntax for dynamic variables is:

```yaml
vars:
  myvar:
    sh: cat myfile.txt
```

andreynering:
Hi @philippgille,

As said by @smyrman, there's a syntax conflict with the `sh:` keyword, so implementation wise we need to find a workaround.

That said, I agree this would be very useful.

This is slightly related to #82

smyrman:
Well, there is no conflict for Booleans, numbers and arrays.

```yaml
vars:
  myvar: <string, bool, number or array>
```

[...]

PR #1146, opened 2023-05-02
add sources special var

### LINKED ISSUE #948: Use sources as variable
```
version: '3'

tasks:
  test:
    cmds:
      - echo {{.sources}}
    sources:
      - *.sh
```
```
version: '3'

tasks:
  test:
    cmds:
      - echo {{.sources}}
    vars:
      SCRIPTS:
    sources:
      - {{.BASH_SCRIPTS}}
```

hernandanielg:
Fixes #948

This is my attempt to implement the sources special variable, it does the trick but I am unsure if it should be this way... please provide some feedback ;)

Example test
```
$ cat ./testdata/special_vars/Taskfile.yml
tasks:
  default:
    cmds:
      - task: print
      - task: included:print

  print:
    cmds:
      - echo root/TASK={{.TASK}}
      - echo root/ROOT_DIR={{.ROOT_DIR}}
      - echo root/TASKFILE_DIR={{.TASKFILE_DIR}}
      - echo root/TASK_VERSION="{{.TASK_VERSION}}"
      - echo root/SOURCES="{{.SOURCES}}"
    sources:
     - '*.sh'

$ go run ./cmd/task --dir ./testdata/special_vars/ -f print
task: [print] echo root/TASK=print
root/TASK=print
task: [print] echo root/ROOT_DIR=/Users/hernandanielg/src/task/testdata/special_vars
root/ROOT_DIR=/Users/hernandanielg/src/task/testdata/special_vars
task: [print] echo root/TASKFILE_DIR=/Users/hernandanielg/src/task/testdata/special_vars
root/TASKFILE_DIR=/Users/hernandanielg/src/task/testdata/special_vars
task: [print] echo root/TASK_VERSION="(devel)"
root/TASK_VERSION=(devel)
task: [print] echo root/SOURCES="install-task.sh"
root/SOURCES=install-task.sh
```


andreynering:
Hi @hernandanielg, thanks for this PR!

As I mentioned on this comment https://github.com/go-task/task/issues/948#issuecomment-1338627486 I think this variable should probably be a slice/array of strings, instead of a string.

This way, the user can use [slim-sprig](https://go-task.github.io/slim-sprig/) to manipulate the list however they want, adding more flexibility to this variable.

andreynering:
For reference, as said on Discord, this may need some refactor work to support types other than string on the variables map.

There's this old issue about it: #140. In theory it would allow even users themselves to declare array, maps, etc.

trulede:
Closing in favour of #2047 which has most of the requested changes.

PR #1163, opened 2023-05-15
feat(prompts): add ability for tasks to prompt user pre execution

### LINKED ISSUE #100: A warning/confirm field that shows a confirm prompt before proceeding

MaxCheetham:
fixes #100 

Adding a `Prompt` filed to allow users to set a message to be shown before task excecution.
Also added an early exit code (205) for when this happens


andreynering:
A section on the Usage page in the documentation could be interesting as well.

Another contributor [wrote this](https://github.com/go-task/task/pull/352/files#diff-72376d0b487d7231cf31959bf266f6aea098e899743bae02e36d176cef4db476) for an implementation that end up not being merged. You could use that as inspiration.

MaxCheetham:
Hey @andreynering, thanks for catching that bug, and also the request for the enchancements.

- I added an `AssumeYes` flag (`--yes` | `-y-`)
- Fixed the bug you found by moving the prompting closer to task execution
    - this has the (upside / downside?) of meaning that tasks can run pre-prompt if they themselves have no prompt. A possible extension to this could be a `skippable` bool in the task file that could allow for a prompted task that was denied to be skipped, however, I felt like the implacations of this change would be too big to include in this PR, and would merit some descussion before hand.
- I added a section to the Usage docs, I will be the first to admit that my technical writing isn't fantastic, so any advise on wording / structure would be taken onboard gladly. 🚀

andreynering:
Thanks @MaxCheetham!

I did  a few improvements on 44aaec86a10dfac0871d677563ff5e80890f46ad. The most important one being a fix that forces the task to fail on a non-terminal environment (imagine CI). Without this, the task would hang forever waiting for someone to type `y`, but that would be impossible on these environments.

PR #1168, opened 2023-05-15
powershell completion: disable file path autocomplete / add flags

trim21:
>
>

Didn't know there is a command `task --list-all --json/--silent` for all available tasks when I submit last PR.

Previous, if your pwd have a dir or file named `gen` or `gen...`, and you have a task named `gen`,
when you type `task gen<tab>`, powershell will fallback to filepath complete, and your input will become `task .\gen`

after this pr, `task gen<tab>` are still `task gen`

File path auto completion still works with `.\<tab>` or `./<tab>`

### DIFF
--- a/completion/ps/task.ps1
+++ b/completion/ps/task.ps1
@@ -54,6 +54,6 @@ Register-ArgumentCompleter -CommandName task -ScriptBlock {
 		}
 
 
-		return $desc + "`n"
+		return $desc
 	}
 }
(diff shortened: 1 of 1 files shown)


trim21:
I try to add short opt but looks like old powershell (5.1, win10 default version) doesn't work will with that, So I'll just add long flags.

trim21:
adding description feature, convert to draft

trim21:
description can also be padded to look better, but maybe leave it to someone who are interested.

trim21:
description feature make it much slower to finish completion ( I guess it's caused by parse json), so I removed it.

andreynering:
Thanks!

PR #1181, opened 2023-05-23
chore: add environment variables to the deployment environment

misitebao:
Works:
- [x] Add environment variables to the deployment environment


andreynering:
Thanks @misitebao!

Issue #1185, opened 2023-05-25
Checksum checker modifies files even if the md5 hash matches


deviantintegral:
> Thanks for your bug report!
>
> Before submitting this issue, please make sure the same problem was
> not already reported by someone else.
>
> Please describe the bug you're facing. Consider pasting example
> Taskfiles showing how to reproduce the problem.

- Task version: Task version: v3.24.0 (h1:0DKsGTmqYbEaACffm9H2BOm8pTcuSnb5SRUl7TCilQ4=) (but confirmed in main)
- Operating System: Linux

When a task is up to date, the checksum file is modified even if the hash is the same from a `sources` configuration:

```
$ stat .task/checksum/override-drupal-composer; task --verbose override:drupal:composer; stat .task/checksum/override-drupal-composer
  File: .task/checksum/override-drupal-composer
  Size: 33        	Blocks: 8          IO Block: 4096   regular file
Device: fd11h/64785d	Inode: 8522386     Links: 1
Access: (0600/-rw-------)  Uid: (  501/  andrew)   Gid: (   20/ dialout)
Access: 2023-05-25 15:39:58.990003715 +0000
-----> Modify: 2023-05-25 15:39:58.980003715 +0000
-----> Change: 2023-05-25 15:39:58.980003715 +0000
 Birth: 2023-05-16 20:16:16.020001609 +0000
task: "override:drupal:composer" started
task: status command  test -f ./vendor/composer/installed.json && grep -q '"dev": true' ./vendor/composer/installed.json
 exited zero
task: Task "override:drupal:composer" is up to date
  File: .task/checksum/override-drupal-composer
  Size: 33        	Blocks: 8          IO Block: 4096   regular file
Device: fd11h/64785d	Inode: 8522386     Links: 1
Access: (0600/-rw-------)  Uid: (  501/  andrew)   Gid: (   20/ dialout)
Access: 2023-05-25 15:39:58.990003715 +0000
-----> Modify: 2023-05-25 15:40:20.630003725 +0000
-----> Change: 2023-05-25 15:40:20.630003725 +0000
 Birth: 2023-05-16 20:16:16.020001609 +0000
```

In our case, we have multiple processes running task at once, and they rely on some earlier code to make sure third-party dependencies are installed and that we don't have multiple composer processes running at once.

I think we should do something like this to prevent rewriting the file if the hash hasn't changed. At least in my initial test, this fixes our issue.

If this approach sounds good, I'm glad to open a PR for this.

```diff
[...]

PR #1194, opened 2023-05-30
fix: only rewrite checksum files if the checksum has changed

deviantintegral:
Resolves https://github.com/go-task/task/issues/1185.

I verified that the test case fails on main as well.


andreynering:
Thanks @deviantintegral!

PR #1198, opened 2023-05-31
feat: v2 deprecation and experiments

pd93:
- Docs
  - Experiments
    - A section on the version 2 schema deprecation.
    - A template for future experiments/deprecations.
  - Experiments / Workflow
    - A subsection describing the workflow/procedure for adding experiments
  - Contribution Guide
    - A small section on experiments linking to the workflow
  - Taskfile versions
    - Added deprecation warnings
    - Reordered the file so that the newest schemas are at the top.
- Added an entry to the changelog to improve visibility of the v2 schema deprecation
- Updated issue templates
  - Use HTML comments so that people don't accidentally submit issues with the `>` quoted text
  - Added `Experiments enabled:` to the bug reports
- Added workflows that comment on an issue when experiment labels are added. Copy for each state is pasted below:

### Proposed

This issue has been marked as an experiment proposal! :test_tube: It will now enter a period of consultation during which we encourage the community to provide feedback on the proposed design. Please see the [experiment workflow documentation](https://taskfile.dev/experiments/workflow) for more information on how we release experiments.

### Draft

This experiment has been marked as a draft! :sparkles: This means that an initial implementation has been added to the latest release of Task! You can find information about this experiment and how to enable it in our [experiments documentation](https://taskfile.dev/experiments). Please see the [experiment workflow documentation](https://taskfile.dev/experiments/workflow) for more information on how we release experiments.

### Candidate

This experiment has been marked as a candidate! :fire: This means that the  implementation is nearing completion and we are entering a period for final comments and feedback! You can find information about this experiment and how to enable it in our [experiments documentation](https://taskfile.dev/experiments). Please see the [experiment workflow documentation](https://taskfile.dev/experiments/workflow) for more information on how we release experiments.

### Stable

This experiment has been marked as stable! :metal: This means that the  implementation is now final and ready to be released. No more changes will be made and the experiment is safe to use in production! You can find information about this experiment and how to enable it in our [experiments documentation](https://taskfile.dev/experiments). Please see the [experiment workflow documentation](https://taskfile.dev/experiments/workflow) for more information on how we release experiments.

### Released

This experiment has been released! :rocket: This means that this is no longer an experiment and is available in the latest major version of Task. Please see the [experiment workflow documentation](https://taskfile.dev/experiments/workflow) for more information on how we release experiments.

### Abandoned

[...]

PR #1199, opened 2023-06-03
feat: add a deprecation warning when running Taskfiles with a v2 schema

pd93:
Added a warning to the output when using a `version: 2` schema. (This functionality was split out of #1198).

image

This might break things for people using v2 schemas in scripts if they update... we could stop output if the `--silent` flag is specified, but this might not stop it from breaking where some tasks have `silent: true`. Maybe we send the output to stderr instead? Thoughts welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,13 +3,18 @@ name: Bug Report
 about: Use this to report bugs and issues
 ---
 
-> Thanks for your bug report!
->
-> Before submitting this issue, please make sure the same problem was
-> not already reported by someone else.
->
-> Please describe the bug you're facing. Consider pasting example
-> Taskfiles showing how to reproduce the problem.
+<!--
+
+Thanks for your bug report!
+
+Before submitting this issue, please make sure the same problem was not
+already reported by someone else.
+
+Please describe the bug you're facing. Consider pasting example Taskfiles
+showing how to reproduce the problem.
+
+-->
 
 - Task version:
-- Operating System:
+- Operating system:
+- Experiments enabled:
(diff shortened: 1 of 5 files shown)


andreynering (approved):
Answering some of your questions:

- Yes, sending to `stderr` would be more appropriate, and mitigate possible breakage in scripts
- Yes, respecting the `--silent` flag be interesting. I wouldn't worry too much about `silent: true` set on tasks, though. v2 has very few users currently.

With these points addressed, I think this is good to go! 👍

andreynering:
One more thing: this is worth a mention on the CHANGELOG, IMO.

> Schema version 2 is now considered deprecated and scheduled to be removed in the next major release.

pd93:
> Yes, sending to `stderr` would be more appropriate, and mitigate possible breakage in scripts

Updated

> Yes, respecting the `--silent` flag be interesting. I wouldn't worry too much about `silent: true` set on tasks, though. v2 has very few users currently.

It's actually kind of annoying to pass in the flags to the version checks function. I could move the check into the main function, but I think with this being output to stderr now, this should be necessary anymore?

> One more thing: this is worth a mention on the CHANGELOG, IMO.

Ah yeah, I forgot to bring this change over from #1198. I've added it now, Thanks

andreynering:
> It's actually kind of annoying to pass in the flags to the version checks function. I could move the check into the main function, but I think with this being output to stderr now, this should be necessary anymore?

Agreed. No need to check `--silent` anymore.

CI broke. Probably easy to fix, just a matter of check `stderr` instead of `stdout` on the test, I believe.

pd93:
> CI broke. Probably easy to fix, just a matter of check stderr instead of stdout on the test, I believe.

Whoops, thank you 😛

PR #1201, opened 2023-06-03
chore(website): update website colors

andreynering:
Make light theme less white and dark theme less black. I find the legibility better with a slightly reduced contrast. This in a work in progress ™️. Opening to PR to get a quick opinion, maybe additional suggestions.

We can keep improving with time. Our theme is very basic currently, mostly identical to Docusaurus' default one.

/cc @pd93 @misitebao 

### Before

Screenshot 2023-06-03 at 15 52 35

### After - Option A

Screenshot 2023-06-03 at 15 52 55

### After - Option B

Screenshot 2023-06-03 at 18 59 34

### Before

Screenshot 2023-06-03 at 15 53 11

### After

Screenshot 2023-06-03 at 15 53 37

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -13,9 +13,49 @@
   --ifm-color-primary-lightest: #5D9993;
   --ifm-code-font-size: 95%;
   --docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.1);
+
+  --ifm-navbar-link-color: #fffdf9;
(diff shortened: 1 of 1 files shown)


pd93:
I really like option B for the light theme. The contrast between the title bar and the main page is a definite improvement. The off-white looks better with a dark title bar too. 

No strong opinion on the dark theme background colour, but I agree that reducing the contract helps the readability.

PR #1202, opened 2023-06-03
docs: add comments from #1115 to faq and styleguide

### LINKED ISSUE #1115: Add option to reuse shell
I think it would be beneficial to allow for execution of an entire task in a single shell.

# Motivation

It is currently very difficult to use Python's `venv`s or `conda`, due to the shell being reset on each command. The recommended way is to use absolute paths inside the venv, such as `venv/bin/python` instead of `source venv/bin/activate;python`. This however increasing the friction for newcomers and not all problems of this kind have similar solutions. 

My current Taskfile, looks like this:

```yaml
[...]
tasks:
  build:
    cmds:
      - mamba env create -f environment.yaml || true
      - mamba env update -f environment.yaml --prune
      - bash -c "
        source conda_init;
        mamba activate csidrl;
        poetry install;
        cd lib/SATNet;
        python setup.py install;
        "
```

Which is ugly and complicated. This approach also complicates cross-platform implementations.

There are other workarounds, such as in #512. There would be other uses of this feature, for example resolving #204.

# Proposal

For a task, add `subshell` boolean. This boolean would be by default `true`, ensuring compatibility. On `false`, this option would make `go-task` invoke all `cmds` in a single `mvdan.cc/sh/v3/interp`. When false, this option would also cause all `task:` cmds to run in the same `interp`. 

The unclear question is a treatment of `deps`. The current behavior would see them running in different shells, which may produce some confusion. The consistent behavior with the rest of this proposal would however sacrifice concurrency for simplicity and make `deps` run linearly after each other in the same shell (**when subshell: false only**). This behavior would also resolve #204. Neither of these options would result in consistent behavior throughout the `go-task` though.

# Advantages

- Enabling or vastly simplifying source-based workflows
- Consistency with (ba/z/)sh scripts
- Resolving the use-case in #204 

## My humble opinion

I believe that each of the advantages by themselves would be a strong-enough for this feature to be accepted, but there are some issues.

Even though this feature may simplify some workflows, the added complexity may not be worth it. The code base as I understand it also *wildly* doesn't support this, so it would be a major change. I am willing to take a crack at it though.

> Neither of these options would result in consistent behavior throughout the `go-task` though.

In my personal opinion, both options are better than the current behavior.

pd93:
Fixes #1115 by adding a section to the FAQ and Styleguide about not being able to reuse shells in a Task's commands.

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -7,9 +7,6 @@ sidebar_position: 6
 
 This page contains a list of frequently asked questions about Task.
 
-- [Why won't my task update my shell environment?](#why-wont-my-task-update-my-shell-environment)
-- ['x' builtin command doesn't work on Windows](#x-builtin-command-doesnt-work-on-windows)
-
 ## Why won't my task update my shell environment?
 
 This is a limitation of how shells work. Task runs as a subprocess of your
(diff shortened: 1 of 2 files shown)


andreynering:
Good @pd93.

I just think that #1115 is still a valid proposal and a possible experiment, so I will keep that issue open for now.

andreynering:
For some reason GitHub does not allow me to remove that issue from the list 🙄 🤦‍♂️. I'll have to re-open it.

PR #1211, opened 2023-06-09
Better docs for `--sort`

pd93:
Added possible values for the `--sort` flag in the API reference and usage text as per a recent question on the Discord server.

Also generalised the usage text as it was out-of-date and a list of possible flags is already printed below.


andreynering:
Thanks!

PR #1214, opened 2023-06-10
Support for experiment env vars

pd93:
This PR adds support for experimental feature environment variables. This follows the design discussed in the [breaking changes proposal](https://github.com/go-task/task/issues/1191).

Environment variables can be specified in 3 ways:

1. Using the relevant environment variable in front of a task command. For
   example, `TASK_X_<FEATURE>=1 task <my-task>`. This is intended for one-off
   invocations of Task to test out experimental features.
1. Using the relevant environment variable in your "dotfiles" (e.g. `.bashrc`,
   `.zshrc` etc.). This is intended for permanently enabling experimental
   features in your environment.
1. Creating a `.env` file in the same directory as your root Taskfile that
   contains the relevant environment variables. e.g.

```shell
# .env
TASK_X_FEATURE=1
```

The dotenv reader will only read environment variables that start with `TASK_X_` to avoid polluting the environment with other variables.

### DIFF
--- a/docs/docs/experiments/experiments.md
+++ b/docs/docs/experiments/experiments.md
@@ -21,13 +21,19 @@ are intended to replace.
 
 You can enable an experimental feature by:
 
-1. Using the `--x-{feature}` flag. This is intended for one-off invocations of
-   Task to test out experimental features. You can also disable a feature by
-   specifying a falsy value such as `--x-{feature}=false`.
-1. Using the `TASK_X_{FEATURE}=1` environment variable. This is intended for
-   permanently enabling experimental features in your environment.
+1. Using the relevant environment variable in front of a task command. For
+   example, `TASK_X_{FEATURE}=1 task {my-task}`. This is intended for one-off
+   invocations of Task to test out experimental features.
+1. Using the relevant environment variable in your "dotfiles" (e.g. `.bashrc`,
+   `.zshrc` etc.). This is intended for permanently enabling experimental
+   features in your environment.
+1. Creating a `.env` file in the same directory as your root Taskfile that
+   contains the relevant environment variables. e.g.
 
-Flags will always override environment variables.
+```shell
+# .env
+TASK_X_FEATURE=1
+```
 
 ## Current Experimental Features and Deprecations
 
@@ -38,11 +44,11 @@ existing Taskfiles to the new behavior.
 
 <!-- EXPERIMENT TEMPLATE - Include sections as necessary...
 
-### ![experiment] <Feature> ([#{issue}](https://github.com/go-task/task/issues/{issue})), ...)
+### ![experiment] {Feature} ([#{issue}](https://github.com/go-task/task/issues/{issue})), ...)
 
-- Flag to enable: `--x-{feature}`
-- Env to enable: `TASK_X_{feature}`
+- Environment variable: `TASK_X_{feature}`
 - Deprecates: {list any existing functionality that will be deprecated by this experiment}
+- Breaks: {list any existing functionality that will be broken by this experiment}
 
 {Short description of the feature}
 
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -1,0 +1,43 @@
+package experiments
+
+import (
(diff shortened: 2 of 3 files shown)


docs/docs/experiments/experiments.md
```diff
@@ -21,13 +21,19 @@ are intended to replace.
...
 
-1. Using the `--x-<feature>` flag. This is intended for one-off invocations of
-   Task to test out experimental features. You can also disable a feature by
-   specifying a falsy value such as `--x-<feature>=false`.
-1. Using the `TASK_X_<FEATURE>=1` environment variable. This is intended for
-   permanently enabling experimental features in your environment.
-
-Flags will always override environment variables.
+1. Using the relevant environment variable in front of a task command. For
+   example, `TASK_X_<FEATURE>=1 task <my-task>`. This is intended for one-off
```
andreynering, line 25: I just pushed a bug fix to `main` that caused a conflict on this PR.

I replaced `<VAR>` occurrences with `{VAR}` because Crowdin sometimes tries to parse `<foo>` as a HTML tag, making the Docusaurus build fail.

internal/experiments/experiments.go
```diff
@@ -0,0 +1,67 @@
...
+	return os.Getenv(envName(xName)) == "1"
+}
+
+func readDotEnv() error {
+	env, err := godotenv.Read()
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	} else if err != nil {
+		return err
+	}
```
andreynering, line 59: `else` not needed when `if` is a `return`.

```suggestion
	if errors.Is(err, os.ErrNotExist) {
		return nil
	}
	if err != nil {
		return err
	}
```

internal/experiments/experiments.go
```diff
@@ -0,0 +1,67 @@
...
+const envPrefix = "TASK_X_"
+
+var Flags struct{}
+
+func Parse() error {
+	if err := readDotEnv(); err != nil {
+		return err
+	}
+	t := reflect.TypeOf(&Flags)
+	v := reflect.ValueOf(&Flags)
```
andreynering, line 22: I'm confusing on what exactly you're trying to accomplish here. Can you elaborate?

Why do we need reflection? Couldn't this be a map/slice or similar just to store the name of the experiments that are enabled?

pd93: No problem, the idea behind this was that we have a single struct definition for all our experiments. e.g.

```go
type Flags struct {
    GentleForce bool `x:"gentle force"`
}
```

I am then using reflection to look at each struct field and its tags to fetch the environment variables. Doing it in this way means that we don't need to add any custom code when experiments are added. They just need to be defined in the struct, in the same way you would for a struct that marshals JSON and are then accessible from anywhere.

I've pushed a commit that changes it to use a map instead (I'll squash this later). This means that users would need to specify the experiment via a string const rather than it being a struct. It's probably a little faster, but to check if an experiment is enabled you have to do:

```go
if experiments.IsEnabled(experiments.TestExperiment) {
	// do experiment stuff
}
```

Instead of 

```go
if experiments.Flags.TestExperiment {
        // do experiment stuff
}

Happy to move forwards with this if you prefer this approach.

Comment (andreynering) [reply internal/experiments/experiments.go]: I now got what you tried to do.

In theory we could also do:

```go
var TestExperiment bool

func init() {
	TestExperiment = IsEnabled("TestExperiment")
}
```

or

```go
func IsEnabledTestExperiment() bool {
	return IsEnabled("TestExperiment")
}
```

I don't have any strong opinion between these three methods, but I think they are better than the first approach of using reflection. Any one of these three styles is fine to me.

andreynering (approved):

internal/experiments/experiments.go
pd93: Actually, I really like the `init()` way. This gives the nicest possible API imo:

```go
if experiments.GentleForce {
    // do stuff
}
```

You also only need to update the file in 2 places this way when adding an experiment. I've updated this PR and the gentle force experiment PR.

Issue #1200, opened 2023-06-03
Gentle force experiment proposal


pd93:
## Context

As discussed in [#189](https://github.com/go-task/task/issues/189), the current behavior of the `--force` flag is to run **all** tasks (i.e. ignoring all fingerprinting checks on the called task _and its dependant tasks_). This can be useful, but I think that most users would expect the force flag to only force the direct task that the user is calling. Implicitly running all dependant tasks could be dangerous and unexpected.

Here's a real world scenario:

If I have a task called `setup` that configures a local k8s cluster and takes a few minutes to complete, I might want to add a `status` check that queries the cluster to see if it is already running. I can then make any tasks that perform operations on the cluster depend on the `setup` task. This way, if the cluster is not running, it will create it before running the called task, but if the cluster is already running, it will skip the `setup` task and run the called task immediately.

If I then create a parent task (let's call it `deploy`), and this Task also has fingerprinting on it, I would expect that if I run `task deploy --force`, it would only force the `deploy` task and not force the `setup` task which takes a long time to complete.

```yaml
version: 3

tasks:
  setup:
    status: # check if k8s cluster is running
    cmds:
      - # steps to run cluster

  deploy:
    deps: [setup]
    status: # check if app is already deployed
    cmds:
      - # steps to deploy app to cluster
```

This example is pretty basic and not particularly dangerous (mostly inconvenient), but I can imagine scenarios where this could accidentally run destructive tasks unintentionally.

## Proposal

Theoretically, we could add a `--force-task` flag which will only force the called task and not its dependant tasks. However, I feel that this is counterintuitive and that it should be "harder" or "more explicit" to force all tasks than it is to force only the called task.

Instead, I propose that we change the behavior of the `--force` flag to only force run the _called_ task and that we add a new flag called `--force-all` which will force run the called task and all of its dependant tasks.

Since this is a breaking change, we would need to add this change as an experiment. The flags to enable this experiment would be:

- Flag to enable: `--x-gentle-force`
- Env to enable: `TASK_X_GENTLE_FORCE=1`

It is important to note that this experimental flag _will not_ actually force a task when used. Instead, it will change the behavior of the existing `--force` flag and **add** the ability to call a new flag `--force-all` which will force all tasks. In other words, to get the new experimental behavior using only flags, you would run a command like this:

- `task --x-gentle-force --force deploy` - Forces only `deploy` task
- `task --x-gentle-force --force-all deploy` - Forces `deploy` task and all of its dependant tasks

[...]

PR #1216, opened 2023-06-11
feat: implement gentle force experiment draft

pd93:
Note: This depends on #1214 

Implements a draft version of the [gentle force experiment](https://github.com/go-task/task/issues/1200). Based on the experiments branch and [PR](https://github.com/go-task/task/pull/1214).

This works by adding a new field to the `taskfile.Call` struct called `Direct` which is only set when a task is called directly from the CLI. Another new field is added to the `Executor` called `ForceAll`. Depending on whether or not the experiment is enabled, we set the Executor's `Force` and `ForceAll` fields and then we will skip fingerprinting if the task is direct and the executor is set to force, OR if the executor is set to force all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@
   - e.g. `taskfile.yml`, `taskfile.yaml`, `taskfile.dist.yml` &
     `taskfile.dist.yaml`
 - Bug fixed were made to the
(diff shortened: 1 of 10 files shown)


andreynering (approved):
Nice work! 👏 👏 👏

PR #1231, opened 2023-06-22
feat: Added support for lazy variables evaluation

tohutohu:
This Pull Request supports delayed evaluation of dynamic variables, adding a `lazy` option to the `vars[]` schema that makes the execution of global dynamic variable shell commands happen only when necessary.

## What this feature
In the current go-task, global dynamic variables always execute their commands. Therefore, with a Taskfile.yaml like the following, running `task depend-on-only-fast-var` always causes a 3-second wait due to `SLOW_VAR`.

```yaml
version: "3"

vars:
  SLOW_VAR:
    sh: sleep 3 && echo "slow"
  FAST_VAR:
    sh: echo "fast"

tasks:
  depend-on-slow-var:
    cmds:
      - echo {{.SLOW_VAR}}
  depend-on-only-fast-var:
    cmds:
      - echo {{.FAST_VAR}}
```

With the functionality added in this PR, dynamic variable evaluations can be delayed until necessary. This means that with a Taskfile.yaml like the following, `task depend-on-only-fast-var` can be executed immediately.

```yaml
version: "3"

vars:
  SLOW_VAR:
    sh: sleep 3 && echo "slow"
    lazy: true
  FAST_VAR:
    sh: echo "fast"
    lazy: true

tasks:
  depend-on-slow-var:
    cmds:
      - echo {{.SLOW_VAR}}
  depend-on-only-fast-var:
    cmds:
      - echo {{.FAST_VAR}}
```

## How 
Instead of passing strings as variables to the text/template, a structure that implements the `Stringer` interface is passed, which executes the command and uses the result as a dynamic variable the first time it's called. That result is then cached in the structure, ensuring that the same command is never executed twice. The Singleflight library guarantees that it will be executed only once even if called concurrently.

## Compatibility with this PR
This PR does not break compatibility. Dynamic variables that are not marked as `lazy: true` will continue to operate as before.


andreynering:
Hi @tohutohu, thanks for opening this PR!

It's an interesting idea to use the `Stringer` interface, indeed.

I haven't reviewed the code yet (there would be some changes), but I want to discuss something first:

What if we apply this for every variable? It doesn't seem like it would be a breaking change to me initially. Behavior would change slightly in that a variable that is unused would not run, but I think it'd be unlikely that this would breaking something for a real user.

your-diary:
Any updates?

---

This PR may solve or at least provide a workaround for these discussions:

- https://github.com/go-task/task/discussions/925

- https://github.com/go-task/task/discussions/708

These are a real blocker for me because I'm accustomed to GitHub Actions, in which we can easily pass the result of a command (*step* in GitHub Actions) to another command (another step).

---

> What if we apply this for every variable?

I second this for better performance (though I haven't checked the implementation of the PR).

PR #1241, opened 2023-06-27
Small prompt fixes

pd93:
This PR fixes a couple of minor output related "problems" with the new `prompt` feature:

- Removes duplicated double quotes. We don't need to add quotes when using `%q`:
- Removes the newline after the prompt is requested. This makes the output a bit tidier:

Before:
image
image

After:
image
image


andreynering:
Thanks @pd93!

PR #1245, opened 2023-06-30
fix: ignore all errors when reading dotenv for experiments

### LINKED ISSUE #1244: if .env is not a file, the program panics
- Task version: 3.27.0
- Operating system: mac os 13.1
- Experiments enabled: no

This is the error statement when Task CLI is executed `panic: read .env: is a directory`

I have an .env directory in a project I'm working on. In go-task 3.27.0 version, go-task expects .env to be a file not a directory, so the presence of a directory with the name .env causes panic.

pd93:
Fixes #1244 

Simply squashes all errors related to reading the dotenv. If its not there, we don't care. We just don't load anything from it.

### DIFF
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -1,7 +1,6 @@
 package experiments
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks! I'm going to release a patch version.

Issue #996, opened 2023-01-30
Vars don't get properly propagated with two levels of `included` nesting


gsantoro:
Hello,
thanks for the amazing tool that Taskfile is. I have a bit of advanced use case, so bear with me.

I am facing an issue with the current setup:
- I have three Taskfiles: A, B, C
- A includes B and calls B:up
- B includes C and calls C:up
- C has a variable that I want to be overridable: `C_VARIABLE: '{{.C_VARIABLE | default "localhost"}}'`

this is the content of A.Taskfile.yml:

```yaml
version: '3'

includes:
  B: 
    taskfile: ./B.TaskFile.yml

tasks:
  up:
    cmds:
      - task: B:up
```

this is the content of B.Taskfile.yml:

```yaml
version: '3'

vars:
  B_VARIABLE: b

includes:
  C: 
    taskfile: ./C.TaskFile.yml
    vars:
      C_VARIABLE: "{{.B_VARIABLE}}"

tasks:
  up:
    cmds:
      - echo "B_VARIABLE {{.B_VARIABLE}}"
      - task: C:up
```

and this is the content of C.Taskfile.yml

```yaml
version: '3'

vars:
  C_VARIABLE: '{{.C_VARIABLE | default "localhost"}}'

tasks:
  up:
    cmds:
      - echo "C_VARIABLE {{.C_VARIABLE}}"
```

If I run the following command it behaves as expected. Meaning that when B:up calls C:up, it overrides the variable C_VARIABLE with the content of B_variable=b

```bash
➜ t -t B.Taskfile.yml up
task: [up] echo "B_VARIABLE b"
B_VARIABLE b
task: [C:up] echo "C_VARIABLE b"
C_VARIABLE b
```

if instead I run the same up task from A

```bash
➜ t -t A.Taskfile.yml up
task: [B:up] echo "B_VARIABLE b"
B_VARIABLE b
task: [B:C:up] echo "C_VARIABLE localhost"
C_VARIABLE localhost
```

the C_VARIABLE is not overridden but it uses the default value instead.

This is not what I was expecting

```bash
➜ t -t A.Taskfile.yml up
task: [B:up] echo "B_VARIABLE b"
B_VARIABLE b
task: [B:C:up] echo "C_VARIABLE b"
C_VARIABLE b
```

Am I doing something wrong? Do I have wrong expectations or is this a bug?

- Task version: v3.20.0
- Operating System: MacOS arm64 architecture

acls:
[...]

Issue #778, opened 2022-06-14
Vars of included Taskfiles in included Taskfiles 


maxwassiljew:
- Task version: Task versionv3.12.1
- Operating System: Linux

I have encountered an inconsistency in vars of included taskfiles and need some assistance. The setup I have looks like this. There is a common-include taskfile and a package1 taskfile. The parent taskfile includes both taskfiles. The package1 also includes common-include.
The problem arises with the variables that are passed. From parent to common-include everything works fine but from package1 to common-include the passed variable is simply ignored. 

### Example Taskfile showing the issue

```bash
❯ tree .
.
├── include
│   └── Taskfile.include.yaml
├── package1
│   └── Taskfile.package1.yaml
└── Taskfile.yaml

2 directories, 3 files
```

./Taskfile.yaml
```yaml
version: '3'

includes:
  included:
    taskfile: ./include/Taskfile.include.yaml
    vars:
      VAR_1: provided-from-parent-var1
  package1:
    taskfile: ./package1/Taskfile.package1.yaml

tasks:
  default:
    cmds:
      - task: included:task1
```

./package1/Taskfile.package1.yaml
```yaml
version: '3'

includes:
  included:
    taskfile: ../include/Taskfile.include.yaml
    vars:
      VAR_1: provided-from-package1-var1

tasks:
  default:
    cmds:
      - task: included:task1
```

./include/Taskfile.include.yaml
```yaml
version: "3"

vars:
  VAR_1: '{{.VAR_1 | default "included-default-var1"}}'

tasks:
  task1:
    cmds:
      - echo "VAR_1 is {{.VAR_1}}"
```

```bash
❯ go-task
task: [included:task1] echo "VAR_1 is provided-from-parent-var1"
VAR_1 is provided-from-parent-var1

❯ go-task package1:default
task: [package1:included:task1] echo "VAR_1 is included-default-var1"
VAR_1 is included-default-var1
```

By executing `go-task` the default task from parent is executed and the VAR_1 has the provided value. By executing `go-task package1:default` the default task from package1 is executed and the VAR_1 does not have the provided value `provided-from-package1-var1` but the default value from the common-include `included-default-var1`.

andreynering:
Fixed in #1256

PR #1256, opened 2023-07-07
fix: propagate include vars in multi-level includes

hudclark:
Previously, included vars would be set to nil upon being included by another taskfile. They are now propagated up during the include process.

This is especially problematic for users who want to use library-like taskfiles in a monorepo.
```
├── common
│   └── docker.yml
├── service-a
│   └── Taskfile.yml 
├── service-b
│   └── Taskfile.yml
└── Taskfile.yaml
```

You can imagine that the contents of `service-a/Taskfile.yml` resemble
```yaml
version: '3'
includes:
  docker:
    taskfile: ../common/docker.yml
    vars:
      DOCKER_IMAGE: service-a
```

Fixes: 
- https://github.com/go-task/task/issues/778
- https://github.com/go-task/task/issues/996

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1798,6 +1798,29 @@ VAR_2 is included-default-var2
 	assert.Equal(t, strings.TrimSpace(buff.String()), expectedOutputOrder)
 }
 
+func TestIncludedVarsMultiLevel(t *testing.T) {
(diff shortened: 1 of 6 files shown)


andreynering:
Nice work @hudclark, thanks a lot!

Issue #1250, opened 2023-07-01
Included Taskfiles have wrong working directory with USER_WORKING_DIR


idelchi:
I am having trouble executing tasks from an included file in the current working directory while using Taskfile.yml. 

My goal is to be able to navigate to any sub-folder, run a task command, which would use the Taskfile.yml at the root, which in turn includes utilities and commands from other folders. 

Here's a minimal example:

Taskfile.yml:
```
version: '3'

includes:
  info:
    taskfile: 'subfolder/Taskfile.yml'
    # dir: '{{.USER_WORKING_DIR}}

tasks:
  default: 
    cmds:
      - echo {{.USER_WORKING_DIR}}
      - pwd

```

subfolder/Taskfile.info.yml:

```
version: '3'

tasks:
  default: 
    cmds:
      - echo {{.USER_WORKING_DIR}}
      - pwd
    # dir: '{{.USER_WORKING_DIR}}'


```

When I'm inside the subfolder and run `task info`, the output is (as expected):

```
/home/user/ws/subfolder
/home/user/ws
```

Adding `dir: '{{.USER_WORKING_DIR}}'` to the include statement in Taskfile.yml, has no effect, while I would have expected it to resolve to /home/user/ws/subfolder.

Adding `dir: '{{.USER_WORKING_DIR}}'` to the task in subfolder/Taskfile.info.yml yields:

```
/home/user/ws/subfolder
/home/user/ws/home/user/ws/subfolder   (?)
```

It seems that `USER_WORKING_DIR` resolves to the correct directory, but the commands are not running there. Perhaps I'm using the functionality incorrectly?

A use-case is where I'm using the "--global" flag and have the aforementioned structure in my $HOME folder. 
I can only get tasks in the 'root' Taskfile to execute in the current working context, but not the included ones. 

- Task version: v3.26.0
- Operating system: Linux / Debian Bookworm
- Experiments enabled: No

idelchi:
The combination 

Taskfile.yml
```
includes:
  info:
    taskfile: 'subfolder/Taskfile.yml'
    dir: '{{.USER_WORKING_DIR}}
```

subfolder/Taskfile.yml
```
tasks:
  default:
    cmds:
      - echo {{.USER_WORKING_DIR}}
      - pwd
```

Will still run the `cmds` in the top-level Taskfile.yml folder, and not in the current working directory.

```
user@host:/var$ task -g info
task: [info:default] echo /var
/var
task: [info:default] pwd
/home/user
```

[...]

Issue #1205, opened 2023-06-04
USER_WORKING_DIR is concatenating "C:/" to the path in Windows


stevensgarcia:
Hi,

I'm using a main Taskfile which includes a vagrant_taskfile.yml.

```yml
# Windows
# C:\Users\Usuario\taskfiles\Taskfile.yml
version: '3'

includes:
  vagrant: 
    taskfile: ./vagrant_taskfile.yml
```

I've experienced this concatenating issue with `{{.USER_WORKKING_DIR}}` before in Linux and solved it using this config:

```yml
# Linux
# /home/user/taskfiles/vagrant_taskfile.yml
ssh:
    dir: "../../../{{.USER_WORKING_DIR}}"
    cmds:
      - vagrant ssh
```

This works perfectly OK in Linux. However in **Windows 10 Pro Powershell** it fails:

```ps
# Windows - PowerShell
# Execution context: C:\Users\Usuario\Documents\Vagrant_Projects\demo

task: [main:vagrant:ssh] vagrant ssh
task: Failed to run task "main:vagrant:ssh": could not stat: CreateFile C:\C:\Users\Usuario\Documents\Vagrant_Projects\demo: The filename, directory name, or volume label syntax is incorrect.
```

:warning: **Please note the concatenated `C:\` at the beginning of the execution path. This is the problem.** :warning:

No matter how I play with the hierarchy path `../../`, those parent dirs will be removed, but that `C:\`  will remain attached to the path, so the command wont work. Not to mention that if I only use `{{.USER_WORKING_DIR}}` it will concatenate the main and current context paths together ending with the old problem.

```yml
# Windows
# C:\Users\Usuario\taskfiles\vagrant_taskfile.yml
ssh:
    dir: "../../../{{.USER_WORKING_DIR}}"
    cmds:
      - vagrant ssh
```
```ps
# Windows - PowerShell
# Execution context: C:\Users\Usuario\Documents\Vagrant_Projects\demo

task: [main:vagrant:ssh] vagrant ssh
task: Failed to run task "main:vagrant:ssh": could not stat: CreateFile C:\Users\Usuario\taskfiles\C:\Users\Usuario\Documents\Vagrant_Projects\demo: The filename, directory name, or volume label syntax is incorrect.
```
**Here you can see both paths got concatenated.** As mentioned in the beginning, this can be solved usign `dir: "../../../{{.USER_WORKING_DIR}}"` in Linux.

Any suggestions to fix this in Windows?

[...]

PR #1274, opened 2023-07-21
fix: wrong working directory with USER_WORKING_DIR in included Taskfi…

andarto:
…les (#1250)

When using `dir: '{{.USER_WORKING_DIR}}'` in an included task, the working directory was set incorrectly by concatenating the base Taskfile directory and the USER_WORKING_DIR.

#1205 might be the same problem and is possibly fixed too, but I haven't tested the patch on Windows.

Also, while this patch fixes the symptom, there a better place to implement the fix would be where the USER_WORKING_DIR gets concatenated in the first place, but I wasn't able the find that in the time a had available. At least this patch and the included test case should point you in the right direction.


andreynering:
Hi @andarto,

Thanks for your contribution!

I have done changes to this fix and opened #1309, so I'm closing this one.

PR #1281, opened 2023-07-26
feat: Add option to flatten included taskfiles #273

marverix:
Hello,
this is only PoC of https://github.com/go-task/task/issues/273. My Golang skills are equal to 0. I would appreciate your guidance and/or if you could make good use of this PoC to create a final solution.

Br,
Marek


pd93:
The implementation of this is incomplete and has several major conflicts (some of which may get worse with upcoming changes). Given the recent discussion in #273, I'm going to close this for now to reduce our backlog of PRs. I'm still open to a working implementation of this in the future.

PR #1288, opened 2023-07-31
fix: defer keyword in json schema

calvinmclean:
This fixes a few issues with JSON schema validation when using the `defer` keyword.
It adds a `defer_call` object that is used alongside `cmd_call` and `for_call`. Then, `defer` is removed from the `cmd_call`.

## Example Taskfile:
```yaml
version: "3"

tasks:
  integration-test:
    desc: Startup docker resources and run integration tests. Cleanup when done.
    dir: ./garden-app
    cmds:
      - docker compose -f ../deploy/docker-compose.yml --profile test up -d
      - defer: docker volume rm deploy_influxdb
      - defer: docker compose -f ../deploy/docker-compose.yml --profile test down
      - sleep 15
      - go test -race -covermode=atomic -coverprofile=integration_coverage.out -coverpkg=./... ./integration_tests
```

## Errors:
```
Incorrect type. Expected "boolean".
Missing property "cmd".
```
Screenshot 2023-07-30 at 21 03 31
Screenshot 2023-07-30 at 21 03 39

### DIFF
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -209,6 +209,9 @@
             "$ref": "#/definitions/3/task_call"
           },
           {
+            "$ref": "#/definitions/3/defer_call"
+          },
+          {
             "$ref": "#/definitions/3/for_call"
           }
         ]
@@ -308,10 +311,6 @@
             "description": "Prevent command from aborting the execution of task even after receiving a status code of 1",
             "type": "boolean"
           },
-          "defer": {
-            "description": "",
-            "type": "boolean"
-          },
           "platforms": {
             "description": "Specifies which platforms the command should be run on.",
             "type": "array",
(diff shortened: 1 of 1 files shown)


ydarma:
Actually the correct type seems to be "#/definitions/3/cmd"

andreynering:
Merged with adjustments at c42bc6914ed8223a0de38b7b1f937c73b17825cd. Thank you both!

We should not use `#/definitions/3/cmd` here because `defer` is only a subset of that. I needed to add support for `defer: { task: foobar }`, though.

mjpieters:
The `defer` call object takes a `vars` keyword too, but the schema is lacking this.

mjpieters:
I've opened #1907 to address this, and to update the schema reference documentation.

PR #1290, opened 2023-08-01
docs: add required version for supporting loops

q0rban:
I was banging my head on this for quite some time, reading the docs and trying to get loops working. Eventually I started looking through the issue queue to see if there were bugs reported on this functionality. After reading through the issue queue, I learned that this feature is new, and therefore not available on the version of task that is installed using homebrew. Since using this feature on older versions of Task reports no errors or warnings, I think this little documentation change would be helpful for others.

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -1001,9 +1001,9 @@ This works for all types of variables.
 
 ## Looping over values
 
-Task allows you to loop over certain values and execute a command for each.
-There are a number of ways to do this depending on the type of value you want to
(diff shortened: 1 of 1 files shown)


andreynering:
@q0rban Thanks for contributing!

And indeed, this lack of errors on old versions can be a problem. What happens is that we don't use the "strict mode" of the YAML library we use because that breaks YAML anchors that some users use. I have to look at it again any day if it's fixed. It would make errors being reported for non-existent keys.

Issue #1305, opened 2023-08-17
Task reported as not up-to-date but not executed


visciang:
- Task version: 3.28.0
- Operating system: MacOs
- Experiments enabled: no

Taskfile:

```yaml
version: "3"

tasks:
  test:
    cmds:
      - cp source.txt generate.txt
    sources:
      - source.txt
    generates:
      - generate.txt
```

Test:

```
touch source.txt

task test
# task run

task --status test
# task is up-to-date

echo "change" > source.txt

task --status test
# task is not up-to-date

task test
#  >>>  task does not run and is reported as up-to-date  <<<

```

Do I miss something in the task file definition?

pasman1:
Look like there is a bug in the option `--status`.
Sometimes it prints nothing, sometimes bad status.

Tested on LinuxMint.

visciang:
I had a very very quick look at the test suite, but I didn't find a basic test on the `source / checksum` feature that captures the basic behaviour, very likely it's because my rusty golang memory.

I see that task_test.go->TestStatusChecksum half of the logic:
no checksum change => no task run

I miss the other part:
checksum change => task run

visciang:
I think I reduce the problem, the bug seems related to the `task --status test` command.
It computes **and updates** the checksum (see `.task/checksum/test`).
I think `task --status` should NOT have side effects.

PR #1307, opened 2023-08-17
fix --status side effect, it should not write the computed checksum file

visciang:
This address #1305.

`task --status` writes the computed checksum on file.
The operation should be side effect free and just report if the task is up-to-date or not.

I'm opening the PR in draft to validate if my guess is correct.

I miss a test on this change, currently I don't think I can write an idiomatic test, maybe @pd93 could jump in for a bit of support / review.
 
Thank you


andreynering:
Hi @visciang, thanks for contributing!

I have opened #1313 with a definitive fix.

PR #1315, opened 2023-08-29
chore: update experiments and deprecation docs

pd93:
Based on feedback from a friend who noted that they saw a drop down for the experiments section of the docs and assumed that the items in the submenu would be the experiments themselves. I agree that this is more intuitive and so have placed each experiment in its own document. The workflow is now in the main experiments doc instead.

I have also separated the deprecations documents from the experiments as while they are sometimes linked. This is not always the case and I think it makes sense to have a dedicated space for deprecations.

Note, the templates are now draft documents. These _shouldn't_ display when deploying to prod because of `draft: true`, but they will display when running locally.


andreynering:
Thanks!

PR #1325, opened 2023-09-08
checksum: Use XXH3 algorithm instead of md5

ReillyBrogan:
XXH3 is substantially faster than md5.

Test:
- Used the Linux source code at v6.2.12
- Had a taskfile configured with all source directories as `sources`
- Ran on a AMD Ryzen 3900XT CPU
- NVME storage
- Host system was Linux Fedora 38 using kernel 6.4.13
- Tests were ran in several configurations:
  - Full page cache results were gathered by running the test task several times in a row until the results became consistent, indicating that the entire source tree was in the page cache. Then the test was ran 5 more times and those results were averaged. This is a best-case result, most of the time go-task will be close to this result.
  - Empty page cache results were gathered by running `sync; echo 3 > /proc/sys/vm/drop_caches` after each run in order to empty out the page cache. Each test was ran 5 times and the results averaged. This is a worst case result, such as if the user runs go-task after booting their machine.

Results:
md5:
- Full: 184.65ms
- Empty: 483.34ms

XXH3:
- Full: 90.29ms (-51.1% over md5)
- Empty: 398.70ms (-17.5% over md5)

BLAKE3 (This is another option that has the property of providing cryptographic hashes at a slightly reduced performance over XXH3. Go-task however likely does not benefit from cryptographically secure hashes)
- Full: 106.21ms (-42.4% over md5)
- Empty: 407.78ms (-15.6% over md5)


ReillyBrogan:
I doubt that it matters, but on Linux the binary size increased by 132KB over the md5 binary, or 1.6% (the BLAKE3 binary increased by 103KB instead, or 1.3%).

Note also that this change will invalidate all existing checksum files, and so deserves a release note accordingly.

ReillyBrogan:
Using a 128KB buffer improved the benchmarks a bit more:

XXH3:
  Full: 80.29 ms (-56.5% over md5)
  Empty: 379.35ms (-21.5% over md5)

andreynering:
Hi @ReillyBrogan,

Great idea, thanks for your contribution!

I'll certainly mention this in the change log. This is a small breaking change, and I don't think it's a problem to release as a minor version.

PR #1327, opened 2023-09-12
Command pre/post/task_separator hooks

### LINKED ISSUE #1013: Interleaved output with pre/post tasks
Hi! I am an owner/author of the tool - https://github.com/wttech/aemc and I am considering using your tool on top of mine.

I am trying to format the output of my tasks to be looking similar as before using your awesome Task tool.

basically, the effect I have now is:

image

I changed the green color to that purple one using env var.
But I also want to put the new line before running each task.

To see sth like this (now is doubled for some reason, single "\n" is more preferable):

image

I achieved that by:

```yaml
output:
  group:
    end: "\n"
```

However, now e.g Docker is no longer outputting colored text, and whole outputs are buffered and printed in the end of running each command. It's essential for me to not wait so long for outputs as I have very long running tasks and just the reason to switch interleaved output mode to grouped only because a need to print tasks outputs in groups separated by that new line is a bit cumbersome.

 I am aware that I could add `echo ""` everywhere but that will not be so nice.  But what if in Task we will have the option to define: 

a) in my case... sth like this:

```
pre_task:
  cmds:
    - echo ""
```

b) or sb else could want even to have:

```
pre_task:
  cmds:
    - echo "started task '{{ .TASK }}'"

post_task:
  cmds:
    - echo "ended task '{{ .TASK }}' in {{ .ELAPSED_TIME }}"
```

also is it possible somehow to hide cmd displayed after task? I mean that one highlighted:

image

I have a quite long/multiline command and I am not sure if displaying it is beneficial

FYI for a many years before I was using Gradle for defining/running tasks. Now I was looking for light alternative and seems that Task might be a good option. Let's see ;)

krystian-panek-vmltech:
fixes #1013 

before change:

image

adding such snippet:

```yml
cmds:
  pre: echo ""
```

will make it look like this: (output nicely grouped even when output style is interleaved)

image

using an output style 'group' with end "\n" does not make sense to me because then I need to wait for the long-running task (even 10 minutes) to see the grouped output. this solves my problem as I have still dynamically printed output  and it is grouped nicely by this pre cmd hook printing empty line exactly as I desired.

### DIFF
--- a/setup.go
+++ b/setup.go
@@ -238,6 +238,7 @@ func (e *Executor) setupDefaults() {
 func (e *Executor) setupConcurrencyState() {
 	e.executionHashes = make(map[string]context.Context)
 
+	e.taskSeparated = new(int32)
 	e.taskCallCount = make(map[string]*int32, e.Taskfile.Tasks.Len())
 	e.mkdirMutexMap = make(map[string]*sync.Mutex, e.Taskfile.Tasks.Len())
 	for _, k := range e.Taskfile.Tasks.Keys() {
--- a/task.go
+++ b/task.go
@@ -138,6 +138,8 @@ func (e *Executor) Run(ctx context.Context, calls ...taskfile.Call) error {
 	return g.Wait()
 }
 
+var taskOutputSeparated = false
+
 // RunTask runs a task by its name
 func (e *Executor) RunTask(ctx context.Context, call taskfile.Call) error {
 	t, err := e.CompiledTask(call)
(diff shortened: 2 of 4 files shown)


krystian-panek-wttech:
this extension does not work perfectly as I wanted. Ideally, the pre/post hooks are executed in the beginning and end of task execution but... after a deep dive into Task implementation, I see that tasks are executed by recursion which makes it really hard to print these new lines correctly... for example the hooks could be called in a following order: pre pre, pre end pre end end end. As an effect then I will see doubled new lines in between task command outputs... :/

krystian-panek-wttech:
OK I achieved the desired effect, please validate :) I even tried providing a long vertical line as a task separator to indicate the purpose of a new option; now tasks outputs could be easily grouped in console by leveraging that task separator

image

andreynering:
Hi @krystian-panek-wttech,

I'm not convinced to add this to the code base. This seems to be more complex than worth it to just have a better log formatting.

If you think we could improve the formatting, like having like breaks, we could consider doing that directly on Task without the need of hooks.

krystian-panek-wttech:
I am ok with doing that without hooks. My main case is to add separators before and after executing tasks with an interleaved output style. If you have an idea how to implement it differently I would be interested. So basically there could be no hooks but only these empty lines before and after the task execution could be written.

PR #1338, opened 2023-09-15
Do not print precondition error if context was aborted

### LINKED ISSUE #1337: False positive precondition fails if context is aborted
I have a bug similar to #597 where you can see precondition false positives error print that can be reproduced with the following tasks:
```
version: '3'

tasks:
  default:
    deps:
      - fail-0s
      - pass-1s
    cmds:
      - "true"
  fail-0s:
    run: always
    preconditions:
      - sh: 'false'
        msg: "precondition should fail"

  pass-1s:
    run: always
    preconditions:
      - sh: 'sleep 1; true'
        msg: "precondition should never fail"
```

output:
```
$ task
task: precondition should fail
task: precondition should never fail
task: precondition not met
```

Context abort due to a failed context causes other conditions to print the error message as well.
I already found the reason, so a PR will follow soon :+1:

- Task version: main (72d77eb)
- Operating system: Arch Linux
- Experiments enabled: no

sylv-io:
Fixes #1337

### DIFF
--- a/precondition.go
+++ b/precondition.go
@@ -21,7 +21,9 @@ func (e *Executor) areTaskPreconditionsMet(ctx context.Context, t *taskfile.Task
 			Env:     env.Get(t),
 		})
 		if err != nil {
-			e.Logger.Errf(logger.Magenta, "task: %s\n", p.Msg)
+			if !errors.Is(err, context.Canceled) {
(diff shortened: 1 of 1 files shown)


andreynering:
Thanks @sylv-io!

PR #283, opened 2020-01-20
Compile tasks before printing help or summary

paulvarache:
Updates the `help.go` file to compile a task before printing its description. This enables templatization for the `-l` flag.

Also updates `summary/summary.go` and `task.go` to compile tasks before printing a summary.

Relates to https://github.com/go-task/task/issues/276

Issue #276, opened 2019-12-12
desc doesnt seem to support templatization


mxrss:
If relevant, include the following information:
- Task version: Task version: 2.8.0
- OS: darwin
- Example Taskfile showing the issue

```

version: '2'

vars:
  Repo: blah
  Tag: confd
  Version: v0.16.0
  FullName: "{{ .Repo }}/{{ .Tag }}:{{ .Version }}"

tasks:
  container:build:
    desc: |
     This will build "{{ OS }}"
    cmds:
      - docker build . -t "{{ .FullName }}"
    silent: true

```

Variable interpolation works well, but it seems that desc does not support auto generating content. I am looking for this to try to make my tasks a little more generic.. 

{{ OS }} or other golang templates should be executed for desc (Description).

andreynering:
Hi @mxrss,

Indeed this is a bug. Thanks for reporting!

PR #1343, opened 2023-09-21
fix: templates on task descriptions

blackjid:
# Issue
When a task description has uses templates, they are not properly compiled when listing the tasks.

This was reported back in https://github.com/go-task/task/issues/276 and then fixed in https://github.com/go-task/task/pull/283

I think then there was a regression in f22389a8

I don't have much experience in go and in this project context, I've tried my best to find the real issue.

### DIFF
--- a/task.go
+++ b/task.go
@@ -458,13 +458,14 @@ func (e *Executor) GetTaskList(filters ...FilterFunc) ([]*taskfile.Task, error) 
 
 	// Compile the list of tasks
 	for i := range tasks {
-		task := tasks[i]
+		idx := i
+		task := tasks[idx]
 		g.Go(func() error {
 			compiledTask, err := e.FastCompiledTask(taskfile.Call{Task: task.Task})
 			if err == nil {
 				task = compiledTask
 			}
-			task = compiledTask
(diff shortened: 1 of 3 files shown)


task.go
```diff
@@ -458,13 +458,14 @@ func (e *Executor) GetTaskList(filters ...FilterFunc) ([]*taskfile.Task, error)
 
 	// Compile the list of tasks
 	for i := range tasks {
-		task := tasks[i]
+		idx := i
+		task := tasks[idx]
 		g.Go(func() error {
 			compiledTask, err := e.FastCompiledTask(taskfile.Call{Task: task.Task})
```
blackjid, line 464: I also think that this `err` could be just `_` and remove the if that follows..

pd93: Agreed, this is not needed.

pd93 (approved):
Looks good to me. Thanks for catching this and finding the cause.

[Something something `GOEXPERIMENT=loopvar`](https://go.dev/blog/loopvar-preview) 😅

andreynering:
Thanks @blackjid!

PR #1344, opened 2023-09-21
feat: unify prompts

pd93:
This allows the `--yes` flag to work for all prompts (including the ones in the remote Taskfiles experiment) and reduces some duplicated prompt code.


wburningham:
@pd93 would you be willing to merge this and tag a new `v3.30.2` version? (I'm at the point where I'd like to test the CI integration within my company)

PR #1345, opened 2023-09-22
feat: prefer remote taskfiles over cached ones

pd93:
As per discussion in https://github.com/go-task/task/pull/1152#issuecomment-1705085513, this PR changes the Remote Taskfiles experiment to prefer remote files over locally cached ones.

Previously, if a user used the `--download` flag to cache a file, this file would be automatically preferred over the remote copy. The only way to get a newer version of that file was to delete the cached version entirely.

The new behaviour always prefers the remote copy and will always attempt to fetch and use it _unless_ the user specifies the `--offline` flag (in which case Task will search for a cached version).

In addition to this, if the network times out, when trying to fetch a remote copy, Task will now search for a cached version and use that instead.

---

My question now is... Do we need `--download` at all? I would argue that we can now make `--download` the default behaviour. i.e. we always cache the file when we download it. If we don't do this, then there is a risk that the cached copy will slowly diverge (if a user doesn't remember to use `--download` periodically when the remote copy changes). I think it's safer to assume that the user always wants the latest version of the file cached.

I'm also considering a `--timeout` flag. I have set the default to 10 seconds (which seems reasonable to me), but I can see CI/script users wanting to adjust this for various reasons.

### DIFF
--- a/docs/docs/experiments/remote_taskfiles.md
+++ b/docs/docs/experiments/remote_taskfiles.md
@@ -74,16 +74,17 @@ you are doing.
 
 ## Caching & Running Offline
 
-If for whatever reason, you don't have access to the internet, but you still
-need to be able to run your tasks, you are able to use the `--download` flag to
-store a cached copy of the remote Taskfile.
+Whenever you run a remote Taskfile, the latest copy will be downloaded from the
+internet and cached locally. If for whatever reason, you lose access to the
+internet, you will still be able to run your tasks by specifying the `--offline`
+flag. This will tell Task to use the latest cached version of the file instead
+of trying to download it. You are able to use the `--download` flag to update
+the cached version of the remote files without running any tasks.
 
-<!-- TODO: The following behavior may change -->
-
-If Task detects that you have a local copy of the remote Taskfile, it will use
-your local copy instead of downloading the remote file. You can force Task to
-work offline by using the `--offline` flag. This will prevent Task from making
-any calls to remote sources.
+By default, Task will timeout requests to download remote files after 10 seconds
+and look for a cached copy instead. This timeout can be configured by setting
+the `--timeout` flag and specifying a duration. For example, `--timeout 5s` will
+set the timeout to 5 seconds.
 
 <!-- prettier-ignore-start -->
 [remote-taskfiles-experiment]: https://github.com/go-task/task/issues/1317
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,16 @@
 
 ## Unreleased
 
+- The
+  [Remote Taskfiles experiment](https://taskfile.dev/experiments/remote-taskfiles)
+  now prefers remote files over cached ones by default (#1317, #1345 by @pd93).
+- Added `--timeout` flag to the
+  [Remote Taskfiles experiment](https://taskfile.dev/experiments/remote-taskfiles)
+  (#1317, #1345 by @pd93).
 - Fix bug where dynamic `vars:` and `env:` were being executed when they should
   actually be skipped by `platforms:` (#1273, #1377 by @andreynering).
-- Fix `schema.json` to make `silent` valid in `cmds` that use `for` (#1385, #1386 by @iainvm).
+- Fix `schema.json` to make `silent` valid in `cmds` that use `for` (#1385,
+  #1386 by @iainvm).
 - Add new `--no-status` flag to skip expensive status checks when running
   `task --list --json` (#1348, #1368 by @amancevice).
 
@@ -12,7 +19,7 @@
 
 - Enabled the `--yes` flag for the
   [Remote Taskfiles experiment](https://taskfile.dev/experiments/remote-taskfiles)
-  (#1344 by @pd93).
+  (#1317, #1344 by @pd93).
 - Add ability to set `watch: true` in a task to automatically run it in watch
   mode (#231, #1361 by @andreynering).
 - Fixed a bug on the watch mode where paths that contained `.git` (like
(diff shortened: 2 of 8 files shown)


caphrim007:
The way I try to frame stuff like this cli-args thing is to consider how both myself and the engineers that I work with would either pass or fail the principle-of-least-astonishment (POLA).

In this example, if I were to look at the following syntax in a taskfile,

```
include:
  my-remote-namespace: https://raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
```

my gut-reaction before any thinking occurs is that the system is going to download that Taskfile.yml. In other words, my personal POLA is that URLs are not much different than filesystems. I point to a thing and it just gets it.

Caching, given the above, is an implementation detail to me as the user; it's invisible. If it works, yay. If it doesn't work, something seems a little slow, but the system still got my file...I wonder why slow...who cares, yay.

The above is what I also seem to have been able to grok from my colleagues.

So in this regard, my 2c, is that the `--download` arg is redundant, and the `--offline` arg is exposing an internal behavior of the system that might happen under normal operating conditions. But, to your points, I might want to deliberately specify it if I'm using Task in situations where robots are involved and I have full control over my environment.

Happy to hear others opinions.

pd93:
Thanks @caphrim007. Really appreciate your thoughts :) I've had a bit more time to think about this.

I think the main reasons for the `--offline` flag are:

1. If a user wants to view/edit a remote Taskfile before execution or maybe for debugging
2. If a user knows that they don't have an internet connection and they don't want to wait for the timeout.

To extend on point 1. If we kept the `--download` flag, but made it so that it _never_ executes a command, then it would further facilitate the ability to "view/edit a remote Taskfile before execution".

So maybe all we need to do is:

1. Change the default behaviour to always download/cache files (as previously stated)
2. Change the `--download` flag so that it never executes tasks

blackjid:
I think I would want this to behave a bit like dependencies...

- **The first run**, download and cache.. (possible log to stdout that downloading is happening)
  - You can always run a command to upgrade the dependencies. e.g `task includes upgrade`
    - There might be an option to set an auto upgrade every x hours, minutes.
- **Next runs**, just use the cache.. until
  - The reference change,, for example the urs change to point to a different branch/tag
  - The autoupgrade ttl time reach and a new download happens

I think this is mostly the same as the "always download and cache" with a configurable timeout. But presented in a different way. At least I'm more used to think in terms of dependencies management that in terms of cache/timeout.

c-ameron:
I like @blackjid's idea! For my context, I am wanting to use these features to have a standard set of taskfile includes across my org.

For me, I would like it to download first, then by default always use the cache. I run taskfile a lot, (multiple times a minute if I'm running a debugging command like `task test -- feature/a` ), so the extra network overhead wouldn't be useful.

I like the idea of having the user force a new download for the cached files. The auto-update-ttl idea is also great. It would allow users to have the fast latency of not fetching every task run, but also allow automatic updates in a soft manner.

Another suggestion, would be to have these as options as keys inside the `.yml` file as well. To me it would be clunky to have to have these all as a flag when running my tasks.
As an example

```yaml
includes:
  my-remote-namespace: https://raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
  offline: true
  auto-update-ttl: weekly
```
Thanks!

pd93:
Hey all. Thanks for the comments and sorry for the lack of progress on the experiment lately. It's been a busy month or so for me!

I've pushed the changes discussed in previous comments and I believe this is ready for a review when @andreynering has some time.

@c-ameron I like the idea of a cache TTL, but I'm going to leave this as out-of-scope for now. I don't think this addition will affect the API and could be easily added later. The same goes for adding the flags as keys in the file. The schema is a bit harder to amend for an experiment if we were to change our minds on anything, so I'd like to concentrate on the fundamentals for now (so that we can deliver this experiment quicker) and then revisit these bits later. That said, please feel free to open issues for these features so that they aren't forgotten.

docs/docs/experiments/remote_taskfiles.md
```diff
@@ -74,16 +74,17 @@ you are doing.
...
-
-<!-- TODO: The following behavior may change -->
-
-If Task detects that you have a local copy of the remote Taskfile, it will use
-your local copy instead of downloading the remote file. You can force Task to
-work offline by using the `--offline` flag. This will prevent Task from making
-any calls to remote sources.
+Whenever you run a remote Taskfile, the latest copy will be downloaded from the
+internet and cached locally. If for whatever reason, you lose access to the
+internet, you will still be able to run your tasks by specifying the `--offline`
```
andreynering, line 79: As we talked on Discord, it'd be interesting to have an `offline: true` setting and a `TASK_OFFLINE=1` env to allow users to set this once and have it always enabled.

Can be on another PR if you prefer, no problem.

docs/docs/experiments/remote_taskfiles.md
```diff
@@ -74,16 +74,17 @@ you are doing.
...
+flag. This will tell Task to use the latest cached version of the file instead
+of trying to download it. You are able to use the `--download` flag to update
+the cached version of the remote files without running any tasks.
```
andreynering, line 82: > You are able to use the `--download` flag to update the cached version of the remote files without running any tasks.

Yes, that's the idea :+1:

andreynering (approved):
:clap: :clap: :clap:

docs/docs/experiments/remote_taskfiles.md
pd93: I think the functionality here is ready, so let's get this merged. I'll work on the schema/env options in another PR as suggested.

@c-ameron Thanks for creating the issues. I've added them to the TODO list in the experiment issue so they're not forgotten.

PR #1356, opened 2023-10-01
fix: exclude other "ignored" files.

butuzov:
Limits ignore files to exact 4 directories used to be ignored before ".git/.hg/node_modules/.task"

Usecase: setting up monitoring for the files that matches ignored directory, but not location inside of it. So for example, we I want (actual case I run into) to monitor "yaml" templates and test them on change, I can't use `.github` directory as it ignored by default because it matches `.git`.

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -176,6 +176,16 @@ func (e *Executor) registerWatchedFiles(w *watcher.Watcher, calls ...taskfile.Ca
 }
 
 func shouldIgnoreFile(path string) bool {
-	return strings.Contains(path, "/.git/") || strings.Contains(path, "/.hg/") ||
-		strings.Contains(path, "/.task/") || strings.Contains(path, "/node_modules/")
+	ignorePaths := []string{
+		"/.task",
+		"/.git",
+		"/.hg",
+		"/.node_modules",
+	}
+	for _, p := range ignorePaths {
+		if strings.Contains(path, fmt.Sprintf("%s/", p)) || strings.HasSuffix(path, p) {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


butuzov:
Looks like test didn't run (so no main tests are running for PR tests) ?

pd93:
> Looks like test didn't run (so no main tests are running for PR tests) ?

The watch tests are disabled in CI because they can be flaky. These are excluded using the build tags at the top of the `watch_test.go` file. See related commit: https://github.com/go-task/task/commit/d6d51a2f8becb72f64cbba7e778c29ba85cefb38.

You can run `task test:all` to include these tests when running locally. On that note, these tests are currently failing because your new method is not exported and the test file is in a test package (`task_test`). I don't think we really want to export the method though, so perhaps we should change the package for `watch_test.go` to `task`.

watch.go
```diff
@@ -176,6 +176,6 @@ func (e *Executor) registerWatchedFiles(w *watcher.Watcher, calls ...taskfile.Ca
 }
 
 func shouldIgnoreFile(path string) bool {
-	return strings.Contains(path, "/.git") || strings.Contains(path, "/.hg") ||
-		strings.Contains(path, "/.task") || strings.Contains(path, "/node_modules")
+	return strings.Contains(path, "/.git/") || strings.Contains(path, "/.hg/") ||
+		strings.Contains(path, "/.task/") || strings.Contains(path, "/node_modules/")
```
pd93, line 180: This change will break in situations where the path ends in exactly `.git` (with no trailing slash). Can I suggest something like this: 

```suggestion
	ignorePaths := []string{
		"/.task",
		"/.git",
		"/.hg",
		"/.node_modules",
	}
	for _, p := range ignorePaths {
		if strings.Contains(path, fmt.Sprintf("%s/", p)) || strings.HasSuffix(path, p) {
			return true
		}
	}
	return false
```

Making this into a loop is also a bit cleaner than the current or statement IMO.

butuzov: Yeah, its actually solves everything!

andreynering:
Thanks @butuzov!

PR #1383, opened 2023-10-29
fix: ensure requires works with dotenv at the task level

### LINKED ISSUE #1362: requires.env seem to run before dotenv files are loaded
- Task version: 3.30.1
- Operating system: MacOS
- Experiments enabled: No

First of all, thanks for making such a neat tool, I found it very easy migrating to go-task from my archaic bash scripts. I did, however, run it one small issue with `requires.env` and `dotenv` task properties. When I load a variable from one of the files defined in `dotenv` array and try to verify that it's present by specifying it in `requires.env` it fails regardless if the variable is present in env file or not. This only happens when both properties are defined on a task level, when `dotenv` is defined at the root the check works as expected.

```yaml
# Taskfile.yml
version: '3'

tasks:
  fail:
    dotenv: ['.env']
    requires:
      vars: [FOO]
    cmds:
      - echo "Works"
```

```sh
# .env
FOO="BAR"
```

yordis:
closes #1362


internal/compiler/v3/compiler_v3.go
```diff
@@ -120,6 +120,9 @@ func (c *CompilerV3) getVariables(t *taskfile.Task, call *taskfile.Call, evaluat
 	if err := t.Vars.Range(taskRangeFunc); err != nil {
 		return nil, err
 	}
+	if err := t.Env.Range(taskRangeFunc); err != nil {
+		return nil, err
+	}
```
yordis, line 125: I am not sure if the order is correct, or if I should call this before `t.Vars` and `call.Vars`

PR #1384, opened 2023-10-30
fix: load dotenv before vars

yordis:


yordis:
It may require https://github.com/go-task/task/pull/1383 to be merge first!

yordis:
Looking at the code, I am not sure if that is intentional or not... I see that https://github.com/go-task/task/blob/8cfac5a25a7a1b9cd0ac1a64e409d91319eea27c/internal/compiler/v3/compiler_v3.go#L148 is not passing any env ... so I am not what would be the intent here

PR #1421, opened 2023-12-03
Support any variables

pd93:
This is the first iteration of the "Any Variables" experiment that allows users to use any type in their Taskfile variables. In this PR:

- Adds the `TASK_X_ANY_VARIABLES` experiment flag
- Renames `Var.Static` -> `Var.Value`
- Changes `Var.Value` to an `any` type
- Unmarshal into `any` type instead of just strings when the experiment flag is enabled.
- Updated schema to support objects and arrays in vars
- Added some experiment docs
- `for` supports arrays when looping over a variable
- Added an error message when trying to use `sh` with the experiment enabled
- Added new syntax for dynamic variables

There are no tests in this PR as its an experiment. However, I have added a Taskfile in `testdata/vars/any` which can be used to try out the functionality:

`go run ./cmd/task --dir testdata/vars/any`

---

The main things I'm looking for feedback on are:

- Bugs/errors (obviously)
- Missing functionality (Is there anything else we could do now that we support `any` variables)
- Dynamic variables
  - Dynamic variables previously used a subkey (`sh`) to define the command. This is incompatible with map variables as they would be defined the same way. This means that dynamic variables are now defined by starting a string variable with the `$` character instead.
  - Do we like/dislike this?
  - An alternative would be to not allow map variables in this experiment and keep the existing `sh` syntax. This would make this change non-breaking and it could be merged into v3 rather than waiting for a major version bump. However, you won't be able to define variables as maps at all.
  - Feedback on how useful map variables are would be help weigh up the pros/cons of each option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Added
+  [Any Variables experiment](https://taskfile.dev/experiments/any_variables)
+  (#1415, #1421 by @pd93).
 - Added `aliases` to `--json` flag output (#1430, #1431 by @pd93)
 
 ## v3.32.0 - 2023-11-29
--- a/args/args.go
+++ b/args/args.go
@@ -18,7 +18,7 @@ func ParseV3(args ...string) ([]taskfile.Call, *taskfile.Vars) {
 		}
 
 		name, value := splitVar(arg)
-		globals.Set(name, taskfile.Var{Static: value})
+		globals.Set(name, taskfile.Var{Value: value})
 	}
 
 	return calls, globals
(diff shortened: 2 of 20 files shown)


pd93:
Done some more bug bashing today 🐞 This is starting to feel somewhat stable from my end.

Had a bizarre GH actions issue with the linter, but updating to the latest `golangci-lint` release seems to have fixed it 🤷

@andreynering Would love to get this experiment merged as a draft for the next release if possible. Any immediate thoughts? I think the main decision for this experiment is going to be whether we should support maps (and remove `sh`) or disallow them for now (which might allow us to release this into v3).

andreynering:
@pd93 Great work so far!

I think the easier decision for now is to remove support for maps so we can release the experiment for now. We can then decide later if we want to add that support back. I'd say that arrays will probably be a lot more used and useful than maps, so supporting that is the priority, anyway.

And yes, I can postpone the next release to tomorrow or Friday if needed, so you have time to address that and it'll be included the release. :wink:

andreynering:
Hmm... once actually trying the Taskfile on `testdata/vars/any`, I get marshal errors on arrays specifically. Other types (ints, bools and maps) work if I comment these lines:

```bash
task: Failed to parse Taskfile.yml:
yaml: line 51: cannot unmarshal !!seq into variable
```

Looks like something easy to fix.

andreynering (approved):

PR #1436, opened 2023-12-21
feat: support looping over map variables

pd93:
Adds support for looping over a map-type variable using the `for` syntax. Since Go maps are unordered, the order of the loop will be random.

This also fixes a bug where specifying a map variable when looping over a variable would give an invalid error about using `sh` with the Any Variables experiment.

### DIFF
--- a/docs/docs/experiments/any_variables.md
+++ b/docs/docs/experiments/any_variables.md
@@ -63,8 +63,7 @@ tasks:
 ```
 
 There are many more templating functions which can be used with the new types of
-variables. For a full list, see the
-[slim-sprig][slim-sprig] documentation.
+variables. For a full list, see the [slim-sprig][slim-sprig] documentation.
 
 ## Looping over variables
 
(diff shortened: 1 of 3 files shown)


vmaerten:
Is it possible to also access the key in the loop ? For exemple with `{{.KEY}}`  or `{{.ITEM.KEY}}` / `{{.ITEM.VALUE}}` ?

andreynering (approved):
👏 👏 👏 

> Since Go maps are unordered, the order of the loop will be random.

Perhaps having a custom `UnmarshalYAML` function to make the keys ordered would be interesting for the users. Many would expect keys order to persist. We already do that for the variables map itself.

Can be done in another PR, though.

pd93:
> Is it possible to also access the key in the loop ? For exemple with `{{.KEY}}` or `{{.ITEM.KEY}}` / `{{.ITEM.VALUE}}` ?

@vmaerten I think this is a good idea. I've added the value to a `{{.KEY}}` variable for now and we'll see what people think.

> > Since Go maps are unordered, the order of the loop will be random.
> 
> Perhaps having a custom `UnmarshalYAML` function to make the keys ordered would be interesting for the users. Many would expect keys order to persist. We already do that for the variables map itself.
> 
> Can be done in another PR, though.

@andreynering This might be a bit of a challenge (though I'm sure it doable). The problem with custom unmarshaling is that we'd have to store the map in an OrderedMap struct. This is going to create issues when trying to access the map via templating as regular templating function and dot syntax is not going to work. We'd need to find a way to store the raw map so that its available for templating, but also store the order so that it can be iterated over correctly when looping. Might require a bit of thought into the best implementation of this.

PR #1438, opened 2023-12-21
feat: added basic merge template function

pd93:
Adds a very basic `merge` template function that will iterate over maps A and B and put all their elements into a new map M. Map B will overwrite map A if a key conflicts. The new map M is returned leaving the original maps unchanged.

Not sure where functions like this should be documented as we seem to have 3 sources for template functions:

- Go templating itself
- Slim Sprig
- Task

### DIFF
--- a/internal/templater/funcs.go
+++ b/internal/templater/funcs.go
@@ -6,10 +6,9 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/davecgh/go-spew/spew"
 	"mvdan.cc/sh/v3/shell"
 	"mvdan.cc/sh/v3/syntax"
-
-	"github.com/davecgh/go-spew/spew"
 
 	sprig "github.com/go-task/slim-sprig/v3"
 )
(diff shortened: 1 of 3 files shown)


internal/templater/funcs.go
```diff
@@ -52,6 +52,16 @@ func init() {
 		"relPath": func(basePath, targetPath string) (string, error) {
 			return filepath.Rel(basePath, targetPath)
 		},
+		"merge": func(a, b map[string]any) map[string]any {
+			m := make(map[string]any)
```
andreynering, line 56: Idea: set the initial capacity to the length of the existing maps:

```suggestion
			m := make(map[string]any, len(a)+len(b))
```

andreynering (approved):
Functions added in Task can be documented in this section: https://taskfile.dev/usage/#gos-template-engine

Perhaps we could move them to the API page as a table and just link on that section...

I noticed `joinPath` and `relPath` were not documented. I'll add them to the existing section for now. (EDIT: done on bb16c3efcad783c4671ccc8b1a60660140cb4fcf).

internal/templater/funcs.go
pd93: I did think about this but didn't add it since if there are duplicate keys in both maps, we will be creating up to 2x more capacity than needed. However, I guess this isn't a huge problem as we're probably optimising for allocations rather than memory efficiency.

I've added this change and also included the new function in the docs.

andreynering: I realized now that we could have made this function variadic, accepting any number of maps instead of just 2. Good that it can still be made without being backwards incompatible.

pd93: That's a great idea. See #1464

PR #1445, opened 2023-12-27
docs: improved experimental docs formatting

pd93:
- Removed an old section from the styleguide that stated that `taskfile.yml` files (with a lowercase 't') are not supported. This is no longer true since #1221 was merged.
- Updated the format of the experiments & deprecations docs to make the warnings much clearer. The production warning from the main experiments page has now been copied into each individual experiment page too so that users see it if they navigate directly to an experiment.

Before

image

After

image


andreynering (approved):
:+1:

PR #1448, opened 2023-12-28
Docker compose support

mgale:
>
>

Define the container setup in task yaml file like:
```
container:
  # Definition needs to be a valid docker-compose definition
  context: default
  definition:
    services:
      container:
        image: ubuntu:latest
        volumes:
          - ./containers/local/config:/local/config:delegated
        environment:
          - FOO:BAR
        command: sleep infinity
        init: true
```

Either force the task to run in a container by setting the param `force_container` to true on the task or using the cli arg `--force-container`

The thought process here is when running locally a container may be optional but might be required when used in a CICD pipeline.


andreynering:
As I also said on https://github.com/go-task/task/pull/974#issuecomment-2029710295, Iḿ closing this on favor of https://github.com/go-task/task/issues/448.

PR #1447, opened 2023-12-28
feat: remove v2 support

pd93:
Removes support for `version: 2` Taskfile schemas as outlined in the [deprecation notice](https://github.com/go-task/task/issues/1197).

The docs have also been updated to reflect this removal.

PR #1450, opened 2023-12-29
refactor: taskfile/ast package

pd93:
I've always found it a little strange that the code that reads a Taskfile is in a `read` package. I'm used to domain-driven packages, so it feels more correct to me to have something like `taskfile.Read(...)` than `read.Taskfile(...)`. This PR is a proposal to rearrange the API a little bit to something I think is a bit nicer.

Instead of the schema being `taskfile.Taskfile`, we have a `taskfile/ast` (Abstract Syntax Tree) package which hosts all the code for parsing and storing the Taskfile syntax. The `taskfile` package is then responsible for interfacing with Taskfiles (`Read(...)`, `Exists(...)`, `Merge(...)` etc).

Note: I based this branch on `remove-v2-support` (#1447), so that needs to be merged first.


andreynering (approved):
Nice refactor! :clap: 

`ast` could also have been called `schema`, but both options are more than fine!

pd93:
> `ast` could also have been called `schema`, but both options are more than fine!

I did consider `schema`, but I felt that `ast` was more accurate as the structures don't always directly correspond to the schema. Rather, they are an abstraction of it. It's also less characters for my lazy fingers to type 🤣 

I don't feel strongly about this though if we want to change it before the next release!

Issue #1065, opened 2023-03-17
Idea: Consider having functions to access variables and envs


andreynering:
This is an idea for the next major version, and would be a breaking change.

Nowadays both variables and envs are accessed as Go template values, like this:

```yaml
{{.MY_VARIABLE}}
{{.MY_ENV}}
```

This have a couple of drawbacks:

1) Variables need to be declared in an specific order if you want to include one into another:

```yaml
FOO: foo
FOOBAR: '{{.FOO}}bar'
FOOBARBAZ: '{{.FOOBAR}}baz'
```

2) Variables and envs are mixed, and this often generate a bit of confusion to users.

3) Shell variables are always computed, even when they are not being used.
    - We also currently cache them, which can be annoying if the user want to calculate it again on every call. On the current implementation removing the cache would not be a good idea, though, because it would calculate these multiple times unnecessarily.

This proposal to have two template functions called `var` and `env` to access these values:

```yaml
MY_VARIABLE: '{{var "MY_VARIABLE"}}'
MY_ENV: '{{env "MY_ENV"}}'
```

As part of these proposal, environment variables would stop being added to variables, so you need to really call `env` if you want an env.

andreynering:
To make this clear: this is an idea, but needs experimentation to see if it would really work as expected.

Also, opinion from users are welcome, specially considering that this is a breaking change and would add work to users that will want to upgrade to the next version.

ghostsquad:
I think this is a great idea. Changing variable access from being a field to a function call enables lazy variables, which is a big pain point right now.

ghostsquad:
> @andreynering what are the problems with:
> 
> > Variables need to be declared in an specific order if you want to include one into another
> 
> This keeps things simple, readable, and maintainable for the user.

I too am interested in this answer. I mentioned in a different issue that task uses a "map" for storing variables, where it really should be using a list/array. Maps inherently don't have an order. Golang even explicitly randomizes map access in loops to discourage any reliance on order. Yet in Task, order of declared keys in variables matters.

andreynering:
@ghostsquad It's not true that Task store vars in a map. It's stored in a ordered map implementation. That's why we're able to make it work correctly.

ghostsquad:
> @ghostsquad It's not true that Task store vars in a map. It's stored in an ordered map implementation. That's why we're able to make it work correctly.

I'm referring to how it's stored in yaml.

shakfu:
[...]

PR #1451, opened 2023-12-29
feat: add ability to disable dynamic variable caching

pd93:
See [comment on discord](https://discord.com/channels/974121106208354339/974121514184085555/1190097443732328588) copied below:

> I came across something strange. 
> ```yml
> test:
>   cmds:
>     - echo {{ .FILE }} == {{ .TEMPFILE }}
>   vars:
>     FILE:
>       sh: mktemp
>     TEMPFILE:
>       sh: mktemp
> ```
> ```
> > task test
> task: [test] echo /var/folders/6s/2k2vd19s7kg20405_blbh7fm0000gn/T/tmp.705g66B7Ku == /var/folders/6s/2k2vd19s7kg20405_blbh7fm0000gn/T/tmp.705g66B7Ku
> /var/folders/6s/2k2vd19s7kg20405_blbh7fm0000gn/T/tmp.705g66B7Ku == /var/folders/6s/2k2vd19s7kg20405_blbh7fm0000gn/T/tmp.705g66B7Ku
> ```
>
> Why FILE and TEMPFILE has the same value?

This is happening because of Task's dynamic variable cache. We should add the ability to disable this on a per-variable basis:

```yml
  without-cache:
    silent: true
    vars:
      FOO:
        sh: mktemp
        cache: false
      BAR:
        sh: mktemp
        cache: false
    cmds:
      - echo {{ .FOO }}
      - echo {{ .BAR }}
```
```
/tmp/tmp.vGKqfv8fm7
/tmp/tmp.4Jtfmihw58
```

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -646,9 +646,9 @@ tasks:
 compare the checksum of the source files to determine if it's necessary to run
 the task. If not, it will just print a message like `Task "js" is up to date`.
 
-`exclude:` can also be used to exclude files from fingerprinting.
-Sources are evaluated in order, so `exclude:` must come after the positive
-glob it is negating.
+`exclude:` can also be used to exclude files from fingerprinting. Sources are
+evaluated in order, so `exclude:` must come after the positive glob it is
+negating.
 
 ```yaml
 version: '3'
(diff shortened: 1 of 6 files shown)


andreynering:
Hey @pd93,

I think this changes needs a bit of thought:

1) I'm not sure, but I think it _may_ happen, with `cache: false`, that a given variable shell will be run multiple times, which will be unexpected (and perhaps even dangerous) to the user. That should at least be checked before merging.
2) I have an idea (#1065) to have "lazy variables" at some point. That would be a overhaul on how vars and envs are handled, in which a variable would only be computed once really requested by the template engine. Depending on how we implement that, that may make `cache: false`  obsolete (because vars would not be cached by default to begin with). So this makes me question: should we introduce an option that may soon become obsolete?

pd93:
@andreynering Yeah, you make some very good points. I'll admit that I cobbled this together quite quickly without giving it too much thought in response to the comments on Discord.

Lazy variables are definitely the way forwards long term. I have given some thought about how this could be done. However, I'm not a massive fan of having to call a function inside of templates to get variable values. I actually think there might be a way to do it without the need for functions which would also keep compatibility with v3. Perhaps this is a discussion for another time/place though.

Let's close this for now 👍

PR #1452, opened 2023-12-30
feat: add spew function to templater

pd93:
Adds the spew function to templater which lets you print the details of a variable. This is really useful for debugging now that we have multiple variable types.


andreynering:
Good.

PR #1456, opened 2024-01-04
refactor: bits from the dag branch

pd93:
This PR is an assortment of cherry-picked changes from the `dag` branch. None of these are directly related to the dag functionality, but were made because they benefit its design. The idea being that when I (eventually) post the dag PR, the changes will be much easier to understand and only related to the dag functionality itself.

- refactor: `IncludedTaskfiles` -> `Includes`
  - The idea of this is just that it keeps the consistency in naming with the key used in the Taskfile schema. It also means less use of variables with the word "taskfile" in them, which can get confusing because it's not actually a taskfile, but a reference to one.
- feat: store namespace directly in `Include`
  - This has benefits in the dag code since the namespace is required when merging the taskfiles in the graph together and the complete `Include` struct can be easily stored on a graph edge. However, it also helps tidy up the code a bit without the dag code.
- feat: `Includes` should use `OrderedMap`
  - Not really related to dag. Just a nice tidy up. No need for multiple ordered map implementations.
- refactor: `orderedmap` package -> `omap`
  - More go-like package name IMO.
- refactor: split `read.go` into `taskfile.go` and `reader.go`
  - This is very specific to the dag changes and is really only there because it will help with reading the diffs as there is a dedicated `Reader` in the `dag` branch and code was moved around.
- refactor: move prompts into consts
- fix: getSpecialVars


andreynering (approved):
:clap: :clap: :clap:

PR #1459, opened 2024-01-07
feat: invert call.Direct

pd93:
The API for `taskfile/ast.Call.Direct` isn't great when using Task as a Go package. Creating a task as if it were being called from the CLI means you have to write something like this:

```go
call := ast.Call{Task: "default", Direct: true}
```

I can't think of many situations where you would want to set `Direct: false` from an API as this mainly intended for internal use when parsing a `ast.Cmd` or `ast.Dep`. Additionally, it's easy to forget to add `Direct: true` and omitting it is probably not the behaviour you want.

If we invert the property name from `Direct` to `Indirect`, the default value becomes `false` and so when calling a task with the API you would simply need to do:

```go
call := ast.Call{Task: "default"}
```

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -13,7 +13,7 @@ func Parse(args ...string) ([]ast.Call, *ast.Vars) {
 
 	for _, arg := range args {
 		if !strings.Contains(arg, "=") {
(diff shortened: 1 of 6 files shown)


andreynering:
Better, indeed. Thanks!

PR #1460, opened 2024-01-09
Remove unused problematic method

AlekSi:
When the [tools.go](https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) way of managing dependencies is used, there are occasional problems with incompatible transitive dependencies. Currently, Task suffers from one.

https://github.com/golang/go/issues/61374 changed the signature of the `slices.SortFunc` function to make it compatible with Go 1.21. Currently, Task uses the old version. That makes it impossible to use Task and another tool that uses a newer version of `golang.org/x/exp/slices` in the same `tools.go` file.

Given that `OrderedMap.SortFunc` is not used anywhere, I propose just to remove it. This way, Task can be used with the latest version of `golang.org/x/exp/slices` and with the one currently specified in Task's `go.mod` file.

### DIFF
--- a/internal/orderedmap/orderedmap.go
+++ b/internal/orderedmap/orderedmap.go
@@ -89,11 +89,6 @@ func (om *OrderedMap[K, V]) Sort() {
 	slices.Sort(om.s)
 }
 
-// SortFunc will sort the map using the given function.
-func (om *OrderedMap[K, V]) SortFunc(less func(i, j K) bool) {
-	slices.SortFunc(om.s, less)
-}
-
 // Keys will return a slice of the map's keys in order.
 func (om *OrderedMap[K, V]) Keys() []K {
 	return om.s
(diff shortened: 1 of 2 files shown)


pd93:
Why not just update the `golang.org/x/exp/slices` package to the latest version instead? When 1.22 releases next month, we will probably drop support for 1.20, migrate to the standard `slices` package and drop `x/exp` entirely. Removing a method from our API doesn't seem worth it IMO - especially for the sake of compatibility with an experimental package that probably shouldn't be used in production anyway.

I've been wanting to remove the dep on `x/exp` for a while since the API is fragile and prone to issues like this. When I added it, I just wanted to try out all the new functionality, but with hindsight, I think this was the wrong thing to do in an open-source project. In the future, I think it's probably smarter for us to bring any functions we need from the `x/exp` package under our own `internal` package until something stable is released.

andreynering:
Hi @AlekSi,

I upgraded the package at #1462.

AlekSi:
I would really appreciate if you tag a new release with that fix :)

PR #1465, opened 2024-01-11
feat: version checks

pd93:
- feat: version checks
- refactor: consistent naming for errors
- feat: update schema to accept semver strings

Task's current version checking is a bit out of date and unused. It's a lot of work for us to maintain a list of features and which schema versions they work in, but it's quite trivial to let a user specify a minimum version of task as their schema version and then error if the version of the binary is lower.

This PR extends the existing `TaskfileVersionNotDefined` error (exit code 107) and renames it to the more generic `TaskfileVersionCheckError`. This error is then used whenever we come across a problem with schema/binary versions. In this PR, this happens in 3 places:

1. When reading a Taskfile, if the `version` key is not provided:

```
task: Missing schema version in Taskfile "/path/to/Taskfile.yml"
```

2. In `doVersionChecks`, if the `version` key is less than v3:

```
task: Invalid schema version in Taskfile "/path/to/Taskfile.yml":
Schema version (2.0.0) no longer supported. Please use v3 or above
```

3. In `doVersionChecks`, if the `version` key is greater than the installed version of Task

```
task: Invalid schema version in Taskfile "/path/to/Taskfile.yml":
Schema version (4.0.0) is greater than the current version of Task (3.33.0)
```

### DIFF
--- a/docs/static/schema.json
+++ b/docs/static/schema.json
@@ -531,14 +531,14 @@
       "properties": {
         "version": {
           "description": "Specify the Taskfile format that this file conforms to.",
-          "anyOf": [
+          "oneOf": [
+            {
+              "type": "string",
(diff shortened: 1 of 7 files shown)


andreynering (approved):
:tada:

PR #1469, opened 2024-01-12
feat: add ROOT_TASKFILE special variable

### LINKED ISSUE #1468: Template variable containing absolute path to root taskfile
My particular use case for wanting to provide this value is that I would like to define a `default` task that will run `task -t "{{.ROOT_FILE}}" --list`, such that calling `task` without a specific task name will print the basic help for the file. There are other possibilities, such as a task that copies some user-created template taskfile into the invocation directory, which could be accomplished via `cp "{{.ROOT_FILE}}" "{{.USER_WORKING_DIR}}"` on linux systems. At present, no such commands are possible.

Currently, the `.ROOT_DIR` variable provides only the directory of the root taskfile, which can be passed to `task -d "{{.ROOT_DIR}}" --list` for that particular use case. However, since the CLI accepts the `-t`/`--taskfile` argument, it is possible for a non-default filename to be used, which would prevent any task commands from definitively knowing the actual file being run.

As a side note, it may also be useful to have such a variable for included task files as well, although I don't have a specific use case for such.

pd93:
Fixes #1468


andreynering (approved):
:rocket:

PR #1476, opened 2024-01-18
Fix issues with JSON Schema + add schema linting to CI

### LINKED ISSUE #1474: Schema for `defer_call` contains invalid JSON Schema, causing some schema validators to fail
This was encountered and reported against `check-jsonschema` (a JSON Schema CLI): https://github.com/python-jsonschema/check-jsonschema/issues/376

That bug report covers two issues:
1. there is invalid/malformed data in the taskfile JSON Schema
2. the python-jsonschema projects do not handle the malformed data with a clear error

I'm pursuing (2) independently. Regarding (1), here's the malformed JSON Schema data:

https://github.com/go-task/task/blob/dc3cf1cb16af4210a67c8d7209e476f760fe1266/docs/static/schema.json#L351-L353

`"string"` is not a valid subschema. Possibly this was meant to be `{"type": "string"}` (?) but it's not obvious.

### Aside: Fixing the structure of `definitions`

The current schema has all `definitions` nested under an intermediate key, i.e. `definitions/3/foo` rather than `definitions/foo`. That weakens the ability of downstream validation of the `definitions` section of the schema.

`definitions` is defined under the Draft 7 metaschema as an object whose values are all, themselves, valid schemas.
The current structure of the schema means that this is validating `#/definitions/3` rather than `#/definitions/defer_call` (which would be much more useful).

It would be nice to move all of the definitions up one level so that metaschema validation can catch issues like this.
(Shameless plug: `check-jsonschema --check-metaschema schema.json` can do this for you. :slightly_smiling_face: )

Under [the latest draft definition of $defs](https://json-schema.org/draft/2020-12/json-schema-core#name-schema-re-use-with-defs), this kind of structure is more clearly spelled out, using "MUST"s to clarify that a schema which doesn't follow this structure is invalid. Under the Draft 7 spec, I think it's pretty clear that this was the intent based on the metaschema, but it wasn't called out explicitly.

-----

I have no direct experience with `task`, so I might not be the right person to modify the schema appropriately.

But I'm happy to submit one or two PRs to
- remove the errant `"string"` in `defer_call`
- remove the `3` from `definitions` nesting (If this is versioning, we could rename all of the defs to end with `v3`, as in `env -> env_v3`)

Just let me know how to proceed!

sirosen:
Resolves #1474.
I've gone ahead and included a CI change to start running `check-jsonschema` against the schema in the docs dir, as mentioned in that issue, since it was part of how I validated this change.

I've tried to chunk the commits for easy review:

- Remove "versioning" from jsonschema 'definitions' (remove the `"3"`)
- Dedent JSON Schema 'definitions' (git diff hates this one)
- Fix malformed `defer` definition in JSON Schema (`"string" -> {"type": "string"}`)
- Add a linting step for the JSON Schema (`.github/workflows/lint.yaml`)

---

Aside:

I've been a little bit biased in check-jsonschema towards developers with workflows like my own and haven't thought hard enough about use-cases like this repo. I'll be thinking harder about providing docker containers to run the tool, perhaps some experimental binary builds (there are tools for this like pyinstaller) and maybe a GitHub Action too.

One thing which GitHub Actions has, which I would really like to be able to offer, is the integration with Dependabot to help you stay up to date as the tool evolves. So that's another angle on this.

I can't promise anything yet, but I'd like to make installing and using it easier for non-Python projects.


andreynering:
@sirosen I believe the CI failed because it missed the `checkout` step.

sirosen:
:facepalm: I feel silly for missing that; good catch! I've just pushed the added step as a fresh commit. I could squash it with the previous one if you have strong feelings about the git history and would like me to do so.

andreynering:
@sirosen Don't worry! A small mistake.

It'd be nice to have that commit squashed, if you don't mind. Also, adopting [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for the commit messages would be great.

sirosen:
I just did a rebase to conventional-commits-ize the branch + squash those two commits. :+1:

PR #1478, opened 2024-01-19
fix: experiment flags not working when .env is not in cwd

pd93:
Updates the way we look for an `.env` file. Currently, if you use the `--dir` or `--taskfile` flags when using `.env` to define which experiments are enabled, Task will look in the wrong directory since the default (`./env`) is relative to the user's current working directory, not the directory containing the Taskfile.


andreynering (approved):
:100:

PR #1481, opened 2024-01-21
refactor: ast merge functions

pd93:
Small refactor to move `func Merge(t1, t2 *ast.Taskfile, ...)` in the `taskfile` package to `func (t1 *Taskfile) Merge(t2 *Taskfile, ...)` in the `ast` package. This is just to make it consistent with the other `ast` structures which each have their own `Merge` function.

It also splits the code that merges tasks together (currently inside the aforementioned Taskfile merger) into it own `func (t1 *Tasks) Merge(t2 Tasks, ...)` so that this is consistent as well.


andreynering (approved):
:hammer:

andreynering:
I rebased to fix a small conflict.

PR #1482, opened 2024-01-23
fix(completions): support lowercase filename for Taskfile

xontab:
Allow Taskfiles starting with lowercase characters in zsh completion since it is supported since v3.27.0 - 2023-06-29


andreynering:
Thanks @xontab!

PR #1483, opened 2024-01-25
feat: read Taskfiles from stdin

### LINKED ISSUE #655: Read Taskfile from stdin
Hi, it'd be cool to do something like this:

```bash
generate_taskfile.sh | task --taskfile - lorem
```

pd93:
Closes #655

Adds a new `StdinNode` that implements the `Node` interface. This is a special-case Node that only gets created if Task detects bytes in stdin when starting. I've refactored the way we create "root" nodes (i.e. nodes without parents) in order to do this. I think the result is a bit nicer and lays the groundwork for root remote nodes too.

I also had to resolve an issue in the reader where an included file's `BaseDir` was only being set for `FileNode`. I have added a `BaseDir()` method to the `Node` interface and this will now be added by any node that returns a non-empty string when calling this method.

Lastly, I have updated the docs with a section on reading from stdin. This also includes a commit that adds some missing syntax highlighters to our docs site and standardises the ` ```shell` code blocks language ID.

### DIFF
--- a/docs/docs/api_reference.md
+++ b/docs/docs/api_reference.md
@@ -11,7 +11,7 @@ toc_max_heading_level: 5
 
 Task command line tool has the following syntax:
 
-```bash
+```shell
 task [--flags] [tasks...] [-- CLI_ARGS...]
 ```
 
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -69,7 +69,7 @@ tasks:
       - ./foo-printer.bash
 ```
 
-```bash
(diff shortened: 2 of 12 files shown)


pd93:
@andreynering Small change in 7abbc15 as per this conversation: https://github.com/go-task/task/issues/655#issuecomment-1910529974

@brianbraunstein Could you please give the changes a go and see if they work as you expect

andreynering:
@pd93 Looks fine to me 👍

PR #1489, opened 2024-01-27
feat: wildcard matching for task names

### LINKED ISSUE #836: RFE: Allow use of wildcard or regex in task names (dynamic task names)
In order enable taskfile to a viable replacement for some more specialized test runners, such tox or nox, we need to allow it to accept task names that are using a pattern (regex or glob), so users would not have to copy/paste task definitions when creating test-matrixes.

A very easy to explain example is that tox accepts tasknames like `py`, `py36`, `py311` without forcing user to explicitly define each value. The idea is that the variadic part of the name ends-up  as being processed as a variable.

In fact its support for this goes even deeper allowing you to use multiple dimensions and have tasks like: `py36-func`, py36-unit`, all without having to define an endless list of combinations.

While supporting multiple dimensions might be seen maybe as too much for taskfile, maybe we should at least allow user to define a taskname like `foo*` and ensure that the effective taskname is available inside a variable. That should allow someone to process the name and infer the matrix variables to use for it, or give an error message if the arguments are not valid.

While the glob version (`py*`) would be easier to implement, the regex would be more flexible as it would allow to validate the task name and even document allowed inputs, like `py(\d+)\-((unit|func))`. Using capture groups we can also extract the dimensions, if using [named capturing groups](https://www.regular-expressions.info/named.html) we could even determine in which variables we could save them.

pd93:
Fixes #836

## Context

This PR is a proposal for a design that would allow basic wildcard matching for task names. At the moment, if users want to make a generic task that performs a common set of operations on many different services or directories (etc) they must either:

1. Pass a variable into Task using environment variables
2. Use `.CLI_ARGS` and the `--` operator

Both of these methods can be clunky to type into a CLI and are not very intuitive. `make` has the option to do wildcard matching (using `%`). This PR is a proposal to do the equivalent in Task.

## Proposal

When a Taskfile has a task with a name that contains a wildcard (`*`) and the user calls `task` with a task name that "matches" a wildcard task, that task will be run. Matches will occur by replacing any instances of `*` with the `(.*)` regex string and doing a regex match. The captured variables will be made available to the Task via the `MATCH` variable in the form of an array. For example:

```yaml
version: 3

tasks:
  # A task that starts any service
  start-*:
    vars:
      SERVICE: "{{index .MATCH 0}}"
    cmds:
      - echo "Starting {{.SERVICE}}"
      - ...
```

```shell
> task start-foo
Starting foo
...
```

## Implementation

The first commit updates all usages of `ast.Call` to be passed by reference. Doing this allows methods that accept a call to update the call's variables. This means we can add any wildcard matches into the `call.Vars` without have to propagate them and change a bunch of function signatures.

Other than this, the implementation is very simple, we are simply overriding the `omap.Get()` method by adding a `Get()` method directly on `ast.Tasks` and then adding a simple `WildcardMatch()` function to `ast.Task` which will work out if a Task matches the call and retrieve the wildcard variables

## Notes

I haven't written any docs yet since I'd like some feedback on the design/implementation. I'll add this once we've decided we're happy with it. What do people think about the naming? i.e. `*` and `MATCH`. There were some other ideas discussed in #836, but I think starting with simple wildcard matching is probably a good start. I personally, would find this to be more than enough for my use cases.

Final note. The `omap.Range()` function doesn't support yielding right now so there is a bit of a hack to get around this (see `TODO` in code). There is an interesting [iterator proposal](https://medium.com/eureka-engineering/a-look-at-iterators-in-go-f8e86062937c) that will be introduced to Golang 1.22 as an experiment. This would be a really good solution to this, but we probably won't be able to add this until it is added to the language permanently (probably 1.23) and we drop support for 1.22 (so around a year from now).

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -1249,8 +1249,8 @@ $ task run-foo-bar
 foo bar
 ```
 
-If multiple matching tasks are found, the first one listed in the Taskfile will
-be used. If you are using included Taskfiles,
+If multiple matching tasks are found, an error occurs. If you are using included
+Taskfiles, tasks in parent files will be considered first.
 
 ## Doing task cleanup with `defer`
 
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -1205,6 +1205,53 @@ tasks:
       - yarn {{.CLI_ARGS}}
 ```
 
+## Wildcard arguments
+
+Another way to parse arguments into a task is to use a wildcard in your task's
+name. Wildcards are denoted by an asterisk (`*`) and can be used multiple times
+in a task's name to pass in multiple arguments.
+
+Matching arguments will be captured and stored in the `.MATCH` variable and can
+then be used in your task's commands like any other variable. This variable is
+an array of strings and so will need to be indexed to access the individual
+arguments. We suggest creating a named variable for each argument to make it
+clear what they contain:
+
+```yaml
+version: '3'
+
+tasks:
+  echo-*:
+    vars:
+      TEXT: '{{index .MATCH 0}}'
+    cmds:
+      - echo {{.TEXT}}
+
+  run-*-*:
+    vars:
+      ARG_1: '{{index .MATCH 0}}'
+      ARG_2: '{{index .MATCH 1}}'
+    cmds:
+      - echo {{.ARG_1}} {{.ARG_2}}
+```
+
+```shell
+# This call matches the "echo-*" task and the string "hello" is captured by the
+# wildcard and stored in the .MATCH variable. We then index the .MATCH array and
+# store the result in the .TEXT variable which is then echoed out in the cmds.
+$ task echo-hello
+hello
+# You can use whitespace in your arguments as long as you quote the task name
+$ task "echo-hello world"
+hello world
+# And you can pass multiple arguments
+$ task run-foo-bar
+foo bar
+```
+
+If multiple matching tasks are found, the first one listed in the Taskfile will
+be used. If you are using included Taskfiles,
+
 ## Doing task cleanup with `defer`
 
 With the `defer` keyword, it's possible to schedule cleanup to be run once the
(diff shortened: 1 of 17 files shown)


vmaerten:
Thanks for this work.
It's a really cool feature and I have some use case of it! 
I think MATCH is a good name

docs/docs/usage.md
```diff
@@ -1205,6 +1205,53 @@ tasks:
...
+# You can use whitespace in your arguments as long as you quote the task name
+$ task "echo-hello world"
+hello world
+# And you can pass multiple arguments
+$ task run-foo-bar
+foo bar
+```
+
+If multiple matching tasks are found, the first one listed in the Taskfile will
+be used. If you are using included Taskfiles,
```
andreynering, line 1253: 1) Would return an error be better if multiple matches are found?
2) Reminder: the second sentence is still incomplete :slightly_smiling_face:

andreynering (approved):
Great work! :rocket: 

I agree that a wildcard is probably enough. Regexp was suggested on the issue, but it'd be too complex IMO, specially when mixing in YAML.

docs/docs/usage.md
pd93: 1. I think that adding this restriction is fine for now. It keeps things simple. However, I can definitely see a world where someone with a complex setup asks for the current behaviour as it would allow you to do something like this:

    ```yml
    version: 3

    tasks:
      foo-*-*:
        ..
      foo-*:
        ..
    ```

    This allows Task to do something different depending on if the user does `task foo-bar` vs `task foo-bar-baz`. This would not be possible if we errored when multiple tasks match since `task-bar-baz` matches both tasks in the file.

    Happy to add the error for now and revisit this later though.

2. Haha, I probably started writing this and then realised I didn't know what the behaviour with includes was, so went to check and never came back. I have completed the sentence now 😆

PR #1491, opened 2024-01-30
docs: remove unnecessary docs for requires

teatimeguest:
It seems to be for an early version of PR #1204.


andreynering:
Thanks @teatimeguest!

PR #1510, opened 2024-02-16
docs: specify that variable declaration order is respected

kirkrodrigues:
https://github.com/go-task/task/commit/6ed30f1addf85dcf015e3586008e3f9bdf59041d ensures that variable declaration order is respected, but this wasn't documented as far as I could tell. This PR adds a note about it to the API reference.


andreynering:
Thanks @kirkrodrigues!

PR #1521, opened 2024-02-27
refactor: move flags into their own package

pd93:
Simple refactor to move flags into their own package.

Just an idea that I was poking around at. It means that all the code to do with flags is isolated and cleans up the `cmd/task.go` file a bit. Definitely subjective though, so it's fine if you're not a fan.


andreynering (approved):
Looks good!

Just needs to fix conflicts before merging.

PR #1523, opened 2024-03-01
feat: taskfile special variable

### LINKED ISSUE #1522: [3.35.0 regression] Precondition scripts no longer working
- Task version: 3.35.0
- Operating system: Linux (Solus)
- Experiments enabled: N/A

I was updating the Solus go-task package to v3.35.0, and while testing it I noticed that there was a regression in that a script that we had as a precondition of a task stopped working. This worked on v3.34.1, so this is a regression specific to v3.35.0

We have a task for publishing packages that looks like the following:
```
  publish:
    desc: Tag and publish a release
    dir: '{{ .USER_WORKING_DIR }}'
    preconditions:
      - sh: test $(git symbolic-ref HEAD 2>/dev/null) = "refs/heads/main"
        msg: Not on main branch
      - sh: "{{ .TASKFILE_DIR }}/common/Scripts/package-publish-safety-catches.sh"
        msg: Failed to pass safety catches
    deps:
      - package-file
    cmds:
      - git push
      - task: push
```

The `package-publish-safety-catches.sh` script is available [here](https://raw.githubusercontent.com/getsolus/packages/main/common/Scripts/package-publish-safety-catches.sh).

When this task is called it now always errors out with the following error:
```
task: Failed to pass safety catches
task: precondition not met
```
When the script is ran directly it runs without issue and exits with a `0` status code (IE, indicating that none of the conditions that it's designed to detect are present). I tested both Fish shell and Bash to see if it made a difference and both had the same error. In case it matters we are building with golang 1.22.0. I also tested the official v3.35.0 binaries from Github and they exhibit the same behavior.

Bisecting this reveals that the bad commit is bb9d58225502953eb32b58525c975f078d3d9c8e.

pd93:
Fixes #1522

Fixes the `TASKFILE_DIR` variable and adds a new `TASKFILE` variable.

### DIFF
--- a/docs/docs/api_reference.md
+++ b/docs/docs/api_reference.md
@@ -130,6 +130,7 @@ There are some special variables that is available on the templating system:
 | `TASK`             | The name of the current task.                                                                                                                            |
 | `ROOT_TASKFILE`    | The absolute path of the root Taskfile.                                                                                                                  |
 | `ROOT_DIR`         | The absolute path of the root Taskfile directory.                                                                                                        |
+| `TASKFILE`         | The absolute path of the included Taskfile.                                                                                                              |
 | `TASKFILE_DIR`     | The absolute path of the included Taskfile directory.                                                                                                    |
 | `USER_WORKING_DIR` | The absolute path of the directory `task` was called from.                                                                                               |
 | `CHECKSUM`         | The checksum of the files listed in `sources`. Only available within the `status` prop and if method is set to `checksum`.                               |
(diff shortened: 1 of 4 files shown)


pd93:
Coincidentally, this was actually part of an upcoming change, so I've just cherrypicked it into a patch.

@ReillyBrogan any chance you could give this change a go for me? It's difficult for me to test your exact scenario, but I think this should fix your issue. If you have Go installed, you can try it by running:

`go install github.com/go-task/task/v3/cmd/task@a28d273`.

ReillyBrogan:
I built v3.35.0 with this patch cherry-picked and it does indeed fix the issue!

Issue #1519, opened 2024-02-27
Wildcards not working


iPenguin:
Hi I'm trying to use the wildcard arguments as [listed in the documentation](https://taskfile.dev/usage/#wildcard-arguments). I've updated to the latest version of taskfile but it's not working.

I also tried to copy the example in the docs and that isn't working either.

Am I missing something obvious?

```
---
version: '3'

tasks:
  echo-*:
    vars:
      TEXT: '{{index .MATCH 0}}'
    cmds:
      - echo {{.TEXT}}

  run-*-*:
    vars:
      ARG_1: '{{index .MATCH 0}}'
      ARG_2: '{{index .MATCH 1}}'
    cmds:
      - echo {{.ARG_1}} {{.ARG_2}}
```

Thanks,
Brian

- Task version: v3.34.1
- Operating system: Linux
- Experiments enabled: no

pd93:
@andreynering This is happening quite often. Maybe we ought to change the CD to only update the docs site on releases for now?

andreynering:
@pd93 Yeah, we should eventually do that.

iPenguin:
Ah, well that explains it, thanks.

If you aren't going to version the docs for a while, maybe you could indicate that a feature is still in progress, or what version it should land if you know.

Thanks,
Brian

FFengIll:
I meet the same question, is there anything updated for it?

pd93:
@FFengIll Wildcard matching was released in [v3.35.0](https://github.com/go-task/task/releases/tag/v3.35.0). Please make sure you're running the latest version. If you're still unable to make it work, please open a new issue.

iPenguin:
> @FFengIll Wildcard matching was released in [v3.35.0](https://github.com/go-task/task/releases/tag/v3.35.0). Please make sure you're running the latest version. If you're still unable to make it work, please open a new issue.

I just downloaded the new version and tried it out. It works well. Thank you!

Issue #1415, opened 2023-11-30
Any variables experiment


pd93:
> [!WARNING]
> All experimental features are subject to breaking changes and/or removal at any time. We strongly recommend that you do not use these features in a production environment. They are intended for testing and feedback only.

> [!NOTE]
> You can view the [Any Variables experiment documentation](https://taskfile.dev/experiments/any-variables) on our website, including instructions on how to enable/disable it.

## Context

This experiment attempts to solve the problems originally described by #140.

Currently, all Task variables are strings. You can define a variable as a YAML `string`, `bool`, `int`, `float` or `null` value. However, when task interprets it, it will be decoded as a string regardless of the input type.

For example, the following Taskfile will _always_ output "foo" even when `BOOL` is `false` because the variable is being interpreted as a string and an `if` statement evaluates to `true` when the string is not empty:

```yaml
version: 3

tasks:
  foo:
    vars:
      BOOL: false
    cmds:
      - '{{ if .BOOL }}echo foo{{ end}}'
```

Lists are also interpreted as strings and this has led to some annoying workarounds in the codebase. For example, the `for` feature is great for running a task multiple times with different variables. However, if you want to loop over an arbitrary list of strings, you have to define the list as a delimiter separated string and then split it by specifying the delimiter in the `for` statement. For example:

```yaml
version: 3

tasks:
  foo:
    vars:
      LIST: 'foo,bar,baz'
    cmds:
      - for:
          var: LIST
          split: ','
        cmd: echo {{ .ITEM }}
```

This could be simplified if we supported variables as lists:

```yaml
version: 3

tasks:
  foo:
    vars:
      LIST: [foo, bar, baz]
    cmds:
      - for:
          var: LIST
        cmd: echo {{ .ITEM }}
```

## Proposal

We propose to change the type of internal Task variables to `any` (otherwise known as an empty `interface{}` in Go). This will allow users to define variables as any type they want, and Task will interpret them properly when used in tasks. The following types should be supported:

- `string`
- `bool`
- `int`
- `float`
- `array`
- `map`

Adding support for these types is relatively simple by itself. However, there a few changes that will be needed to make the rest of Task's features work nicely with the new variable types:

- [ ] `for` should support iterating over arrays (and maybe maps?) https://github.com/go-task/task/pull/1436
- [ ] `sh` needs to be removed and replaced with a new syntax for dynamically defined variables (see [backwards compatibility](#backwards-compatibility))
- [ ] ...

## Backwards Compatibility

[...]

PR #1525, opened 2024-03-02
chore: deploy docs only when a new version is released

vmaerten:
Hi !

Documentation is released each time a commit is pushed to main. This can lead to misunderstanding to users, they can think that all features are available on the latest Task's version which is not always true.
(see : https://github.com/go-task/task/issues/1415#issuecomment-1895470072 https://github.com/go-task/task/issues/1519#issuecomment-1966814452)

As suggested in the last comment, we could deploy the doc only when a new version is released.

What I did ? I removed the upload-source-documents's workflow and added the doc deployment in release's workflow

I also checked, the upload-source-documents's workflow has been trigger manually once so it should not be an issue if we drop the "workflow dispatch" 
cf : 
image


andreynering:
Hi @vmaerten!

Thanks for your contribution!

I'm going to close this in favor of #1527 which will be a more complete solution to this problem. 🙂

PR #1527, opened 2024-03-03
chore: docs versioning and i18n

pd93:
This PR solves a couple of issues with documentation versioning and internationalisation.

The problems:

1. When we make changes to our docs, it updates the website immediately.
   - This has led to confusion in threads where users think that a feature is available because it is on the website. e.g.
     - https://github.com/go-task/task/issues/1415#issuecomment-1895470072
     - https://github.com/go-task/task/issues/1519#issuecomment-1966814452
3. Our i18n translations are stored in the Git repository
   - This encourages users to make changes to other languages via Git (This should be done via Crowdin)
   - Automatically generated "sync crowdin" PRs need to be constantly reviewed and merged
   - Its a _lot_ of duplication in the repository
   - It creates a lot of changes when used in conjunction with versioning (see solution for problem 1)

The solutions:

1. We could theoretically avoid deploying the website until we cut a new release (this was proposed in #1525). However, this means that the docs for our "next" version are not very discoverable and it's more difficult to spot errors before a release is cut. A better solution is to use Docusaurus's built-in versioning functionality to create two versions of our documentation:

- `latest`: Represents the latest released version of Task and will only be updated when a release is made. This will be the default when opening our website
- `next`: Represents all the tip of our `main` branch and contains all the latest changes. Users would have to explicitly select "Next 🚧" from the versions drop-down on the website to view this. Docusaurus also automatically adds a banner to the top of every page warning users that this functionality has not been released.

This PR also updates the release tool so that the "latest" version is automatically updated when a maintainer create a new release.

2. The solution to this is simple. Remove and gitignore the `docs/i18n` folder and remove the `sync-translated-documents` workflow from the repository. Instead, we should update the command that Netlify runs when it builds the documentation to pull the translations from Crowdin. This is how the [Docusaurus documentation actually suggests this should be done](https://docusaurus.io/docs/i18n/crowdin#automate-with-ci).

Note that this PR puts in place all the changes required for this to work (I think?). However, the command will need updating in Netlify to actually apply this change, otherwise the site will be deployed without any translations. The command will need updating to `npx crowdin download && npx docusaurus build`

- chore: remove all translations
- chore: remove sync crowdin workflow
- chore: gitignore i18n directories
- chore: add crowdin:sync script to package.json
- feat: add "latest" version
- feat: update release tool to update latest doc version
- fix: logo in next/intro docs and spacing


andreynering:
@pd93 Awesome work! Thanks for taking the time to do these improvements.

I have two small details to discuss:

1) We have this bug where depending on the width of the browser window, we have too little space for all that stuff on the top bar. This already happens on our current site, but got a little worse with yet another menu on the top bar.

Screenshot 2024-03-10 at 11 06 47

We don't necessarily need to fix on this PR (can be in another one), but I would consider 1) make these social network links into icon-only links (the [Docusaurus website](https://docusaurus.io/) itself does this); or 2) Remove Twitter, Mastodon and Discord and keep only GitHub (and perhaps Discord) which is/are the most important.

2) How will Crowdin / translations work with the next website? Should we remove this language menu for Next?

Screenshot 2024-03-10 at 11 13 55

pd93:
> 1. We have this bug where depending on the width of the browser window, we have too little space for all that stuff on the top bar. This already happens on our current site, but got a little worse with yet another menu on the top bar.

I had noticed this, but was going to raise another issue/discussion as there are a few things we could do to improve the navbar. I generally agree with your first choice of moving towards icons instead of text links. There are also some other things I had considered:

- We also have some things duplicated in the sidebar and top bar. We could probably keep this as `Docs`, `Blog` and `Donate`
- Remove the Donate page from the docs and make it its own proper [page](https://docusaurus.io/docs/creating-pages).
- Move the docs under the `/docs` path (we could create some temporary [redirects in Netlify](https://docs.netlify.com/routing/redirects/) to aid migration.
- Unrelated, but I'd like to consider renaming the top level `docs` folder in the repo to `website` as the website encompasses more than just documentation now. This follows the structure in Docusaurus' own repo. The `docs/docs` hierarchy also stutters a bit.

 
> 2. How will Crowdin / translations work with the next website? Should we remove this language menu for Next?
>

andreynering (approved):
The site is looking great, thanks!

I'd just change the icons to be in the following order (most important first: GitHub, Discord, Twitter, Mastodon.

> Unrelated, but I'd like to consider renaming the top level docs folder in the repo to website as the website encompasses more than just documentation now. This follows the structure in Docusaurus' own repo. The docs/docs hierarchy also stutters a bit.

It looks like a good idea to me!

PR #1537, opened 2024-03-05
feat(taskfile): add post tasks feature

peterparser:
Hi, while using go-task I found that a post-tasks like feature would be cool, in my case it would be nice to use it as a teardown for my test environments.

I used `posts` as a keyword for implementing this feature, if you find something more suitable I will be more than happy to refactor it.

Thank you!


andreynering:
Hi @peterparser,

Why do you think `posts` would be useful?

You can already add extra commands to your `cmds` list. And to ensure something is run even after a fail, we already have [`defer`](https://taskfile.dev/usage/#doing-task-cleanup-with-defer).

So, I'm inclined to not have this so we don't have an overlap of features.

pd93:
No response from the auth for 3 months. From what I can tell the functionality the author was after is already available in task.

PR #1556, opened 2024-03-20
use path/filepath instead of path

mattn:
Do not use path package for operating system paths.


andreynering:
Nice to see you here @mattn, and thanks for contributing!

Issue #1240, opened 2023-06-26
Lazy global dynamic variable


lockieluke:
Only load global variables with `SH` when the task needs it:
```yml
version: '3'
vars:
  MESSAGE:
    sh: sleep 5 && echo "WORLD"

tasks:
  print:
    cmds:
      - echo "HELLO {{.MESSAGE}}"

  print-no-wait:
    cmds:
      - echo "PRINT THIS INSTANTLY"
```
Other tasks that don't need the variable should not have it loaded

user2410:
I have been using this workaround for the issue.

```yaml
version: '3'

vars:
  MASTER_IP_CMD: |
    awk '/ master /{print $1}' ./machines.txt

tasks:
  build:
    desc: "Build the cluster"
    cmds:
      - ...
      - vagrant up # This produces ./machines.txt
      - ...

  certs:gen:
    desc: "Generate certificates for components"
    vars:
      MASTER_IP:
        sh: |
          {{.MASTER_IP_CMD}}
    cmds:
      # Your commands here

  dist:certs:
    desc: "Distribute certificates to VMs"
    vars:
      MASTER_IP:
        sh: |
          {{.MASTER_IP_CMD}}
    cmds:
      # Your commands here
```

Because the MASTER_IP variable is only available after `vagrant up`, resolving the global variable before running `task build` will result in error.
This workaround avoids evaluating the global dynamic variable by storing the command as a plain string, deferring its execution until a task runs and actually needs it.

Issue #1030, opened 2023-03-02
Taskfile Scoped Variables


ameergituser:
I would like Taskfile scoped variables so that the behaviour of the parent taskfile or the included child taskfiles is not implicitly changed just by including a taskfile. This topic was touched on a few times with @andreynering and the latest discussion is in #1006. 

You need this in use cases whereby multiple separate Taskfile projects are chained together in a parent/child relationship. And in this case, you don't own or necessarily know the inner details of these lower-layer included taskfiles. 

Currently, you have to know every single variable that is globally scoped so that you don't use the same variable name in the parent or child taskfiles. It kind of defeats the purpose of taskfile inclusion to provide abstraction. 

The worst is that if you don't go through every taskfile, you won't know when you are using a variable that changes the behaviour somewhere in some taskfile unintentionally.  

@andreynering Thanks for the comment on #1006. I agree that for the next major version making variables scoped by Taskfile is a reasonable consideration and design decision, which would make maintaining taskfiles simpler as the behaviour would be a lot more predictable. 

1. Do you think it's feasible to add a taskfile setting to explicitly enable scoped variables in the current version? 
2. And for the future major version, this could be the default? and perhaps allow variable scope/overriding as an explicit feature?

apgrucza:
I agree that variables scoped by Taskfile should be the default. Although we could add an option to explicitly allow variable sharing on particular includes, it's not ideal as the included task might still want to have a subset of its variables remain private. It's better for the included Taskfile to decide what's public and what's private. This could be achieved by introducing new `public` and/or `private` attributes.

I've noticed a lot of issues raised regarding the behaviour of variables and Taskfile inclusion. Examples are #454, #696, #778 and #996. It would be good to find a solution that addresses all of these issues. There seems to be two reasons that people use includes:

1. to pull in tasks from the included file
2. to pull in variables from the included file

I wonder whether it makes sense to decouple these and come up with a separate mechanism for sharing variables across Taskfiles.

As an alternative to the public/private idea, we could support a new file type (e.g. `Taskvars.yml`). This would basically be a cut-down Taskfile that only allows `vars`, `env` and `dotenv` attributes. The file would be loaded with a new attribute (e.g. `var-files: ['Taskvars.yml']`). Many other tools support loading variables from a file in this fashion.

[...]

Issue #852, opened 2022-08-31
Make run:once work for dependencies shared amongst included taskfiles


Place1:
I'm currently working in a mono-repo of sorts and each of our services/libraries has a Taskfile with a "build" task.

When a service depends on a library we use "includes" and "deps" to make sure the library is built before the service is built.

We also have a root Taskfile that includes all service Taskfiles and calls their build task in a deps array.

This works really well but we've found that when a library is referenced by multiple services it will be built multiple times by the root Taskfile.

We also found that the "sources" feature doesn't work for tasks shared between included Taskfiles.

Here's the dry run from the root Taskfile showing the library getting built twice:

```bash
$ task build --dry
task: [service-a:library:build] run-build.sh 
task: [service-b:library:build] run-build.sh 
task: [service-a:build] run-build.sh 
task: [service-b:build] run-build.sh
```

and here's what a service Taskfile looks like

```yml
version: '3'

includes:
  web:
    taskfile: ../library/Taskfile.yml
    dir: ../library

tasks:
  build:
    run: once
    deps: [library:build]
    cmds:
      - run-build.sh
    sources:
      - src/**/*
```

and the library Taskfile

```yml
version: '3'

tasks:
  build:
    run: once
    cmds:
      - run-build.sh
    sources:
      - src/**/*
```

It seems like the include namespacing causes `task` to treat every task as a unique one when multiple Taskfiles include a shared task.

We also noticed that if we run a "task build" in service-a and then service-b that the `.task/` checksum cache is not shared and so the library is built twice. The same applies for any order of builds, i.e. building the library then the services.

Ideally the includes form a directed acyclic graph (DAG) and `task` will run the dependencies in order and only once (when run:once is set), with the task's checksum cache being considered.

Is this something in scope for improvement in taskfile v3?

TomerBin:
Hey all :)
Did anyone come up with a solution or workaround for that issue?

pd93:
Responding to [this comment](https://github.com/go-task/task/pull/1563#issuecomment-2103378594)

> Is #852 resolved with this?

@ReillyBrogan No, the DAG changes do not resolve this issue currently. This is because we are still merging into `ast.Taskfile` once the DAG reader is done reading files. This will eventually go away and resolve this issue. However, we can do something to fix it in the meantime.

See #1655 for details. Feedback on the PR would be appreciated from anyone interested in this issue.

PR #1563, opened 2024-03-25
feat: directed acyclic graph (DAG)

pd93:
This PR overhauls how Task reads, parses and merges Taskfiles.

## An example project

The following document will get quite technical, so I'm going to use an example to help explain. Imagine a project with the following Taskfile structure:

```
├── services
│   ├── serviceA
│   │   └── Taskfile.yml
│   ├── serviceB
│   │   └── Taskfile.yml
│   └── serviceC
│       └── Taskfile.yml
├── taskfiles
│   └── utils.yml
└── Taskfile.yml
```

We have three "services", each with their own Taskfile, and a root Taskfile that includes all of them. This allows us to call `task` from the project root and get access to all the tasks in each of the services. In addition to this, we also have a `taskfiles` directory with a `utils.yml` file that contains some common tasks that are shared between the services and the tasks in the root file.

This is a fairly typical, if not simple, project structure for a Taskfile-based project and on the surface, Task handles it pretty well. However, as we are about to see, under the hood, Task does not process this very efficiently, and when you add more services and more shared utilities, or run on hardware with limited processing capability, the inefficiencies can become problematic.

## The current process

For the sake of brevity, I'm not going to write up the contents of each file. Instead, I am going to illustrate the `includes` by drawing a **graph**:

```mermaid
flowchart TD
    R(Taskfile.yml)
    A(services/serviceA.yml)
    B(services/serviceB.yml)
    C(services/serviceC.yml)
    U(taskfiles/utils.yml)
    R --> A
    R --> B
    R --> C
    R --> U
    A --> U
    B --> U
    C --> U
```

You can see that our root Taskfile (`Taskfile.yml`) is including each of the service Taskfiles and the utilities Taskfile. Each of the service Taskfiles is then also including the utilities Taskfile. The total setup consist of 5 files.

When Task is called, the first thing it does it look for the root Taskfile (aka. the "entrypoint"). The contents of this files are then read into memory and parsed using the `taskfile/ast` package. So far, so good.

Once the entrypoint has been parsed, we can now start to evaluate any `includes` that may be specified in the file. In this example, we have 4 includes:

- `services/serviceA.yml`
- `services/serviceB.yml`
- `services/serviceC.yml`
- `taskfiles/utils.yml`

Now, for each of these taskfiles, we will recursively repeat the process of reading the files into memory and parsing them. Once all the includes have been followed, we then unwind the stack and merge the ASTs together. This results in one giant Taskfile the contains all the tasks from all Taskfiles in the project.

If you haven't spotted it yet, there is a major problem with this approach. Let's jump into the code and add a log line at the beginning of each call to the function that reads a Taskfile that print the name of the file being read. Here are the results:

```
Taskfile.yml
services/service1/Taskfile.yml
taskfiles/utils.yml
services/service2/Taskfile.yml
taskfiles/utils.yml
services/service3/Taskfile.yml
taskfiles/utils.yml
taskfiles/utils.yml
```

As you can see, despite only having 5 files in the project, Task has read a file 8 times. The `taskfiles/utils.yml` file has been read 4 times! The number of times a file is read will correspond directly to the total number of `includes` in all your Taskfiles combined (or the number of arrow heads in the graph diagram) +1 for the entrypoint.

## Analysing the problem

There are many simple solutions to the problem described above. For example, we could just add a simple file cache so that a file is not read multiple times. However, I think that the problems with the current approach go deeper than just the inefficiency of reading files multiple times.

I briefly mentioned that the ASTs are merged together before any tasks are run. There are a few disadvantages to this approach:

- Merging is very memory intensive and involves deep copying structures between Taskfiles.
- The code is complex and difficult to maintain.
  - This increases the chance of bugs.
  - and increases the barrier to entry for contributors.
- You lose information about the location of data when merging.
  - The includes the scopes of variables and tasks.
  - Currently, we attempt to save some of this information by storing it in the merged Taskfile, but this just adds to the complexity/maintenance cost of the code.
- Most of the time, you don't actually need the entire Taskfile tree to be read.
  - Often, we just want to run a Task in the root Taskfile. Why bother reading and merging all the other Taskfiles?

It doesn't matter how good our file caching is if we end up calling a merge method for each and every include. Finding a way to store the entire Taskfile tree in memory without merging them together would be a much better solution.

## A new approach

Taskfiles are no different to other programming languages or config-based tools that have `includes` or `imports`. We usually refer to the structures that store dependencies as graphs (as-per the example diagram). Specifically, these are dependency graphs, which are a type of directed acyclic graph (or DAG for short).

DAGs have a number of properties that make them very useful:

- As the name suggests, they are acyclic, which means we get cycle prevention for free. This solves issues like [this](https://github.com/go-task/task/pull/1332) which occur because of our naive approach to cycle detection.
- It gives us file caching for free. A DAG allows us to store/cache the AST of a files once it has been read. Since the key of the node in the graph is the file's location, this will be retrieved next time its needed rather than being read from the filesystem again.
- DAGs are a well understood data structure and there are many algorithms and libraries available for working with them.
- They allow us to easily output and visualise the Taskfile tree for debugging or illustration purposes.
- Most importantly, they allow us to store the entire Taskfile tree in memory without merging them together. This means that we can solve all of the issues mentioned in the previous section.

## This PR

An important point to make - **This PR does not remove AST merging**. The DAG implementation has taken _a lot of work_ to complete. The first DAG changes were made by me nearly a year ago and I have been slowly cherry-picking refactors and improvements from the `dag` branch into `main`. This PR represents the final piece of work to bring the DAG implementation to release.

However, this is still just a stepping stone. Currently, the DAG is still resolved into a single AST before tasks are run. The next step would be to refactor the code that fetches/compiles tasks to run off the DAG instead of the merged AST. From here, there are a number of things that we can do...

## Future work

This work enables or makes easier a number of future improvements:

- Fetching/compiling tasks from the DAG instead of the merged AST.
- Removing merging entirely.
- Scoped variables (#1030).
- Lazy loading of variables (#1065, #1240).
- Taskfile tree visualisation.
- DAG for individual tasks?
  - This might be helpful for some of the other features mentioned above, or it might just be useful for illustration purposes. This was briefly discussed in #1234.


andreynering (approved):
I know it look a lot of time. Thanks for your perseverance! This is going to be a great improvement. :clap: :clap: :clap:

marco-m:
@pd93 kudos, great work!

ReillyBrogan:
Is #852 resolved with this?

pd93:
> Is #852 resolved with this?

@ReillyBrogan Unfortunately not. I've responded in #852 to keep everything together.

PR #1584, opened 2024-04-09
fix: fix bug for files with special chars `&()`

### LINKED ISSUE #1551: Having `&` in Directory name causign problems.
- Task version: 3.35.1
- Operating system: darwin 13.6.5 (m1)
- Experiments enabled: no idea.

Running task in watch mode with next configuration
```yaml
# https://taskfile.dev

version: '3'

vars:
  GREETING: Hello, World!

tasks:
  default:
    sources:
    - ./**/*.*
    - ./*.* 
    method: timestamp
    cmds:
      - ./update_me_please
    silent: true
```

Causes it to fail with proximatly next message

```
/Path/Teka/Father&Sone/first_path.txt: 1:64: & is not a valid word
```
So having "&" in folder is a problem...

related - https://github.com/go-task/task/discussions/1069

andreynering:
Fixes #1551

### DIFF
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -103,6 +103,9 @@ func IsExitError(err error) bool {
 func Expand(s string) (string, error) {
 	s = filepath.ToSlash(s)
 	s = strings.ReplaceAll(s, " ", `\ `)
+	s = strings.ReplaceAll(s, "&", `\&`)
+	s = strings.ReplaceAll(s, "(", `\(`)
+	s = strings.ReplaceAll(s, ")", `\)`)
 	fields, err := shell.Fields(s, nil)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


internal/execext/exec.go
```diff
@@ -103,6 +103,9 @@ func IsExitError(err error) bool {
 func Expand(s string) (string, error) {
 	s = filepath.ToSlash(s)
 	s = strings.ReplaceAll(s, " ", `\ `)
+	s = strings.ReplaceAll(s, "&", `\&`)
+	s = strings.ReplaceAll(s, "(", `\(`)
+	s = strings.ReplaceAll(s, ")", `\)`)
```
vergenzt, line 108: Should this not do something along the lines of "shell quote the string"? Instead of escaping individual characters one at a time?

kasadaamos: yeah - something like https://pkg.go.dev/github.com/frioux/shellquote@v0.0.2 seems more complete

PR #1587, opened 2024-04-09
fix: crowdin progress

pd93:
Fixes the percentages in the locale dropdown on the website.


andreynering:
It looks like the build failed. It may not be related to this change, though. Perhaps try a rebase?

https://app.netlify.com/sites/taskfile/deploys/661d2bcac480240078f7fc4e

andreynering:
@pd93 This change breaks the local server/build when `CROWDIN_PERSONAL_TOKEN` is not set. Ideally contributors should be able to compile the server as well, so we need a fallback.

Anyway, translations are simply not working at all, and I decided to disable it, at least temporarily: #1617.

pd93:
> Anyway, translations are simply not working at all, and I decided to disable it, at least temporarily

@andreynering No problem. I'll mark this as a draft for now and it can be revisited once translations are working again. I'll try to find some time to look into those issues too.

pd93:
We can keep the branch around, but there is no need to keep this PR open for now.

PR #1603, opened 2024-04-18
feat: add alias `q` for template function `shellQuote`

### LINKED ISSUE #1601: Add `q` alias for `shellQuote` template function?
The project I'm working on right now works with a lot of files with special characters in their names. I'm passing these filenames around as variables and referring to them a lot in shell commands.

I just realized I can use the `shellQuote` function to quote things... but unfortunately that makes my commands a lot longer since I'm referring to 2-3 filename variables in each command (having to write temporary files in a pipeline into a directory containing special chars), so I'm having to repeat `{{shellQuote .VAR}}` a lot.

It'd be SOOOO nice to be able to just do `{{q .VAR}}`! Without (IMO) really any loss of clarity. Since the primary / only(?) way of actually _operating_ on the contents of variables is via shell commands, it seems appropriate to me that the `shellQuote` function should have such a short alias.

Thoughts?

vergenzt:
Resolves #1601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   @pd93).
 - Fix error when a file or directory in the project contained a special char
   like `&`, `(` or `)` (#1551, #1584 by @andreynering).
+- Added alias `q` for template function `shellQuote` (#1601 by @vergenzt)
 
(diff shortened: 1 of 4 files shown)


andreynering:
Thanks @vergenzt!

PR #1607, opened 2024-04-21
feat: add support for multiple includes on a graph edge

### LINKED ISSUE #1606: Cannot include multiple time the same Taskfile with differents namespaces
Hi,

Following this [comment](https://github.com/go-task/task/issues/1588#issuecomment-2060054471), I've opened an issue to track this.

I am using this pattern a lot, in almost in every project (around 20). The idea is to have on Taskfile per language and standardise all Tasks for all projects.

Here the repo to reproduce : https://github.com/vmaerten/taskfile-demo
You can run `go run ./cmd/task --dir ./taskfile-demo deploy/dev -f   `

I have this error : 
```
❯ go run ./cmd/task --dir ../../perso/taskfile-demo deploy/dev                                                                                                                                                                                                                                                      20:36:48
panic: interface conversion: interface {} is ast.Include, not *ast.Include

goroutine 7 [running]:
github.com/go-task/task/v3/taskfile.(*Reader).include.func1.1()
	/home/valentin/dev/fork/task/taskfile/reader.go:152 +0x8a8
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/home/valentin/go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78 +0x56
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
	/home/valentin/go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:75 +0x96
exit status 2


```

I've given a talk at devoxxfr last thurday presenting Task (using this [repo](https://github.com/vmaerten/taskfile-demo)), and when discussing with attendees, multiple people have the same use case.
I think @davinkevin has the same use case.

The next release will most likely break some setup. Is it possible to get DAG working with this setup ?

- Task version: using last commit on main ([d01b3c8979a7245b04dbc80a2b7cb14885714368](https://github.com/go-task/task/commit/d01b3c8979a7245b04dbc80a2b7cb14885714368))
- Operating system: Linux - Fedora
- Experiments enabled: Remote

pd93:
Fixes #1606

Allows someone to create multiple includes to/from the same Taskfiles. This was restricted when we moved to the DAG reader implementation. The DAG library we use does not support adding multiple edges between nodes. However, the problem is easily solved by changing the edge data from `*ast.Include` to `[]*ast.Include`. This also allows us to easily calculate the weight (number of connections) of any edge by calculating the length of the slice.

I've also undone the changes to some of the tests that used to use multiple includes that were changed when the DAG reader was implemented.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -19,7 +19,7 @@ const (
 	CodeTaskfileCacheNotFound
 	CodeTaskfileVersionCheckError
 	CodeTaskfileNetworkTimeout
-	CodeTaskfileDuplicateInclude
+	_ // CodeTaskfileDuplicateInclude
 	CodeTaskfileCycle
 )
 
--- a/errors/errors_taskfile.go
+++ b/errors/errors_taskfile.go
@@ -3,7 +3,6 @@ package errors
 import (
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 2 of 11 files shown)


errors/errors.go
```diff
@@ -19,7 +19,7 @@ const (
 	CodeTaskfileCacheNotFound
 	CodeTaskfileVersionCheckError
 	CodeTaskfileNetworkTimeout
-	CodeTaskfileDuplicateInclude
+	_ // CodeTaskfileDuplicateInclude
```
andreynering, line 22: We could remove this unused line.

andreynering (approved):

errors/errors.go
pd93: Because of how `iota` works, removing this line would change the number assigned to `CodeTaskfileCycle`. It's unlikely that anyone is checking against this specific code in scripts, but if they are then removing this line would break it. I feel like its safer to just deprecate this error. We can maybe assign a different error to it in the future.

andreynering: Got it. I believe we can just keep it for now 👍

PR #1610, opened 2024-04-22
feat(remote): replace env variable in include remote URL

vmaerten:
I've added the support of env variable in include remote. All env variable presents in the URL (${ENV}) will be replaced.

Related to : https://github.com/go-task/task/discussions/1420#discussioncomment-7752658

### DIFF
--- a/website/docs/experiments/remote_taskfiles.mdx
+++ b/website/docs/experiments/remote_taskfiles.mdx
@@ -48,13 +48,13 @@ tasks:
 and you run `task my-remote-namespace:hello`, it will print the text: "Hello
 from the remote Taskfile!" to your console.
 
-You can also reference environment variable in your URL to have authentication, for example :
+URL is processed by the templating system, so you can reference environment variable in your URL to have authentication, for example :
 
 ```yaml
 version: '3'
 
 includes:
-  my-remote-namespace: https://${TOKEN}@raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
+  my-remote-namespace: https://{{.TOKEN}}@raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
 ```
 
 `TOKEN=my-token task my-remote-namespace:hello` will be resolved by Task to `https://my-token@raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml`
--- a/website/docs/experiments/remote_taskfiles.mdx
+++ b/website/docs/experiments/remote_taskfiles.mdx
@@ -48,6 +48,17 @@ tasks:
 and you run `task my-remote-namespace:hello`, it will print the text: "Hello
 from the remote Taskfile!" to your console.
 
+You can also reference environment variable in your URL to have authentication, for example :
+
+```yaml
+version: '3'
+
+includes:
+  my-remote-namespace: https://${TOKEN}@raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
+```
+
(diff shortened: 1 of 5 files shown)


website/docs/experiments/remote_taskfiles.mdx
```diff
@@ -48,6 +48,17 @@ tasks:
 and you run `task my-remote-namespace:hello`, it will print the text: "Hello
 from the remote Taskfile!" to your console.
 
+You can also reference environment variable in your URL to have authentication, for example :
+
+```yaml
+version: '3'
+
+includes:
+  my-remote-namespace: https://${TOKEN}@raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
```
pd93, line 57: I'm not a fan of adding a second syntax for templating variables. Particularly one that only applies to remote nodes. Since `includes` already gets processed through the templating system, we just need to add environment variables to the set of variables that are given to the templater and everything should just work. Something along these lines:

```diff
diff --git a/taskfile/reader.go b/taskfile/reader.go
index e67bb11f..1bf65598 100644
--- a/taskfile/reader.go
+++ b/taskfile/reader.go
@@ -11,6 +11,7 @@ import (
        "gopkg.in/yaml.v3"
 
        "github.com/go-task/task/v3/errors"
+       "github.com/go-task/task/v3/internal/compiler"
        "github.com/go-task/task/v3/internal/filepathext"
        "github.com/go-task/task/v3/internal/logger"
        "github.com/go-task/task/v3/internal/templater"
@@ -97,9 +98,18 @@ func (r *Reader) include(node Node) error {
 
        // Loop over each included taskfile
        _ = vertex.Taskfile.Includes.Range(func(namespace string, include *ast.Include) error {
+               vars := compiler.GetEnviron()
+               err = vertex.Taskfile.Vars.Range(func(k string, v ast.Var) error {
+                       vars.Set(k, ast.Var{Value: v})
+                       return nil
+               })
+               if err != nil {
+                       return err
+               }
+
                // Start a goroutine to process each included Taskfile
                g.Go(func() error {
-                       cache := &templater.Cache{Vars: vertex.Taskfile.Vars}
+                       cache := &templater.Cache{Vars: vars}
                        include = &ast.Include{
                                Namespace:      include.Namespace,
                                Taskfile:       templater.Replace(include.Taskfile, cache),
```

^^^ very untested code...

vmaerten: Thanks a lot for your review @pd93 ! Indeed you're right, I didn't know the "include" was processed by the templating system.

I used `vars.Merge` instead of doing it manually. I've also added a test in `includes_interpolation` part

pd93: Good shout using `vars.Merge` and thanks for adding tests. Other than the linting issues, I'm happy with this now. Please make sure you format the files using [`gofumpt`](https://github.com/mvdan/gofumpt). If using VSCode, you can do this by putting the following in the `.vscode/settings.json` file (which is gitignored):

```json
    "go.useLanguageServer": true,
    "gopls": {
        "formatting.local": "github.com/go-task",
        "formatting.gofumpt": true
    },
```

vmaerten: Sorry I forgot the lint!
I've run `task lint:fix`
It should be good (I've run the same command as in the CI to verify, and it's OK)

pd93 (approved):
LGTM. Thanks for this! 👍

PR #1618, opened 2024-04-24
feat: warn about move from any variables to map variables

pd93:
Prints a warning when `TASK_X_ANY_VARIABLES` is set informing the user that it is no longer required. It will pointer users who require map variables to the new `TASX_X_MAP_VARIABLES` experiment instead.


andreynering (approved):
👏

PR #1624, opened 2024-04-26
feat: task executable variable

### LINKED ISSUE #1616: Add TASK_EXECUTABLE_PATH variable
Following https://discord.com/channels/974121106208354339/1210203447979155518 

Adding the Taskfile binary executable absolute path could be beneficial in cases where the default `task` command is either not in the $PATH or has a different name altogether.

pd93:
Fixes #1616

I've gone with `TASK_EXE` instead of `TASK_EXECUTABLE` as suggested simply because its a bit shorter. Thoughts welcome.

### DIFF
--- a/internal/compiler/compiler.go
+++ b/internal/compiler/compiler.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


andreynering (approved):
Nice!

I did a quick test on Windows and noticed it requires `filepath.ToSlash` to work properly, so I updated the branch with this fix.

PR #1628, opened 2024-04-29
Fix reference to GOOS and GOARCH link

Pistonight:
## Issue
The link for GOOS and GOARCH values in the documentation are broken. The golang/go repo uses `master` as the default branch

## Fix
Change `main` to `master`

## Validation
`task website` and verified link now works and points to https://github.com/golang/go/blob/master/src/go/build/syslist.go

## Note
I wasn't sure if the `versioned_docs` need to be updated. The local build seems to be referencing that, so I updated it as well


andreynering:
Thanks @Pistonight!

PR #1633, opened 2024-05-05
feat: experiment taskfile envs take precedence over os envs

vmaerten:
Linked to : 
- https://github.com/go-task/task/issues/1038

I am totally open to rework this, including add new template function (as discussed in discord)

### DIFF
--- a/website/docs/experiments/env_precedence.mdx
+++ b/website/docs/experiments/env_precedence.mdx
@@ -1,0 +1,65 @@
+---
+draft: false # Hide in production
+slug: '/experiments/env-precedence'
+---
+
+# Env Precedence (#1038)
+
+:::caution
+
+All experimental features are subject to breaking changes and/or removal _at any
+time_. We strongly recommend that you do not use these features in a production
+environment. They are intended for testing and feedback only.
+
+:::
+
+:::warning
+
+This experiment breaks the following functionality:
+
+- environment variable will take precedence over OS environment variables
+
+:::
+
+:::info
+
+To enable this experiment, set the environment variable: `TASK_X_ENV_PRECEDENCE=1`.
+Check out [our guide to enabling experiments ][enabling-experiments] for more
+information.
+
+:::
+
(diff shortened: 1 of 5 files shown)


vmaerten:
I managed to understand why, it's because the Init() from experiment is called before my SetEnv. I managed to make this test pass by instantiating an experiment in the test.

It's ready to review

mgbowman:
@andreynering @pd93 sorry to nag but my org could _really_ use this functionality right now. Any chance one of you guys could review / comment on this?

website/docs/experiments/env_precedence.mdx
```diff
@@ -0,0 +1,65 @@
...
+tasks:
+  default:
+    env:
+      KEY: 'other'
+    cmds:
+      - echo "$KEY"
+```
+Running `KEY=some task` before this experiment, the output would be `some`, but after this experiment, the output would be `other`.
+
+If you still want to get the OS variable, you can use the template function env like follow : `{{env.OS_VAR}}`.
```
andreynering, line 48: I think it would be interesting to suggest the use of ` default` as a way of allowing an override while falling back to a given env:

```YAML
MY_ENV: '{{.MY_ENV | default "fallback"}}'
```

Also, looks like there's a typo on this line:

```suggestion
If you still want to get the OS variable, you can use the template function env like follow : `{{env "OS_VAR"}}`.
```

andreynering (approved):
Looks good to me.

This has conflicts to be fixed, and I'll also ask a review from @pd93 to get an extra opinion.

vmaerten:
Thanks for the typo @andreynering 
I've rebased and fixed conflict
I've also added a small paragraph about default

pd93 (approved):
No further comments from me for now. Let's get some feedback from some users and we can make decisions from there.

mgbowman:
Awesome 🚀

When can we expect an official release containing this new experiment?

vmaerten:
Thanks for your interest in this feature.

We cannot provide ETA (you can find more information in our [doc](https://taskfile.dev/faq/#when-will-feature-be-released--etas))

PR #1634, opened 2024-05-05
fix(remote): do not display prompt if it's empty

vmaerten:
I think there is a bug introduced during the [PR introducing DAG](https://github.com/go-task/task/pull/1563).

The prompt is still displayed when it's empty.

cf : 
image


pd93:
So confused how this has crept back in... I fixed this in Janurary! 😆 (#1487). Oh well. Thanks for the fix 👍

Issue #1317, opened 2023-08-29
Remote Taskfiles experiment


pd93:
> [!WARNING]
> All experimental features are subject to breaking changes and/or removal at any time. We strongly recommend that you do not use these features in a production environment. They are intended for testing and feedback only.

> [!NOTE]
> You can view the [Remote Taskfiles experiment documentation](https://taskfile.dev/experiments/remote-taskfiles) on our website, including instructions on how to enable/disable it.

## Context

This experiment attempts to solve the problems originally described by #770.

Currently, Task only supports running Taskfiles located on the local filesystem. This makes it difficult to reuse Taskfiles across various repositories and systems without making copies of a file. It would be useful to have a way to run a Taskfile located on a remote system or run a local Taskfile that includes one on a remote system.

## Proposal

This experiment would allow users to run a Taskfiles located elsewhere, such as via HTTP or in a remote Git repository. These remote Taskfiles can either be run directly from the CLI or included by another Taskfile.

We're looking to gather feedback on the current draft, so please feel free to leave your thoughts/comments here.

Decision log / TODOs:

- [x] Refactor nodes as an interface #1316
- [x] HTTP nodes #1152
- [x] Decide between [go-getter] or a native implementation
  - No go-getter for now. We can re-evaluate later
- [x] How to handle security (remote files that change etc)
  - See https://taskfile.dev/experiments/remote-taskfiles/#security
- [x] Caching (`--download`/`--offline` flags or similar)
  - See https://taskfile.dev/experiments/remote-taskfiles/#caching--running-offline
- [x] Implement `--yes` flag on prompts #1344
- [x] Decide if we should use remote/local files first if both are available #1345
- [x] Consider storing checksums/cached files with a more user-friendly filename #1636
- [x] Command to clear the remote cache #1639
- [x] Check that `local` -> `remote` -> `remote:local` includes work correctly #1347
- [x] Redact credentials in remote URIs #2100
- [x] Checksum Pinning #2222
- [x] Support for self-signed certificates #2242 
- [x] Auto update TTL for cached files #1402
- [x] Add remote flags as config in Taskfile #1403
- [x] Decide how special variables should look for remote files #2267
- [x] Running root remote Taskfiles #1347
  - e.g. `task test --taskfile https://raw.githubusercontent.com/go-task/task/main/Taskfile.yml`
- [x] Git Nodes
  - e.g. `git@example.com/foo/bar?ref=main//Taskfile.yml` #1652

[...]

PR #1636, opened 2024-05-08
feat(remote): prefix checksums/cached files with the filename

vmaerten:
I've introduce a new concept in node, the `filename`. This will allow us to store remote checksums/cached files with a more user-friendly filename

Linked to : 
- https://github.com/go-task/task/issues/1317

### DIFF
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -50,9 +50,9 @@ func (c *Cache) key(node Node) string {
 }
 
 func (c *Cache) cacheFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.yaml", c.key(node)))
+	return filepath.Join(c.dir, fmt.Sprintf("%s.%s.yaml", node.Filename(), c.key(node)))
 }
 
 func (c *Cache) checksumFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.checksum", c.key(node)))
+	return filepath.Join(c.dir, fmt.Sprintf("%s.%s.checksum", node.Filename(), c.key(node)))
 }
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -50,9 +50,19 @@ func (c *Cache) key(node Node) string {
 }
 
 func (c *Cache) cacheFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.%s.yaml", node.Filename(), c.key(node)))
+	return c.filePath(node, "yaml")
 }
 
 func (c *Cache) checksumFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.%s.checksum", node.Filename(), c.key(node)))
+	return c.filePath(node, "checksum")
+}
+
+func (c *Cache) filePath(node Node, suffix string) string {
+	lastDir, filename := node.FilenameAndLastDir()
+	prefix := filename
+	// Means it's not "", nor "." nor "/", so it's a valid directory
+	if len(lastDir) > 1 {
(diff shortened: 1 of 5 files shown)


pd93:
Thanks for working on this 🚀 Just a thought on the requirements for this. This was originally raised as a feature to make it easier to distinguish cached files from one another. Suppose we have a setup like the following:

```yaml
version: 3

includes:
  a: https://my.remote.com/a/Taskfile.yml
  b: https://my.remote.com/b/Taskfile.yml

...
```

This implementation will create two cached files:

- `Taskfile.yml.<hash(a)>.yaml`
- `Taskfile.yml.<hash(b)>.yaml`

These still aren't very distinguishable. Not sure if we want to distinguish them further or if this is "good enough". No strong opinions here - just food for thought.

vmaerten:
I totally agree with you! I did not think about this configuration.

The configuration I had in mind was : one folder with multiple Taskfile in it, but we could support multiple folders with one taskfile in it.

First solution I had in mind was to put all the path with `-` as separator in the file, in your example it would be `a-Taskfile.yml.<hash(a)>.yaml` 
But the "big" path, I would not fit well, for example (like in our [docs](https://taskfile.dev/experiments/remote-taskfiles/)) some people (including me) will host the remote Taskfile in Github, leading to huge path (`/my-org/my-repo/main/Taskfile.yml`).
So this solution is not optimal.


So, my proposition here is to include only the last directory in the filename, following this pattern `<lastDir>-<filename>.<hash>.yaml`
```yml
version: 3

includes:
  a: https://my.remote.com/a/Taskfile.yml
  b: https://my.remote.com/b/Taskfile.yml
  long: https://raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
  long_with_dir: https://raw.githubusercontent.com/my-org/my-repo/main/directory/Taskfile.yml
  no_dir: https://my.remote.com/Taskfile.yml
...
```
Would be resolved to : 

- `a-Taskfile.yml.<hash>.yaml`
- `b-Taskfile.yml.<hash>.yaml`
- `main-Taskfile.yml.<hash>.yaml`
- `directory-Taskfile.yml.<hash>.yaml`
- `Taskfile.yml.<hash>.yaml`


This way we could support almost all the cases.

What is your opinion about it ? Add the last dir in the filename or keep it simple with only the filename ? 
 I don't know if people include multiple taskfiles from different origins.

pd93:
> First solution I had in mind was to put all the path with `-` as separator in the file, in your example it would be `a-Taskfile.yml.<hash(a)>.yaml` But the "big" path, I would not fit well

@vmaerten Agreed, this was my first thought too, but I think having very long file names defeats the point of making this more readable. It can also cause problems in Windows where there are path length restrictions on some versions.

> So, my proposition here is to include only the last directory in the filename, following this pattern `<lastDir>-<filename>.<hash>.yaml`

I think this is probably good enough for now. This is still an experiment, so we can always revisit it later if we have a better idea without worrying about breaking things.

taskfile/cache.go
```diff
@@ -50,9 +50,19 @@ func (c *Cache) key(node Node) string {
...
 
 func (c *Cache) cacheFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.yaml", c.key(node)))
+	return c.filePath(node, "yaml")
 }
 
 func (c *Cache) checksumFilePath(node Node) string {
-	return filepath.Join(c.dir, fmt.Sprintf("%s.checksum", c.key(node)))
+	return c.filePath(node, "checksum")
+}
```
vmaerten, line 58: extract a method here because it was almost the same code

vmaerten:
We are on the same page so :)
I've updated the code

pd93 (approved):

PR #1639, opened 2024-05-09
feat(remote): add a command to clear the cache

vmaerten:
I've added a command to clear the cache for remote files.

I've made the choice to disallow using download and clear-cache at the same time because clear-cache should be used as standalone.

We could allow using both flags at the same time, but we need to define the bahavior : 
- Clear cache override download (so we clear the cache but not download)
- Download override clear cache (so we download but not clear cache)

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -146,7 +146,7 @@ func run() error {
 	}
 
 	if flags.ClearCache {
-		cache, err := taskfile.NewCache(e.TempDir)
+		cache, err := taskfile.NewCache(e.TempDir.Remote)
 		if err != nil {
 			return err
 		}
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -150,8 +150,7 @@ func run() error {
 		if err != nil {
 			return err
 		}
-		err = cache.Clear()
-		return err
+		return cache.Clear()
 	}
 
 	if (listOptions.ShouldListTasks()) && flags.Silent {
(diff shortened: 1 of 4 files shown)


cmd/task/task.go
```diff
@@ -145,6 +144,15 @@ func run() error {
 		return nil
 	}
 
+	if flags.ClearCache {
+		cache, err := taskfile.NewCache(e.TempDir)
```
andreynering, line 148: This will conflict with #1661, so we'll need to adapt it here or there, whether PR is merged last.

cmd/task/task.go
```diff
@@ -145,6 +144,15 @@ func run() error {
...
+		if err != nil {
+			return err
+		}
+		err = cache.Clear()
+		return err
```
andreynering, line 153:
```suggestion
		return cache.Clear()
```

andreynering (approved):

cmd/task/task.go
vmaerten: Yes I know, to be honest I would prefer to merge #1661 before then I'll rebase this one.
If it's ok for you

cmd/task/task.go
vmaerten: Good catch thanks

cmd/task/task.go
vmaerten: Rebase done

pd93 (approved):

Issue #1640, opened 2024-05-09
Regression - env var yaml integer no longer exported to enviroment


ameergituser:
**input**
```
version: '3'

tasks:
  default:
    env:
      CC: x86_64-w64-mingw32-gcc
      CGO_ENABLED: 1
      GOOS: windows
      GOARCH: amd64
    cmds:
      - echo CC=$CC CGO_ENABLED=$CGO_ENABLED GOOS=$GOOS GOARCH=$GOARCH

```
**output: Task version: v3.36.0**
```
CC=x86_64-w64-mingw32-gcc CGO_ENABLED=1 GOOS=windows GOARCH=amd64
```

**output: Task version: v3.37.0**
```
CC=x86_64-w64-mingw32-gcc CGO_ENABLED= GOOS=windows GOARCH=amd64
```

**Comment:**
Note `CGO_ENABLED` env var was not set for v3.37.0
Converting the integer var into a string works around this issue.

- Operating system: Linux dev 5.15.0-106-generic 116-Ubuntu 
- Experiments enabled: No

andreynering:
Hi @ameergituser,

This was fixed by @vmaerten on #1641 and I'll release a patch in minutes.

PR #1641, opened 2024-05-09
fix: handle int and float env variable by converting them to string

vmaerten:
Hey @pd93,

I've tried a solution to fix : 
- https://github.com/go-task/task/issues/1640

It's because since the any vars, some vars are detected as int or float and do not get added to the env map.
So we can let them passthrough and convert them with `%v`

I've also added a test to convert this case

### DIFF
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -13,19 +13,25 @@ func Get(t *ast.Task) []string {
 	}
 
 	environ := os.Environ()
-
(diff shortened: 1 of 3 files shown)


andreynering:
Thanks @vmaerten!

I did an additional fix to make booleans work as well: bfc9d7847dc2a635a46230ec3f94998f2caa37af

v3.37.1 will be released soon.

PR #1650, opened 2024-05-10
chore: update actions version because node 16 is deprecated

vmaerten:
Node 16 is deprecated so I've bump checkout / setup-go and setup-arduino to the lastest version


andreynering:
Awesome, thanks @vmaerten!

PR #1652, opened 2024-05-11
feat(remote): support include git remote

vmaerten:
I've work on add the support of Git in the remote experimentation.
I've used the go-git package, which is Git implementation in pure go.
I've also used git-urls to help me to parse all URLs.

Some informations : 
- I did not handle the custom timeout for now
- We support SCP / SSH / HTTPS URL
- It seems that SSH takes more time than HTTPS
- Git remote entrypoint (--taskfile) works
- include local -> remote -> remote:local works

I'll work next on perf optimisation. I've in mind to mutualise git clone to avoid cloning multiple times the same repo.

I've added tests. I'll add test of node_http in another PR. 

EDIT : The link with to the [preview](https://deploy-preview-1652--taskfile.netlify.app/next/experiments/remote-taskfiles/#git-nodes)


vmaerten:
I've tested it and a friend also tested it. I am open to gather feedback on it :)

pd93:
@vmaerten Thanks for working on this! This is a very popular request. I will definitely give this a go in the next couple of days. Finally some nice weather in the UK, so I'll be out enjoying the sun as much as possible this weekend! 🌞

It might be worth putting a message on the remote experiment issue so that people who are subscribed over there can try it too. I have updated the OP there to link back here.

vmaerten:
I really enjoy working on Task. Let me know if you have questions regarding my implementation
Sure, enjoy the sun, I did the same this weekend in France! :slightly_smiling_face: 

Good idea! I've just done it :slightly_smiling_face:

vmaerten:
@blackjid 's comment : 

> That is awesome @vmaerten thank for the work...
>
> Why did you choose to set the filename and the git reference in the same querystring in the url? Why not just having a 
 different querystring for the filename? https://github.com/foo/bar.git?ref=main&filename=directory/Taskfile.yml
>
> Or if you don't want the extra query string, a more readable form I think could be to add the path to the file just after the  .git https://github.com/foo/bar.git/directory/Taskfile.yml?ref=main maybe with a double slash // like terraform and  others does.. https://github.com/foo/bar.git//directory/Taskfile.yml?ref=main
> 
> Sorry if this is not the place or time to give feedback. Maybe discord?

First of all, thanks for the feedback.
To be honest, I did not find a consensus for this kind of URL in the community, so I took the example URL  in the issue ([here](https://github.com/go-task/task/issues/1317#issue-1871470772))

Following your propostions, I like the one with the query param representing the filename. In the other hand, if the community already knows the "terraform" format, it could be good to adopt it.

Let's discuss with @pd93 and @andreynering

adlnc:
Hi, 

I find this work awesome ! Just wanted to mention that I've been able to run this on my side using http over public repo, private repo using token auth and ssh. All behaved as expected.

This, added to the remote file feature will definitely help saving a tremendous amount of time sharing and managing taskfiles. I'm really excited to have this available in a few. 

Thanks for this work.

vmaerten:
Thanks for checking this! 
We will definitely move forward on this!

jylenhofdecat:
Any update ?

taskfile/node.go
```diff
@@ -47,11 +49,15 @@ func NewNode(
...
+	}
+	if isGit {
+		node, err = NewGitNode(entrypoint, dir, insecure, opts...)
+	} else if getScheme(entrypoint) == "http" || getScheme(entrypoint) == "https" {
 		node, err = NewHTTPNode(l, entrypoint, dir, insecure, timeout, opts...)
-	default:
-		// If no other scheme matches, we assume it's a file
+	} else {
 		node, err = NewFileNode(l, entrypoint, dir, opts...)
 	}
```
pd93, line 63: I preferred this as a switch tbh. Could we just update `getScheme` to return "git" if we detect a git URL?

pd93 (commented):
Generally pretty happy with this. Just one minor comment.

w.r.t @blackjid's comment:

There have been a lot of comments in the past about reusing the [go-getter](https://github.com/hashicorp/go-getter) syntax as it seems pretty popular in the Go community. We're not using go-getter right now, but using a compatible syntax would make it easier to transition to it in the future if we wanted to for any reason.

vmaerten:
I need to rebase due to recent merged PR
I will do it

taskfile/node.go
vmaerten: Good catch! I totally agree with you!

vmaerten:
@pd93 I've rebased to add the function for prefixing the cache (done here: https://github.com/go-task/task/pull/1636).
I've also taken your review into account.
For some reason, the CI is not triggered. :'(

taskfile/node_git_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert.Equal(t, "Taskfile.yml", node.path)
+	assert.Equal(t, "ssh://git@github.com/foo/bar.git?ref=main//Taskfile.yml", node.rawUrl)
+	assert.Equal(t, "ssh://git@github.com/foo/bar.git", node.URL.String())
+	entrypoint, err := node.ResolveEntrypoint("common.yml")
+	assert.NoError(t, err)
+	assert.Equal(t, "ssh://git@github.com/foo/bar.git?ref=main//common.yml", entrypoint)
+}
+
+func TestGitNode_sshWithDir(t *testing.T) {
+	node, err := NewGitNode("git@github.com:foo/bar.git?ref=main//directory/Taskfile.yml", "", false)
```
blackjid, line 22: > but using a compatible syntax would make it easier to transition to it in the future if we wanted to for any reason.

I did't thought about it, that would be a compelling reason to use that syntax.

I don't think this new change syntax is compatible...
```
git@github.com:foo/bar.git?ref=main//directory/Taskfile.yml
```
should be
```
git@github.com:foo/bar.git//directory/Taskfile.yml?ref=main
```

so `ref` is directly mapped to a git refeference, could be a branch or a tag for example.
id's say you'd need to split the url in `//` to get the git url at `0` and the filename at `1`

blackjid (commented):
thanks @vmaerten !

oliver-helix:
What am I doing wrong?

```sh
$ export TASK_X_REMOTE_TASKFILES=1
$ task --list --taskfile git@github.com/$ORGANIZATION/$REPO?ref=main//taskfile.yml
stat git@github.com/$ORGANIZATION/$REPO?ref=main//taskfile.yml: no such file or directory
```

The following code shows that the repository exists ...
```sh
$ git clone git@github.com:$ORGANIZATION/$REPO.git --branch main && cp $REPO/taskfile.yml .
```

vmaerten:
For now, you need to put the schema in front of the URL.
Either `ssh://git@github.com/foo/bar.git?ref=main//directory/Taskfile.yml` or `https://github.com/foo/bar.git?ref=main//directory/Taskfile.yml`
I'll impletement later the URL without the scheme

taskfile/node_git_test.go
blackjid: hey @vmaerten thanks for all the work on this!

just want to add that with this syntax, the ref query string should be optional, because, it will only reference to a branch or tag, and without it it would be the default branch.

vmaerten: Hey!
You're right! I'll work on this :)

vmaerten:
@blackjid & @pd93 Following this comment : https://github.com/go-task/task/pull/1652#discussion_r1659114619
I've changed the url to be go-getter compatible. I've also put the ref optional, default to the default branch

blackjid:
what do you think about this? https://github.com/go-task/task/pull/1774
clearly there is need to something...

vmaerten:
@blackjid I've merged this one to move forward on this feature

Let's talk about go-getter or not, in the remote issue or the PR

PR #1655, opened 2024-05-13
fix: run once in shared dependencies

### LINKED ISSUE #852: Make run:once work for dependencies shared amongst included taskfiles

pd93:
Fixes #852

This PR changes the hash we use when a task is marked as `run:once`. Previously we used the full name of the task. This is fine most of the time, but when a task in an included Taskfile is included multiple times, the name will contain a namespace that is different for each task. Using the example given in #852, the hashes produced are:

```
service-a:build
service-b:build
service-b:library:build #  <-- this...
service-a:library:build #  <-- is different to this
```

As a result `library:build` is run twice, even though it is marked as `run: once`.

In this PR, we are creating a hash that is a combination of the file location and the name of the task _without a namespace_. This means that the hash is unique for each task, but will be the same when a task is included in different namespaces.

```
/path/to/project/service-a/Taskfile.yml:build
/path/to/project/service-b/Taskfile.yml:build
/path/to/project/library/Taskfile.yml:library:build  <-- this...
/path/to/project/library/Taskfile.yml:library:build  <-- is the same as this
```

---

> [!NOTE]  
> As discussed below, we're no longer focussing on the second problem in this PR. We're open to further discussion in another issue for anyone coming across this in the future.

#852 mentions a secondary problem:

> We also noticed that if we run a "task build" in service-a and then service-b that the `.task/` checksum cache is not shared and so the library is built twice.

You can see this illustrated below:

```sh
❯ task build   
task: [service-a:library:build] echo "build library" # <-- first run - library builds
build library
task: [service-b:build] echo "build b" # <-- library is run:once so service-b doesn't build it again
task: [service-a:build] echo "build a"
build a
build b
```

```sh
❯ task build
task: Task "service-a:library:build" is up to date # <-- second run - library is up to date so doesn't build
task: Task "service-a:build" is up to date
task: Task "service-b:build" is up to date
```

```sh
❯ task build
task: [service-b:library:build] echo "build library" 
# ^ service-b goes first this time (because the order of deps is non-deterministic)
# but it doesn't share a hash with service-a so it builds the library again
build library
task: Task "service-b:build" is up to date
task: Task "service-a:build" is up to date
```

We need to determine the correct behaviour:

1. Working as intended - Tasks should always have independent caches as even shared tasks can have different outputs depending on the context. The fact that the library is built again sometimes is a side effect of the non-deterministic order of the tasks.
1. Always share a cache between shared tasks - This seems dangerous as variables that are passed into a task can change its output. This might lead to a task not running when it should.
1. Share a cache when `run: once` is specified - This still has the same problem as above. However, you could argue that the user has made a mistake by adding `run: once` when the result of a task could be different depending on the context.

If we choose 2 or 3, then what should the task hash be?

- `<location>:service-a:build` (probably not)
- `<location>:service-b:build` (also probably not)
- Some combination of the two
- Something else entirely?

I would appreciate any thoughts/feedback on these options - or other options entirely.

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -4,15 +4,12 @@ import (
 	"fmt"
 
 	"github.com/go-task/task/v3/internal/hash"
+	"github.com/go-task/task/v3/internal/slicesext"
 	"github.com/go-task/task/v3/taskfile/ast"
 )
 
 func (e *Executor) GetHash(t *ast.Task) (string, error) {
-	r := t.Run
-	if r == "" {
-		r = e.Taskfile.Run
-	}
-
+	r := slicesext.FirstNonZero(t.Run, e.Taskfile.Run)
 	var h hash.HashFunc
 	switch r {
 	case "always":
--- a/internal/hash/hash.go
+++ b/internal/hash/hash.go
@@ -15,7 +15,7 @@ func Empty(*ast.Task) (string, error) {
 }
 
 func Name(t *ast.Task) (string, error) {
-	return t.Task, nil
+	return fmt.Sprintf("%s:%s", t.Location.Taskfile, t.LocalName()), nil
 }
 
 func Hash(t *ast.Task) (string, error) {
--- a/internal/slicesext/slicesext.go
+++ b/internal/slicesext/slicesext.go
@@ -18,3 +18,13 @@ func UniqueJoin[T cmp.Ordered](ss ...[]T) []T {
 	slices.Sort(r)
 	return slices.Compact(r)
 }
+
+func FirstNonZero[T comparable](values ...T) T {
+	var zero T
+	for _, v := range values {
+		if v != zero {
+			return v
+		}
+	}
+	return zero
+}
--- a/task_test.go
+++ b/task_test.go
@@ -1664,6 +1664,26 @@ func TestRunOnlyRunsJobsHashOnce(t *testing.T) {
 	tt.Run(t)
 }
 
+func TestRunOnceSharedDeps(t *testing.T) {
+	const dir = "testdata/run_once_shared_deps"
+
+	var buff bytes.Buffer
+	e := task.Executor{
+		Dir:      dir,
+		Stdout:   &buff,
+		Stderr:   &buff,
+		ForceAll: true,
+	}
+	require.NoError(t, e.Setup())
+	require.NoError(t, e.Run(context.Background(), &ast.Call{Task: "build"}))
+
+	rx := regexp.MustCompile(`task: \[service-[a,b]:library:build\] echo "build library"`)
+	matches := rx.FindAllStringSubmatch(buff.String(), -1)
+	assert.Len(t, matches, 1)
+	assert.Contains(t, buff.String(), `task: [service-a:build] echo "build a"`)
+	assert.Contains(t, buff.String(), `task: [service-b:build] echo "build b"`)
+}
+
 func TestDeferredCmds(t *testing.T) {
 	const dir = "testdata/deferred"
 	var buff bytes.Buffer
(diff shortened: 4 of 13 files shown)


vmaerten:
In my option, the number 2 is really dangerous. I would prefer a task running twice even it's not needed, than a task not running at all when it should

it's a tough topic. I would go for the 1 or 3
I do not have strong opinion on this but I would go for the 1 because there is no risk that the task is not run (even it should be)

pd93:
> In my option, the number 2 is really dangerous. I would prefer a task running twice even it's not needed, than a task not running at all when it should

Agreed

> it's a tough topic. I would go for the 1 or 3 I do not have strong opinion on this but I would go for the 1 because there is no risk that the task is not run (even it should be)

I've given this more thought and I'm inclined to agree for now. I can see the potential upsides and downsides of option 3, but for now I think leaving this PR as-is and fixing the first issue is a safe bet. I'm still open to discussion about the best way to solve the second problem in another issue.

vmaerten (commented):
What do you think about adding a test for this usecase ? 

I manage to test it (to ensure it work as expected)

I've added it to the `run` testdata

Something like this : 

library.yml
```yml
version: 3

tasks:
  build:
    run: once
    cmds:
      - echo 'Building library...'

```

service.yml
```yml
version: 3

includes:
  library:
      taskfile: library.yml

tasks:
  build:
    deps: [library:build]
    cmds:
      - echo 'Building service {{.NAME}}...'

```

The root taskfile: 

```yml
includes:
  service-a:
    taskfile: service.yml
    vars:
      NAME: "service-a"
  service-b:
    taskfile: service.yml
    vars:
      NAME: "service-b"

tasks:
  build:
    silent: false
    cmds:
      - task: service-a:build
      - task: service-b:build
```

Then we need to verify that `Building library...` is printed only once.

I can add or you can add, or you can discard my comment. It's your call here :)

Thanks for the work :)

andreynering (approved):
Nice! 👏 

I agree with @vmaerten that adding a test before merging would be nice.

vmaerten (approved):

vmaerten (approved):

Issue #1637, opened 2024-05-08
align target hints by column


dmitry-mightydevops:
- Task version:
```sh
➜ task --version
Task version: v3.36.0 (h1:XVJ5hQ5hdzTAulHpAGzbUMUuYr9MUOEQFOFazI3hUsY=)
```
- Operating system:

```sh
➜ uname -a          
Linux dmitry-desktop 6.8.5-101.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Apr 11 19:59:26 UTC 2024 x86_64 GNU/Linux
```
- Experiments enabled:
N/A

### Problem

![image](https://github.com/go-task/task/assets/70822710/cae38f6b-4bf3-4daa-9124-6a6e176b0fff)

would like to align the hints on the right side by the same column and remove unnecessary empty space lines between targets listed in the output

dmitry-mightydevops:
here is the taskfile

```yaml
version: '3'

tasks:
  check-file-existence:
    desc: "Check if the specified file exists"
    vars:
      file: '{{.file}}'
    cmds:
      - test -f "{{.file}}" || (echo "The file {{.file}} does not exist" && exit 1)
    silent: true

  encrypt-string:
    desc: |
      Encrypt secret string with AWS KMS keys to be used in terraform, example: `task secrets:encrypt-string -- secret-text-here`
    vars:
      secret: '{{ index ($x:=splitArgs .CLI_ARGS) 0 }}'
    requires:
      vars: [ENV]
    cmds:
      - |
        echo :OK"
    silent: true

  decrypt-string:
    desc: |
      Decrypt string with AWS KMS keys to be used in terraform, example: `task secrets:encrypt-string -- encrypted-text-here`
    vars:
      secret: '{{ index ($x:=splitArgs .CLI_ARGS) 0 }}'
    requires:
      vars: [ENV]
    cmds:
      - |
        echo :OK"
    silent: true

  encrypt-file:
    desc: |
      Encrypt secret with AWS KMS keys to be used in terraform, example: `task secrets:encrypt-file -- apps/spiderkeeper/envs/prod.secrets.yaml`
    vars:
      file: '{{ index ($x:=splitArgs .CLI_ARGS) 0 }}'
    requires:
      vars: [ENV]
    cmds:
      - task: check-file-existence
        vars: {file: "{{.file}}"}
      - |
        echo :OK"
    silent: true

  decrypt-file:
    desc: |
[...]

Issue #1350, opened 2023-09-26
Text of multi line description not aligned when running `task --list`


JakeSummers:
_This feature request is really an exact duplicate of #133 but I don't have permission to reopen the issue_

It seems like 10% of my targets are complex and require multiple lines to explain what they do.  It would be awesome if multi-line descriptions could be supported.  

Current behaviour: 

Running `task --list`, for this file: 

```yaml
version: '3'

tasks:
  long-doc:
    desc: |
      Most tasks don't require a long message but
      10% of targets need a longer message.  It would
      be preferable if the alignment was more crisp.
    cmds:
      - echo 'Other conversations!'

  foo:
    desc: hi
    summary: Using short is best
    cmds:
      - echo "hi"
```

Results in this: 

```
$ task --list
task: Available tasks for this project:
* foo:            hi
* long-doc:       Most tasks don't require a long message but
10% of targets need a longer message.  It would
be preferable if the alignment was more crisp.
```

It would be much more readable if it did this: 

```
$ task --list
task: Available tasks for this project:
* foo:            hi
* long-doc:       Most tasks don't require a long message but
                  10% of targets need a longer message.  It would
                  be preferable if the alignment was more crisp.
```

I know that the summary feature #180 has been added since #133.  This is a nice feature but in my view, it plays a different role.  :)

PR #1656, opened 2024-05-13
fix: list-task with multiline desc

vmaerten:
Linked to : 
- https://github.com/go-task/task/issues/1637

I have multiple solutions in mind
- First solution : remove all \n so we display one line
image

- Second solution : Display only the first line. We could considere for longer desc, we should use summary and not desc
image

- Third solution: Display "as it" with multiline
image

I would considere that desc should be short and summary for long description. So, I would go for the first one.


pd93:
Also related:

- #1350

I generally agree that descriptions should be short and kept to a single line. However, I suspect we will have users complain about options 1 and 2 as it restricts their options. A lot of people use the `--list` functionality as their docs and I can understand that restricting this to 1 line is annoying.

I've had a ton of issues using `text/tabwriter` in the past and gave up on trying to make it work. (for example `task --experiements` is still not aligned correctly). If you're able to find a good generalised solution for alignment, I'd be very interested.

vmaerten:
@pd93 I agree, I was not aware of this issue.

I've changed to the solution 3. I've also fixed the issue with experiments display. Now, all experiments are aligned in a table.
image

It seems tabwrite does not work well with colors (cf : https://stackoverflow.com/questions/35398497/how-do-i-get-colors-to-work-with-golang-tabwriter / https://github.com/fatih/color/issues/175 / https://github.com/golang/go/issues/12073)

That why, it can be trick, as we display multiple colors in the same table.

It may be worthy to take a look at https://github.com/juju/ansiterm/
I'll take a look if I can find something that can handle both tabs and colors

vmaerten:
I've push a last version, using https://github.com/Ladicle/tabwriter as a drop in replacement of tabwriter but supporting color

It's way easier to use (no more headache debugging the tabwrite, yes I have a headache now :joy: ) than the text/tabwriter

testdata/desc/Taskfile.yml
```diff
@@ -0,0 +1,17 @@
+version: 3
+
+
```
andreynering, line 3: Let's remove these duplicated empty lines.

```suggestion

```

testdata/desc/Taskfile.yml
```diff
@@ -0,0 +1,17 @@
...
+tasks:
+  build:
+    aliases:
+      - b
+    desc: |
+      Multi-line escription with alias which is super long long long long long long
+      another line
+      third line long long long long long long long long
+
+
```
andreynering, line 13:
```suggestion

```

andreynering (commented):
I think I prefer option 1, forcing descriptions to be a single line. We have `summary` for longer documentation.

Thoughts?

vmaerten:
I totally agree with you @andreynering. I did this at the first time before the @pd93's comment
We could write in the doc that desc will be printed in one line. If you want multiple lines you need to use summary

pd93:
I'm happy to go with option 1 if you both agree. I do think we will get some complains about this though. I generally prefer to encourage users to doing something a specific way

vmaerten:
Option 1 has been chosen.
Here is the result:

image

pd93 (approved):

PR #1657, opened 2024-05-13
feat: improved prompt and release task

pd93:
This PR does a few things - mostly as I found a few issues while creating an improved release task:

- Added missing `additionalProperties: false` keys in `schema.json`
- Moved prompt check after preconditions and fingerprinting
  - Annoying to be prompted if the task will be stopped anyway
- Stop prompt from showing if `--dry` flag is enabled
  - Again, no value to having a prompt if the task won't execute anyway
- Removed `\n` from prompt message so that the response is on the same line instead of by itself
  - Just a visual thing really. Looks a bit nicer
- **A significantly improved release tool**
  - Uses wildcards (e.g. `task release:patch` instead of `task release -- patch`)
  - Lots of docs to help in the `--summary` and printed after the task completes
  - Prompt and preconditions to make sure it's not run accidentally.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -123,10 +123,53 @@ tasks:
     cmds:
       - go install github.com/goreleaser/goreleaser@latest
 
-  release:
+  release:*:
     desc: Prepare the project for a new release
+    summary: |
+      This task will do the following:
+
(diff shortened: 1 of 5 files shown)


andreynering (approved):
Great improvements! 👏 👏 👏

PR #1661, opened 2024-05-16
feat(remote): global tempDir when the path is absolute

vmaerten:
This PR allow to have a global temp directory when using remote taskfile.

Considering a setup with multiple different projects, all using the same remote taskfile. if /  when the remote taskfile changes, we need to approve the prompt for all projects.
At the beginning, I thought changing the TASK_X_TEMP_DIR to `~/.task`, but I've seen that it's suffixed with the dir so I won't work.

The fingerprint dir should stay scoped

This PR changes the behavior only for remote. If the path is absolute or start with `~` then we do not suffix it.

For the implementation, I've chosen to create a struct and store the two paths in it. We could also in the future add an env variable like `TASK_X_REMOTE_DIR`. If you think it can be a nice addition, I can implement it in this PR 

Any feedback on this feature are appreciated :)

### DIFF
--- a/website/docs/experiments/remote_taskfiles.mdx
+++ b/website/docs/experiments/remote_taskfiles.mdx
@@ -112,7 +112,12 @@ and look for a cached copy instead. This timeout can be configured by setting
 the `--timeout` flag and specifying a duration. For example, `--timeout 5s` will
 set the timeout to 5 seconds.
 
+By default, the cache is stored in the Task temp directory, represented by the `TASK_TEMP_DIR` [environment variable](../reference/environment.mdx)
+You can override the location of the cache by setting the `TASK_REMOTE_DIR` environment variable. This way, you can share the cache between different projects.
+
+
 {/* prettier-ignore-start */}
 [enabling-experiments]: ./experiments.mdx#enabling-experiments
 [man-in-the-middle-attacks]: https://en.wikipedia.org/wiki/Man-in-the-middle_attack
 {/* prettier-ignore-end */}
+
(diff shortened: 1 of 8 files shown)


andreynering (commented):
@vmaerten Good catch, this makes sense! The implementation looks good to me as well.

I think a `TASK_REMODE_DIR` would be a good addition to this change. If given, use that, otherwise fallback to `TASK_TEMP_DIR` as you already did.

vmaerten:
@andreynering thanks! I've added the new env variable.
For now I did not put it in the `Environment Reference` doc section because it's linked to the experimentation but I've added a small section in [caching](https://deploy-preview-1661--taskfile.netlify.app/next/experiments/remote-taskfiles/#caching--running-offline)

Let me know if you think that adding it in environment reference is needed.

It would be amazing if we could get this merged before the next version

website/docs/experiments/remote_taskfiles.mdx
```diff
@@ -112,7 +112,12 @@ and look for a cached copy instead. This timeout can be configured by setting
...
 
+By default, the cache is stored in the Task temp directory, represented by the `TASK_TEMP_DIR` [environment variable](../reference/environment.mdx)
+You can override the location of the cache by setting the `TASK_REMOTE_DIR` environment variable. This way, you can share the cache between different projects.
+
+
 {/* prettier-ignore-start */}
 [enabling-experiments]: ./experiments.mdx#enabling-experiments
 [man-in-the-middle-attacks]: https://en.wikipedia.org/wiki/Man-in-the-middle_attack
 {/* prettier-ignore-end */}
+
```
andreynering, line 123: Unneeded extra blank lines.

```suggestion
By default, the cache is stored in the Task temp directory, represented by the `TASK_TEMP_DIR` [environment variable](../reference/environment.mdx)
You can override the location of the cache by setting the `TASK_REMOTE_DIR` environment variable. This way, you can share the cache between different projects.

{/* prettier-ignore-start */}
[enabling-experiments]: ./experiments.mdx#enabling-experiments
[man-in-the-middle-attacks]: https://en.wikipedia.org/wiki/Man-in-the-middle_attack
{/* prettier-ignore-end */}
```

andreynering (approved):
> Let me know if you think that adding it in environment reference is needed.

@vmaerten Yes, please add to this section: https://taskfile.dev/api/#env

pd93 (approved):

PR #1663, opened 2024-05-19
fix: version check

### LINKED ISSUE #872: Specify minimum required task verison.
While the `version:` key is nice to specify the Taskfile schema version, it would be nice to have some option to define a _minimum_ required task version to run a task.

For example, a recent change introduced `.ROOT_DIR`, which is incredibly handy for how I leverage task for my team, but many of my devs are stumbling over that as they haven't bothered to upgrade task yet.

I propopse that in in taskfiles, the `version:` key be expanded to check for minimum task version.

Example:

```
version: 3
``` 

would be compatible with all 3.x taskfile releases, but...

```
version: 3.15.2
```

would cause `task` to exit non-zero if the task version isn't _at least_ 3.15.2.

vmaerten:
Fixes #872

Following this PR : 
- https://github.com/go-task/task/pull/1465

https://github.com/go-task/task/blob/a74b0bc679e770179c43e265f134feaa3b1b733e/setup.go#L241-L253
This does not work when the sum is displayed in the version
Cf : 
image
And
https://github.com/go-task/task/blob/a74b0bc679e770179c43e265f134feaa3b1b733e/internal/version/version.go#L21-L23

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -5,21 +5,25 @@ import (
 	"runtime/debug"
 )
 
-var version = ""
+var (
+	version = ""
+	sum     = ""
+)
 
-func GetVersion(withSum bool) string {
-	if version != "" {
-		return version
-	}
-
+func init() {
 	info, ok := debug.ReadBuildInfo()
 	if !ok || info.Main.Version == "" {
-		return "unknown"
+		version = "unknown"
+	} else {
+		version = info.Main.Version
+		sum = info.Main.Sum
 	}
+}
 
-	ver := info.Main.Version
-	if info.Main.Sum != "" && withSum {
-		ver += fmt.Sprintf(" (%s)", info.Main.Sum)
-	}
-	return ver
+func GetVersion() string {
+	return version
+}
+
+func GetVersionWithSum() string {
+	return fmt.Sprintf("%s (%s)", version, sum)
 }
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -7,7 +7,7 @@ import (
 
 var version = ""
 
-func GetVersion() string {
+func GetVersion(withSum bool) string {
 	if version != "" {
 		return version
 	}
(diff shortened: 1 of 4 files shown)


internal/version/version.go
```diff
@@ -7,7 +7,7 @@ import (
 
 var version = ""
```
pd93, line 8: I don't think this variable is currently set/used?

Might be nice to do something like the following:

```go
var (
    version = ""
    sum = ""
)

func init() {
  	info, ok := debug.ReadBuildInfo()
	if !ok || info.Main.Version == "" {
		version = "unknown"
	}
    version = info.Main.Version
	sum = info.Main.Sum
}
```

This way, we're not fetching the version info every time `GetVersion` is called.

internal/version/version.go
```diff
@@ -7,7 +7,7 @@ import (
...
 
-func GetVersion() string {
+func GetVersion(withSum bool) string {
```
pd93, line 10: If following the previous suggestion, I think I would also prefer two separate functions. This feels more Go-like and readable in calling code to me than parameter flags:

```go
func GetVersion() string {
    return version
}

func GetVersionWithSum() string {
    return fmt.Sprintf("%s (%s)", version, sum)
}
```

pd93 (commented):
Thanks for fixing this. Bit of an oversight on my part when doing #1465. A couple of minor suggestions

internal/version/version.go
vmaerten: I had the same question when doing this.
Good to me :)

internal/version/version.go
vmaerten: This way it's cleaner, thanks!

vmaerten:
It's done @pd93 
Thanks a lot for your review, especially about "Go-like code" Even if we are using Go at work, it's always good to know 
good practices

pd93 (approved):
Looks good to me. Thanks!

internal/version/version.go
stefanb: Please note that building from source tarball without any git info still requires setting the version via ldflags.
See
* https://github.com/go-task/task/issues/1709

PR #1666, opened 2024-05-20
docs: taskfile versions

pd93:
Pairs nicely with the changes in #1663.

This updates the desperately out-of-date Taskfile Versions doc and removes most of the old information about version 1 & 2. I don't think this information is often used and we don't really have any docs explaining how schema versions _currently_ work in v3. I think this is a good place for them to go.

As always, thoughts/comments/suggestions welcome


andreynering (approved):
I agree with the proposal!

Issue #1484, opened 2024-01-26
Allow `defer` to know if the task succeded or failed


andreynering:
Originally asked in #1472 by @your-diary.

A possible way to allow this is to expose one/some variables to `defer`. One possibility is to expose a `DEFER_EXIT_CODE` var which would contain `0` for success or the exit code of the task that failed.

Perhaps we could also have `DEFER_SUCCESS` (true/false) and `DEFER_FAIL` (true/false) to make a simpler check easier? (Not sure, `DEFER_EXIT_CODE` is probably enough).

dorimon-1:
@andreynering Could you explain a bit deeper how you would the yml to be?
I thought about 
```
  task-2:
    - defer_failure: echo Failure!
    - defer_failure: { task: "task-1", vars: { PARAM: "failure" } }
    - defer_success: echo Success!
    - defer_success: { task: "task-1", vars: { PARAM: "success" } }
    - echo 'cmd ran'
    - exit 1
```

your-diary:
@dorimon-1 
One can ignore my original suggestion. It is just an example.

As for your suggestion, if we add `defer_success` and `defer_failure` while keeping `defer` keyword, we can make it backward-compatible.

KJone1:
Exactly what I have been searching for.
 
I think @dorimon-1 idea looks neat and allows for backward compatibility with regular defer. 
Also having env var for defer status code could be helpful for some ppl.
If someone could implement this it would be much appreciated 😉.

andreynering:
This was just implemented on #1762.

PR #1673, opened 2024-05-26
feat: expose task's exit code to env var 'DEFER_EXIT_CODE'

dorimon-1:
#### Issue: #1484 
Exposes environment variable **DEFER_EXIT_CODE** once a task is finished, to allow you to use it in defer statements.
In case there was an error (exit code != 0), it wont be overridden by a successful task, but it will be overridden when a different task fails with a different code.


website/versioned_docs/version-latest/api_reference.mdx
```diff
@@ -137,6 +137,7 @@ There are some special variables that is available on the templating system:
 | `TIMESTAMP`        | The date object of the greatest timestamp of the files listed in `sources`. Only available within the `status` prop and if method is set to `timestamp`. |
 | `TASK_VERSION`     | The current version of task.                                                                                                                             |
 | `ITEM`             | The value of the current iteration when using the `for` property. Can be changed to a different variable name using `as:`.                               |
+| `DEFER_EXIT_CODE`  | The exit code of the last failed task. Can be used inside a defer statement to perform different operations according to the exit code.                  |
```
pd93, line 140: Any reason why this couldn't just be `EXIT_CODE`? "Defer" is the place this variable would be used rather than a description of what the variable holds.

dorimon-1: No reason actually though so too, but ended up using DEFER_EXIT_CODE because it what was requested at the issue.

website/package-lock.json
pd93, line 1: Could you please delete this. We use `yarn` instead of `npm`

dorimon-1: My bad, generated it for some weird reason.
removed

pd93 (approved):
Looks good to me. Thanks!

andreynering (changes requested):
Hi @dorimon-1. Thanks for your pull request, it is much appreciated!

I'm requesting changes `EXIT_CODE` should be a regular variable and not an environment variable. This would make it consistent with the other [special variables](https://taskfile.dev/api/#special-variables).

dorimon-1:
Also fixed conflicts

website/versioned_docs/version-latest/reference/templating.mdx
```diff
@@ -115,7 +115,7 @@ special variable will be overridden.
 | `TIMESTAMP`        | The date object of the greatest timestamp of the files listed in `sources`. Only available within the `status` prop and if method is set to `timestamp`. |
 | `TASK_VERSION`     | The current version of task.                                                                                                                             |
 | `ITEM`             | The value of the current iteration when using the `for` property. Can be changed to a different variable name using `as:`.                               |
-
+| `EXIT_CODE`	     | The exit code of the last failed task. Can be used inside a defer statement to perform different operations according to the exit code.                 	|
```
vmaerten, line 118: This should be done only in `docs/reference/templating.mdx` and reverted from this file
versioned_docs is generated and will be overriden

vmaerten (changes requested):

vmaerten (approved):

andreynering:
Hey @dorimon-1!

There were further changes required for this to work as desired. I did those changes on #1762 and just merged it.

Basically, we neede to make that variable as scoped as possible, i.e., available only to that specific `defer:` command and not globally.

PR #1675, opened 2024-05-29
change install from npm scripts postinstall to optionalDependencies

sirenkovladd:
>
>

Fix: https://github.com/go-task/go-npm/issues/11

Change install from npm
postinstall script -> optionalDependencies

`postinstall` can be a vulnerable script, which is why some installations ignore all hooks
it also forces the installer to execute code making the installation slower and providing a bad experience

it is better to use `optionalDependencies`, which will allow you to install a binary image for a specific operating system

this will be faster and clearer, and will also prevent evil code to execute on your machine


sirenkovladd:
not sure if you made the right GitHub action, better check it carefully

Kenneth-Sills:
I agree that we should move to this approach. It also addresses a couple of issues:

1. Right now, Task requires root permissions to install and [always installs globally](https://github.com/go-task/go-npm/issues/6). If you install it local to a project, that project's dependencies will silently overwrite any existing Task installation. It will even cause conflicts with official package installations. You can always download the binaries, but it's an unnecessary barrier to entry that makes it harder to throw together a prototype, show coworkers, and push for adoption. 
2. The cleanup paradigm `go-npm` uses is out of date and since NPM 7 [removed `*uninstall` lifecycle scripts](https://docs.npmjs.com/cli/v10/using-npm/scripts#a-note-on-a-lack-of-npm-uninstall-scripts), it leaves orphaned binaries lying around.

packages/task-release/release.js
```diff
@@ -0,0 +1,86 @@
...
+
+const packageJson = readFileSync("package.json", "utf-8");
+
+async function createFolder(platform) {
+  const folder = `task-${platform.os}-${platform.arch}`
+
+  console.debug(`Creating folder ${folder}`)
+  await mkdir(`${folder}/bin`, { recursive: true })
+  console.debug(`Copying files to ${folder}`)
+  await writeFile(`${folder}/package.json`, packageJson.replace(/{platform}/g, platform.os).replace(/{arch}/g, platform.arch).replace(/{version}/g, mainPackage.version))
```
trim21, line 58: Maybe use `.replaceAll('{platform}', ...)`? don't this we need regex here.

sirenkovladd: What are the advantages of using replaceAll instead of replace with a global flag?

trim21: readers don't need to parse regex in mind

caarlos0:
fwiw, I'm implementing this in goreleaser as well... for what I can tell, this does not seem to work.

The platform-specific dependency is downloaded, but it seems the binary is never put in the right places.

`npmx <the thing>` also don't work 🤔

vmaerten:
Thanks for your PR
I am closing this because we use goreleaser now

PR #1677, opened 2024-06-01
Fix and modify requires behavior.

### LINKED ISSUE #1676: Empty required variable still passing the required check?
### Discussed in https://github.com/go-task/task/discussions/1620

<div type='discussions-op-text'>

<sup>Originally posted by **usersina** April 26, 2024</sup>
- Task version: `v3.36.0`
- Operating system: `Arch Linux x86_64 6.6.28-1-lts`
- Experiments enabled: `No`

Given the following `Taskfile.yml`:

```yml
version: "3"

silent: true

vars:
  IMAGE_NAME: my-web-app
  CONTAINER_REGISTRY: # docker.io

tasks:
  check-missing:
    desc: Test the task variables
    requires:
      vars: [CONTAINER_REGISTRY]
    cmds:
      - echo "CONTAINER_REGISTRY is {{.CONTAINER_REGISTRY}}"
```

My `check-missing` is not working as intended.

Actual output
> CONTAINER_REGISTRY is 

Expected output 
> task: Task "check-missing" cancelled because it is missing required variables: CONTAINER_REGISTRY

This works as intended if I simply delete `CONTAINER_REGISTRY` or comment it out, but why isn't it working if it's empty?</div>

trulede:
Fix and modify the behavior of `requires`.

Fixes #1676 (discussion #1620)
Modifies/updates #1204

Requires was not working in the case that a variable was set to nil (see discussion/issue).

Additional, the behavior of requires is modified so that an zero length (empty) string value fails the required check. In more complex use cases, with sub-tasks and variable templates, the expectation is that requires will ensure that a variable is set to _something_. In such cases, an empty string is not considered as _something_ and so the feature is not useful.

An alternative would be to have a way to set a variable value to `nil` via the template mechanism.

The change is tested with the following taskfile (modified from the original discussion).

```
version: "3"

silent: true

vars:
  IMAGE_NAME: my-web-app
  CONTAINER_REGISTRY: 
  # CONTAINER_REGISTRY: # comment
  # CONTAINER_REGISTRY: ''
  # CONTAINER_REGISTRY: ' '

tasks:
  check-missing:
    desc: Test the task variables
    requires:
      vars: [CONTAINER_REGISTRY]
    cmds:
      - echo "CONTAINER_REGISTRY is {{spew .CONTAINER_REGISTRY}}"
```

### DIFF
--- a/requires.go
+++ b/requires.go
@@ -19,9 +19,22 @@ func (e *Executor) areTaskRequiredVarsSet(ctx context.Context, t *ast.Task, call
 
 	var missingVars []string
 	for _, requiredVar := range t.Requires.Vars {
-		if !vars.Exists(requiredVar) {
-			missingVars = append(missingVars, requiredVar)
+		v := vars.Get(requiredVar)
+
+		// Check and continue on positive conditions.
+		if v.Value != nil {
+			switch val := v.Value.(type) {
+			case string:
+				if len(val) > 0 {
+					continue
+				}
+			default:
+				continue
+			}
 		}
+
+		// The required variable is not available.
+		missingVars = append(missingVars, requiredVar)
 	}
 
 	if len(missingVars) > 0 {
--- a/website/docs/usage.mdx
+++ b/website/docs/usage.mdx
@@ -911,13 +911,14 @@ tasks:
 ### Ensuring required variables are set
 
 If you want to check that certain variables are set before running a task then
-you can use `requires`. This is useful when might not be clear to users which
+you can use `requires`. This is useful when it might not be clear to users which
 variables are needed, or if you want clear message about what is required. Also
 some tasks could have dangerous side effects if run with un-set variables.
 
 Using `requires` you specify an array of strings in the `vars` sub-section under
 `requires`, these strings are variable names which are checked prior to running
-the task. If any variables are un-set the the task will error and not run.
+the task. If any of the listed variables are either un-set or zero length strings
+then the task will error and not run.
 
 Environmental variables are also checked.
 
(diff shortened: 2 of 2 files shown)


trulede:
Further to the change in behaviour, please see the results of this experiment where the templating system is considering a zero length string as "not set".

```yaml
version: "3"

vars:
  A: fubar
  B: ''
  C:

tasks:
  check-set:
    desc: Test the task variables
    vars:
      A_VAR:  '{{.A | default "A NOT SET"}}'
      B_VAR:  '{{.B | default "B NOT SET"}}'
      C_VAR:  '{{.C | default "C NOT SET"}}'
    cmds:
      - echo "A is {{.A_VAR}}"
      - echo "B is {{.B_VAR}}"
      - echo "C is {{.C_VAR}}"
```

produces this output

```bash
$ task check-set -f -v
task: [/mnt/c/Users/trule] Not found - Using alternative (Taskfile.yml)
task: "check-set" started
task: [check-set] echo "A is fubar"
A is fubar
task: [check-set] echo "B is B NOT SET"
B is B NOT SET
task: [check-set] echo "C is C NOT SET"
C is C NOT SET
task: "check-set" finished
```

This PR modifies the requires behaviour to match that of the templating system, which is I think essential for _requires_ to be usable. Currently, we do this `FOO: '{{.FOO| default "FOO_NOT_SPECIFIED"}}'` and **_hope_** that the user will notice the strange message in the output ... we tried _requires_, but it did not work. The concise error message that _requires_ produces would be very welcome, both for developers and users.

PR #1702, opened 2024-06-30
Add TTL to taskfile.Cache

### LINKED ISSUE #1402: Feature Request:  Have an auto update ttl for downloaded taskfiles
As discussed [here](https://github.com/go-task/task/pull/1345#issuecomment-1791151450), it would be great to have an auto-cache-ttl setting for downloaded taskfiles.

Context:
I am wanting to use the remote features to have a standard set of taskfile includes across my org.

For me, I would like it to download first, then by default always use the cache. I run taskfile a lot, (multiple times a minute if I'm running a debugging command like `task test -- feature/a` ), so having extra network overhead wouldn't be useful.
I understand that there is work on on offline flag, but using that would mean my taskfiles would need manual updating.
Having some kind of auto-update-ttl would allow users to have the fast latency of not fetching every task run, but also allow automatic updates in a soft manner.

Thanks for all your hard work on this great tool!

jlucktay:
# Description

This PR implements part of the [remote Taskfiles experiment](https://taskfile.dev/experiments/remote-taskfiles/) (#1317) that is currently ongoing.

## Implementation notes

- `taskfile.Cache` has a new private `time.Duration` field for the TTL setting.
  - The [default TTL setting](https://github.com/jlucktay/task/blob/3c9dab3ecf1e9b4bab16a093c870a7ff55341ba2/taskfile/cache.go#L13-L14) for the cache is 24 hours. Should this be shorter/longer? 4 hours? 1 week?

- The `taskfile` package has some new exports:
  - the default TTL value described above, in a `const`.
  - a static `error`: `ErrExpired`.
  - a `CacheOption` type, and a `WithTTL` func returning this type, to help with creation of new `Cache` structs.

- The main `task` executable has a new flag (currently gated by the `TASK_X_REMOTE_TASKFILES` experiment) to override the default cache TTL setting, outlined below.

# Flow diagram

There is [a flowchart sketched out here on this branch](https://github.com/jlucktay/task/blob/issue-1402/add-ttl-to-taskfile-cache/taskfile/README.md) to describe the overall flow.

# New CLI behaviour

```
$ TASK_X_REMOTE_TASKFILES=1 task --help 
…
Options:
…
      --remote-cache-ttl duration   TTL of remote Taskfiles downloaded into the local cache. (default 24h0m0s)
```

Resolves #1402.

### DIFF
--- a/signals_test.go
+++ b/signals_test.go
@@ -20,9 +20,7 @@ import (
 	"time"
 )
 
-var (
-	SLEEPIT, _ = filepath.Abs("./bin/sleepit")
-)
+var SLEEPIT, _ = filepath.Abs("./bin/sleepit")
 
 func TestSignalSentToProcessGroup(t *testing.T) {
 	task, err := getTaskPath()
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -60,7 +60,9 @@ func (c *Cache) write(node Node, b []byte) error {
 }
 
 func (c *Cache) read(node Node) ([]byte, error) {
-	fi, err := os.Stat(c.cacheFilePath(node))
+	cfp := c.cacheFilePath(node)
+
+	fi, err := os.Stat(cfp)
 	if err != nil {
 		return nil, fmt.Errorf("could not stat cached file: %w", err)
 	}
@@ -70,7 +72,7 @@ func (c *Cache) read(node Node) ([]byte, error) {
 		return nil, ErrExpired
 	}
 
-	return os.ReadFile(c.cacheFilePath(node))
+	return os.ReadFile(cfp)
 }
 
 func (c *Cache) writeChecksum(node Node, checksum string) error {
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -2,24 +2,51 @@ package taskfile
 
 import (
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 )
+
+// defaultCacheTTL is one day (24 hours).
+const defaultCacheTTL = time.Duration(time.Hour * 24)
 
 type Cache struct {
 	dir string
+	ttl time.Duration
 }
 
-func NewCache(dir string) (*Cache, error) {
+// ErrExpired is returned when a cached file has expired.
+var ErrExpired = errors.New("task: cache expired")
+
+type CacheOption func(*Cache)
+
+func NewCache(dir string, opts ...CacheOption) (*Cache, error) {
 	dir = filepath.Join(dir, "remote")
 	if err := os.MkdirAll(dir, 0o755); err != nil {
 		return nil, err
 	}
-	return &Cache{
+
+	cache := &Cache{
 		dir: dir,
-	}, nil
+		ttl: defaultCacheTTL,
+	}
+
+	// Apply options.
+	for _, opt := range opts {
+		opt(cache)
+	}
+
+	return cache, nil
+}
+
+// WithTTL will override the default TTL setting on a new Cache.
+func WithTTL(ttl time.Duration) CacheOption {
+	return func(cache *Cache) {
+		cache.ttl = ttl
+	}
 }
 
 func checksum(b []byte) string {
@@ -33,6 +60,16 @@ func (c *Cache) write(node Node, b []byte) error {
 }
 
 func (c *Cache) read(node Node) ([]byte, error) {
+	fi, err := os.Stat(c.cacheFilePath(node))
+	if err != nil {
+		return nil, fmt.Errorf("could not stat cached file: %w", err)
+	}
+
+	expiresAt := fi.ModTime().Add(c.ttl)
+	if expiresAt.Before(time.Now()) {
+		return nil, ErrExpired
+	}
+
 	return os.ReadFile(c.cacheFilePath(node))
 }
 
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -10,8 +10,8 @@ import (
 	"time"
 )
 
-// defaultCacheTTL is one day (24 hours).
-const defaultCacheTTL = time.Duration(time.Hour * 24)
+// DefaultCacheTTL is one day (24 hours).
+const DefaultCacheTTL = time.Duration(time.Hour * 24)
 
 type Cache struct {
 	dir string
@@ -31,7 +31,7 @@ func NewCache(dir string, opts ...CacheOption) (*Cache, error) {
 
 	cache := &Cache{
 		dir: dir,
-		ttl: defaultCacheTTL,
+		ttl: DefaultCacheTTL,
 	}
 
 	// Apply options.
--- a/taskfile/cache.go
+++ b/taskfile/cache.go
@@ -10,7 +10,7 @@ import (
 	"time"
 )
 
-// DefaultCacheTTL is one day (24 hours).
+// DefaultCacheTTL is used when a value is not explicitly given to a new Cache.
 const DefaultCacheTTL = time.Duration(time.Hour * 24)
 
 type Cache struct {
--- a/taskfile/cache_internal_test.go
+++ b/taskfile/cache_internal_test.go
@@ -81,16 +81,16 @@ func TestCacheInsideTTL(t *testing.T) {
 }
 
 func TestCacheOutsideTTL(t *testing.T) {
-	// Prime a new Cache with a TTL of one second.
-	cache, fileNode := primeNewCache(t, t.TempDir(), WithTTL(time.Second))
+	// Prime a new Cache with an extremely short TTL.
+	cache, fileNode := primeNewCache(t, t.TempDir(), WithTTL(time.Millisecond))
 
 	// Write some bytes for the cached file.
 	writeBytes := []byte("some bytes")
 	err := cache.write(fileNode, writeBytes)
 	require.NoError(t, err, "writing bytes to cache")
 
-	// Sleep for two seconds so that the cached file is outside of TTL.
-	time.Sleep(2 * time.Second)
+	// Sleep for 5ms so that the cached file is outside of TTL.
+	time.Sleep(5 * time.Millisecond)
 
 	// Reading from the cache after sleeping past the end of TTL should get an error.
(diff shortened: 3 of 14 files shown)


taskfile/cache.go
```diff
@@ -33,6 +60,16 @@ func (c *Cache) write(node Node, b []byte) error {
 }
 
 func (c *Cache) read(node Node) ([]byte, error) {
+	fi, err := os.Stat(c.cacheFilePath(node))
```
ccoVeille, line 63: Maybe your could store the cacheFilePath in a variable, so you don't compute it twice in this method

taskfile/cache_internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return cache, fileNode
+}
+
+func TestCache(t *testing.T) {
+	cache, fileNode := primeNewCache(t, t.TempDir())
+
+	// Attempt to read from cache, then write, then read again.
+	_, err := cache.read(fileNode)
+	require.ErrorAs(t, err, &os.ErrNotExist, "reading from cache before writing should match error type")
```
ccoVeille, line 57: I'm surprised, I would have expected an ErrorIs here


```suggestion
	require.ErrorIs(t, err, os.ErrNotExist, "reading from cache before writing should match error type")
```

os.ErrNotExist is a var not a type.

You maybe have used ErrorAs with `os.PathError`

taskfile/cache_internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// Prime a new Cache with a TTL of one second.
+	cache, fileNode := primeNewCache(t, t.TempDir(), WithTTL(time.Second))
+
+	// Write some bytes for the cached file.
+	writeBytes := []byte("some bytes")
+	err := cache.write(fileNode, writeBytes)
+	require.NoError(t, err, "writing bytes to cache")
+
+	// Sleep for two seconds so that the cached file is outside of TTL.
+	time.Sleep(2 * time.Second)
```
ccoVeille, line 93: It's wait too long for tests.

Define a shorter TTL with withttl, then reduce here

taskfile/cache_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/go-task/task/v3/taskfile"
+)
+
+func TestNewCache(t *testing.T) {
+	cacheDir := t.TempDir()
+	_, err := taskfile.NewCache(cacheDir)
+	require.NoErrorf(t, err, "creating new cache in temporary directory '%s'", cacheDir)
+}
+
+func TestNewCacheWithTTL(t *testing.T) {
```
ccoVeille, line 18: You don't have a need for this, I think.

You could update TestNewCache to accept a variadic and pass it to NewCache

taskfile/cache_internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+var discardLogger = &logger.Logger{
+	Stdout: io.Discard,
+	Stderr: io.Discard,
+}
+
+func primeNewCache(t *testing.T, tempDir string, cacheOpts ...CacheOption) (*Cache, *FileNode) {
+	t.Helper()
+
+	cache, err := NewCache(tempDir, cacheOpts...)
```
ccoVeille, line 31: Do you use this testing helper with another folder than the.TmpDir() ?

If not then you could remove the tempdir argument and use 


```suggestion
	cache, err := NewCache(t.TempDir(), cacheOpts...)
```

jlucktay: I was taking `tempDir` as an argument there because I re-use it on the next line, and calling it twice would give two different directories. I suppose I'll take it out of L32?

taskfile/cache_test.go
jlucktay: It could be a table-drive test, yeah. I'll boil it down a bit.

taskfile/cache_internal_test.go
jlucktay: I think I was using `ErrorAs` for a different error type, before a refactor, yes. Will switch it over, thanks.

jlucktay:
Thank you for the notes @ccoVeille 🙂 🙏  
I'll give things a once-over soon.

taskfile/cache_internal_test.go
ccoVeille: Move this 

```go
        require.NoErrorf(t, err, "creating new cache in temporary directory '%s'", tempDir)
```

To

NewCache by using a fmt.Errorf to add the temporary Dir/file you were trying to create

ccoVeille: Something like this in NewCache

```go
         if err := os.MkdirAll(dir, 0o755); err != nil { 
                  return nil, fmt.Errorf("creating cache dir %s: %w", dir, err)
          }
```

taskfile/cache.go
jlucktay: 8e45a612c1a92ed43962400ec9192e77be0c370b will resolve.

taskfile/cache_internal_test.go
jlucktay: b9ad0bff3c213f87b75aed719bf8adce8d0f9540 will resolve.

taskfile/cache_test.go
jlucktay: 0eadafa8d6fc309d694fef68c3b282180db86b95 should resolve.

taskfile/cache_internal_test.go
jlucktay: ea229890a392eb4615386110adecaf22478c54bf will resolve.

ccoVeille (approved):

ccoVeille (approved):

website/versioned_docs/version-latest/experiments/remote_taskfiles.mdx
```diff
@@ -104,21 +104,33 @@ Whenever you run a remote Taskfile, the latest copy will be downloaded from the
...
 variable. This way, you can share the cache between different projects.
 
+The default TTL for remote Taskfiles in the cache is 24 hours. This can be
+overridden by the `--remote-cache-ttl` flag outlined above. The value must be
+specified as a duration, e.g. `30s`, `15m`, `4h`. More information on how
+[durations are parsed is available here](https://pkg.go.dev/time#ParseDuration).
+
 {/* prettier-ignore-start */}
 [enabling-experiments]: ./experiments.mdx#enabling-experiments
 [man-in-the-middle-attacks]: https://en.wikipedia.org/wiki/Man-in-the-middle_attack
```
vmaerten, line 136: This should be done only in `webiste/docs` and reverted from this file.
`webiste/versioned_docs` is a generated folder and will be overriden

vmaerten (changes requested):

website/versioned_docs/version-latest/experiments/remote_taskfiles.mdx
jlucktay: Noted, thanks! 
When I edited the other file and ran a preview on my local, it didn't show any of the changes. 
I'll move them over when I get back to my desk.

vmaerten: Yeah it's because we have two version of the doc : 
- The stable's one (the one you see when you go to taskfile.dev, or when running the server locally. It point to the latest release
- The "next's one" which point to main

To access to the "next's one", you need to click in the header bar : 
image

signals_test.go
```diff
@@ -20,9 +20,7 @@ import (
 	"time"
 )
 
-var (
-	SLEEPIT, _ = filepath.Abs("./bin/sleepit")
-)
+var SLEEPIT, _ = filepath.Abs("./bin/sleepit")
```
ccoVeille, line 23: While you are simply applying a gofumpt on this.

I'm surprised to see:
- the error is ignored
- this variable name is in uppercase
- the variable is apparently used only in one test.

I would expect this variable to be used only where needed and to check the possible error

taskfile/cache.go
```diff
@@ -2,24 +2,51 @@ package taskfile
...
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
+// DefaultCacheTTL is one day (24 hours).
```
ccoVeille, line 13:
```suggestion
// DefaultCacheTTL is used when WitTTL is not used
```

There is no need add a comment that could be outdated next time the value is changed. I suggest a better comment

website/versioned_docs/version-latest/experiments/remote_taskfiles.mdx
jlucktay: e309b73f455d88cd133f6730e9b09ed3fd0c5063 will resolve.

taskfile/cache.go
jlucktay: 741743b7137328a81c95ff33b82f0bd02ebb581f will resolve.

taskfile/cache.go
```diff
@@ -2,24 +2,51 @@ package taskfile
...
+const DefaultCacheTTL = time.Duration(time.Hour * 24)
+
 type Cache struct {
 	dir string
+	ttl time.Duration
 }
 
-func NewCache(dir string) (*Cache, error) {
+// ErrExpired is returned when a cached file has expired.
+var ErrExpired = errors.New("task: cache expired")
```
vmaerten, line 22: It should go in the `errors` package.
You need to be a bit more specific in the error message, for example write wich file is expired

vmaerten (commented):
Hi !
Thanks for working on this :fire:  To me 24h is too short.
I'll take my example but it can work for other.

We include only taskfiles we own in our company, so I know when I need to fetch a new version. I am always running with offline mode to avoid unecessary http call (and it's faster to not fetch X differents remote taskfiles).

I think we could add an env variable to control the cache duration (this way we do not need to worry about passing the flag each time we download a remote file)

One more thing, the chart does not reflect the real behavior. 
image
As we said, the "offline" first will be done in another PR, you need to update the chart

pd93:
@jlucktay Thanks for all your work on this, but I'm going to close this in favor of #2176 which should implement most of the things you wanted from this PR. Would love any input you have on the changes there though.

PR #1704, opened 2024-07-01
feat: add option to declare an included Taskfile as flatten

vmaerten:
It's one of the most upvoted feature ! So I gave it a try

Closes : https://github.com/go-task/task/issues/273


.editorconfig
```diff
@@ -8,6 +8,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 indent_style = tab
 
-[*.{md,yml,yaml,json,toml,htm,html,js,css,svg,sh,bash,fish}]
+[*.{mdx,yml,yaml,json,toml,htm,html,js,css,svg,sh,bash,fish}]
```
ccoVeille, line 11:
```suggestion
[*.{md,mdx,yml,yaml,json,toml,htm,html,js,css,svg,sh,bash,fish}]
```

No?

task_test.go
```diff
@@ -1216,6 +1216,44 @@ func TestIncludesInternal(t *testing.T) {
...
+func TestIncludesFlatten(t *testing.T) {
+	const dir = "testdata/includes_flatten"
+	tests := []struct {
+		name           string
+		taskfile       string
+		task           string
+		expectedErr    bool
+		expectedOutput string
+	}{
+		{"included flatten", "Taskfile.yml", "gen", false, "gen from included\n"},
```
ccoVeille, line 1228: Please use fields name for this struct.

It could lead to errors otherwise, fields can be reordered, value could be assigned yo the wrong field

task_test.go
```diff
@@ -1216,6 +1216,44 @@ func TestIncludesInternal(t *testing.T) {
...
+				Dir:        dir,
+				Entrypoint: dir + "/" + test.taskfile,
+				Stdout:     &buff,
+				Stderr:     &buff,
+				Silent:     true,
+			}
+			err := e.Setup()
+			if test.expectedErr {
+				require.Error(t, err)
+				assert.Equal(t, test.expectedOutput, err.Error())
```
ccoVeille, line 1247:
```suggestion
				assert.ErrorContains(t, err, test.expectedOutput)
```
Or

```suggestion
				assert.EqualError(t, err, test.expectedOutput)
```

taskfile/ast/taskfile.go
```diff
@@ -55,8 +55,8 @@ func (t1 *Taskfile) Merge(t2 *Taskfile, include *Include) error {
 	}
 	t1.Vars.Merge(t2.Vars, include)
 	t1.Env.Merge(t2.Env, include)
-	t1.Tasks.Merge(t2.Tasks, include, t1.Vars)
-	return nil
+	err := t1.Tasks.Merge(t2.Tasks, include, t1.Vars)
```
ccoVeille, line 58: If you use only for returning it. So, without wrapping value on error, then you could simply


```suggestion
	return t1.Tasks.Merge(t2.Tasks, include, t1.Vars)
```

ccoVeille (commented):
Please accept my humble code review

.editorconfig
vmaerten: In this PR (#1432) we moved from .md to .mdx but the editor config was not updated
Can do another PR if needed

pd93: Only the docs are mdx. There is still md elsewhere (readme, changelog etc). No reason not to list both here though. No worries about a separate PR.

vmaerten: You're right! I'll add both :)

ccoVeille:
You are welcome

andreynering (changes requested):
Hey @vmaerten, thanks for working on this!

I noticed a bug: `dir:` is not being respected on `includes:` when `flatten: true` is given.

```yaml
version: '3'

tasks:
  pwd:
    desc: Print working directory
    cmds:
      - pwd
```

```yaml
includes:
  included:
    taskfile: ./TaskfileIncluded.yml
    dir: ./website
    flatten: true
```

With `flatten: true`, it prints the root directory instead of the specified one (`dir: ./website` in my example).

Also @pd93 mentioned here (https://github.com/go-task/task/issues/273#issuecomment-2124856987) that he has refactors in mind that could affect this work. Worth getting an opinion before merging.

vmaerten:
Thank @andreynering for the review. I've forgot the AdvancedImport's part
It's now fixed and work as intended. Let me know if there are some usecase I've forgotten

vmaerten:
> > Thank @andreynering for the review. I've forgot the AdvancedImport's part It's now fixed and work as intended. Let me know if there are some usecase I've forgotten
> 
> Awesome to see this feature in a PR, I've been eager to build something with this.
> 
> An usecase that I would benefit from, would be if I could specify `override` as an option to `flatten`. If set, the inheriting Taskfile's cmd would take precedence over the included files set to flatten. I could then provide sets of Task "templates" but my users could override specific cmds if they need to.

I agree but for now, I would keep this feature as simple as possible
We can iterate through this after ffc

pd93 (approved):
> Also @pd93 mentioned here ([#273 (comment)](https://github.com/go-task/task/issues/273#issuecomment-2124856987)) that he has refactors in mind that could affect this work. Worth getting an opinion before merging.

I have some WIP code that removes the merging altogether. This feature will need to be reworked into that at some point. However, I've not had the time to work on it recently and I don't see any reason to delay this functionality given its popularity.

The PR seems well thought out and reimplementing it if merging is removed should be simple enough.

1 minor comment, but otherwise happy. Thanks for working on this @vmaerten! 🚀

errors/errors_task.go
```diff
@@ -80,6 +80,19 @@ func (err *TaskNameConflictError) Code() int {
 	return CodeTaskNameConflict
 }
 
+type TaskNameFlattenConflictError struct {
+	TaskName string
+	Include  string
+}
+
+func (err *TaskNameFlattenConflictError) Error() string {
+	return fmt.Sprintf(`task: Found multiple tasks (%s) included by "%s""`, err.TaskName, err.Include)
```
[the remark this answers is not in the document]
vmaerten: Do you think about `CodeTaskNameConflict` or another one ?
At first I used `CodeTaskNameConflict` but I found more readable to distinct both. That being said, it would make sense, so I'll give a try to adjust the message

PR #1708, opened 2024-07-03
fix: ROOT_TASKFILE when Entrypoint is not set

### LINKED ISSUE #1706: .ROOT_TASKFILE in root taskfile returns the root directory instead
- Operating system: "Ubuntu 22.04.4 LTS (on WSL2)
- Experiments enabled: N/A

```
❯ cat Taskfile.yml
version: '3'

tasks:
  who-am-i:
    cmds:
      - 'echo "Working dir: {{.USER_WORKING_DIR}}"'
      - 'echo "Task: {{.TASK}}"'
      - 'echo "Taskfile: {{.TASKFILE}}"'
      - 'echo "Root Taskfile: {{.ROOT_TASKFILE}}"'

❯ task who-am-i
task: [who-am-i] echo "Working dir: /home/chris/repos/repo-tools-playground"
Working dir: /home/chris/repos/repo-tools-playground
task: [who-am-i] echo "Task: who-am-i"
Task: who-am-i
task: [who-am-i] echo "Taskfile: /home/chris/repos/repo-tools-playground/Taskfile.yml"
Taskfile: /home/chris/repos/repo-tools-playground/Taskfile.yml
task: [who-am-i] echo "Root Taskfile: /home/chris/repos/repo-tools-playground"
Root Taskfile: /home/chris/repos/repo-tools-playground
```

I would have expected:
```
Root Taskfile: /home/chris/repos/repo-tools-playground/Taskfile.yml
```

(I searched for `ROOT_TASKFILE` in the issues and didn't see any related issues)

vmaerten:
fixes #1706

### DIFF
--- a/internal/compiler/compiler.go
+++ b/internal/compiler/compiler.go
@@ -180,10 +180,14 @@ func (c *Compiler) ResetCache() {
 }
 
 func (c *Compiler) getSpecialVars(t *ast.Task) (map[string]string, error) {
+	entrypoint := c.Entrypoint
+	if entrypoint == "" {
+		entrypoint = "Taskfile.yml"
+	}
 	return map[string]string{
 		"TASK":             t.Task,
 		"TASK_EXE":         filepath.ToSlash(os.Args[0]),
-		"ROOT_TASKFILE":    filepathext.SmartJoin(c.Dir, c.Entrypoint),
+		"ROOT_TASKFILE":    filepathext.SmartJoin(c.Dir, entrypoint),
 		"ROOT_DIR":         c.Dir,
 		"TASKFILE":         t.Location.Taskfile,
 		"TASKFILE_DIR":     filepath.Dir(t.Location.Taskfile),


vmaerten:
I though put a default value here 
```golang
	entrypoint := flags.Entrypoint
	if entrypoint == "" {
		entrypoint = "Taskfile.yml"
	}

```
but it would break 
```golang

func getDefaultDir(entrypoint, dir string) string {
	// If the entrypoint and dir are empty, we default the directory to the current working directory
	if dir == "" {
		if entrypoint == "" {
			wd, err := os.Getwd()
			if err != nil {
				return ""
			}
			dir = wd
		}
		return dir
	}

	// If the directory is set, ensure it is an absolute path
	var err error
	dir, err = filepath.Abs(dir)
	if err != nil {
		return ""
	}

	return dir
}

```

I don't know which solution do you prefer

internal/compiler/compiler.go
```diff
@@ -180,10 +180,14 @@ func (c *Compiler) ResetCache() {
 }
 
 func (c *Compiler) getSpecialVars(t *ast.Task) (map[string]string, error) {
+	entrypoint := c.Entrypoint
+	if entrypoint == "" {
+		entrypoint = "Taskfile.yml"
```
andreynering, line 185: Actually, the default is not necessarily `Taskfile.yml`. We have a list of possible 8 file names that are detected and work by default:

https://github.com/go-task/task/blob/88b095020e14defcce4645e72f503eb5d880f78d/taskfile/taskfile.go#L19-L28

In the `taskfile` package we have code that will try to detect it. It may be a bit tricky, but in order to be reliable we need to reuse that code somehow.

andreynering (changes requested):

internal/compiler/compiler.go
vmaerten: I totally agree I missed that part
What I did here, export taskfile.go in a new package in order to have access to `DefaultTaskfiles`
Then I check (as it's done on `Exists` func) if the file exist. I did not reuse the Exist function because we do not want to log smth

vmaerten: I checked and it works with `Taskfile.dist.yml`

internal/compiler/compiler.go
```diff
@@ -180,10 +181,20 @@ func (c *Compiler) ResetCache() {
...
+	entrypoint := c.Entrypoint
+	if entrypoint == "" {
+		for _, defaultName := range taskfile.DefaultTaskfiles {
+			alt := filepathext.SmartJoin(c.Dir, defaultName)
+			if _, err := os.Stat(alt); err == nil {
+				entrypoint = alt
+				break
+			}
+		}
+	}
```
andreynering, line 193: Hey @vmaerten,

Sorry for the long delay to review.

This looks mostly good, but I'd consider extracting this block into a separate function to make things cleaner. Or perhaps, we could use the existing `Exists` function if easy enough.

andreynering (approved):

internal/compiler/compiler.go
vmaerten: No worries!
I've extracted a function and used it in both method (exists and getSpecialVar)
Way cleaner thx!

pd93:
@vmaerten I just ran into this again while looking at #2267. Any chance you could rebase your changes? I think we should get this merged.

Issue #1406, opened 2023-11-21
Auto complete hangs with remote taskfiles


wburningham:
When including a remote taskfile the autocomplete hangs. 

I can't publish my taskfile because of company policy, but my setup is including a single remote ~650 line taskfile with 5 tasks that takes 1.2s to download (timing wget).

Steps to reproduce:

- Create a remote and local taskfile
- Enable the remote taskfile experiment and include the remote taskfile
- Setup autocompletion per the docs
- Attempt to type `task <foo prefix>` and notice that it hangs

- Task version: `v3.30.1`
- Operating system: `macOS 14.1.1`
- Experiments enabled: `TASK_X_REMOTE_TASKFILES: 1`

vmaerten:
Hello ! 
With the latest refactor, we optimized the way remote files are fetched (#1713) so it should be better.
At work, I set `TASK_OFFLINE` so this way I always use the cached copy and the autocomplete is instant. If I need to download a fresh version, I use ` task --offline=false --download`

vmaerten:
We also have a caching strategy allow you to automatically cache the remote taskfiles and read them locally

PR #1713, opened 2024-07-07
refactor: check if the remote exists just before reading it

vmaerten:
Hey!

## Context

When we have multiple remote Taskfiles, I noticed it can takes time to download all of them, so I added `--offline`.
But it was still taking times (either in running tasks, either in the autocomplete) so I took a look.

## Problem

It appears that we check if the file exist in the remote even if the `Offline` flags is set, this cause unnecessary HTTP requests (at the end it takes more time than needed)

## Solution

Move this check in `Read` method, this way it'll run only if there is no `--offline` .
I don't know if store the logger in the struct can causes performance issues, I did not notice that.

NOTE: as `Read` is executed before `ResolveEntrypoint` it shoud not cause any issues. I have a remote Taskfile included in a remote one and it's working

It can also improve preformance for this issue https://github.com/go-task/task/issues/1406

## Benchmark locally
I've ran some local benchmark in my projects : 

image
 vs 
image

EDIT:
Thanks to @ccoVeille, it appears that the timeout error does not work on main. cf : 
image
with the fix : 
image

PS: we may need to discuss about putting `--offline` by default in the autocomplete files or at least provide an env variable to do it easily
PS2: if this PR is merged, will backport to my `remote-git` PR


taskfile/node_http.go
```diff
@@ -60,6 +56,14 @@ func (node *HTTPNode) Remote() bool {
 }
 
 func (node *HTTPNode) Read(ctx context.Context) ([]byte, error) {
+	url, err := RemoteExists(ctx, node.logger, node.URL)
+	if err != nil {
+		return nil, err
+	}
+	node.URL = url
+	if errors.Is(ctx.Err(), context.DeadlineExceeded) {
```
ccoVeille, line 64: Shouldn't this one be before the previous if?

I mean you should  be testing this no ?


```suggestion
	if errors.Is(err, context.DeadlineExceeded) {
```

Said otherwise, is there any case where RemoteExist will not return an error when the context is expired?

ccoVeille (changes requested):

taskfile/node_http.go
vmaerten: You're right, it should be the previous if but it's a bit more complexe to fix than apply your commit
After checking, the Timeout error does not work on main.
Thanks for reporting this

taskfile/reader.go
```diff
@@ -208,7 +208,7 @@ func (r *Reader) readNode(node Node) (*ast.Taskfile, error) {
 		// Read the file
 		b, err = node.Read(ctx)
 		// If we timed out then we likely have a network issue
-		if node.Remote() && errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		if node.Remote() && errors.Is(err, &errors.TaskfileNetworkTimeoutError{}) {
```
ccoVeille, line 211: This should be an errors.As

taskfile/node_http.go
```diff
@@ -60,6 +56,11 @@ func (node *HTTPNode) Remote() bool {
 }
 
 func (node *HTTPNode) Read(ctx context.Context) ([]byte, error) {
+	url, err := RemoteExists(ctx, node.logger, node.URL, node.timeout)
```
ccoVeille, line 59: You removed the context.WithTimeout

You could put it back here before this

Because the timeout should be used with the head and the get


```suggestion
  ctx, cancel := context.WithTimeout(ctx, node.timeout)
  defer cancel()
	url, err := RemoteExists(ctx, node.logger, node.URL)
```

You are not using the timeout right now in remote except for the error

vmaerten: The timeout is set in the reader : 

```golang
ctx, cf := context.WithTimeout(context.Background(), r.timeout)
defer cf()

// Read the file
b, err = node.Read(ctx)
```

vmaerten:
Ok third attempt I think I have to desired behavior:
- If the request timeout && --download : error
- If the request timeout, use cached copy : 
  - If the cache is here = OK
  - If not = error
- if the request does not timeout = OK 
@pd93  Let me know if I missed something in the behavior

ccoVeille (approved):

taskfile/node_http.go
ccoVeille: Then you should catch the CTX error in the caller ? This way, you don't introduce the timeout parameter, that is only used for reporting errors,no?

Maybe I missed something again 😅😁

pd93 (approved):
> Let me know if I missed something in the behavior

What you described sounds right to me.

> PS: we may need to discuss about putting `--offline` by default in the autocomplete files or at least provide an env variable to do it easily

Agreed. If we haven't fetched the file yet then I'm assuming we haven't prompted the user to allow the remote file either, so we probably shouldn't fetch it for completions. I don't think a file should enter the user's filesystem until they have accepted the prompt.

> PS2: if this PR is merged, will backport to my `remote-git` PR

❤️

PR #1716, opened 2024-07-08
feat: add TASK_OFFLINE env and expose it as a special variable

### LINKED ISSUE #1470: Feature Request: provide a bool variable to tasks that represents if the —offline flag is used
My company is working on moving over to taskfiles for 200+ builds/apps/libraries. We rely on remote taskfiles to deduplicate build scripts. We’d like to execute logic based on whether or not a consumer uses the —offline flag. 

Examples:

1. prevent a task from running if using offline mode. Certain tasks require the latest remote taskfiles 
2. If offline disable commands in tasks that would make HTTP requests (docker pull)
3. Prompt a user for confirmation before running tasks that could benefit from pulling the latest remote taskfile

vmaerten:
Two things in this PR (very correlated) : 
- Add a `TASK_OFFLINE` env variable to control `--offline` flag. Is it useful when you want to use `--offline` all the time. And also, It will be picked up by the autocomplete without modifying the autocomplete file (related to #1713)
The CLI flags takes precedences over the env variable, this way we can override it as needed
- Added a special variable `CLI_OFFLINE` to expose this flag to the templating system (closes #1470)


internal/flags/flags.go
```diff
@@ -77,7 +79,10 @@ func init() {
 		log.Print(usage)
 		pflag.PrintDefaults()
 	}
-
+	offline, err := strconv.ParseBool(env.GetEnvOrDefault("TASK_OFFLINE", "false"))
+	if err != nil {
+		offline = false
+	}
```
pd93, line 85: Since 1.22 we can use the standard `cmp` package for this.

```suggestion
	offline, err := strconv.ParseBool(cmp.Or("TASK_OFFLINE", "false"))
	if err != nil {
		offline = false
	}
```

pd93 (changes requested):
This looks good to me! Just one minor requested change. In the future, I think it would be good to add more overridable env vars.

internal/flags/flags.go
vmaerten: Oh I didn't know! Thanks

vmaerten:
> In the future, I think it would be good to add more overridable env vars.

I agree. I think, I'll take some times to draft a PR to add more override like this

taskfile/ast/task.go
```diff
@@ -83,6 +83,7 @@ func (t *Task) WildcardMatch(name string) (bool, []string) {
 		return false, wildcards
 	}
 
+	fmt.Printf("wildcards: %v\n", wildcards)
```
pd93, line 86: Assuming leftover from debugging?

internal/fingerprint/sources_checksum.go
```diff
@@ -32,7 +32,6 @@ func (checker *ChecksumChecker) IsUpToDate(t *ast.Task) (bool, error) {
 	if len(t.Sources) == 0 {
 		return false, nil
 	}
-
```
pd93, line 35: nit: No need to remove the whitespace here. No changes in this file now.

taskfile/ast/task.go
vmaerten: Yeah. I feel a bit ashamed :grimacing:

pd93 (approved):
Nice one, thanks!

PR #1669, opened 2024-05-23
Add verbose/silent variables

Vince-Smith:
Fixes #1480

PR #1719, opened 2024-07-09
ci: add new job to check if versioned_docs has been modified

vmaerten:
A small proposal: check in the CI if files under versioned_docs has been modified.

We had the issue twice (https://github.com/go-task/task/pull/1669 && https://github.com/go-task/task/pull/1673) so I think doing this check in the CI can be useful
EDIT : this one too : https://github.com/go-task/task/pull/1702
Here is what it looks like : 

image

I've put it under `lint` because I did not want to create another workflow file

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -41,3 +41,18 @@ jobs:
 
       - name: check-jsonschema (metaschema)
         run: check-jsonschema --check-metaschema website/static/schema.json
+  check_doc:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pd93 (approved):
This is a really good idea. Thanks for catching the problem in those PRs. Are there any docs changes that have been overridden as a result of this?

vmaerten:
@pd93 AFAIK, only for https://github.com/go-task/task/pull/1669, I made the change in #1717 (and @andreynering merged it)

PR #1724, opened 2024-07-15
feat: go version 1.22

mefuller:
bump required version of go to 1.22


pd93:
Not quite ready for this yet. In https://github.com/go-task/template/pull/3#issuecomment-2228625895 I was referring to the dependency on `go-task/template`. I've updated this here: 4123ffc

PR #1725, opened 2024-07-17
feat: Add template variable for `watch` trigger

tploss:
When running a task in `watch` mode it can be useful to know which file triggered the task run.


trulede:
The watcher implementation has changed significantly. If you wanted to open a new PR it might be necessary to consider a list of changed files (i.e. the var is a list of strings), since several changes may occur before the task calls are cancelled/restarted.

PR #1740, opened 2024-08-07
Update usage.mdx - Fix DEAD (404 Page not found) link 

loloMD:
Fix DEAD link in `PLatform specific tasks and commands`.

Link to the valid `GOOS` and `GOARCH` as defined in Go language

### DIFF
--- a/website/docs/usage.mdx
+++ b/website/docs/usage.mdx
@@ -506,7 +506,7 @@ be skipped, and no error will be thrown.
 
 The values allowed as OS or Arch are valid `GOOS` and `GOARCH` values, as
 defined by the Go language
-[here](https://github.com/golang/go/blob/master/src/go/build/syslist.go).
+[here](https://github.com/golang/go/blob/e705a2d16e4ece77e08e80c168382cdb02890f5b/src/go/build/build.go#L38).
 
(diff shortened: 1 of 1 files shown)


website/docs/usage.mdx
```diff
@@ -506,7 +506,7 @@ be skipped, and no error will be thrown.
 
 The values allowed as OS or Arch are valid `GOOS` and `GOARCH` values, as
 defined by the Go language
-[here](https://github.com/golang/go/blob/master/src/go/build/syslist.go).
+[here](https://github.com/golang/go/blob/e705a2d16e4ece77e08e80c168382cdb02890f5b/src/go/build/build.go#L38).
```
pd93, line 509: I don't think we should point at a specific commit. A branch or tag would be preferable. In fact, maybe we should point at the `pkg.go.dev` docs instead of the code: https://pkg.go.dev/go/build#pkg-types.

andreynering: We used to have a comment with a list of possible GOOS and GOARCH values in that file. It seems that such list is now gone.

I feel that we should maybe change to have a list on our side, but we need to find the original list to copy it.

andreynering:
Closing in favor of #1747

PR #1744, opened 2024-08-08
Set desc to cmd on short form tasks

titpetric:
For a taskfile with short form tasks like this one:

```
---
version: "3"

tasks:
  build: docker compose build
  up: docker compose up -d --wait
  logs: docker compose logs etl -f
  down: docker compose down
  mysql: docker compose exec -it etl-db mysql -u etl -petl etl
```

No tasks with descriptions are available. This change automatically sets the description to the aliased command.

```
# task -t Taskfile.services.yml 
* build:       docker compose build
* down:        docker compose down
* logs:        docker compose logs etl -f
* mysql:       docker compose exec -it etl-db mysql -u etl -petl etl
* up:          docker compose up -d --wait
task: Task "default" does not exist
```

A particular case for this is also that i want `services:*` targets be automatically listed in main taskfile. If a one liner becomes reasonably complex, we still have the option to rewrite it to full form.

### DIFF
--- a/taskfile/ast/task.go
+++ b/taskfile/ast/task.go
@@ -95,6 +95,7 @@ func (t *Task) UnmarshalYAML(node *yaml.Node) error {
 		if err := node.Decode(&cmd); err != nil {
 			return errors.NewTaskfileDecodeError(err, node)
 		}
+		t.Desc = cmd.Cmd
 		t.Cmds = append(t.Cmds, &cmd)
 		return nil
 


vmaerten:
Hello!
Thanks for your PR.
For me it's a breaking changes. This would modify the bahavior of `task -l` which is used by the autocomplete, and some people parse the command's output.
With more than 10 tasks, the `task -l` would take a lot of space in the terminal

If you want to have task listed without description, you can use `task -a` which will list all task even without description.
If you want to see which cmd(s) will be executed, you can run `task --summary TASK`
Last but not least, you can still provide a description with 
```
version: "3"

tasks:
  build:
	cmd: docker compose build
	desc: docker compose build
```
Yes it's a bit longer than the short version but it let use choose if they want or not the description

PR #1745, opened 2024-08-08
Better aliasing for includes

titpetric:
I want to import all the targets for an included taskfile into the current namespace.

```
includes:
  services:
    taskfile: Taskfile.services.yml
    aliases: [""]
```

The aliases feature exists, however when setting it to empty it still includes the `:`, making all of the alias still have the delimiter as the prefix. The change checks for an empty namespace, and removes a redundant sprintf call in the same function.

Before the change:

```
* services:build:       docker compose build                                       (aliases: :build)
* services:down:        docker compose down                                        (aliases: :down)
* services:logs:        docker compose logs etl -f                                 (aliases: :logs)
* services:mysql:       docker compose exec -it etl-db mysql -u etl -petl etl      (aliases: :mysql)
* services:up:          docker compose up -d --wait                                (aliases: :up)
```

After the change:

```
* services:build:       docker compose build                                       (aliases: build)
* services:down:        docker compose down                                        (aliases: down)
* services:logs:        docker compose logs etl -f                                 (aliases: logs)
* services:mysql:       docker compose exec -it etl-db mysql -u etl -petl etl      (aliases: mysql)
* services:up:          docker compose up -d --wait                                (aliases: up)
```

### DIFF
--- a/taskfile/ast/tasks.go
+++ b/taskfile/ast/tasks.go
@@ -156,8 +156,8 @@ func (t *Tasks) UnmarshalYAML(node *yaml.Node) error {
 }
 
 func taskNameWithNamespace(taskName string, namespace string) string {
-	if strings.HasPrefix(taskName, NamespaceSeparator) {
+	if namespace == "" || strings.HasPrefix(taskName, NamespaceSeparator) {
 		return strings.TrimPrefix(taskName, NamespaceSeparator)
 	}
-	return fmt.Sprintf("%s%s%s", namespace, NamespaceSeparator, taskName)
+	return namespace + NamespaceSeparator + taskName
 }


titpetric:
This has some issues:

```
root@carbon:~# task up -v
task: [redacted] Not found - Using alternative (Taskfile.yml)
task: "up" started
task: "up" finished
root@carbon:~/empire/2024-github-etl# task -l
* migrate:              Migrate schema
* up:                   Bring up env
* ingest:github:        Ingest github repository
* ingest:gomod:         Ingest repository go.mod files
* services:build:       docker compose build                                       (aliases: build)
* services:down:        docker compose down                                        (aliases: down)
* services:logs:        docker compose logs etl -f                                 (aliases: logs)
* services:mysql:       docker compose exec -it etl-db mysql -u etl -petl etl      (aliases: mysql)
* services:up:          docker compose up -d --wait                                (aliases: up)
```

Somehow just the alias `up` doesn't get picked up, invoking `task services:up` works. Other aliases seem to work.

Firesphere:
I was literally looking for a solution like this yesterday!

pd93:
I think flattening is the behavior you're looking for: #273, #1704

This has been long requested and is probably coming soon. I don't think `alias` is the right way to do this. In fact, we should probably raise an error when someone does `alias: ""`.

PR #1746, opened 2024-08-09
Recompile task after runDeps to reload envs set by deps

### LINKED ISSUE #1381: dotenv files are loaded before task dependencies are ready
Consider the following example where `make-env` task produces a `.env` file and `my-task` task depends on `make-env` and loads `.env`. I would expect to see `MY_VAR = MY_VALUE` after running the task my-task command, but it is `MY_VAR =  ` instead. It seems like it loads the dotenv file first (but does not fail because the file does not exist yet?) and then executes all dependencies.

```yaml
---
version: 3

tasks:
  make-env:
    cmds:
      - echo "MY_VAR=MY_VALUE" >> ./.env
  my-task:
    deps:
      - make-env
    dotenv:
      - ./.env
    cmds:
      - echo "MY_VAR = $MY_VAR"
```

- Task version: v3.31.0 (h1:o6iyj9gPJXxvxPi/u/l8e025PmM2BqKgtLNPS2i7hV4=)
- Operating system: macOS 13.6 (22G120)
- Experiments enabled: no

hori-ryota:
fix #1381

I would like to create an environment variable file using deps, read it with dotenv, and use it in cmds.

### tested by followings

1. generate `Taskfile.test.yml` (copy from #1381 )

```yaml
version: 3

tasks:
  make-env:
    cmds:
      - echo "MY_VAR=MY_VALUE" >> ./.env
  my-task:
    deps:
      - make-env
    dotenv:
      - ./.env
    cmds:
      - echo "MY_VAR = $MY_VAR"
```

2. run following

```bash
rm -f .env && go run ./cmd/task -t Taskfile.test.yml my-task
```

result:

(before):

```
task: [make-env] echo "MY_VAR=MY_VALUE" >> ./.env
task: [my-task] echo "MY_VAR = $MY_VAR"
MY_VAR =
```

(after):

```
task: [make-env] echo "MY_VAR=MY_VALUE" >> ./.env
task: [my-task] echo "MY_VAR = $MY_VAR"
MY_VAR = MY_VALUE
```

### DIFF
--- a/task.go
+++ b/task.go
@@ -195,6 +195,13 @@ func (e *Executor) RunTask(ctx context.Context, call *ast.Call) error {
 		if err := e.runDeps(ctx, t); err != nil {
 			return err
 		}
+		if len(t.Deps) > 0 {
+			// Recompile the task after running dependencies to reload environment variables from dotenv files which may have been set up during the dependency execution
+			t, err = e.CompiledTask(call)
+			if err != nil {
+				return err
+			}
+		}
 
 		skipFingerprinting := e.ForceAll || (!call.Indirect && e.Force)
 		if !skipFingerprinting {


hori-ryota:
@andreynering @pd93 If this approach doesn't seem optimal, I am considering setting up a custom system to reload dotenv and execute the tasks. I would appreciate your thoughts on whether this might be a viable direction.

hori-ryota:
By the way, dotenv: is checked before `deps:` and used in `deps:` executions, whereas env: seems to be used only in `cmds:` .

hori-ryota:
@andreynering @pd93 What do you think about this pull request?

andreynering:
Hey @hori-ryota, I know we've been slow on reviewing pull requests. We all have full time jobs and a real life happening. I added myself as a reviewer to remember to take a look when possible.

hori-ryota:
Thank you! Please let me know if there's anything else I should do or improve. I really appreciate this excellent tool!

andreynering:
@hori-ryota In the meantime, can you rebase this PR with `main`?

hori-ryota:
rebased.

task.go
```diff
@@ -195,6 +195,13 @@ func (e *Executor) RunTask(ctx context.Context, call *ast.Call) error {
 		if err := e.runDeps(ctx, t); err != nil {
 			return err
 		}
+		if len(t.Deps) > 0 {
+			// Recompile the task after running dependencies to reload environment variables from dotenv files which may have been set up during the dependency execution
+			t, err = e.CompiledTask(call)
+			if err != nil {
+				return err
+			}
+		}
```
andreynering, line 204: This introduces a small performance penalty, and is something that is not needed 99% of the time.

I'm not so sure that it's worth doing for this rare edge case. In this case, it could be better to just run both tasks separately so you ensure that the `.env` file exists before running the actual task.

```bash
task setup && task actual-task
```

I suppose that your `.env` should be generated only once, right? I wouldn't expect a `.env` file to have to be generated on every run.

hori-ryota: Thank you for your feedback. In a large monorepo like ours, each module has its own Taskfile and setup. Being able to run a single task and have all necessary setups execute via dependencies greatly simplifies our workflow. Without this, we’d need to run separate setup tasks for each module, making management and documentation more complicated.

Additionally, our .env requirements change frequently as the team iterates. Running a lightweight, idempotent setup every time ensures our environment is always up-to-date and stable, reducing errors and confusion in the long run. This approach helps keep the entire project’s operations more efficient and maintainable.

trulede:
@hori-ryota I'm interested to take a look at this PR, or create a new version of it. There are a few issues open which I think this could help address.

EvansJahja:
@hori-ryota 
some thoughts
1. do you really need to use a 
.env file? If it were possible to output to a variable (or env var) would it solve your problem?
2. I wonder if this interacts with sources and generates, in order to make sure the system is aware about the refresh

hori-ryota:
@EvansJahja Thanks for the thoughts!

1. On using a `.env` file vs. variables:
The main reason I chose `.env` is that I want those environment variables to be usable outside of Task as well — for example with `direnv` and other tools that naturally consume `.env` files. In our monorepo, we have a common `.env` with shared variables, and each microservice adds its own `.env` on top. Other approaches might work too, but using `.env` feels the most natural fit for this workflow right now.

We use Task as a thin task runner that bundles multiple operations. For instance, we define a `setup` task and a `run-service` task, and put `setup` in the `deps` of `run-service`. That way, a newcomer can just run `run-service` and each microservice starts up with everything ready.

2. On interaction with `sources` / `generates`:
From what I checked, the recompile happens before the fingerprint check, so it looks fine. Even if it turned out to cause problems, I think that would be a separate issue worth solving on its own. If the `.env` reload direction itself gets approved, I'm happy to make sure the implementation plays well with `sources` / `generates`.

hori-ryota:
@trulede Thanks for stepping in!

I'm aware that simply adding a recompile is a somewhat rough solution, and I think there are other approaches worth exploring. I'd have no objection to closing this PR in favor of a new version. Happy to help wherever I can.

Task has grown a lot in functionality, and I feel the processing hierarchy has become a bit complex in places. It might be a good opportunity to reorganize and document that hierarchy as part of the work.

(That complexity reflects years of great feature growth. Huge thanks to the long-time maintainers. :smile:)

PR #1748, opened 2024-08-09
feat: add override as an option to flatten

elocke:
I got excited about the new flatten feature in https://github.com/go-task/task/pull/1704 so this extends that work.

Here's my attempt of what could be extending it to override included & flattened tasks. I have an use case for ci similar to https://github.com/go-task/task/issues/273#issuecomment-565959588. 

It's my first contribution here so let me know what you need.

```txt
Taskfile.yml
- build [echo "override"]

Taskfile.included.yml
- build ["echo build"]
- test ["echo test"]

> task build
      "override"

> task test
      "test"
```


pd93:
I'm concerned that the behavior here in confusing. The Taskfile might look like this:

```yml
includes:
  included:
    taskfile: "./included"
    flatten: true
    override: true
```

It is not clear to me which file is taking priority. I would suggest from this syntax that since the `override` property is specified alongside the included taskfile that the included file would take override the parent, but looking at your code/tests, the opposite is true.

Maybe something along these lines is clearer:

```yml
includes:
  included:
    taskfile: "./included"
    flatten: true
    prefer: "parent"   # or "child"
```

@vmaerten as the implementer of flattening, do you have any opinions? I personally don't have a use for this feature, so it's hard for me to be objective.

vmaerten:
@pd93 I agree with you about the confusion but it's not my main concern.

If we consider one root taskfile that includes two taskfiles, both flatten but one override and the other not, like this : 

```yml
version: '3'

includes:
  one:
    taskfile: ./one.yml
    flatten: true
    override: true
  two:
    taskfile: ./two.yml
    flatten: true

```

one and two have at least one task in common.

Example : 

```yml
version: '3'

task:
  build:
    cmd: echo "build"
```

Depending on which is included first, the behavior won't be the same. And, personally, I think it's an issue.

(This usecase is also valid, with both as override, which one will be taken into account ? One or two ?)

That's why I did not implement this feature at first.

I am not against this feature but I would like to see a real use case for this

PR #1757, opened 2024-08-13
bug: Inconsistent current directory resolution depending on include order

pbitty:
It seems there's a race condition somewhere in the `Reader.include()` code.

I was exploring the behaviour of the `includes.*.dir` options and I wrote some tests.  When I did this, I found there two different behaviours that it would switch between intermittently.  I tried to capture what I think is the expected behaviour in the tests and when you run the tests they will pass about 50% of the time.

I found that if I force the "include loop" to not have any concurrency, the tests pass 100% of the time.

I have tried to play around with the variable assignments and mutations that happen inside `include`, to see if I can isolate the cause, but no luck yet.  In this PR I am posting only the tests and the concurrency fix (commented out), to highlight the problem and see if anyone has any ideas as to what's causing it.

I will continue to explore it - but if anyone has any ideas, let me know.


taskfile/reader.go
```diff
@@ -96,6 +96,13 @@ func (r *Reader) include(node Node) error {
 	// Create an error group to wait for all included Taskfiles to be read
 	var g errgroup.Group
 
+	// It seems there's some data race related to the Go routines concurrenty
+	// processing multiple includes in a single Taskfile.  Uncommenting this makes
+	// the tests pass, which suggests a data race as the culprit.
+	//
+	// Uncomment this to make path-resolution bug go away
+	// g.SetLimit(1)
```
pbitty, line 104: Uncommenting this will stop the bug from happening.

pbitty:
I've been able to narrow down the cause to the way the includes are parsed.  I haven't figured out the root cause, but it seems when that having multiple includes with the same `entrypoint` has something to do with it.

By disabling concurrency and putting the include with an empty `dir` at the end of the list, the directory resolution always works.

I will keep investigating, but if anyone spots the issue in the meantime let me know.

pbitty:
@andreynering @vmaerten , I believe I have figured out the issue.  The relative path resolution in HTTP nodes only works in some scenarios because it always looks "one folder up".  I believe the desired result is to make all paths relative to the parent.

If the parent is a File node, it will stop at the file node.  If the parent is a HTTP node, that node will have taken the value from its parent, and so on.

Does this sound right to you?

I am still familiarizing myself with the graph merging logic, so it's possible I am misunderstanding what is happening.  One thing I still don't fully understand is why the ordering matters.  It seems that the first include in `child-taskfile2.yml` dictates the `dir` for tasks in the remaining includes (since they all point to the same Taskfile in the graph).  How this happens is not entirely clear to me, but it's the behaviour I am seeing from the outside.

pbitty:
I'm still not entirely sure what the expected behaviour is.  I've posted a [question about it in Discourse](https://discord.com/channels/974121106208354339/974121514184085555/1283079003326120036) - will gladly move the discussion here if that's more appropriate.

pbitty:
Whoops, I did it again - I forgot to enable remote taskfile support in the tests.  If no objections, I will cherry-pick the change I did in https://github.com/go-task/task/pull/1771/commits/ec6476195dc26517fa96d89fefdfee67fd74ba27 and then resolve any merge conflicts if both PRs are to be merged.

testdata/includes_http/root-taskfile-remotefile-empty-dir-1st.yml
pbitty, line 1: Without the fix, tests on this file will fail.

vmaerten (approved):
Hey !
I finally got the time to check it
Indeed you are right, when a remote include another remote, the dir is not correct
Thanks a lot for adding test on remote task!

vmaerten:
@pd93 I'll leave this PR open for a few days before merging it, in case you want to review it.

pd93:
@vmaerten Thanks! It's at the top of my list

andreynering:
It looks good to me, so I'm merging.

Thanks @pbitty!

PR #1758, opened 2024-08-14
Update minimum go version

pd93:
[Go 1.23 has been released](https://go.dev/blog/go1.23), so it's time to bump our minimum version again.

Previously, we supported 1.21 and 1.22. This PR bumps the minimum version to 1.22 and drops support for 1.21. Our tests/linter will now run against 1.22 and 1.23.

Bumping the minimum version allows us to use some of the [features introduced in 1.22](https://go.dev/doc/go1.22). There are a number of changes are relevant to this project, including:

- [Ranging over integers directly](https://go.dev/ref/spec#For_range)
  - No more `for i := 0; i < len(s); i++` loops. We can now do `for i := range len(s)`.
- [Changes to how loop variables are set](https://go.dev/blog/loopvar-preview)
  - No need to do weird things like `idx := i` in closures anymore. It just works.
  - I've updated a test to ensure we're getting the correct behavior.
- The addition of [`cmp.Or`](https://pkg.go.dev/cmp#Or).
- The addition of [`slices.Concat`](https://pkg.go.dev/slices#Concat).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.55.2
+          version: v1.60.1
 
   lint-jsonschema:
     runs-on: ubuntu-latest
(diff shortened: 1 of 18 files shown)


andreynering (approved):
Awesome, thanks!

vmaerten (approved):
I saw it on Twitter. I will need to update projects at work too!
Nice work thanks :fire:

PR #1767, opened 2024-08-20
feat: loop over a matrix

### LINKED ISSUE #1766: Feature: matrix support
Hello,

It would be nice to be able to do something like this:

```yaml
env:
  APP: myapp
  CGO_ENABLED: 0
  GOARCH:
    sh: go env GOARCH
  GOOS:
    sh: go env GOOS

tasks:
  build:
    desc: Build application
    cmd: go build -o build/${APP}-${GOOS}-${GOARCH} .

  build:all:
    desc: Build for all targets
    matrix:
      ARCH: [amd64, arm64]
      OS: [linux, darwin, windows]
    env:
      GOARCH: "{{.matrix.ARCH}}"
      GOOS: "{{.matrix.OS}}"
    cmds:
      - task: build
```

Previous issue on the subject: https://github.com/go-task/task/issues/675

Current workaround:

```yaml
  build:all:
    desc: Build for all targets
    cmds:
      - for: [linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, windows/amd64, windows/arm64]
        cmd: >-
          export GOOS=$(echo {{.ITEM}} | cut -d/ -f1);
          export GOARCH=$(echo {{.ITEM}} | cut -d/ -f2);
          go build -o build/${APP}-${GOOS}-${GOARCH} .
```

pd93:
Closes #1766 

This seemed like low-hanging fruit. Very simple implementation. For now it will only accept raw values (no templating or reference resolving or anything like that). This should serve as a good first pass and we can build on it in the future if needed.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -2338,6 +2338,10 @@ func TestForCmds(t *testing.T) {
 			expectedOutput: "a\nb\nc\n",
 		},
 		{
+			name:           "loop-matrix",
+			expectedOutput: "windows/amd64\nwindows/arm64\nlinux/amd64\nlinux/arm64\ndarwin/amd64\ndarwin/arm64\n",
+		},
+		{
 			name:           "loop-sources",
 			expectedOutput: "bar\nfoo\n",
 		},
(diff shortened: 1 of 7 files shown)


andreynering (approved):
👏

PR #1771, opened 2024-08-21
refactor: re-organize node loading code to make it easier to follow

pbitty:
While working in this part of the code I found it difficult to follow the different code paths when loading
a node's content.  I realize readability is subjective, but I am suggesting this change in case you also find it improves readability and maybe maintainability.

In this change I am attempting to make the branches of the decision tree return early as much as possible.  This way we discard each possibility as we work our way down, in this order:

1. Non-remote nodes
2. In offline mode, use cached remote nodes
3. On network timeout, use cached remote nodes
4.
    a. Use fetched remote node
    b. On checksum change, handle writing to cache

Tests are added to ensure this functionality does not break.

I broke up the changes into logical commits that are fairly isolated.  It might be easier to review them individually.

What do you think?

(I got into this code path while exploring how to incorporate go-getter into the remote taskfile feature (as proposed in [this comment](https://github.com/go-task/task/issues/1317#issuecomment-2274443297)).  Thanks for maintaining this tool!)


testdata/includes_remote/Taskfile.yml
```diff
@@ -0,0 +1,4 @@
+version: '3'
+
+includes:
+  first: "{{ .FIRST_REMOTE_URL }}"
```
andreynering, line 4: Missing ending line break. Also, removing spaces to adhere to convention.

```suggestion
  first: "{{.FIRST_REMOTE_URL}}"
```

testdata/includes_remote/first/Taskfile.yml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+
+includes:
+  second: "{{ .SECOND_REMOTE_URL }}"
```
andreynering, line 4:
```suggestion
  second: "{{.SECOND_REMOTE_URL}}"
```

testdata/includes_remote/first/Taskfile.yml
```diff
@@ -0,0 +1,11 @@
...
+
+tasks:
+  write-file:
+    requires:
+      vars: [CONTENT, OUTPUT_FILE]
+    cmd: |
+      echo "{{ .CONTENT }}" > "{{ .OUTPUT_FILE }}"
```
andreynering, line 11:
```suggestion
      echo "{{.CONTENT}}" > "{{.OUTPUT_FILE}}"
```

testdata/includes_remote/first/second/Taskfile.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+
+tasks:
+  write-file:
+    requires:
+      vars: [CONTENT, OUTPUT_FILE]
+    cmd: |
+      echo "{{ .CONTENT }}" > "{{ .OUTPUT_FILE }}"
```
andreynering, line 8:
```suggestion
      echo "{{.CONTENT}}" > "{{.OUTPUT_FILE}}"
```

andreynering (changes requested):
Hi @pbitty!

I like the idea, but there a couple of thing that needs to be addressed before merging:

1) Tests are failing unless `TASK_X_REMOTE_TASKFILES=1` is set
2) Can you re-write the commit messages to comply with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)?

pbitty:
> Hi @pbitty!
> 
> I like the idea, but there a couple of thing that needs to be addressed before merging:

Thanks, will do!

> 
> 1. Tests are failing unless `TASK_X_REMOTE_TASKFILES=1` is set

Ha, I missed this because I have set it permanently in my shell environment.  I figure the best place to add it is in the Go tests themselves.  Let me know if you have another idea.

task_test.go
```diff
@@ -2548,3 +2656,18 @@ func TestReference(t *testing.T) {
...
+// because the experiment settings are parsed during experiments.init(), before any tests run.
+func overrideExperimentValue(t *testing.T, e *experiments.Experiment, enabled bool, val string) {
+	prev := *e
+	*e = experiments.Experiment{
+		Name:    prev.Name,
+		Enabled: enabled,
+		Value:   val,
+	}
+	t.Cleanup(func() { *e = prev })
+}
```
pbitty, line 2673: @andreynering  Are you okay with this work-around?  I figure anything deeper would require changing the `experiments` package and it doesn't seem worth it to accommodate tests.

pbitty:
@andreynering , I did a force-push in order to alter the commit messages.  I could also squash all the commits down to a single one, and call it
```
refactor: Re-organize node loading code to make it easier to follow
```

or something along those lines.

vmaerten:
Hey @pbitty !
Thanks for your contribution!
Like @andreynering, I like the idea, I find this more readable. 
Can you also modify the PR title to follow Conventional Commits? because we will squash and merge.

I would like to merge https://github.com/go-task/task/pull/1713 first as it solves one issue (Timeout message is not correct)
What is your opinion about it @andreynering ?

task_test.go
vmaerten: I think this is good.
Imho, we do not need the `enabled` param, it'll be always true, as all experiments are disabled by default

pbitty: Good point.  I will adjust it.

vmaerten (approved):
I like the idea! Thanks a lot for the tests :fire:

andreynering (approved):
Looks good. It just needs a rebase so we are able to merge it (there are conflicts).

Regarding conventional commits, just note that messages should usually be all lowercase, i.e. `fix: Extract ...` should be `fix: extract ...`, etc.

pbitty:
Just noting that I haven't forgotten about this, just trying to find some time this week to do it.

vmaerten:
Thanks!

PR #1782, opened 2024-09-02
fix(completion): display aliases in fish completion

### LINKED ISSUE #1781: `fish` completions don't report `aliases`
Hey team, 👋 

A small feature request, because I use (and love) task for a long time, and I leverage completions a lot in my day to day work.

However, I'm not able to use `aliases` the same way, because they are ignored completly from `completions` script for my `shell` (that is `fish`). 

See, with this following file: 

```
version: 3

tasks:

  foo:
    aliases:
      - bar
    cmd: echo bar

  another: echo another
```

I have the following result: 

image

The `bar` task, even if available, is totally ignored. 

Can we do something about it?

vmaerten:
Fixes #1781 

Before : 
task_fish

After : 
task_fish_fix

@davinkevin It would be amazing if you could test it before we release this. I test it locally with the docker container it sounds fine

### DIFF
--- a/completion/fish/task.fish
+++ b/completion/fish/task.fish
@@ -10,7 +10,7 @@ function __task_get_tasks --description "Prints all available tasks with their d
   end
 
   # Grab names and descriptions (if any) of the tasks
-  set -l output (echo $rawOutput | sed -e '1d; s/\* \(.*\):\s*\(.*\)\s*(aliases.*/\1\t\2/' -e 's/\* \(.*\):\s*\(.*\)/\1\t\2/'| string split0)
+  set -l output (echo $rawOutput | sed -e '1d; s/\* \(.*\):\s*\(.*\)\s*(\(aliases.*\))/\1\t\2\t\3/' -e 's/\* \(.*\):\s*\(.*\)/\1\t\2/'| string split0)
   if test $output
     echo $output
   end


davinkevin:
I have the same result as you and it looks good to me! 

I was surprised the completion was displayed like this, but it works like expected 😇

PS: I hope those completions will, one day, be part of the product and not volatile scripts maintained aside from the source code.

davinkevin:
Interesting 🤔 

with this file: 

```
version: 3

tasks:

  aaa:
    aliases:
      - bbb
      - ccc
      - ddd
    cmd: echo aaa

  bbb2: echo bbb2
  ccc: echo ccc
```

The completion choose `bbb2` if I type `task b` + `tab`, instead of asking to choose between `bbb2` and `bbb`. 

wdyt?

completion/fish/task.fish
```diff
@@ -10,7 +10,7 @@ function __task_get_tasks --description "Prints all available tasks with their d
   end
 
   # Grab names and descriptions (if any) of the tasks
-  set -l output (echo $rawOutput | sed -e '1d; s/\* \(.*\):\s*\(.*\)\s*(aliases.*/\1\t\2/' -e 's/\* \(.*\):\s*\(.*\)/\1\t\2/'| string split0)
+  set -l output (echo $rawOutput | sed -e '1d; s/\* \(.*\):\s*\(.*\)\s*(\(aliases.*\))/\1\t\2\t\3/' -e 's/\* \(.*\):\s*\(.*\)/\1\t\2/'| string split0)
```
pd93, line 13: Happy with this as a fix, but I think it would be good to move this complexity to Task itself sometime rather than relying on invoking `sed`. Looking at this expression gives me a headache 😆

vmaerten: I totally agree with you, I would even argue that fish / bash / zsh does not support the same "feature" in the auto complete
As you said, we could move all complexity in our go code

vmaerten: I've tested it in zsh, the behavior is the same, aliases are displayed but not "searchable"

pd93:
> PS: I hope those completions will, one day, be part of the product and not volatile scripts maintained aside from the source code.

For the record. I just merged #1157 which is the first step towards this. We can now start the process of generating the scripts instead of using the templates.

vmaerten:
@pd93 What do you think about merging this to have  the same behavior as zsh, and, in the meantime, working on generating the scripts in golang ?

pd93 (approved):
> What do you think about merging this to have  the same behavior as zsh, and, in the meantime, working on generating the scripts in golang ?

@vmaerten Yeah I'm fine with this. This change is still an improvement, even if there are still bigger changes to come.

PR #1784, opened 2024-09-02
fix: matrix loops should be deterministic

pd93:
I noticed that the unit tests for the matrix implementation are flaky. This is because the order of the input maps is not preserved. e.g.

```yaml
  loop-matrix:
    cmds:
      - for:
          matrix:
            OS: ["windows", "linux", "darwin"]
            ARCH: ["amd64", "arm64"]
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
```

can result in both:

```txt
windows/amd64
windows/arm64
linux/amd64
linux/arm64
darwin/amd64
darwin/arm64
```

and:

```txt
windows/amd64
linux/amd64
darwin/amd64
windows/arm64
linux/arm64
darwin/arm64
```

I have made it so that the output is always the former example.

### DIFF
--- a/taskfile/ast/for.go
+++ b/taskfile/ast/for.go
@@ -5,12 +5,13 @@ import (
 
 	"github.com/go-task/task/v3/errors"
 	"github.com/go-task/task/v3/internal/deepcopy"
+	"github.com/go-task/task/v3/internal/omap"
 )
 
 type For struct {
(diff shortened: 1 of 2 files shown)


andreynering (approved):
Good catch!

PR #1701, opened 2024-06-28
fix ordered map race

trim21:
fix https://github.com/go-task/task/issues/1605

suppress https://github.com/go-task/task/pull/1608

add lock to all ordered map

Issue #1605, opened 2024-04-19
fatal error: concurrent map writes


trim21:
- Task version: v3.36.0
- Operating system: windiws
- Experiments enabled: no

please try this branch https://github.com/go-task/task/pull/1701 if you encounter this issue.

I'm running job with `--watch`

```text
fatal error: concurrent map writes

goroutine 10 [running]:
github.com/go-task/task/v3/internal/omap.(*OrderedMap[...]).Set(...)
        github.com/go-task/task/v3@v3.36.0/internal/omap/orderedmap.go:67
github.com/go-task/task/v3.(*Executor).GetTask(0xc00010f7c0, 0xc000071020)
        github.com/go-task/task/v3@v3.36.0/task.go:425 +0x2e8
github.com/go-task/task/v3.(*Executor).compiledTask(0xc00010f7c0, 0xc0001d3e08?, 0x1)
        github.com/go-task/task/v3@v3.36.0/variables.go:30 +0x39
github.com/go-task/task/v3.(*Executor).CompiledTask(...)
        github.com/go-task/task/v3@v3.36.0/variables.go:21
github.com/go-task/task/v3.(*Executor).registerWatchedFiles.func1(0xc000037280?)
        github.com/go-task/task/v3@v3.36.0/watch.go:127 +0x5e
github.com/go-task/task/v3.(*Executor).registerWatchedFiles(0xc00010f7c0, 0xc000037280, {0xc00006c298, 0x1, 0x0?})
        github.com/go-task/task/v3@v3.36.0/watch.go:176 +0xb3
github.com/go-task/task/v3.(*Executor).watchTasks.func3()
        github.com/go-task/task/v3@v3.36.0/watch.go:95 +0x72
created by github.com/go-task/task/v3.(*Executor).watchTasks in goroutine 1
        github.com/go-task/task/v3@v3.36.0/watch.go:92 +0x515

goroutine 1 [sleep]:
time.Sleep(0x12a05f200)
        runtime/time.go:195 +0x126
github.com/radovskyb/watcher.(*Watcher).Start(0xc000037280, 0x12a05f200)
        github.com/radovskyb/watcher@v1.0.7/watcher.go:608 +0x105
github.com/go-task/task/v3.(*Executor).watchTasks(0xc00010f7c0, {0xc00006c298, 0x1, 0x1})
        github.com/go-task/task/v3@v3.36.0/watch.go:102 +0x525
[...]

PR #1797, opened 2024-09-09
feat: use external package for ordered maps

pd93:
This PR completely removes and replaces our native `omap` package with a 3rd-party one. There are a few reasons for this change:

- It is not very performant compared to other implementations.
- An internal implementation leads to very bespoke code whereas using a generic package enforces a cleaner separation between the omap implementation detail and Task features.
- One less thing for us to maintain. reuse > rewrite.

This PR also makes way for the use of iterators (which we cannot use until we bump our minimum version to 1.23). A working PoC of this is available in #1798.

This also has a direct knock-on effect to #1605 - though falls short of fixing it. @trim21 has #1701 open to address this. I'm afraid that this will need rethinking as the fix was in our custom `omap` implementation previously. Now that we have a bit more separation, we will need to bring the fixes into our implementations of `Tasks`, `Includes` and `Vars`. This falls inline with the [thoughts of the creator](https://github.com/elliotchance/orderedmap/issues/35#issuecomment-1852820693) of the `orderedmap` package we are moving to. That is, concurrency should be fixed by the implementation, not the ordered map itself. This is similar to how a Go `map` does not have built-in concurrency protection.

---

As for the code, I chose [elliotchance/orderedmap](https://github.com/elliotchance/orderedmap) for a couple of reason over other packages:

- It maintains `O(1)` over `Set`, `Get`, `Delete` and `Len` operations.
- Supports iterators
- Good API design
- Well maintained and reliable author

I have opened a couple of issues on the project that would ease our integration. However, neither of there are dealbreakers - Just nice-to-have:

- https://github.com/elliotchance/orderedmap/issues/43
- https://github.com/elliotchance/orderedmap/issues/44).

---

One additional note. We now use custom unmarshalling on all implementations of ordered maps (previously this was only on `Includes`). This is because the `orderedmap` package does not natively support is. It may do in the future, but this doesn't matter to us as the custom unmarshal functions actually tidy up some of our code. Generally speaking, I have tried to tidy up and make all the implementations of ordered maps in our code a bit more consistent.


vmaerten (approved):
Great description :clap: and great PR. It fixes the issue and tidy up our code.
I totally agree with you on those 3 points. Also the author seems to be reactive (as he already answered one of your issue)

Globally happy with this changes.

Looking forward to be able to use the iterator :smile:

trim21:
Please don't do it by just changing all `internal/omap` to new package.

This ordered maps doesn't have mutex and is not goroutine safe, so we still need to add mutex to fix [fatal error: concurrent map writes](https://github.com/go-task/task/issues/1605) in the future, and that's why we change omap in the first please.

This means we will still need to add mutex to ordered map struct in the future. and it's much easier to add it with `internal/omap/`.

If you don't want to add mutex first, we still should change internal implment of `internal/omap` instead of removing it.

pd93:
@trim21 I'm working on a separate branch to add the mutexes. I didn't want to make this PR even larger than it already is though.

I want to get both merged before we release again.

pd93:
> swap internal implement of omap will only make it smaller, not bigger.

I mean I don't want to make this PR bigger by adding the mutexes to the implementations now. It will make the refactor and the fix harder to review. I intend to release both at the same time.

I do not want to have an internal omap package for the reasons stated in the PR description.

andreynering (approved):
👏

PR #1798, opened 2024-09-09
feat: iterators

pd93:
This is a follow-up to #1797 and #1972 which adds support for iterators in our codebase. This tidies up a lot of code (We have a _lot_ of `Range` functions) and makes control-flow a bit easier in loops.

I have also slightly refactored the `sort` package to be slightly more generic and use a function type rather than an interface. The interface was not necessary and increased the code complexity.

~~This is a Go 1.23 feature which requires a bump to our minimum Go version hence the draft PR (Obviously the tests will fail as Go 1.24 does not currently exist).~~ Done.


taskfile/ast/graph.go
```diff
@@ -116,14 +116,14 @@ func (tfg *TaskfileGraph) Merge() (*Taskfile, error) {
 		return nil, err
 	}
 
-	_ = rootVertex.Taskfile.Tasks.Range(func(name string, task *Task) error {
+	// TODO: I don't think this is necessary anymore
```
vmaerten, line 119: Do you want to resolve / check this TODO in another PR ?

vmaerten (approved):

taskfile/ast/graph.go
pd93: This was a note to myself and I just forgot to come back to it. `Task.Task` is populated when the `Tasks` parent type is decoded so this is no longer needed. I've rebased the PR and removed this section.

PR #1804, opened 2024-09-12
Fix bug for running tasks twice

### LINKED ISSUE #1498: `run: once` not working for a task included multiple times
- Task version:
```
task --version    
Task version: v3.34.1 (h1:yAAxUM54zoaHv+OtDnGgkWSVeiRuaOCn1lPUXPQQA0o=)
```
- Operating system:
```
lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: Fedora
Description:    Fedora release 38 (Thirty Eight)
Release:        38
Codename:       ThirtyEight
```
- Experiments enabled: none

When multiple task files include the same file and then depend on the same task, the depended on task gets executed twice even when `run: once` is set.

```yaml
./tools/input

./tools/taskfile.yml
version: "3"

tasks:
  exe:
    run: once
    sources:
      - ./input
    generates:
      - ./output
    cmds:
      - touch output

./lib/a/taskfile.yml
version: "3"

includes:
  tools:
    taskfile: ../../tools
    dir: ../../tools
    internal: true

tasks:
  lib:
    deps:
      - tools:exe
    cmds:
      - touch a

./lib/b/taskfile.yml
version: "3"

includes:
  tools:
    taskfile: ../../tools
    dir: ../../tools
    internal: true

tasks:
  lib:
    deps:
      - tools:exe
    cmds:
      - touch b

./taskfile.yml
version: "3"

tasks:
  all:
    deps:
      - a:lib
      - b:lib

includes:
  a:
    taskfile: ./lib/a
    dir: ./lib/a

  b:
    taskfile: ./lib/b
    dir: ./lib/b
```

```
task all
task: [a:tools:exe] touch output
task: [b:tools:exe] touch output
task: [b:lib] touch b
task: [a:lib] touch a
```

In the setup I have 4 taskfiles. tools/taskfile.yml defines building a tool that is needed to build the a and b lib. lib/a/taskfile.yml and lib/b/taskfile.yml both include tools/taskfile.yml and depend on the building of the tool. The root taskfile.yml includes lib/a and lib/b and defines an all task to build both lib/a and lib/b.

Even with `run: once` set in the task that builds the tool, the task gets run twice.

Note that if the taskfiles are collapsed into a single task file, this issue is not present.
```yaml
version: "3"

tasks:
  tools:exe:
    run: once
    sources:
      - ./tools/input
    generates:
      - ./tools/output
    cmds:
      - touch ./tools/output

  a:lib:
    deps:
      - tools:exe
    cmds:
      - touch lib/a

  b:lib:
    deps:
      - tools:exe
    cmds:
      - touch lib/b

  all:
    deps:
      - a:lib
      - b:lib
```
```
task all
task: [tools:exe] touch ./tools/output
task: [a:lib] touch lib/a
task: [b:lib] touch lib/b
```

libre-man:
This fixes #1498 by keeping track of the namespaces of a task in an array, instead of just the last one.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -37,6 +37,18 @@ type SyncBuffer struct {
 	mu  sync.Mutex
 }
 
+func (sb *SyncBuffer) String() string {
+	sb.mu.Lock()
+	defer sb.mu.Unlock()
+	return sb.buf.String()
+}
+
+func (sb *SyncBuffer) Reset() {
+	sb.mu.Lock()
+	defer sb.mu.Unlock()
+	sb.buf.Reset()
+}
+
 func (sb *SyncBuffer) Write(p []byte) (n int, err error) {
 	sb.mu.Lock()
 	defer sb.mu.Unlock()
@@ -179,7 +191,7 @@ func TestSpecialVars(t *testing.T) {
 	for _, dir := range []string{dir, subdir} {
 		for _, test := range tests {
 			t.Run(test.target, func(t *testing.T) {
-				var buff bytes.Buffer
+				var buff SyncBuffer
 				e := &task.Executor{
 					Dir:    dir,
 					Stdout: &buff,
@@ -188,7 +200,7 @@ func TestSpecialVars(t *testing.T) {
 				}
 				require.NoError(t, e.Setup())
 				require.NoError(t, e.Run(context.Background(), &ast.Call{Task: test.target}))
-				assert.Equal(t, test.expected+"\n", buff.String())
+				assert.Equal(t, test.expected+"\n", buff.buf.String())
 			})
 		}
 	}
@@ -286,7 +298,7 @@ func TestStatus(t *testing.T) {
 		}
 	}
 
-	var buff bytes.Buffer
+	var buff SyncBuffer
 	e := &task.Executor{
 		Dir: dir,
 		TempDir: task.TempDir{
@@ -322,7 +334,7 @@ func TestStatus(t *testing.T) {
 	require.NoError(t, e.Run(context.Background(), &ast.Call{Task: "gen-bar"}))
 
 	// We're silent, so no output should have been produced.
-	assert.Empty(t, buff.String())
+	assert.Empty(t, buff.buf.String())
 
 	// Now, let's remove source file, and run the task again to to prepare
 	// for the next test.
@@ -371,7 +383,7 @@ func TestStatus(t *testing.T) {
 func TestPrecondition(t *testing.T) {
 	const dir = "testdata/precondition"
 
-	var buff bytes.Buffer
+	var buff SyncBuffer
 	e := &task.Executor{
 		Dir:    dir,
 		Stdout: &buff,
@@ -486,7 +498,7 @@ func TestStatusChecksum(t *testing.T) {
 				require.Error(t, err)
 			}
 
-			var buff bytes.Buffer
+			var buff SyncBuffer
 			tempdir := task.TempDir{
 				Remote:      filepathext.SmartJoin(dir, ".task"),
 				Fingerprint: filepathext.SmartJoin(dir, ".task"),
@@ -1735,7 +1747,7 @@ func TestRunOnlyRunsJobsHashOnce(t *testing.T) {
 func TestRunOnceSharedDeps(t *testing.T) {
 	const dir = "testdata/run_once_shared_deps"
 
-	var buff bytes.Buffer
+	var buff SyncBuffer
 	e := task.Executor{
 		Dir:      dir,
 		Stdout:   &buff,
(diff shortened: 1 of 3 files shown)


libre-man:
I'm not entire sure what the test failure is here, can someone shed some light on that? Locally all tests pass.

pd93:
It's not clear to me what is being fixed here since #1498 is already resolved. Could you please provide an example that reproduces your specific issue. Please also read the discussion in #1655 as a decision was made to not change some behaviour on purpose.

libre-man:
I created this PR before that ticket was closed. The issue here is when we have a structure like this
```
FileA |
      Includes -> FileB
                | Includes -> FileC
                            | -> Has Task A
      Includes -> FileC
                | -> Has Task A
```

If now we call Task A through FileA -> FileB -> FileC and through FileA -> FileC we get Task A to run multiple times. This is especially and issue when Task A is _not_ idempotent or not safe to run concurrently. A prime example of this is installing something with `npm ci`.

task_test.go
```diff
@@ -37,6 +37,18 @@ type SyncBuffer struct {
...
+	defer sb.mu.Unlock()
+	return sb.buf.String()
+}
+
+func (sb *SyncBuffer) Reset() {
+	sb.mu.Lock()
+	defer sb.mu.Unlock()
+	sb.buf.Reset()
+}
+
```
pd93, line 51: I don't think these are necessary. You can just call `buff.buf.Reset()` like you have done elsewhere. The only operation that needs a mutex is `Write()`.

taskfile/ast/task.go
```diff
@@ -219,7 +221,7 @@ func (t *Task) DeepCopy() *Task {
 		Platforms:            deepcopy.Slice(t.Platforms),
 		Location:             t.Location.DeepCopy(),
 		Requires:             t.Requires.DeepCopy(),
-		Namespace:            t.Namespace,
+		Namespace:            append([]string{}, t.Namespace...),
```
pd93, line 224: This change isn't necessary.

```suggestion
		Namespace:            t.Namespace,
```

pd93:
Here is a reproduction of your issue as I understand it:

```yaml
# Taskfile.yml
version: "3"

includes:
  b: ./b.yml
  c: ./c.yml

tasks:
  default:
    deps:
      - abc
      - ac

  abc:
    cmds:
      - task: b:c:task-a

  ac:
    cmds:
      - task: c:task-a
```

```yaml
# b.yml
version: "3"

includes:
  c: ./c.yml
```

```yaml
# c.yml
version: "3"

tasks:
  task-a:
    run: once
    sources:
      - ./input
    generates:
      - ./output
    cmds:
      - touch output
```

```shell
❯ task -d ./tmp/1804 abc
task: [b:c:task-a] touch output

❯ task -d ./tmp/1804 ac
task: [c:task-a] touch output          # Runs task again
```

This is currently working as intended (See discussion in #1655), although we have left the door open to discussion. We made the decision that we would not prevent these tasks from running since they have different calling paths. It is possible that the tasks can contain different data etc and it is safer to run a task twice than risk it not running when it should have.

This needs more discussion before it can be merged, but I have left a couple of comments. @vmaerten, @andreynering, do you have any additional thoughts beyond what has been previously discussed? Note that this implementation appears to follow option 3 as described in #1655 i.e. Whether the task is run or not is only affected when `run: once` is enabled. The behavior does not appear to change if this is omited.

@libre-man please feel free to speak up if you disagree with any of the comments made previously. Also the PR could probably do with a new test that replicates this issue to ensure that we don't get regressions in the future it this gets merged.

taskfile/ast/task.go
libre-man: But this is not an array, which would mean that a deep copy would need to deep copy the array too right?

libre-man:
The main issue is that `run: once` as I see it is used to say "I only want this test to be run once", which it currently doesn't do. I would even say that for `run: when_changed` the current behaviour of saying that different calling paths are different tasks don't make a lot of sense, however I might be wrong of course. The task, in my opinion, should care where it is called from, only which arguments it gets.

My use case is this: I have a file `Taskfile.setup.yml` that is imported by many other taskfiles. It is responsible for setting up stuff like Python, pnpm, smithy. Making all those scripts safe to be run concurrently is not really feasible, I would need to use file system locks and it would get way too complicated very quickly. With the current implementation there is no way around this, and it kinda breaks my encapsulation that every task is responsible for setting up their own dependencies. 

If the change gets the OK I'll polish the PR with tests too, I'm waiting to do that until the OK so I don't waste any time on that if it won't get merged anyway.

taskfile/ast/task.go
pd93: Ah yes sorry, you are right. We have `deepcopy.Slice()` for this.

```suggestion
		Namespace:            deepcopy.Slice(t.Namespace),
```

erictooth:
Just to add to the discussion, I'm running into what sounds like a similar (the same?) situation where some taskfiles get included by multiple other taskfiles, and despite them having `run: once`, the same task (`pkg_a:build`) runs multiple times in parallel:

`pkg_b:pkg_a:build`
`pkg_c:pkg_b:pkg_a:build`
`pkg_d:pkg_c:pkg_b:pkg_a:build`

The only workaround I've been able to think of is renaming the task `pkg_a_build` and then including the pkg_a taskfile with `flatten: true`. This causes the task to be properly deduplicated across the chained includes and then it runs only once.

vmaerten:
Thanks for your PR, this has been fixed

libre-man:
This is actually not fixed it seems, we are still seeing this issue

@vmaerten which commit is meant to fix this.

PR #1807, opened 2024-09-14
Fix zsh completion

shirayu:
The call of `_arguments` causes an error like this and the completion does not work in zsh.

```console
$  zsh --version
zsh 5.9 (x86_64-debian-linux-gnu)

$  eval "$(task --completion zsh)"
_arguments:comparguments:327: can only be called from completion function
```

This PR fixes it.
I refereed to [mise's completion](https://github.com/jdx/mise/blob/c73ecd1aa55f6e2c29da5d7314a207af4bb2010f/completions/_mise) and this fix seems to work fine.

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -39,26 +39,36 @@ function __task_list() {
     _describe 'Task to run' scripts
 }
 
-_arguments \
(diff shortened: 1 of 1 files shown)


vmaerten:
Thanks for your contribution, I've created https://github.com/go-task/task/pull/1809 to fix also powershell (and clean up a bit zsh's completion)
I'll credit you in the changelog

Issue #1795, opened 2024-09-09
Wildcards and fingerprinting: consider to automatically concat the wildcard parameter on the cache key


andreynering:
Discussed in https://github.com/go-task/task/discussions/1794

PR #1808, opened 2024-09-15
feat: improve fingerprint and output with wildcard

vmaerten:
Fixes : 
- https://github.com/go-task/task/issues/1795

It also improve the output, now the task with wildcard replaced are displayed : 

image

I was struggled a bit with the naming. I choose `Fullname` but feel free if you have a better idea.


task.go
```diff
@@ -447,6 +446,7 @@ func (e *Executor) GetTask(call *ast.Call) (*ast.Task, error) {
 			call.Vars = &ast.Vars{}
 		}
 		call.Vars.Set("MATCH", ast.Var{Value: matchingTasks[0].Wildcards})
+		matchingTasks[0].Task.FullName = call.Task
```
andreynering, line 449: It looks like we'd be getting a race condition here, no? Parallel calls to this task would be overiding the same attribute.

Tricky to implement this in another way, though, I know...

vmaerten: You are totally right. The same task with a wildcard, used twice in a dep (ran in parallel) the first one will be overridden by the last

Put this in draft until I find a solution for it

Thanks!

vmaerten:
Finally took the time to finish this!
I've also fix another issue with this

PR #1809, opened 2024-09-15
fix(completions): zsh and powershell completions

### LINKED ISSUE #1796: New PowerShell recommended profile does not seem to work on Linux
I put `Invoke-Expression (&task --completion powershell)` on the profile, then I see when starts pwsh (PowerShell 7.4.5):

```log
Line |
   2 |  Invoke-Expression (&task --completion powershell)
     |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot convert 'System.Object[]' to the type 'System.String' required by parameter
     | 'Command'. Specified method is not supported.
```

- Task version: v3.39.0
- Operating system: Debian/Linux
- Experiments enabled: ?

vmaerten:
Fixes #1796
If fixes for powershell but also for zsh

Based on #1807 , and github cli (`echo "$(gh completion -s zsh)"`)
I removed local that were not used.

I  also checked for fish and bash, it works well.

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -1,6 +1,5 @@
 #compdef task
-
-local context state state_descr line
+compdef _task task
 typeset -A opt_args
 
 _GO_TASK_COMPLETION_LIST_OPTION="${GO_TASK_COMPLETION_LIST_OPTION:---list-all}"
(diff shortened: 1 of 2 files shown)


pd93 (approved):

pd93:
@vmaerten Do you think it is worth doing a patch release for this?

vmaerten:
@pd93 As the "old" method still works well, I would say it can wait

We also could remove this option in the website waiting for the new release (3.40). This way, our website will be up to date.

But I am ok with releasing a patch, up to you (and @andreynering)

PR #1810, opened 2024-09-15
fix: special variables are defined with dotenv at task level

### LINKED ISSUE #1232: [BUG] Enable dotenv will make go template pipeline be invalid.
- Task version: v3.25.0 (h1:XtIEtWpQSbIf2FGU+44/QB1whFf2sBUB/02W7ZQSz0U=)
- Operating system: Windows 10
- Experiments enabled: false

---

## How to reproduce it?

When I don't include dotenv key in config file, all things work well:

```yaml
version: '3'

vars:
  ROOT: 
    sh: echo {{toSlash .ROOT_DIR}}
  DOCKER: 
    sh: echo {{.ROOT}}/docker

tasks:
  foo:
    cmds: 
      - echo {{.DOCKER}}
```

```shell
# in my directory: D:/Reops/test
$ task foo
task: [foo] echo D:/Repos/test/docker
D:/Repos/test/docker
```

But when I add `dotenv` keys in config file with an **empty** `.env` file(created but no content):

```yaml
version: '3'

dotenv: 
  - .env

vars:
  ROOT: 
    sh: echo {{toSlash .ROOT_DIR}}
  DOCKER: 
    sh: echo {{.ROOT}}/docker

tasks:
  foo:
    cmds: 
      - echo {{.DOCKER}}
```

then run task again and the console will show:

```
$ task foo
template: :1:15: executing "" at <.ROOT_DIR>: invalid value; expected string
```

vmaerten:
fixes #1232 

What caused the bug? In dotenv, we use the templater (to allow the use of variables in dotenv's path), and if the user includes a special variable in their template, there was a template error because special variables were not defined.

Special variables that are not linked to a task and/or a call are always available.

### DIFF
--- a/internal/compiler/compiler.go
+++ b/internal/compiler/compiler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


andreynering (approved):
👏

Issue #1803, opened 2024-09-11
[v3.39.0] defer no longer evaluates variables


robcao:
`defer` is no longer evaluating variables in task 3.39.0. In the sample `Taskfile.yml` below, I've created a simple reproduction.

To reproduce:

Run `task my-repro` with the taskfile below.

See the following output is 

```bash
my-value

```

Expected behavior (+ the behavior on v3.38.0):

```bash
my-value
my-value
```
```yaml
# https://taskfile.dev

version: '3'
silent: true
tasks:
  my-repro:
    cmds:
      - defer: echo {{ .MY_VAR }}
      - echo {{ .MY_VAR }}
    vars:
      MY_VAR: my-value

```

- Task version: 3.39.0
- Operating system: wsl2/ubuntu22.04, Shell: GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
- Experiments enabled:

robcao:
I believe this was the change that introduced the behavior change: https://github.com/go-task/task/pull/1762

I pulled down the commit before the change: `git co 35119c12ab222fc5a2a1d28536f5e575136e103e`, and validated that the command works as expected on this commit via `task my-repro`.

Pulling down the change: `git co b259edeb65e24f28c41604b9819869aca082d3b6` and running the `task my-repro` shows the erroneous behavior

mgbowman:
`+1` 

* Task version: `v3.39.0 (h1:Loe6ppP1x38Puv1M2wigp91bH/garCt0vLWkJsiTWNI=)`
* Operating system: `Linux ubuntu20 6.8.0-40-generic #40~22.04.3-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 30 17:30:19 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
* Experiments enabled: `ENV_PRECEDENCE GENTLE_FORCE REMOTE_TASKFILES`

mgbowman:
This seems like a major regression that should have been caught by a test case.

All my `defer: rm -f "{{.VAR}}"` statements that are now evaluating to `rm -f ""` 🙃

kevintijssen:
Just updated to Task version: v3.39.1 (h1:yIaLD5TZWXU/SdmadBReQKu8jblRxzRT+lJG4xC3CMw=) but I still have the same issue

mgbowman:
[...]

PR #1814, opened 2024-09-18
fix: interpolate vars in defer

vmaerten:
Fixes https://github.com/go-task/task/issues/1803

@andreynering I am not sure about this part.

If it does not look good, feel free to re-do it from scratch


andreynering:
@vmaerten Thanks!  I just pushed a change to make the `e.GetTask(call)` call more scoped.

Issue #1822, opened 2024-09-19
Cannot use splitArgs and splitLines in for-loops


stawii:
Both `splitArgs` and `splitLines` functions returns `[]string` type, which is not `[]any`, causing error:
```
loop var must be a delimiter-separated string, list or a map
```

Simple test:
```yaml
  test-fails:
    vars:
      SOME_LINES: |-
        hello world!
        lorem ipsum
        dolor sit amet

      LINES_LIST:
        ref: splitLines .SOME_LINES
    cmds:
      - cmd: echo LINE {{ q .ITEM }}
        for:
          var: LINES_LIST
```

Probably wrapping with ANY (pun intended) list function will do the trick. I've checked with:
```yaml
      LINES_LIST:
        ref: reverse (reverse (splitLines .SOME_LINES))
```
and:
```yaml
      LINES_LIST:
        ref: compact (splitLines .SOME_LINES)
```
both are working and result is as expected:
```raw
task: "test-fails" started
task: [test-fails] echo LINE 'hello world!'
LINE hello world!
task: [test-fails] echo LINE 'lorem ipsum'
LINE lorem ipsum
task: [test-fails] echo LINE 'dolor sit amet'
LINE dolor sit amet
task: "test-fails" finished
```

Please check:
* https://github.com/go-task/task/blob/58ab26c4abcaff48d99aeaf7d333b3e1e08d0438/variables.go#L312
* https://github.com/go-task/task/blob/58ab26c4abcaff48d99aeaf7d333b3e1e08d0438/internal/templater/funcs.go#L26
* https://github.com/go-task/task/blob/58ab26c4abcaff48d99aeaf7d333b3e1e08d0438/internal/templater/funcs.go#L45

I assume changing case stanza from `case []any:` to `case []any, []string:` will solve the issue - but I'm not sure about side effects. EDIT: it doesn't work without type casting, proper code in #1823.

- Task version: 3.39.2
- Operating system: MacOS
- Experiments enabled: no

PR #1823, opened 2024-09-19
Fix: Cannot use splitArgs and splitLines in for-loops

stawii:
Solves https://github.com/go-task/task/issues/1822


pd93 (changes requested):
The fix looks good to me. 1 small request though:

Can we add a case for `[]int` too since we have functions like `until` and `untilStep` which will have the same issue.

pd93 (approved):
Looks good, thanks!

PR #1824, opened 2024-09-20
fix: snapshot builds and wrong winget version

caarlos0:
refs https://github.com/goreleaser/goreleaser/issues/5142

docs should be better, but its not a good idea to use `tag` in the snapshot version template

this should fix the issue you're having


andreynering:
Thanks @caarlos0!

PR #1827, opened 2024-09-20
feat: option to ensure variable is within the list of values

vmaerten:
An extension of requires vars. 

Added the ability check if a variable is one of allowed values.  It only works for string right now.

It useful when we call a task from the CLI using a var. For example : 

```yaml
version: '3'
tasks:
  deploy:
    cmds:
      - echo "deploying to {{.ENV}}"
    requires:
      vars:
        - name: ENV
          enum: [dev, beta, prod]
```

I want to be sure that the user can only deploy to those envs.


mgbowman:
First off, awesome idea! 💪🏻

Maybe it's just me, but `allowed_values` seems a bit verbose. 

What do you think about `values` or `choices` or `options`?

pd93:
> Maybe it's just me, but allowed_values seems a bit verbose.
>
> What do you think about values or choices or options?

jsonschema uses `enum` for this: https://json-schema.org/understanding-json-schema/reference/enum

vmaerten:
I like both `values` and `enum`

pd93:
I'm fine with either. The only thing I will say in favor of `enum` or `choices` etc is that it makes it obvious that there is a constraint. `values` could be misinterpreted. i.e. are they allowed values? default values, will it loop over them etc.

andreynering:
Between these options, I like `enum` the most.

vmaerten:
I've asked to chatGPT (and some friends)

```
If clarity is your priority, allowed_values is solid.
If you're looking for something shorter but still clear, choices or enum are great picks. Enum has the advantage of being standardized in some communities.
```

Naming is hard (and cannot be modified after), we need to choose it carefully.

I think I'll go for `enum`

andreynering (approved):

PR #1830, opened 2024-09-22
fix: checksum override when passed via ldflags

grawlinson:
PR #1711 fixed variable `version` being overridden, this PR ensures that the same does not happen to `sum`.

EDIT: This patch is being shipped in the `go-task` package for Arch Linux.


vmaerten (approved):
Thanks!

PR #1833, opened 2024-09-23
fix(remote): wait for prompt in the reader

### LINKED ISSUE #1832: With multiple remote Taskfile, the confirmation prompt is not blocking
If we have multiple remote taskfile to confirm, all go routine run in parallele and the confirmation prompt is not blocking. This lead to a weird behavior : 

image

- Task version: 3.39.2
- Operating system: Linux
- Experiments enabled: Remote

vmaerten:
Fixes #1832 
I've choose to re-use the mutex on the graph.

Maybe you have a better solution @pd93 

image

### DIFF
--- a/taskfile/reader.go
+++ b/taskfile/reader.go
@@ -244,13 +244,15 @@ func (r *Reader) readNode(node Node) (*ast.Taskfile, error) {
 				// If there is a cached hash, but it doesn't match the expected hash, prompt the user to continue
 				prompt = fmt.Sprintf(taskfileChangedPrompt, node.Location())
 			}
-			r.graph.Lock()
+
 			if prompt != "" {
+				r.graph.Lock()
 				if err := r.logger.Prompt(logger.Yellow, prompt, "n", "y", "yes"); err != nil {
+					r.graph.Unlock()
 					return nil, &errors.TaskfileNotTrustedError{URI: node.Location()}
 				}
+				r.graph.Unlock()
 			}
-			r.graph.Unlock()
 
 			// If the hash has changed (or is new)
 			if checksum != cachedChecksum {
--- a/taskfile/reader.go
+++ b/taskfile/reader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/dominikbraun/graph"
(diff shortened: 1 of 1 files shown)


taskfile/reader.go
```diff
@@ -244,10 +244,14 @@ func (r *Reader) readNode(node Node) (*ast.Taskfile, error) {
 				// If there is a cached hash, but it doesn't match the expected hash, prompt the user to continue
 				prompt = fmt.Sprintf(taskfileChangedPrompt, node.Location())
 			}
+
 			if prompt != "" {
+				r.graph.Lock()
```
pd93, line 249: > I've choose to re-use the mutex on the graph.
>
> Maybe you have a better solution @pd93

We probably shouldn't lock all graph writes just for the prompts IMO. For example if the prompt for Taskfile A is accepted, we shouldn't block writing it to the graph because the a prompt for Taskfile B has not been accepted yet. A separate prompt `sync.Mutex` is probably a good idea.

Independently, we might want to consider changing the regular graph mutex to a `sync.RWMutex` at some point. Reads don't need to block one another. I'm not sure why I didn't do this originally.

taskfile/reader.go
```diff
@@ -244,10 +247,14 @@ func (r *Reader) readNode(node Node) (*ast.Taskfile, error) {
 				// If there is a cached hash, but it doesn't match the expected hash, prompt the user to continue
 				prompt = fmt.Sprintf(taskfileChangedPrompt, node.Location())
 			}
+
 			if prompt != "" {
+				r.promptMutex.Lock()
 				if err := r.logger.Prompt(logger.Yellow, prompt, "n", "y", "yes"); err != nil {
+					r.promptMutex.Unlock()
```
vmaerten, line 254: I thought about using defer, but the return is just after this. Not sure if it's more clear using defer or not

taskfile/reader.go
vmaerten: > A separate prompt sync.Mutex is probably a good idea.

:+1: It was my second idea
I've done It

> Independently, we might want to consider changing the regular graph mutex to a sync.RWMutex at some point. Reads don't need to block one another. I'm not sure why I didn't do this originally.

Yeah agree, let's refactor this in another PR :)

taskfile/reader.go
```diff
@@ -50,14 +52,15 @@ func NewReader(
...
-		logger:   logger,
+		graph:       ast.NewTaskfileGraph(),
+		node:        node,
+		insecure:    insecure,
+		download:    download,
+		offline:     offline,
+		timeout:     timeout,
+		tempDir:     tempDir,
+		logger:      logger,
+		promptMutex: sync.Mutex{},
```
pd93, line 63: Nice. I like that this is on the reader instead of the graph 💯

taskfile/reader.go
vmaerten: I like your arguments, I've tested it and it works

Thanks a lot for your reviews :heart:

pd93 (approved):

Issue #1819, opened 2024-09-19
Flaky TestForDeps/loop-matrix test


Ikke:
- Task version: 3.39.x
- Operating system: Alpine Linux
- Experiments enabled: None

I'm maintaining the Alpine Linux package for task. While upgrading to 3.39.1 and 3.39.2, the TestForDeps/loop-matrix failed occasionally with:

```
=== RUN   TestForDeps/loop-matrix
    task_test.go:2495: 
        	Error Trace:	github.com/go-task/task/v3/task_test.go:2495
        	Error:      	"linux/arm64\ndarwin/amd64\ndarwin/arm64\nwindows/amd64\nlinux/amd64windows/arm64\n\n" does not contain "linux/amd64\n"
        	Test:       	TestForDeps/loop-matrix
```

Retrying the build fixes it. This may indicate some kind of race condition. The failure did happen on somewhat slower arches (armv7, riscv64).

pd93:
Strange, the test uses a `SyncBuffer` that should only be able to write one thing at a time, so it's odd that you're getting `linux/amd64windows/arm64\n\n` in your output instead of `linux/amd64\nwindows/arm64\n`. I suppose new newline is written separately somewhere. I'm guessing the quick fix is to remove the `\n` from the test, but this feels like sweeping the issue under the rug.

Do you know if this can happen in versions prior to 3.39.1 or was it introduced in that version?

Ikke:
On 2.38.0, running test with -count 100:

```
    task_test.go:2399:
                Error Trace:    /home/buildozer/aports/community/go-task/src/task-3.38.0/task_test.go:2399
                Error:          "21\n\n3\n" does not contain "2\n"
                Test:           TestForDeps/loop-different-tasks
--- FAIL: TestForDeps (0.08s)
```

On 2.37.0 additionally:

```
=== RUN   TestForDeps/loop-explicit
    task_test.go:2326:
                Error Trace:    /home/buildozer/aports/community/go-task/src/task-3.37.0/task_test.go:2326
                Error:          "ba\n\nc\n" does not contain "b\n"
                Test:           TestForDeps/loop-explicit
```

Different subtests, but similar issues. So not introduced in 3.39

pd93:
Thanks for checking. I will investigate

pd93:
Just leaving this as a note to myself. This just happened in our CI https://github.com/go-task/task/actions/runs/10962583048/job/30444328303

pbitty:
[...]

PR #1839, opened 2024-09-26
fix: Flaky TestForDeps tests

pbitty:
The executor output for parallel task (and dep) execution can come out one character at a time, so even though the tests use `SyncWriter`, the writer's lock doesn't prevent interleaving.

Using `group` output forces the task output to be buffered and kept together before being printed.

Fixes https://github.com/go-task/task/issues/1819.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -40,6 +40,7 @@ type SyncBuffer struct {
 func (sb *SyncBuffer) Write(p []byte) (n int, err error) {
 	sb.mu.Lock()
 	defer sb.mu.Unlock()
+	fmt.Fprintln(os.Stderr, "len:", len(p))
 	return sb.buf.Write(p)
 }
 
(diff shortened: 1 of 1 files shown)


pbitty:
With the debug print output:
* Prior to the fix:
```
% go test -run TestForDeps/loop-explicit -count 1
len: 1
len: 1
len: 1
len: 1
len: 1
len: 1
PASS
ok  	github.com/go-task/task/v3	0.802s
```

* After the fix:
```
% go test -run TestForDeps/loop-explicit -count 1
len: 0
len: 2
len: 0
len: 2
len: 0
len: 2
PASS
ok  	github.com/go-task/task/v3	0.590s
```

The 2-character output of each task gets lumped together into a single `Write()` call.

andreynering:
Thanks @pbitty!

PR #1842, opened 2024-09-27
fix: Print dotenv file path when there is an error reading file

### LINKED ISSUE #1261: Unusable error message for invalid .env file
If you have an invalid .env file, when trying to run `task` you get the error:
```
Can't separate key from value
```

Nothing else in the console. There is no indication as to what this refers to.

pbitty:
Fixes #1261.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1657,6 +1657,18 @@ func TestDotenvHasEnvVarInPath(t *testing.T) {
 	tt.Run(t)
 }
 
+func TestTaskDotenvParseErrorMessage(t *testing.T) {
+	e := task.Executor{
+		Dir: "testdata/dotenv/parse_error",
+	}
+
+	path, _ := filepath.Abs(filepath.Join(e.Dir, ".env-with-error"))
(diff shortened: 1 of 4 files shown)


task_test.go
```diff
@@ -1657,6 +1657,18 @@ func TestDotenvHasEnvVarInPath(t *testing.T) {
...
+func TestTaskDotenvParseErrorMessage(t *testing.T) {
+	e := task.Executor{
+		Dir: "testdata/dotenv/parse_error",
+	}
+
+	path, _ := filepath.Abs(filepath.Join(e.Dir, ".env-with-error"))
+	expected := fmt.Sprintf("error reading env file %s:", path)
+
+	err := e.Setup()
+	require.ErrorContains(t, err, expected)
```
pbitty, line 1669: I used `ErrorContains()` and only included the part of the error message that _we_ add, so that the tests wouldn't depend on the exact wording coming from the `dotenv` library.

vmaerten:
Thanks!

Issue #1843, opened 2024-09-29
fish/zsh shell completions suggest task names even for CLI_ARGS


boiledfroginthewell:
- Task version: Task version: v3.39.2 (h1:Zt7KXHmMNq5xWZ1ihphDb+n2zYLCo4BdRe09AnMMIgA=)
- Operating system: Manjaro Linux / MacOS 14
- Experiments enabled:

Task names are suggested after `--` by tab completion. This is inconvenient when I want to pass file paths.

![image](https://github.com/user-attachments/assets/eff9bc10-d1be-4bde-be09-d784213a6965)

PR #1844, opened 2024-09-29
fix: CLI_ARGS completion for fish and zsh (#1843)

boiledfroginthewell:
fixes: #1843

### DIFF
--- a/completion/zsh/_task
+++ b/completion/zsh/_task
@@ -14,6 +14,13 @@ function __task_list() {
     taskfile=${(Qv)opt_args[(i)-t|--taskfile]}
     taskfile=${taskfile//\~/$HOME}
 
+    for arg in "${words[@]:0:$CURRENT}"; do
+        if [[ "$arg" = "--" ]]; then
+            # Use default completion for words after `--` as they are CLI_ARGS.
+            _default
+            return 0
+        fi
+    done
 
     if [[ -n "$taskfile" && -f "$taskfile" ]]; then
         enabled=1
@@ -40,6 +47,7 @@ function __task_list() {
 
 _task() {
     _arguments \
+        -S \
         '(-C --concurrency)'{-C,--concurrency}'[limit number of concurrent tasks]: ' \
         '(-p --parallel)'{-p,--parallel}'[run command-line tasks in parallel]' \
         '(-f --force)'{-f,--force}'[run even if task is up-to-date]' \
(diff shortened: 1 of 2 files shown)


completion/zsh/_task
```diff
@@ -55,13 +63,13 @@ _task() {
         '(-t --taskfile)'{-t,--taskfile}'[specify a different taskfile]:taskfile:_files' \
         '(-v --verbose)'{-v,--verbose}'[verbose mode]' \
         '(-w --watch)'{-w,--watch}'[watch-mode for given tasks, re-run when inputs change]' \
+        '(operation)*: :__task_list' \
         + '(operation)' \
```
boiledfroginthewell, line 67: Task-name completions didn't work except for the first positional argument in zsh, but this has been fixed as well.

(Argument specifications after `+ '(operation)'` become mutually exclusive options. In the previous code, `*: :__task_list` also prevented other task names from being suggested when one task name is specified in a command line.)

completion/zsh/_task
```diff
@@ -55,13 +63,13 @@ _task() {
...
-            {-a,--list-all}'[list all tasks]' \
-            {-i,--init}'[create new Taskfile.yml]' \
-            '(-*)'{-h,--help}'[show help]' \
-            '(-*)--version[show version and exit]' \
-            '*: :__task_list'
+            '(*)'{-l,--list}'[list describable tasks]' \
+            '(*)'{-a,--list-all}'[list all tasks]' \
+            '(*)'{-i,--init}'[create new Taskfile.yml]' \
+            '(- *)'{-h,--help}'[show help]' \
+            '(- *)--version[show version and exit]' \
```
boiledfroginthewell, line 72:
```diff
- (-*)--version
+ (- *)--version
```

I suppose `(- *)` is the intended one. This prevents all options and positional arguments from being suggested when the `--version` option is supplied.

(The parentheses specify groups of arguments to prevent suggestions. [`-` and `*` are special group names for options and positional arguments, respectively](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#index-_005farguments). I think the `-*` group is not defined here.)

vmaerten:
Thanks for the PR! I know I'm more than a year late, but would you be willing to rebase and resolve the conflicts, @boiledfroginthewell?

boiledfroginthewell:
Rebased and resolved! Thanks for looking at it.

Issue #1841, opened 2024-09-27
[Bug:] Custom task directory (set with `dir`) is not created before initializing dynamic variable


ColorfulRhino:
Consider the following task using a custom `dir`:

```YAML
version: '3'

tasks:
  test:
    dir: test-dir
    vars:
      TEST_VAR: "This is a test"
    cmds:
      - echo "This is a test. Using TEST_VAR with value: ${TEST_VAR}"
```

The above task will create the directory `test-dir` before executing the command if it is not already created.

However, when using a dynamic variable instead, like this:

```YAML
version: '3'

tasks:
  test:
    dir: test-dir
    vars:
      TEST_VAR:
        sh: curl https://taskfile.dev
    cmds:
      - echo "This is a test. Using TEST_VAR with value: ${TEST_VAR}"
```

It fails to execute when the directory `test-dir` does not yet exist:
`task: Command "curl https://taskfile.dev" failed: chdir /home/user/test-dir: no such file or directory`

**Expected behaviour:**
The directory should be created *before* trying to execute the dynamic variable initialization command.

---

- Task version: v3.39.2
- Operating system: Ubuntu 24.04
- Experiments enabled: no

trulede:
Why can't the curl command be the first of the task cmds? I guess this is just an example, is there a real use-case?

```
version: '3'

tasks:
  test:
    dir: test-dir
    vars:
      TEST_VAR: https://taskfile.dev
    cmds:
      - curl ${TEST_VAR}"
      - echo "This is a test. Using TEST_VAR with value: ${TEST_VAR}"
```

trulede:
OK, this is actually working. The problem is that you are not referencing the variable correctly - you are referencing an envar that is not set. 

It should be:

`- curl {{.TEST_VAR}}`

PR #1848, opened 2024-09-30
Create directory before evaluating dynamic arguments

404Wolf:
Create the task's directory before evaluating dynamic arguments, so that they can use the directory if they need to.

I'm new here and to this repo, so hopefully I conformed to the guidelines, but if not let me know!

Fix for https://github.com/go-task/task/issues/1841

### DIFF
--- a/task.go
+++ b/task.go
@@ -168,6 +168,11 @@ func (e *Executor) splitRegularAndWatchCalls(calls ...*ast.Call) (regularCalls [
 // RunTask runs a task by its name
 func (e *Executor) RunTask(ctx context.Context, call *ast.Call) error {
 	t, err := e.FastCompiledTask(call)
+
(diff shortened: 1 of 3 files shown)


task.go
```diff
@@ -168,6 +168,11 @@ func (e *Executor) splitRegularAndWatchCalls(calls ...*ast.Call) (regularCalls [
 // RunTask runs a task by its name
 func (e *Executor) RunTask(ctx context.Context, call *ast.Call) error {
 	t, err := e.FastCompiledTask(call)
+
+	if err := e.mkdir(t); err != nil {
+		e.Logger.Errf(logger.Red, "task: cannot make directory %q: %v\n", t.Dir, err)
+	}
+
```
trulede, line 175: The code placed here means that directories are created _before_ any checks that might prevent a task from running.

PR #1859, opened 2024-10-09
feat: exclude task from being included

vmaerten:
Allow some tasks to be excluded.

This can be useful in combination with the flatten feature, as it enables the exclusion of conflicting tasks.

It would supersede this PR:

> [!NOTE]
> I've also fixed some formatting in the flatten section in the website


vmaerten:
@elocke do you think it could solve your usecase ?

andreynering (approved):

PR #1866, opened 2024-10-16
feat: allow providing single or multi prompts

### LINKED ISSUE #1861: Multiple prompts
I'd love to be able to give "prompt" an array of prompts.

We have a task that releases to production. Right now we have one prompt that contains multiple questions and checks for the user. 

I'd like to split those questions into multiple.

So instead of 

```
version: '3'
tasks:
  deploy:production:
    desc: Release to prod
    prompt: "Do you wish to release to version {{.RELEASE_VERSION}}? Have you remembered to log into harbor today? Have you announced the release on Slack and aligned with team Foo?"
```

I'd like to do this

```
version: '3'
tasks:
  deploy:production:
    desc: Release to prod
    prompt:
    - Do you wish to release to version {{.RELEASE_VERSION}}? 
    - Have you remembered to log into harbor today? 
    - Have you announced the release on Slack? 
    - Have you aligned with team Foo? 
```

mfbmina:
Hello, it closes #1861!

At the issue, we decided to make `prompt` accept one or more values. As the `go-yaml` does not support this, the solution is described in this issue: https://github.com/go-yaml/yaml/issues/100

A new `Prompt` type was introduced and it handle its serialization to allow one or more values.


taskfile/ast/prompt.go
```diff
@@ -0,0 +1,21 @@
...
+
+	var s string
+	err = unmarshal(&s)
+	if err != nil {
+		return err
+	}
+
+	*p = []string{s}
+	return nil
+}
```
vmaerten, line 21: You should use the same pattern as we have everywhere else (see include.go for example)

```suggestion
func (p *Prompt) UnmarshalYAML(node *yaml.Node) error {
	switch node.Kind {
	case yaml.ScalarNode:
		var str string
		if err := node.Decode(&str); err != nil {
			return errors.NewTaskfileDecodeError(err, node)
		}
		*p = []string{str}
		return nil
	case yaml.SequenceNode:
		var list []string
		if err := node.Decode(&list); err != nil {
			return errors.NewTaskfileDecodeError(err, node)
		}
		*p = list
		return nil
	}
	return errors.NewTaskfileDecodeError(nil, node).WithTypeMessage("prompt")
}

```

website/static/schema.json
```diff
@@ -59,8 +59,11 @@
...
         },
         "prompt": {
-          "description": "A prompt that will be presented before a task is run. Declining will cancel running the current and any subsequent tasks.",
-          "type": "string"
+          "description": "One or more prompts that will be presented before a task is run. Declining will cancel running the current and any subsequent tasks.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         },
```
vmaerten, line 67: Changing the schema like that is a breaking change.
You need to keep the old type. You can use `oneOf` property to define two types

Otherwise, we won't be able to validate current Taskfile (see : 
image
)

vmaerten (changes requested):
Hello!

Thanks for working on this !! :fire:  I like the feature

I've left some comments!

vmaerten (approved):
lgtm! 
Thanks.
Waiting for another review before merging :)

mfbmina:
Hey guys! Can I help with something to get it reviewed?

pd93 (approved):
> Hey guys! Can I help with something to get it reviewed?

No help needed, all approved but unmerged PRs will get looked at before a release happens. That said, this looks fine to me. Thanks for the contribution!

MalteMagnussen:
Thank you very much!

PR #1868, opened 2024-10-17
added dockerfile and multiarch build job in goreleaser workflow. fixe…

### LINKED ISSUE #1801: Distribute as Docker Image
I think it would be helpful to also distribute this tool as a multi-platform Docker image.

My main usecase for this is using that image as a makeshift `build-step` when building out larger images. Like [this](https://github.com/mirceanton/gitops-toolkit/blob/main/Dockerfile#L97), for example.

Since we're already using `goreleaser` to create the release artifacts, it would be a fairly simple change to instruct it to also build and push the docker images as well.
Something like [this](https://github.com/mirceanton/talswitcher/blob/main/.goreleaser.yaml#L54-L96) should do the trick, with an extra docker login step in the goreleaser workflow.

maehldom:
…s #1801

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,3 +24,30 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{secrets.GH_PAT}}
+  
+  docker:
+    runs-on: ubuntu-latest
+    needs: goreleaser
+    if: ${{ secrets.DOCKER_HUB_USERNAME && secrets.DOCKER_HUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      - name: Build and Push Docker image
+        uses: docker/build-push-action@v6
+        platforms: linux/amd64,linux/arm64
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: |
+            go-task/task:latest
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.23.2-alpine
+
+COPY . /app
+
+WORKDIR /app
+
+RUN /app/install-task.sh -b /usr/local/bin
+
+ENTRYPOINT ["task"]


.github/workflows/release.yml
```diff
@@ -24,3 +24,30 @@ jobs:
...
+
+      - name: Build and Push Docker image
+        uses: docker/build-push-action@v6
+        platforms: linux/amd64,linux/arm64
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: |
+            go-task/task:latest
```
mircea-pavel-anton, line 53: Since this workflow is triggered on tag creation, it would be a good idea to tag the created image with the version tag that was created in git. That way we can have matching tags for the docker images and the git repo.

We can still leave in a `latest` tag, but I consider actual version tags to be a requirement to properly use this image

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.23.2-alpine
+
+COPY . /app
+
+WORKDIR /app
+
+RUN /app/install-task.sh -b /usr/local/bin
```
mircea-pavel-anton, line 7: My main gripe with this is that is is, at least as far as I understand, quite impossible for docker to cache any layers to speed up builds.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.23.2-alpine
```
mircea-pavel-anton, line 1: This would technically only be the `build` stage. For the final "production" stage, there is no need to have `golang` as a base. We could do `alpine` or something smaller

.github/workflows/release.yml
```diff
@@ -24,3 +24,30 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{secrets.GH_PAT}}
+  
+  docker:
+    runs-on: ubuntu-latest
+    needs: goreleaser
```
mircea-pavel-anton, line 30: any reason we need to wait for the `goreleaser` job? I don't see any hard dependency and we could save some time by executing these jobs in parallel

mircea-pavel-anton:
`goreleaser` already does cross-compilation for multi-arch builds and supports creating docker images too. In my experience, it is also much faster at it. As far as I can tell, this does not implement multi-arch support.

@maehldom is there any reason for going with this approach instead of going with `goreleaser`?

trulede:
@andreynering @mircea-pavel-anton @maehldom - if you are still caring about this topic ...

https://github.com/trulede/task-docker

I've setup a repo to demonstrate a decoupled approach to building docker images. Decoupled as in not part of the Task repo with a rational to allow development of Docker images independent of Task. The repo currently builds a scratch image containing only a statically compiled Task binary which can be used in multi-stage docker builds. That image is built for a few platforms; darwin and windows should be possible too.

The images are pushed to ghcr, and could be pushed to other registries too.

I have a few other ideas for dockerfiles, using volume mount strategies, to make a Task image that is actually useful beyond deployment/installation. I use a similar pattern for running containerised workflows via Task.

PR #1869, opened 2024-10-17
fix: forward env to RunCommand when evaluating sh vars

### LINKED ISSUE #630: Dynamic variable execution environment does not inherit globally set environment variables
It may be as designed, but it's been bugging me for a while and creates quite a bit of confusion. So if nothing else, I'll document the behavior here.

See the following example:

```yaml
version: '3'
env:
  NAME: "giselle"
tasks:
  default:
    vars: 
      person:
        sh: echo $NAME
    cmds:
      - echo "Hello $NAME"
      - echo "Hello {{ .person }}"
    silent: true
```
Output:
```
Hello giselle
Hello
```

That makes dynamic variables a bit useless in places where you need to pass auth or other relevant info via env.

### LINKED ISSUE #1742: Env access for sh vars
I declared global env in a taskfile like so:

```
---
version: "3"

env:
  DB_DRIVER: mysql
  DB_DSN: "etl:etl@tcp(localhost:3306)/etl"

tasks:
  default:
    desc: "Fill repos with details"
    vars:
      repos:
        sh: env | grep DB_ || true
    cmds:
      - env | grep DB || true
      - for: { var: repos, as: repo, split: "\n" }
        silent: true
        cmd: echo "X{{.repo}}"
```

The expected output of this would be:

```
DB_DRIVER=mysql
DB_DSN=etl:etl@tcp(localhost:3306)/etl
XDB_DRIVER=mysql
XDB_DSN=etl:etl@tcp(localhost:3306)/etl
```

Actual output of this is:

```
DB_DRIVER=mysql
DB_DSN=etl:etl@tcp(localhost:3306)/etl
X
```

Seems the vars `sh` are not picking up the taskfile env which would be my expectation...

rohm1:
Fixes #630


pd93:
@rohm1 For whatever reason GitHub refused to run the tests on a bunch of PRs. Can you please rebase and push your changes again?

vmaerten (approved):

pd93:
Thanks @rohm1! 🚀

Issue #1735, opened 2024-08-04
Feature Request: Allow .CHECKSUM and .TIMESTAMP variables to be used in cmds


andreynering:
### Discussed in https://github.com/go-task/task/discussions/1734

<div type='discussions-op-text'>

<sup>Originally posted by **pavetok** August  4, 2024</sup>
Hi everyone,

First of all, thanks for the great tool!

I want to build docker image only when sources changed, but `{{.CHECKSUM}}` empty in cmds :(

Why is that?

```yaml
image:
  sources:
    - ./**
  status:
    - docker image inspect my/app:{{.CHECKSUM}}
  cmds:
    - docker build . -t my/app:latest -t my/app:{{.CHECKSUM}}
```</div>

PR #1872, opened 2024-10-21
feat: Make CHECKSUM and TIMESTAMP vars available in cmds commands

niklasr22:
Hello, this is my attempt to implement #1735.

### DIFF
--- a/variables.go
+++ b/variables.go
@@ -127,6 +127,23 @@ func (e *Executor) compiledTask(call *ast.Call, evaluateShVars bool) (*ast.Task,
 		}
 	}
 
+	if len(origTask.Sources) > 0 {
+		timestampChecker := fingerprint.NewTimestampChecker(e.TempDir.Fingerprint, e.Dry)
+		checksumChecker := fingerprint.NewChecksumChecker(e.TempDir.Fingerprint, e.Dry)
+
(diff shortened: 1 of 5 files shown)


variables.go
```diff
@@ -228,7 +181,7 @@ func (e *Executor) compiledTask(call *ast.Call, evaluateShVars bool) (*ast.Task,
 		}
 	}
 
-	if len(origTask.Status) > 0 {
```
vmaerten, line 231: I'm not a big fan of those two nested if statements.
`TIMESTAMP` and `CHECKSUM` should only be available if there are sources defined. I'd replace `len(origTask.Status) > 0` with `len(origTask.Source) > 0` and move it above `cmds`.

vmaerten (commented):
Hello
Thanks for your PR, I've left some comments!

variables.go
niklasr22: True, good idea. I changed it like you said but now it also has the effect that .TIMESTAMP and .CHECKSUM are available in preconditions and deps. Do you want to keep it like that?

vmaerten: Indeed this is wanted :slightly_smiling_face:

vmaerten:
Lint job is failing. a `t.Parallele` is missing in your test
After that it should be good

vmaerten (approved):
Thanks for your contribution!

PR #1875, opened 2024-10-21
chore: Distribute as docker image

mircea-pavel-anton:
This is currently a WIP to solve https://github.com/go-task/task/issues/1801

## Some notes

- I chose debian as a base for the docker image instead of alpine, for example, such that I can load the bash completion
- I set the workdir to `/workspace` to facilitate mounting volumes into the container via ad-hoc docker commands, i.e.:

  ```bash
  docker run -it -v $PWD:/workspace ghcr.io/go-task/task:v3.39.2-amd64 lint # equivalent of `task lint`
  ```
- I set the cmd instead of the entrypoint to the `task` executable such that it is easier to just exec into the container like so:

  ```bash
  docker run -it -v $PWD:/workspace ghcr.io/go-task/task:v3.39.2-amd64 /bin/bash
  ```
  
- the current implementation pushes both to dockerhub and ghcr.io
- the current implementation pushes:
  - version tagged images (`ghcr.io/go-task/task:v3.39.2`)
  - version and arch tagged images (`ghcr.io/go-task/task:v3.39.2-amd64`)
  - latest images (`ghcr.io/go-task/task:latest`)
  
## TODO

- update release workflow to login to ghcr and/or dockerhub
- figure out bash completion


mircea-pavel-anton:
@pd93 @andreynering 

What do you think about this implementation?

I moved to using `scratch` as a base for the image and put in only the binary and the shell completions so that they can both be easily imported.

I also added some documentation about the changes, though I decided to keep it rather minimal. Not sure about the `Note` section. I'm open to suggestions.

I also modified the goreleaser to only push to GHCR (initially it was configured to push to docker.io as well).

P.S. Ignore the devcontainer file, that was only for my local testing. I will drop that commit once this PR is ready

.goreleaser.yml
```diff
@@ -131,3 +131,51 @@ winget:
...
           name: winget-pkgs
           branch: "bump-task-to-{{.Tag}}"
+
+dockers:
+  - image_templates:
+      - "ghcr.io/go-task/task:v{{.Version}}-amd64"
+    use: buildx
+    dockerfile: Dockerfile
+    extra_files:
+      - completion/bash/task.bash
```
pd93, line 141: Direct usage of these files is [deprecated](https://taskfile.dev/deprecations/completion-scripts/).

pd93 (commented):
We should probably also add [additional tags](https://goreleaser.com/customization/docker/#keeping-docker-images-updated-for-current-major) for `latest`, `3`, `3.40` etc.

.goreleaser.yml
mircea-pavel-anton: Oh damn, I completely missed that. Will fix!

pd93: Yeah unfortunately, they are still used in a few places. This is why we deprecated them instead of removing them. We need to slowly remove the references to them. We could probably still add completions to the image by writing a `.bashrc` during the build and using the new method

mircea-pavel-anton: I thought about generating the completions in the container, but then I decided against it.

It doesn't make much sense to generate the bash completions given the image is based on scratch IMHO, and even then, if we generate and include one set of completions we might as well generate all.

I think that, given the completions can be generated from the binary, it is just as simple to add a `RUN` command in your dockerfile to set up your required completions.

There are already instructions for generating the completions so that should have it covered. At the end of the day, the container image is a means to distribute the binary, right?

Let me know what you think!

website/docs/installation.mdx
```diff
@@ -220,6 +220,19 @@ If you want to install Task in GitHub Actions you can try using
 
 This installation method is community owned.
 
+### Docker
+
+If you are building a docker image that, among other utilities, must contain `Task` as well, you can copy over both the binary as well as the appropriate bash completion for your shell from the official `ghcr.io/go-task/task` docker image:
+
+> Note: The image is available only for `amd64` and `arm64v8` architectures and only for Task versions starting at `v3.40.2`
```
andreynering, line 227:
```suggestion
:::info

The image is available only for `amd64` and `arm64v8` architectures and only for Task versions starting at v3.40.2.

:::
```

andreynering (commented):
Haven't attempted to build it yet, but added some comments for now.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mircea-pavel-anton: you can ignore this file as i will remove it before merging. This was so i can get a quick dev environment going since I don't have golang installed on my machine directly.

andreynering (approved):
Hi @mircea-pavel-anton, this looks good!

It still misses the login step, right? Anything needed from my part to move this forward?

mircea-pavel-anton:
Hi @andreynering !

Sorry for the delay, I've been a bit busy. I have added the login step and rebased my branch. It should be all good now

.goreleaser.yml
```diff
@@ -131,3 +131,54 @@ winget:
...
+      - "--label=org.opencontainers.image.title={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--label=org.opencontainers.image.url=https://taskfile.dev"
+      - "--label=org.opencontainers.image.source=https://github.com/go-task/task"
+      - "--label=org.opencontainers.image.licenses=MIT"
+      - "--label=org.opencontainers.image.authors=The Task authors <task@taskfile.dev>"
+      - "--label=org.opencontainers.image.description=Task is a task runner / build tool that aims to be simpler and easier to use than, for example, GNU Make."
+  - image_templates:
+      - "ghcr.io/go-task/task:v{{.Version}}-arm64v8"
```
reneleonhardt, line 152: I wouldn't append the variant for arm64, there is only one, so it's practically nowhere used:
https://github.com/search?q=repo%3Acontainerd%2Fcontainerd+arm64v8&type=code
When platforms are built separately then the common suffixes are `amd64` and `arm64`.

Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM scratch
```
reneleonhardt, line 1: distroless nonroot is usually a better choice than scratch:
https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/Dockerfile

/task would make it easier to COPY than /usr/local/bin/task in an image which is completely empty... 😅

mircea-pavel-anton:
@reneleonhardt thanks for the feedback! I added your suggestions to this PR and made sure my branch is updated

@andreynering @pd93 anything else I can help with to move this forward?

Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/static-debian12:nonroot
+COPY task /task
+CMD ["/task"]
```
reneleonhardt, line 3: Probably better if you would actually run this container for whatever reasons 😅
```suggestion
ENTRYPOINT ["/task"]
```

website/docs/installation.mdx
```diff
@@ -201,6 +201,23 @@ If you want to install Task in GitHub Actions you can try using
...
+If you are building a Docker image that, among other utilities, must contain `task` as well, you can copy over both the binary as well as the appropriate Bash completion for your shell from the official `ghcr.io/go-task/task` Docker image:
+
+:::info
+
+The image is available only for `amd64` and `arm64v8` architectures and only for Task versions starting at v3.40.2.
+
+:::
+
+```Dockerfile
+FROM ubuntu:22.04
```
reneleonhardt, line 215: Back to the future... 🚀
```suggestion
FROM ubuntu:24.04
```

Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/static-debian12:nonroot
+COPY task /task
```
reneleonhardt, line 2: I don't know if the build is "hardened", let's just make sure that users can read and execute automatically
```suggestion
COPY --chmod=555 task /task
```

.goreleaser.yml
```diff
@@ -131,3 +131,54 @@ winget:
           owner: go-task
           name: winget-pkgs
           branch: "bump-task-to-{{.Tag}}"
+
+dockers:
```
reneleonhardt, line 135: `docker` would probably sound better 😄
```suggestion
docker:
```

Dockerfile
mircea-pavel-anton: Since its based on distroless,  would you ever run it by itself though? I think this would maybe make sense with a usable base. Dunno

website/docs/installation.mdx
mircea-pavel-anton: At this point we might as well put `latest` as a tag rather than keep updating the example, I'd say.

Dockerfile
mircea-pavel-anton: Since its based on distroless,  would you ever run it by itself though? I think this would maybe make sense with a usable base. Dunno

website/docs/installation.mdx
reneleonhardt: Of course, it's just an example anyway. Convention is not to use latest, the user should override anyway.

mircea-pavel-anton:
Since there is not much activity on this (or similar) PRs, I decided to just package it myself.

I set up this repo: https://github.com/mirceanton/container-images
Which, among other images, builds and pushes `ghcr.io/mirceanton/taskfile`

Tags are set up on the container to follow the git tags in this repo. It is all automated using renovate. I am also pushing multiple tags, i.e.:
- `vX`
- `vX.Y`
- `vX.Y.Z`
- `latest`
- `sha-<GIT_SHA>`
- `date-<BUILD_DATE>`

I will leave this open fwiw, but my goal has been more or less achieved with that repository

PR #1882, opened 2024-10-25
feat: Add parallel test execution to improve runtime

christiandins:
- added linter to ensure parallel tests
- added t.Parallel and t.Helper accordingly

Old:
ok      github.com/go-task/task/v3      3.205s

New:
ok      github.com/go-task/task/v3      2.739s


andreynering:
Good addition. Thanks @christiandins!

PR #1883, opened 2024-10-25
chore: add linter misspell

christiandins:
- find commonly misspelled english words

The past participle for "override" is "overridden", both in American and in British English: 
https://www.oxfordlearnersdictionaries.com/definition/english/override?q=override

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
     - goimports
     - gofmt
     - gofumpt
+    - misspell
 
 linters-settings:
   goimports:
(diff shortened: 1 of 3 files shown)


pd93 (commented):
There are a lot of opinionated style changes in here. Are you using a tool to do this or are they manually changed? I'm not a fan of merging style changes without an agreed upon tool/standard because everyone has an opinion and code tends to flip between styles in PRs unnecessarily.

My assumption is that the changes are based on line length, but its not clear what the rules are or whether they have been applied everywhere or just the files you're changing.

christiandins:
Hi, I’m using [https://github.com/segmentio/golines](https://github.com/segmentio/golines), which runs automatically every time I save a file. You’re absolutely right that these are unnecessary changes, and I’ll revert them accordingly. However, I can definitely recommend *golines* for improved readability! :)

pd93 (approved):

alexandear (approved):

PR #1893, opened 2024-10-29
docs: add an example of a `default` value usage in vars

iamkarlson:
It's rather often than not you need to use a default value for a variable, and then adjust it to your needs. There's no documentation can be found on how to achieve that so this PR should address this issue


vmaerten:
Thanks!

vmaerten (approved):

PR #1894, opened 2024-10-29
feat(remote): add tests for git remote taskfiles

pbitty:
This uses a similar approach as the HTTP remote taskfile tests,
leveraging the `git` tool's built-in support for hosting a local HTTP
server via CGI.

I was inspired by the HTTP tests and realized the same thing could work for git.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1116,7 +1116,7 @@ func TestIncludesRemote(t *testing.T) {
 			secondRemote: "./second/Taskfile.yml",
 		},
 		{
-			firstRemote:  srv.URL + "/repo.git//" + dir + "/first/Taskfile.yml?ref=main",
+			firstRemote:  srv.URL + "/repo.git//" + dir + "/first/Taskfile.yml?ref=HEAD",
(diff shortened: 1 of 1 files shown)


task_test.go
```diff
@@ -1086,12 +1089,23 @@ func TestIncludesRemote(t *testing.T) {
 
 	dir := "testdata/includes_remote"
 
-	srv := httptest.NewServer(http.FileServer(http.Dir(dir)))
+	cwd, _ := os.Getwd()
+	gitHandler, gitSupported := createGitHTTPHandler(t, cwd)
+	if !gitSupported {
+		t.Log("git tests not supported and will be skipped")
+	}
```
pbitty, line 1096: I put this as a guard for now, but I would expect pretty much every CI and dev environment will have `git` installed.  If you all agree, I will remove this guard to reduce noise in the code.

pbitty:
I'm going to mark this as draft while I sort out the CI issues.  I need to find the right git ref to use.

pbitty:
Ok, fixed now.

PR #1902, opened 2024-11-02
Enforce validation for taskfile field in includes

### LINKED ISSUE #1881: Incorrect/confusing `include cycle detected` error message after a silly mistake
- Operating system: MacOS M1 Sonoma

I did a silly mistake on a Taskfile and got a misleading message as a response, so it took me a while to figure out exactly what the issue was. 

### Minimum Reproducible example

I create two taskfiles in the same directory with the following names:

```yaml
# Taskfile.yml
version: '3'

includes:
  common: ./TaskfileCommon.yml

tasks:
  test:
    cmds:
      - task: common:test
```

And

```yaml
# TaskfileCommon.yml
version: '3'

includes: # This was the silly mistake, this should be `vars` not `includes`
  GOBIN:
    sh: echo $(go env GOPATH)/bin

tasks:
  test:
    cmds:
```

Then I tried running:

```
task common:test
```

And the error message I got back was this:

> task: include cycle detected between /Users/vingarcia/temp/tfile/Taskfile.yml <--> /Users/vingarcia/temp/tfile/TaskfileCommon.yml

I am not sure how easy it is to fix this issue and I understand this might not be a high priority bug, but I thought it would be worth mentioning anyway.

Great tool, thanks for making it and keeping it open source =]

semihbkgr:
fixes #1881

This makes the `taskfile` field within `include` required, disallowing `null` (unspecified) or empty values, even if inlined.

I’d like to clarify two points, @pd93:

1- Would it be preferable to return a `TaskfileDecodeError` in the `UnmarshalYAML` function when parsing the yaml into `Include`? This approach could make the error message more descriptive.

Screenshot from 2024-11-02 22-51-17

However, I aligned this error with the existing one we return when the `Taskfile` is not existing ("no such file or directory"), which is why the error is returned after the parsing step.

2- Should we differentiate between null and empty string values, or should we treat them the same and return an error for both?


pd93:
> 1- Would it be preferable to return a `TaskfileDecodeError` in the `UnmarshalYAML` function when parsing the yaml into `Include`? This approach could make the error message more descriptive.
 
> However, I aligned this error with the existing one we return when the `Taskfile` is not existing ("no such file or directory"), which is why the error is returned after the parsing step.

The distinction here is whether or not `taskfile:` is required in the schema or whether it is valid to be empty in the schema, but causes an error when processing.

Opinion: I think we should error in the `UnmarshalYAML` method and return a `TaskfileDecodeError` _and_ we should update the `schema.json` to mark `taskfile:` as a required field when using an `include:` object. I'm not sure if jsonschema can specify non-empty on a scalar value? but if it can, we should do that too. @andreynering @vmaerten feel free to weigh-in.

> 2- Should we differentiate between null and empty string values, or should we treat them the same and return an error for both?

I can't see a situation where an empty string is useful.

semihbkgr:
@pd93 We can use `"minLength": 1` in the json schema to ensure the taskfile value is not empty.

vmaerten:
I totally agree with you @pd93 !

vmaerten:
Hey @semihbkgr,
when you have time, can you rebase so we can move forward on this ? :slightly_frowning_face:

taskfile/ast/include.go
```diff
@@ -113,6 +113,11 @@ func (includes *Includes) UnmarshalYAML(node *yaml.Node) error {
 			keyNode := node.Content[i]
 			valueNode := node.Content[i+1]
 
+			// Ensure the include value is not null, as it must be either a string or an include object.
+			if valueNode.Kind == yaml.ScalarNode && valueNode.Tag == "!!null" {
```
semihbkgr, line 117: The `null` check is placed before decoding the `valueNode` because the `Decode` method from the yaml package does not invoke the `UnmarshalYAML` method if the node is `null`. This ensures that `null` values are caught early and handled with a proper error message, as the `UnmarshalYAML` logic would be skipped for such nodes.

taskfile/ast/include.go
```diff
@@ -156,7 +163,13 @@ func (include *Include) UnmarshalYAML(node *yaml.Node) error {
 		if err := node.Decode(&includedTaskfile); err != nil {
 			return errors.NewTaskfileDecodeError(err, node)
 		}
-		include.Taskfile = includedTaskfile.Taskfile
+		if includedTaskfile.Taskfile == nil {
```
trulede, line 166: I think the check should be 

if includedTaskfile.Taskfile == ""  {
     return ... "Long form imports require that taskfile be specified."
}


Edit: updated to reflect schema, taskfile is required, dir is optional.

taskfile/ast/include.go
```diff
@@ -133,18 +138,20 @@ func (includes *Includes) UnmarshalYAML(node *yaml.Node) error {
...
+		if str == "" {
+			return errors.NewTaskfileDecodeError(nil, node).WithMessage("inline taskfile value in includes cannot be empty")
+		}
 		include.Taskfile = str
 		return nil
 
 	case yaml.MappingNode:
 		var includedTaskfile struct {
-			Taskfile string
+			Taskfile *string
```
trulede, line 154: Related to comment on R166, changing this to a *string is not necessary. 

imports.taskfile needs to be set. Exposing internal details via "negative" messaging can also be confusing (what is null, I did not set null anywhere, what does _that_ mean) ... simple is you need to set "this" or "that".

trulede (changes requested):

PR #1879, opened 2024-10-24
feat: add silent for defer

danilobuerger:
Fixes #1877

PR #1907, opened 2024-11-05
docs: Document defer task options

mjpieters:
- Fix the `defer_call` JSON schema to include the `vars` mapping.
- Update the schema reference documentation to cover `defer: {task: ..., vars: {...}}`.

### DIFF
--- a/website/docs/changelog.mdx
+++ b/website/docs/changelog.mdx
@@ -5,6 +5,11 @@ sidebar_position: 14
 
 # Changelog
 
+## Unreleased
+
+- Document deferred tasks in the schema reference and fix an error in the JSON
+  schema for such tasks (#1907 by @mjpieters).
+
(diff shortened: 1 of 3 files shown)


mjpieters:
Do let me know if #1879 is likely to be accepted, as that would require updates to this PR (we can then simplify the JSON schema to use `"anyOf": [{"type": "string"},{"$ref": "#/definitions/task_call"}]`, and `` [`Dependency`](#dependency) `` in the schema reference documentation).

vmaerten:
Hello ! 
I've merged #1879

mjpieters:
> Hello ! I've merged #1879

Excellent, thanks for the heads-up. I've reworked this MR to take that into account.

andreynering:
Thanks @mjpieters!

PR #1915, opened 2024-11-11
fix(json-schema): add missing `platforms` property to cmds `for`

dkarter:
# Motivation

I've been getting errors from my editor that using `platforms` as a sibling of `for` under `cmds` is not supported, when it is in fact supported and respected by task

I believe this was missed on #980

# Changes

- This change simply adds a small fix to the json schema to denote that
`platforms` is allowed with `for`

# Screenshots / Testing Instructions

🖼️ Before
CleanShot 2024-11-10 at 19 35 59@2x

🖼️ After
CleanShot 2024-11-10 at 19 35 34@2x

### DIFF
--- a/website/static/schema.json
+++ b/website/static/schema.json
@@ -396,6 +396,13 @@
         "vars": {
           "description": "Values passed to the task called",
           "$ref": "#/definitions/vars"
+        },
+        "platforms": {
+          "description": "Specifies which platforms the command should be run on.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       },
       "oneOf": [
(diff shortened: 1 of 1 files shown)


andreynering (approved):

vmaerten (commented):
First of all, thanks for this!
Small comment, duplicate 3 times this key is not the way to go. Can you declare it under `definitions` then use it as a ref ? 

like : 

```
        "platforms": {
          "ref": "#/definitions/platforms"
        }
```

dkarter:
@vmaerten great suggestion! Fixed in [4065c3e](https://github.com/go-task/task/pull/1915/commits/4065c3e054e6ba9a678e253159fbcfa447877bd5)

vmaerten:
Thanks

PR #1924, opened 2024-11-19
fix: propagation of SIGTERM and SIGINT to Task Subprocesses

### LINKED ISSUE #1408: Inadequate Propagation of SIGTERM and SIGINT to Task Subprocesses
Script: `signal-test.sh`
```sh
#!/bin/sh

# Set up signal traps
trap 'echo "Received SIGINT"' SIGINT
trap 'echo "Received SIGTERM"' SIGTERM

# Infinite loop to keep the script running
while true; do
    sleep 1
done

```

Task Configuration: `Taskfile.yml`
```
version: "3"

tasks:
  signal-test:
    cmds:
      - sh ./signal-test.sh

```

Given the process structure:
```
# ps output for reference
PID   PPID  CMD
2       1    task signal-test
3       2    sh ./signal-test.sh

```

When executing kill -INT 2, the following output is observed:
```sh
kill -INT 2
# task: Signal received: "interrupt"
# However, the signal is not propagated to `signal-test.sh`
```

It is critical to note that the SIGINT signal fails to reach the subprocess. This issue is particularly problematic when using Taskfile within a Docker container. Both `docker stop` and CTRL-C interruption during `docker run` attempt to send SIGINT or SIGTERM to the Taskfile process directly. The described problem prevents Docker from achieving a 'graceful exit'.

- Task version: Task version: v3.31.0 (h1:o6iyj9gPJXxvxPi/u/l8e025PmM2BqKgtLNPS2i7hV4=)
- Operating system: Mac
- Experiments enabled:

vendelin8:
Fixes #1408

### DIFF
--- a/signals.go
+++ b/signals.go
@@ -15,11 +15,15 @@ const interruptSignalsCount = 3
 // time to do cleanup work.
 func (e *Executor) InterceptInterruptSignals() {
 	ch := make(chan os.Signal, interruptSignalsCount)
-	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(ch, os.Interrupt, syscall.SIGTERM, syscall.SIGURG)
 
 	go func() {
 		for i := range interruptSignalsCount {
 			sig := <-ch
+			if sig != os.Interrupt && sig != syscall.SIGTERM {
+				i--
+				continue
+			}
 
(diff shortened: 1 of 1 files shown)


vendelin8:
checkable with https://github.com/vendelin8/go-task-docker-signal

PR #1932, opened 2024-11-27
feat(fingreprinting): add .taskignore

### LINKED ISSUE #1753: Sources should not include files in .gitignore
I'm using this task to lint JSON files:
```yaml
  json:
    desc: Lint JSON files
    sources:
      - ./**/*.json
      - exclude: .venv/**/*.json
    cmds:
      - for: sources
        cmd: jq empty {{.ITEM}}
      - for: sources
        cmd: prettier --check --parser json {{.ITEM}}
```

And it would be nice if `sources` could respect the `.gitignore` content instead of having to exclude all ignored patterns.

This is especially needed as i'm using this task from a Remote Taskfiles #1317 and cannot easily edit it.

GeneTakavic24:
Fixes #1753


GeneTakavic24:
Gretting, and thanks for a wonderful project! 

The reasoning behind this PR is as follows:

### Design

Since the project already heavily relies on glob semantics, aligning the same semantics in the Taskfile's `exclude`: section and .`taskignore `can provide the following benefits:

1. **Unified Pattern Usage**: End users can use patterns interchangeably across the `exclude`: section and `.taskignore`.

- The exclude: section currently does not support the same negation pattern semantics as .gitignore.
- If .taskignore strictly follows .gitignore semantics, it might cause confusion and a disjointed interface for users needing consistent formatting across both.

- On the other hand, modifying exclude to follow .gitignore semantics would introduce breaking changes, as exclude no longer adheres to the glob pattern syntax.

2. **Change scope is smaller**

- Parsing .gitignore even with the usage of a library such as [lib](https://github.com/go-task/task/issues/1753#issuecomment-2282816103) provided might need to change the Glob structure: 
as well as fingerprinting functionality to combine both exisiting glob parsing and .gitignore-like parsing.
 https://github.com/go-task/task/blob/c28eb204fb404ea141e5b68c2765d7386a166780/taskfile/ast/glob.go#L9-L12

website/docs/usage.mdx
```diff
@@ -789,6 +789,37 @@ tasks:
       - public/bundle.css
 ```
 
+It is also possible to exclude files from fingerprinting globally by creating `.taskignore`
```
adamdicarlo0, line 792: nit: I would add "a" -- that is, change

> [...] creating `.taskignore` file [...]

to

> [...] creating a `.taskignore` file [...]

testdata/taskignore/.taskignore
```diff
@@ -0,0 +1 @@
+./**/*.json
```
adamdicarlo0, line 1: Suggest adding a comment line to make sure comments are correctly processed:

```
# ./dont_ignore.txt
```

taskfile/taskignore.go
```diff
@@ -0,0 +1,67 @@
...
+	if len(bytes) == 0 {
+		return nil
+	}
+
+	lines := strings.Split(string(bytes), "\n")
+	var validGlobs []string
+
+	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if trimmed != "" && !strings.HasPrefix(trimmed, "#") {
```
adamdicarlo0, line 61: Users might expect that end-line comments (e.g., `./**/*.json # ignore pesky JSON files`) are allowed, as they tend to be allowed in similar file formats.

testdata/taskignore/.taskignore
adamdicarlo0: Also some blank lines to make sure trimming also works correctly.

adamdicarlo0 (commented):
Looks pretty good to me.

If `exclude` supports negations, I would clarify in the docs and code whether negations are allowed in the `.taskignore` file (and unit test it).

I'm not sure whether they'd be supported, from looking at the code (I don't recall whether `exclude` supports them).

NB: I am not a maintainer or (at least not yet) even a contributor to the project!

thcrt:
It would be great to see this merged!

maciej-lech:
Any plans for merging this?

PR #1938, opened 2024-12-06
fix: disable version check for use as an external library

leaanthony:
This PR adds an internal flag to the executor to disable version checks. This is to ensure that any application that uses Task as a library doesn't error when Task checks the buildinfo. 
This PR also renames `close` to `closer` in `task.go` as it clashes with the keyword `close.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -58,6 +58,9 @@ func run() error {
 	dir := flags.Dir
 	entrypoint := flags.Entrypoint
 
+	// Initialise version
+	ver.Init()
+
 	if flags.Version {
 		fmt.Printf("Task version: %s\n", ver.GetVersionWithSum())
 		return nil
@@ -132,8 +135,9 @@ func run() error {
 		Stdout: os.Stdout,
 		Stderr: os.Stderr,
 
-		OutputStyle: flags.Output,
-		TaskSorter:  taskSorter,
+		OutputStyle:        flags.Output,
+		TaskSorter:         taskSorter,
+		EnableVersionCheck: true,
 	}
 	listOptions := task.NewListOptions(flags.List, flags.ListAll, flags.ListJson, flags.NoStatus)
 	if err := listOptions.Validate(); err != nil {
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -10,7 +10,7 @@ var (
 	sum     = ""
 )
 
-func init() {
+func Init() {
 	info, ok := debug.ReadBuildInfo()
 	if !ok || info.Main.Version == "" {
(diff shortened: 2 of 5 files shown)


andreynering (approved):

pd93 (commented):
Thanks for this @leaanthony.

Just a thought before this is merged. I don't think we should require users of our library to set a `DisableVersionCheck` flag to make Task work as package. I feel like this should be the default behaviour and our CLI entrypoint should enable it instead (basically, the reverse of this PR).

leaanthony:
I actually agree with you. This got me over the line but yeah adding the in the CLI and not `init()` works better

leaanthony:
@pd93 - Updated the PR to run the version initialisation on cmli startup. Tidied the version output if there is no sum.

internal/version/version.go
```diff
@@ -1,33 +1,32 @@
...
-		version = "unknown"
-	} else {
-		if version == "" {
-			version = info.Main.Version
-		}
-		if sum == "" {
-			sum = info.Main.Sum
-		}
+		return
 	}
```
pd93, line 16: We need to keep these empty checks as these variables are pre-populated by `ldflags` in some package managers. The default value also needs to be empty for the same reason.

There have been quite a few accidental changes around this function. We should probably add some comments here to explain this too.

leaanthony: Coming at it fresh it does seem a bit strange 😅 The reason for the change was a failing test that assumed it would be "unknown" but would return blank. What's the best approach to fix that? Set it in Init if it's not set in the buildinfo?

pd93: This is because our tests treat Task as a package, not a CLI. This is a bit awkward to fix, but I think we could add a setting in the executor called `EnableVersionCheck` (Similar to the `DisableVersionCheck` you had previously. This setting would then need to be added to the CLI entrypoint and any relevant tests that need the version check functionality.

internal/version/version.go
```diff
@@ -10,7 +10,7 @@ var (
 	sum     = ""
 )
 
-func init() {
+func Init() {
```
pd93, line 13: Reverting this change should fix your tests

cmd/task/task.go
```diff
@@ -58,6 +58,9 @@ func run() error {
 	dir := flags.Dir
 	entrypoint := flags.Entrypoint
 
+	// Initialise version
+	ver.Init()
```
pd93, line 62: No longer required after the discussed changes

andreynering:
Thanks @leaanthony. Manually merged as 69f5714e453c783b874e2d309c78852182bf6a49.

Issue #1955, opened 2024-12-13
Unable to use Requires to limit Wildcards


ionfury:
From the docs I expect to be able to use [requires](https://taskfile.dev/usage/#ensuring-required-variables-have-allowed-values) and [wildcard arguments](https://taskfile.dev/usage/#wildcard-arguments) together to limit the possible wildcards.  Below is an example, following from the initial wildcard PR https://github.com/go-task/task/pull/1489:

```yaml
---
version: "3"

tasks:
  start-*:
    desc: "Starts a service.  Options: web, db"
    vars:
      SERVICE: "{{ index .MATCH 0 }}"
    cmds:
      - echo "Starting {{.SERVICE}}"
    requires:
      vars:
        - name: SERVICE
          enum: ["web", "db"]
```

And the output:

```sh
❯ task start-a
task: Task "start-*" cancelled because it is missing required variables: SERVICE
```

```sh
❯ task start-web
task: Task "start-*" cancelled because it is missing required variables: SERVICE
```

Removing the `requires` block results in the following (correct) output:

```sh
❯ task start-web
task: [start-*] echo "Starting web"
Starting web
```

I would expect `requires` to validate that `SERVICE` is in `["web", "db"]`.  If this assumption or my implementation is incorrect, please forgive me.

- Task version: 
```sh
Task version: 3.40.1 ()
```
- Operating system:
```sh
ProductName:            macOS
ProductVersion:         14.4.1
BuildVersion:           23E224
```
- Experiments enabled: none

PR #1960, opened 2024-12-15
fix: task level vars are not ignored in requires

vmaerten:
fixes : 
- https://github.com/go-task/task/issues/1955
- https://github.com/go-task/task/issues/1768

AFAIK, there is no reason to pass nil to vars in `compiledTask`
Am I missing something here ?


variables.go
```diff
@@ -57,7 +57,7 @@ func (e *Executor) compiledTask(call *ast.Call, evaluateShVars bool) (*ast.Task,
 		Dir:                  templater.Replace(origTask.Dir, cache),
 		Set:                  origTask.Set,
 		Shopt:                origTask.Shopt,
-		Vars:                 nil,
```
pd93, line 60: I've noticed this before and wondered the same thing. However. it never caused any issues until now. I have no idea if there is a reason to assign `nil` here. @andreynering are you aware of anything?

Otherwise LGTM.

testdata/requires/Taskfile.yml
```diff
@@ -10,6 +10,14 @@ tasks:
         - foo
     cmd: echo "{{.foo}}"
 
+  var-defined-in-task:
+    vars:
+      foo: bar
+    requires:
+      vars:
+        - foo
+    cmd: echo "{{.foo}}"
```
andreynering, line 19: Just to follow our style convention:

```suggestion
  var-defined-in-task:
    vars:
      FOO: bar
    requires:
      vars:
        - FOO
    cmd: echo "{{.FOO}}"
```

variables.go
andreynering: I think it was just not used at all before, because all the templating happens inside function. Looks fine to set it here.

andreynering (approved):

testdata/requires/Taskfile.yml
vmaerten: You're right!

andreynering:
FYI: Looks like there was a previous PR that addresses this issue: #1769

vmaerten:
> FYI: Looks like there was a previous PR that addresses this issue: #1769

Oh my bad you're right!  I initially focused on #1955, but then I realized it would also address #1768.

andreynering:
@vmaerten No problem. It's fine to merge this one, but we'll need to close that PR.

Good reminder that when adding a line with `Co-authored-by: Name <email@email.com>` on the commit message, GitHub gives credit to the person as a co-author. Sometimes I use that when working on top of someone else's work.

Issue #1959, opened 2024-12-15
special variable for absolute path of task directory


vmaerten:
### Discussed in https://github.com/go-task/task/discussions/1957

<div type='discussions-op-text'>

<sup>Originally posted by **lesomnus** December 14, 2024</sup>
Assume I run following taskfile at `/home/lesomnus/foo`:
```yaml
version: "3"

tasks:
  hello:
    dir: bar
    cmds:
      - echo $(pwd)
```
It will print  `/home/lesomnus/foo/bar`.
However, I believe it should be templated rather than executing `pwd`.

There was an issue about it at #525 but is closed as duplication of #215 but #215 does not state that.

Maybe I can:
```yaml
version: "3"

tasks:
  gen-proto:
    dir: proto
    vars:
      PWD:
        sh: pwd
    cmds:
      - echo {{.PWD}}
```
but it is too verbose.

Is there already a special variable for this?</div>

PR #1961, opened 2024-12-15
feat: add TASK_DIR special variable

vmaerten:
fixes 
- https://github.com/go-task/task/issues/1959

I think it's a good addition

### DIFF
--- a/internal/compiler/compiler.go
+++ b/internal/compiler/compiler.go
@@ -192,14 +192,15 @@ func (c *Compiler) getSpecialVars(t *ast.Task, call *ast.Call) (map[string]strin
 		"TASK_VERSION":     version.GetVersion(),
 	}
 	if t != nil {
-		TaskDir := t.Dir
-		if !filepath.IsAbs(t.Dir) {
-			TaskDir = filepathext.SmartJoin(c.Dir, t.Dir)
-		}
-		maps.Copy(allVars, map[string]string{"TASK": t.Task, "TASKFILE": t.Location.Taskfile, "TASKFILE_DIR": filepath.Dir(t.Location.Taskfile), "TASK_DIR": TaskDir})
+	if t != nil {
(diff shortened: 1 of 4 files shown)


internal/compiler/compiler.go
```diff
@@ -192,7 +192,11 @@ func (c *Compiler) getSpecialVars(t *ast.Task, call *ast.Call) (map[string]strin
...
 	if t != nil {
-		maps.Copy(allVars, map[string]string{"TASK": t.Task, "TASKFILE": t.Location.Taskfile, "TASKFILE_DIR": filepath.Dir(t.Location.Taskfile)})
+		TaskDir := t.Dir
+		if !filepath.IsAbs(t.Dir) {
+			TaskDir = filepathext.SmartJoin(c.Dir, t.Dir)
+		}
+		maps.Copy(allVars, map[string]string{"TASK": t.Task, "TASKFILE": t.Location.Taskfile, "TASKFILE_DIR": filepath.Dir(t.Location.Taskfile), "TASK_DIR": TaskDir})
 	}
 	if call != nil {
 		maps.Copy(allVars, map[string]string{"ALIAS": call.Task})
```
pd93, line 202: - No real reason to use `maps.Copy` here over manually assigning. I think doing it manually scales better in terms of readability.
- `filepathext.SmartJoin` already does an Abs check. I don't think we need to do it twice.
- Also prefer camelCase for `taskDir`.

```suggestion
	if t != nil {
		allVars["TASK"] = t.Task
		allVars["TASK_DIR"] = filepathext.SmartJoin(c.Dir, t.Dir)
		allVars["TASKFILE"] = t.Location.Taskfile
		allVars["TASKFILE_DIR"] = filepath.Dir(t.Location.Taskfile)
	}
	if call != nil {
		allVars["ALIAS"] = call.Task
	}
```

pd93 (approved):

PR #1976, opened 2024-12-30
chore(requires): skip unneeded variable evaluation for `requires`

andreynering:


requires.go
```diff
@@ -7,20 +7,15 @@ import (
...
+func (e *Executor) areTaskRequiredVarsSet(t *ast.Task) error {
 	if t.Requires == nil || len(t.Requires.Vars) == 0 {
 		return nil
 	}
 
-	vars, err := e.Compiler.GetVariables(t, call)
-	if err != nil {
-		return err
-	}
-
```
andreynering, line 19: Let me know if I'm missing anything, but I think we can just skip evaluation completely for performance. For `requires`, it's enough to know if it is set or not.

pd93: Can't think of any reason. This is a nice optimisation 👍

pd93 (approved):

vmaerten (approved):
Good catch! 
I've been thinking, and testing it. I think you're right, it's not needed

PR #1980, opened 2024-12-31
refactor: signal handling to improve clarity and correctness

EinoPlasma:
Renaming: The constant interruptSignalsCount has been renamed to maxInterruptSignals to better reflect its purpose, which is to define the maximum number of interrupt signals that can be handled.

Loop Fix: The loop condition has been corrected to iterate from 0 to maxInterruptSignals, ensuring that the loop runs the expected number of times.

### DIFF
--- a/signals.go
+++ b/signals.go
@@ -8,20 +8,20 @@ import (
 	"github.com/go-task/task/v3/internal/logger"
 )
 
-const interruptSignalsCount = 3
+const maxInterruptSignals = 3
 
 // NOTE(@andreynering): This function intercepts SIGINT and SIGTERM signals
 // so the Task process is not killed immediately and processes running have
 // time to do cleanup work.
 func (e *Executor) InterceptInterruptSignals() {
-	ch := make(chan os.Signal, interruptSignalsCount)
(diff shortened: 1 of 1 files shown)


signals.go
```diff
@@ -8,20 +8,20 @@ import (
...
 // so the Task process is not killed immediately and processes running have
 // time to do cleanup work.
 func (e *Executor) InterceptInterruptSignals() {
-	ch := make(chan os.Signal, interruptSignalsCount)
+	ch := make(chan os.Signal, maxInterruptSignals)
 	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
 
 	go func() {
-		for i := range interruptSignalsCount {
+		for i := 0; i < maxInterruptSignals; i++ {
```
pd93, line 21: These loops are [functionally the same](https://goplay.tools/snippet/nyFZquJ-MXh) and I personally prefer the newer `for i := range maxInterruptSignals` syntax.

No objection to the renaming.

vmaerten: I agree

andreynering: I'll update the code after merging.

PR #1981, opened 2024-12-31
docs: package api doc

### LINKED ISSUE #1958: Breaking changes in v3, taskfile/read pkg removed
Previously a taskfile/read package was provided, which i imported and used for taskfile. This API isn't available today anymore, and a v4 is not tagged.

pd93:
Closes #1958

Adds a new doc that states that the package API exists, but is not stable. I have some ideas about how to make the package API more user friendly and will extend this document when we have things that we think are a bit more stable.

### DIFF
--- a/website/docs/reference/environment.mdx
+++ b/website/docs/reference/environment.mdx
@@ -1,6 +1,6 @@
 ---
 slug: /reference/environment
-sidebar_position: 4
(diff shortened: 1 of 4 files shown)


vmaerten (approved):
LGTM

I took the liberty to remane the PR title

Issue #1978, opened 2024-12-30
Support experiment enablement in Taskfile schema


nathanperkins:
### Description

Experiments like #1038 cause major changes to the behavior for a task. Given that the taskfile would have to be written to make use of this experiment, it is dangerous that the experiment has to be enabled using env vars for every engineer and every automation flow which runs that task. If the task were accidentally run without the experiment, it could lead to undesired behavior, like deploying remote environments with settings based on a developer's local environment.

The risk of dangerous, undesired behavior makes it practically impossible to use the experiment, so we have to continue using a workaround.

Ideally, we would be able to enable this experiment within the taskfile itself so that we have consistent behavior across our team and automation without any risk.

pd93:
(Was just about to reply in #1038), but may as well have this discussion here now.

This is why we have option 3 in our [docs for enabling experiments]( https://taskfile.dev/experiments/#enabling-experiments). This allows you to commit the `.env` file which enables it for anyone checking out your repo. I do understand that a `.env` file isn't always ideal though.

My main concern with adding them to the Taskfile schema itself is that it would (ironically) be a backwards-compatible change to remove the experiments later which we need to be able to do.

Maybe a compromise could be to add support for a dedicated `.task-experiments.yaml` file or something similar.

nathanperkins:
>This is why we have option 3 in our [docs for enabling experiments](https://taskfile.dev/experiments/#enabling-experiments). This allows you to commit the .env file which enables it for anyone checking out your repo. I do understand that a .env file isn't always ideal though.

Ah, I see. I thought the `.env` file would only be used if we sourced it ourselves or if we included it in `dotenv` but I didn't want to replace any other vars.

Based on my experimentation, it seems like the experiment vars are loaded from `.env` regardless of whether you put it in `dotenv`, while other variables are ignored unless you have it in `dotenv`. It might be worth clarifying this further in the docs.

If so, that's good enough for my team.

>My main concern with adding them to the Taskfile schema itself is that it would (ironically) be a backwards-compatible change to remove the experiments later which we need to be able to do.

That makes sense but also maybe this kind of change should break visibly. I'd rather task exit with an error than for it to do something unexpected.

> Maybe a compromise could be to add support for a dedicated .task-experiments.yaml file or something similar.

That sounds good, although I'm fine with the current `.env` solution as I understand it.

nathanperkins:
Possibly related: https://github.com/go-task/task/issues/1979

pd93:
[...]

Issue #1979, opened 2024-12-30
Exit with error when an experiment is specified


nathanperkins:
### Description

Given that experiments can cause major changes in task behavior (like [env precedence](https://taskfile.dev/experiments/env-precedence), it'd be nice to see a strict option (or maybe default behavior) where specifying an experiment that is unsupported or removed results in task exiting immediately with error.

This would avoid a dangerous situation where we get different, dangerous behavior based on what version of taskfile is running the task.

vmaerten:
I am not sure if we want to develop something for this because Experiments were not designed to be used in production and we do not want to develop large feature for this.
However, we could imagine make experiment available in the templating system, so you could use it in preconditions

pd93:
So this was discussed a bit further in https://github.com/go-task/task/pull/1982#issuecomment-2568194128 and the subsequent comments. Thinking about it further, if a user has (for example), the Remote Taskfile experiments enabled and we then made the experiment GA, do we want to suddenly start erroring, even though their Taskfile is valid just because they have an old flag enabled?

Maybe we can print a warning saying the flag is no longer required, but erroring a Taskfile that previously worked seems annoying and would force the user to go fix the problem immediately or roll back the Task version.

To be clear, experiments like env precedence are breaking changes and cannot be made GA until a major version is cut - in which case the user will be expected to make some changes anyway.

nathanperkins:
If changing from an experiment to native support or removing the experiment requires a major version change, a warning sounds reasonable to me :)

Mostly I'd want to avoid surprises with behavior in our CI from simply using a different minor / patch version.

pd93:
> the thing I'd be more concerned about (on the env var exp) is a team member using an old version which isn't aware and doesn't support the option or the GA version.

Just to ease your concerns - We will bump the schema version to `4` when we bump a major version, so you won't be able to run a v4 Taskfile using v3 of Task and vice versa anyway. In other words (IMO anyway), a v3 schema's behaviour should never change unless an experiment is enabled.

Issue #1916, opened 2024-11-11
Support additional metadata in Taskfile YAML schema


timrulebosch:
We are using Taskfiles as input for an automation system (a VSCode extenstion). For that to operate we need to know additional domain specific information about the tasks in the Taskfile. Currently, the Taskfile schema prevents adding such metadata.

Suggestion is to add a metadata object to the schema for both the Taskfile (global) and individual tasks (per task). Something like this:

```
---
version: '3'

vars:
  foo: bar

metadata:
  annotations:
    generator:
      repo: https://github.com/go-foo/foo
      version: v1.2.3
  labels:
    internal-only: true

tasks:
  my-task:
    desc: This task is generated by foo for project fubar.
    metadata:
       annotations: 
          generator: foo
       labels:
         project: fubar
    cmds:
      - echo "generated by foo"
```

If there is interest we can provide a PR as the basis for implementation.

timrulebosch:
We decided to augment our Taskfile.yml with an additional file "Metadata.yml" (out of necessity, not because we wanted to). With that we can describe the tasks in a way that integrates with the VSCode plugin.

I think that describing this metadata within the Taskfile schema would be more practical, **and would also allow some level of automated documentation generation**.


As an example, here is a fragment of the metadata, and the associated Taskfile:

```
metadata:
  models:
    fmimcl:
      name: fmimcl
      displayName: dse.fmi.mcl
      path: dse/fmimcl
      workflows:
        - generate-fmimcl
        - patch-signalgroup
  tasks:
    generate-fmimcl:
      vars:
        FMU_DIR:
          required: true
          hint: URI identifying an FMU (select from 'uses' or manually enter a path).
        OUT_DIR:
          required: true
          hint: Directory where the model should be created (sim relative).
          default: out/model
        MCL_PATH:
          required: true
          hint: Path where the FMI MCL shared library is located.
        FMU_MODELDESC:
          required: false
          hint: Path where the FMI modelDescription.xml should be created.
          default: '{{.FMU_DIR}}/modelDescription.xml'
        SIGNAL_GROUP:
          required: false
          hint: Path where the associated Signal Group should be created.
          default: '{{.OUT_DIR}}/signalgroup.yaml'
```

```
tasks:

  generate-fmimcl:
    desc: Generate an FMI MCL from an existing FMU.
    run: always
    dir: '{{.USER_WORKING_DIR}}'
    label: dse:fmi:generate-fmimcl
    vars:
[...]

PR #1982, opened 2024-12-31
feat: add a new .taskrc.yml to enable experiments

vmaerten:
- #1978 

I like this idea ! 
This file should override the environment variable.

### DIFF
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -9,8 +9,6 @@ import (
 
 	"gopkg.in/yaml.v3"
 
-	"gopkg.in/yaml.v3"
-
 	"github.com/Ladicle/tabwriter"
 	"github.com/joho/godotenv"
 	"github.com/spf13/pflag"
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -2,6 +2,8 @@ package experiments
 
 import (
 	"fmt"
+	"gopkg.in/yaml.v3"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -11,8 +13,12 @@ import (
 )
 
 const envPrefix = "TASK_X_"
+const defaultConfigFilename = ".task-experiments.yml"
 
 // A set of experiments that can be enabled or disabled.
+type ExperimentConfigFile struct {
+	Experiments map[string]string `yaml:",inline"`
+}
 var (
 	GentleForce     Experiment
 	RemoteTaskfiles Experiment
@@ -24,8 +30,11 @@ var (
 // An internal list of all the initialized experiments used for iterating.
 var xList []Experiment
 
+var ExperimentConfig ExperimentConfigFile
+
 func init() {
 	readDotEnv()
+	ExperimentConfig = readConfig()
 	GentleForce = New("GENTLE_FORCE", "1")
 	RemoteTaskfiles = New("REMOTE_TASKFILES", "1")
 	AnyVariables = New("ANY_VARIABLES")
@@ -53,7 +62,7 @@ func getEnv(xName string) string {
 	return os.Getenv(envName)
 }
 
-func getEnvFilePath() string {
+func getFilePath(filename string) string {
 	// Parse the CLI flags again to get the directory/taskfile being run
 	// We use a flagset here so that we can parse a subset of flags without exiting on error.
 	var dir, taskfile string
@@ -64,18 +73,18 @@ func getEnvFilePath() string {
 	_ = fs.Parse(os.Args[1:])
 	// If the directory is set, find a .env file in that directory.
 	if dir != "" {
-		return filepath.Join(dir, ".env")
+		return filepath.Join(dir, filename)
 	}
 	// If the taskfile is set, find a .env file in the directory containing the Taskfile.
 	if taskfile != "" {
-		return filepath.Join(filepath.Dir(taskfile), ".env")
+		return filepath.Join(filepath.Dir(taskfile), filename)
 	}
 	// Otherwise just use the current working directory.
-	return ".env"
+	return filename
 }
 
 func readDotEnv() {
-	env, _ := godotenv.Read(getEnvFilePath())
+	env, _ := godotenv.Read(getFilePath(".env"))
 	// If the env var is an experiment, set it.
 	for key, value := range env {
 		if strings.HasPrefix(key, envPrefix) {
@@ -83,3 +92,18 @@ func readDotEnv() {
 		}
 	}
 }
+
+func readConfig() ExperimentConfigFile {
+	var cfg ExperimentConfigFile
+	filename := getFilePath(defaultConfigFilename)
+	}
+	if err != nil {
+	content, err := os.ReadFile(filename)
+		return ExperimentConfigFile{}
+
+	if err := yaml.Unmarshal(content, &cfg); err != nil {
+	}
+		return ExperimentConfigFile{}
+
+}
+	return cfg
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -5,6 +5,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
+	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v3"
@@ -22,7 +24,7 @@ var defaultConfigFilenames = []string{
 }
 
 type experimentConfigFile struct {
-	Experiments map[string]string `yaml:"experiments"`
+	Experiments map[string]int `yaml:"experiments"`
 }
 var (
 	GentleForce     Experiment
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -17,12 +17,12 @@ import (
 const envPrefix = "TASK_X_"
 
 var defaultConfigFilenames = []string{
-	".task-experiments.yml",
-	".task-experiments.yaml",
+	".taskrc.yml",
+	".taskrc.yaml",
 }
 
 type ExperimentConfigFile struct {
-	Experiments map[string]string `yaml:",inline"`
+	Experiments map[string]string `yaml:"experiments"`
 }
 var (
 	GentleForce     Experiment
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -2,18 +2,24 @@ package experiments
 
 import (
 	"fmt"
-	"gopkg.in/yaml.v3"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/yaml.v3"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/Ladicle/tabwriter"
 	"github.com/joho/godotenv"
 	"github.com/spf13/pflag"
 )
 
-const envPrefix = "TASK_X_"
-const defaultConfigFilename = ".task-experiments.yml"
+const (
+	envPrefix             = "TASK_X_"
+	defaultConfigFilename = ".task-experiments.yml"
+)
 
 // A set of experiments that can be enabled or disabled.
 type ExperimentConfigFile struct {
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -32,17 +32,19 @@ var (
 )
 
 // An internal list of all the initialized experiments used for iterating.
-var xList []Experiment
-var experimentConfig experimentConfigFile
+var (
+	xList            []Experiment
+	experimentConfig experimentConfigFile
+)
 
 func init() {
 	readDotEnv()
 	experimentConfig = readConfig()
-	GentleForce = New("GENTLE_FORCE", "1")
-	RemoteTaskfiles = New("REMOTE_TASKFILES", "1")
+	GentleForce = New("GENTLE_FORCE", 1)
+	RemoteTaskfiles = New("REMOTE_TASKFILES", 1)
 	AnyVariables = New("ANY_VARIABLES")
-	MapVariables = New("MAP_VARIABLES", "1", "2")
-	EnvPrecedence = New("ENV_PRECEDENCE", "1")
+	MapVariables = New("MAP_VARIABLES", 1, 2)
+	EnvPrecedence = New("ENV_PRECEDENCE", 1)
 }
 
 // Validate checks if any experiments have been enabled while being inactive.
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -21,7 +21,7 @@ var defaultConfigFilenames = []string{
 	".taskrc.yaml",
 }
 
-type ExperimentConfigFile struct {
+type experimentConfigFile struct {
 	Experiments map[string]string `yaml:"experiments"`
 }
 var (
@@ -34,6 +34,7 @@ var (
 
 // An internal list of all the initialized experiments used for iterating.
 var xList []Experiment
+var experimentConfig experimentConfigFile
 
 var ExperimentConfig ExperimentConfigFile
 
@@ -98,8 +99,8 @@ func readDotEnv() {
 	}
 }
 
-func readConfig() ExperimentConfigFile {
-	var cfg ExperimentConfigFile
+func readConfig() experimentConfigFile {
+	var cfg experimentConfigFile
 
 	var content []byte
 	var err error
@@ -112,10 +113,11 @@ func readConfig() ExperimentConfigFile {
 	}
 
 	if err != nil {
-	content, err := os.ReadFile(filename)
-		return ExperimentConfigFile{}
+		return experimentConfigFile{}
+	}
 
 	if err := yaml.Unmarshal(content, &cfg); err != nil {
+		return experimentConfigFile{}
 	}
 		return ExperimentConfigFile{}
 
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -14,12 +14,13 @@ import (
 	"github.com/spf13/pflag"
 )
 
-const (
-	envPrefix             = "TASK_X_"
-	defaultConfigFilename = ".task-experiments.yml"
-)
+const envPrefix = "TASK_X_"
 
-// A set of experiments that can be enabled or disabled.
+var defaultConfigFilenames = []string{
+	".task-experiments.yml",
+	".task-experiments.yaml",
+}
+
 type ExperimentConfigFile struct {
 	Experiments map[string]string `yaml:",inline"`
 }
@@ -99,8 +100,17 @@ func readDotEnv() {
 
 func readConfig() ExperimentConfigFile {
 	var cfg ExperimentConfigFile
-	filename := getFilePath(defaultConfigFilename)
+
+	var content []byte
+	var err error
+	for _, filename := range defaultConfigFilenames {
+		path := getFilePath(filename)
+		content, err = os.ReadFile(path)
+		if err == nil {
+			break
+		}
 	}
+
 	if err != nil {
 	content, err := os.ReadFile(filename)
 		return ExperimentConfigFile{}
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
 	"gopkg.in/yaml.v3"
 
 	"github.com/joho/godotenv"
@@ -21,6 +23,7 @@ var defaultConfigFilenames = []string{
 
 type experimentConfigFile struct {
 	Experiments map[string]int `yaml:"experiments"`
+	Version     *semver.Version
 }
 
 var (
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -2,16 +2,12 @@ package experiments
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
-	"slices"
-	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v3"
 
-	"github.com/Ladicle/tabwriter"
 	"github.com/joho/godotenv"
 	"github.com/spf13/pflag"
 )
@@ -26,6 +22,7 @@ var defaultConfigFilenames = []string{
 type experimentConfigFile struct {
 	Experiments map[string]int `yaml:"experiments"`
 }
+
 var (
 	GentleForce     Experiment
 	RemoteTaskfiles Experiment
@@ -38,11 +35,9 @@ var (
 var xList []Experiment
 var experimentConfig experimentConfigFile
 
-var ExperimentConfig ExperimentConfigFile
-
 func init() {
 	readDotEnv()
-	ExperimentConfig = readConfig()
+	experimentConfig = readConfig()
 	GentleForce = New("GENTLE_FORCE", "1")
 	RemoteTaskfiles = New("REMOTE_TASKFILES", "1")
 	AnyVariables = New("ANY_VARIABLES")
@@ -121,7 +116,6 @@ func readConfig() experimentConfigFile {
 	if err := yaml.Unmarshal(content, &cfg); err != nil {
 		return experimentConfigFile{}
 	}
-		return ExperimentConfigFile{}
 
-}
 	return cfg
+}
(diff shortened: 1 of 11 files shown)


website/docs/experiments/experiments.mdx
pd93, line 1: Instead of duplicating the existing section, I think we can just say

> Creating a `.env` file or `.task-experiments.yml` in the same directory...

> ...If you commit either of these files... 

and then maybe add something like

> If both files are present, the values in the `.task-experiments.yml` file will take precedence.

We can provide examples of both files still.

internal/experiments/experiments.go
```diff
@@ -32,8 +41,11 @@ var (
 	EnvPrecedence   Experiment
 )
 
+var ExperimentConfig ExperimentConfigFile
```
pd93, line 44: I don't think this needs to be exported. The resolved values are already available to any user of the package.

internal/experiments/experiments.go
```diff
@@ -8,14 +8,23 @@ import (
...
 	"github.com/joho/godotenv"
 	"github.com/spf13/pflag"
 
 	"github.com/go-task/task/v3/internal/logger"
 )
 
-const envPrefix = "TASK_X_"
+const (
+	envPrefix             = "TASK_X_"
+	defaultConfigFilename = ".task-experiments.yml"
```
pd93, line 22: Given that we support `yaml` and `yml` for the main Taskfile. We might consider supporting both here too. I think a wider discussion about the file name would be good too. The current name was just a random idea. I'd like to gather some other ideas too and pick the best one(s).

trulede:
Did you ever consider multidoc yaml files? Then you can have many schemas within a single Taskfile.yml, and solve other interesting problems related with advanced Task integrations (see #1916).

Changes in and around reader.go, but nothing significant.

```yaml
---
version: 3
tasks: ...
---
task-experiments:
  TASK_X_FOO: BAR
---
foreign-schema:
  ignored-by-task: ...
```

The current mechanism for enabling experiments is a bit cumbersome. It would be ideal to have the option to _also_ configure this directly in the Taskfile. Additional files is perhaps not helping so much, for instance with remote taskfiles where the remote taskfile would need one of the experiments enabled.

website/docs/experiments/experiments.mdx
vmaerten: Updated ! 
I've added <Tab> to provide both examples

internal/experiments/experiments.go
vmaerten: Agree

internal/experiments/experiments.go
vmaerten: Also agree!

vmaerten:
Great review as always @pd93 thanks!
Creating the PoC didn’t take me too long, so I am to rework and/or discard if we do not want

Regarding the filename, how would you like to proceed? Should I post a message on Discord to discuss it?

@trulede, about configuring in the Taskfile itself, @pd93 already answered here https://github.com/go-task/task/issues/1978#issuecomment-2566004396

It can be tricky to manage if, for instance, three includes define the same experiments (e.g., ANY_MAPS) but with different values.

trulede:
> @trulede, about configuring in the Taskfile itself, @pd93 already answered here [#1978 (comment)](https://github.com/go-task/task/issues/1978#issuecomment-2566004396)
> 
I was suggesting using a multi-doc yaml file, containing the Taskfile (schema) _and other schemas in subsequent documents_ contained within the same yaml file. It just something to consider as you can place multiple docs/schemas in a single YAML file.

The question I might have in respect to this PR is why not use the existing mechanism `dotenv` where I can give the "experiments" file any name (and path) I want?  Yes, the `.env` file was problem, and any new name for the file is not going to be any better.


There is a more fundamental problem here; if I design a Taskfile to use an experimental feature, then that feature may no longer optional, so I need a way to ensure that its enabled. Adding a **new** schema to the YAML file which contains the taskfile schema would be a handy way to achieve that.

Adding a `.env` or `.task-experiments.yml` file suggests to me that a more general problem is emerging, and that a more durable solution could also be possible. Embedding new schemas in the Task file, and/or loading those configuration schemas from the traditional locations (e.g. /usr, /usr/local, ~/ or the taskfile.yml itself) would create a more fine grained configuration system ... or at least the potential for one.

pd93:
> I was suggesting using a multi-doc yaml file, containing the Taskfile (schema) _and other schemas in subsequent documents_ contained within the same yaml file. It just something to consider as you can place multiple docs/schemas in a single YAML file.

Although putting multiple schemas in a single YAML file is possible, I'm not keen on this approach. Experiments were not designed to be used in production (hence the name) and I have no intention to design large features just to support cases where users have experiments enabled across teams. I would prefer to keep experimental code far away from the schema as experiments are inherently very changeable.

However, I do understand that people will and are using experiments across teams and that there needs to be _some_ way (as @nathanperkins requested) to allow this. `.env` works fine for this and remains a very simple feature, but may conflict with other config files in a user's repo. This is why a Task-specific alternative was mentioned.

> The question I might have in respect to this PR is why not use the existing mechanism `dotenv` where I can give the "experiments" file any name (and path) I want? Yes, the `.env` file was problem, and any new name for the file is not going to be any better.

Can you explain (other than file conflicts) why you think the `.env` file is a problem?

Another reason that using the schema (including the dotenv) is not practical is because the Taskfile is not read/parsed and the dotenvs are not loaded until a significant amount of the code has executed. We may need/want to have experimental flags for things that run before a Taskfile is parsed. This is why the experiment flags are one of the first things to be evaluated in our code.

> There is a more fundamental problem here; if I design a Taskfile to use an experimental feature, then that feature may no longer optional, so I need a way to ensure that its enabled. Adding a **new** schema to the YAML file which contains the taskfile schema would be a handy way to achieve that.

Maybe I'm misunderstanding, but as far as I can tell loading an extra file such as `.env` or `.task-experiments.yml` which is committed to the same repo fixes this problem.

> Adding a `.env` or `.task-experiments.yml` file suggests to me that a more general problem is emerging, and that a more durable solution could also be possible. Embedding new schemas in the Task file, and/or loading those configuration schemas from the traditional locations (e.g. /usr, /usr/local, ~/ or the taskfile.yml itself) would create a more fine grained configuration system ... or at least the potential for one.

Loading from `/usr` or `/usr/local` would not solve sharing experimental Taskfiles. I'm open to a more granular configuration scheme with hierarchy etc, but this feels like a separate feature (albeit one that can also enable experiments if we want).

pd93:
> Regarding the filename, how would you like to proceed? Should I post a message on Discord to discuss it?

I think we can just leave this issue open for a bit (no rush to merge) and let @andreynering and other members of the community leave some feedback on the change and filenames. Feel free to put a message on Discord that links people here, but not everyone has a Discord account, so GitHub is a more transparent place for this feedback.

vmaerten:
> I think we can just leave this issue open for a bit (no rush to merge) and let @andreynering and other members of the community leave some feedback on the change and filenames. 

Of course, no need to rush at all :slightly_smiling_face: 

For the filename, I can think at : 
1. `.task-experiments.yaml`
2. `.task-config.yaml`
3. `.task-settings.yaml`

Feel free to comment and suggest filename

andreynering:
Hey guys,

What problem are we solving by having a separate config file? Can't the user just commit the `.env` file into the repository? I'm not saying I'm against the idea, just trying to understand if it is worth the added complexity.

If we decide to proceed, I strongly suggest `.taskrc.yml`, becase it's shorter and the `rc` suffix is a very stablished name for config files in the Unix world. Linters in the JS world also use that (`.eslintrc.json`, for example).

/cc @vmaerten @pd93

andreynering:
Also, a more generic name means we can expand to contain more settings in the future if needed, so having a specific name like `.task-experiments.yaml` I wouldn't recommend.

pd93:
> What problem are we solving by having a separate config file? 

The only problem I really have with `.env` is the fact that people may be using it for something else already and not want to pollute it with our settings. This all spun out of https://github.com/go-task/task/issues/1978#issuecomment-2566004396.
 
> If we decide to proceed, I strongly suggest `.taskrc.yml`

Not against this given the industry standard, though I always found this precedent weird since `rc` is supposed to be short for `runcom` or `run commands` which a settings file does not do 🤷

andreynering:
From [Claude](https://claude.ai/) (pasting just out of curiosity):

> **In configuration files like `.bashrc`, what does `rc` means?**
>
> The "rc" suffix in configuration filenames like `.bashrc` comes from Unix history. It stands for "run commands" or "runtime configuration." This naming convention dates back to the early days of Unix, specifically to the "/etc/rc" script that was used during system startup in 1977's UNIX 7th Edition.
> 
> The "rc" suffix became a convention for naming configuration files after that, and you can see it in many other examples:
> - .vimrc (Vim configuration)
> - .zshrc (Zsh shell configuration)
> - .npmrc (NPM configuration)
> - .screenrc (GNU Screen configuration)
> 
> This convention has persisted through decades of Unix and Linux development, though it's interesting to note that some modern programs opt for different naming schemes like using .config directories or .toml/.yaml files instead.
>
> **It's also used by some linters, right?**
>
> Yes, many linters and code quality tools use the "rc" suffix for their configuration files. Some common examples include:
> 
> - `.eslintrc` - ESLint configuration (JavaScript/TypeScript linting)
> - `.prettierrc` - Prettier configuration (code formatting)
> - `.babelrc` - Babel configuration (JavaScript transpiling)
> - `.stylelintrc` - Stylelint configuration (CSS linting)
> 
> Many of these tools also support alternative formats with additional extensions like `.eslintrc.json`, `.eslintrc.js`, or `.eslintrc.yaml`, giving developers flexibility in how they want to write their configurations. However, the base "rc" suffix remains as a nod to the Unix heritage we discussed earlier.

---

> Having said that, we could just set the schema to a `map[string]int` for the experiments and ignore invalid keys.

Yes, in theory we could just ignore any values. That said, I like the @nathanperkins idea (mentioned [here](https://github.com/go-task/task/issues/1978#issuecomment-2566011382) and [here](https://github.com/go-task/task/issues/1979)) that we would error if an unrecognized / retired experiment is enabled, to avoid surprises to the user.

vmaerten:
> Not against this given the industry standard, though I always found this precedent weird since rc is supposed to be short for runcom or run commands which a settings file does not do 🤷

From ChatGTP, it seems it also mean "runtime configuration" . We also have `.npmrc`

> The term rc in file names like .eslintrc.json stands for "run commands" or "runtime configuration".
It indicates that the file contains configuration settings for a specific tool or application.

---

> Also, a more generic name means we can expand to contain more settings in the future if needed, so having a specific name like .task-experiments.yaml I wouldn't recommend.

I also put this idea on the table here https://github.com/go-task/task/issues/1978#issuecomment-2566424671

 > The experiments specific file was really just to avoid backwards incompatible changes with schemas. For example, if we include settings in the same file (regardless of filename), we will occasionally be removing valid settings from the config file. Having said that, we could just set the schema to a map[string]int for the experiments and ignore invalid keys.

Yes, I started type as `map[string]int`. We could have a more global filename with a key like `experiments` typed as `map[string]int`, and ignore irrelevant keys / values

vmaerten:
Ahah @andreynering just did the same with ChatGPT :joy:

pd93:
@andreynering @vmaerten Interesting to see your AI tools of choice 😆 

TIL it stands for runtime configuration too. I had only ever heard the runcom thing. In that case, this makes a lot more sense and I'm totally fine with it.

Sounds like the consensus is `.taskrc.yml` and/or `.taskrc.yaml` then? We can reuse this going forwards for other thing if we want and for now it can just contain an experiments key of `map[string]int`.

I fine with erroring on unrecognised experiments. There is actually a [leftover line](https://github.com/go-task/task/blob/fe9f4897024cfcba25a7f25ea34a2dd2d9c37b3c/cmd/task/task.go#L147-L149) that does this for the old `ANY_VARIABLES` experiments anyway.

vmaerten:
> Sounds like the consensus is .taskrc.yml and/or .taskrc.yaml then? We can reuse this going forwards for other thing if we want and for now it can just contain an experiments key of map[string]int.

Yes, at least for me!

> I fine with erroring on unrecognised experiments. There is actually a [leftover line](https://github.com/go-task/task/blob/fe9f4897024cfcba25a7f25ea34a2dd2d9c37b3c/cmd/task/task.go#L147-L149) that does this for the old ANY_VARIABLES experiments anyway.

I'm fine with it but it could / should be done in another PR

vmaerten:
So for example, to enable Any maps second version, we could do : 
`.taskrc.yml` : 
```yaml
experiments:
  ANY_MAPS: 2

```

pd93:
Question: Do we want a schema `version` like we have in the main Taskfile. If so, should it match the Taskfile schema version? i.e.

```yml
version: 3

experiments:
  MAP_VARIABLES: 2
```

My opinion is probably yes to both of these. Main reason being that it might be confusing to have different versions for different things.

andreynering:
> Question: Do we want a schema `version` like we have in the main Taskfile?

Good point. I'm not sure, because I don't expect this schema to change much, if at all.

pd93:
> Good point. I'm not sure, because I don't expect this schema to change much, if at all.

Haha, I'm gunna screenshot this comment and save it for a couple of years 😉

vmaerten:
Haha, for now, we'll only have experiments in it. Maybe we can delay the decision until we add other things?

vmaerten:
I've updated the code based on what we discussed. Currently, the parsing is done in experiment, but we'll move it to a more global scope if we decide to use this file for other settings.

I've also created a JSON Schema hosted directly in our website.

For now, I did not add version but it can be challenged

trulede:
If you put an "experiment" into the Task executable, it's not an experiment, it's a feature. I understand the intention, but in doing so its created a config issue, and a difficult user experience. A plugin mechanism might be better, and open up Task to external developers.

Does Task really need a config? IMO nothing is being achieved with the current mechanism. If I design a task which uses an "experimental" feature, then I want to use that feature. Having some kind of enabling mechanism is just complicating the use of Task. Nothing else.

In any case, I suggest following the traditional patterns for config files (git does config this way). Start with something general (`taskconfig`) and then consider where to put it. No suggesting to implement it, just as a way of thinking about the problem.

* /usr/local/etc/taskconfig - system wide
* ~/.taskconfig - global for the user
* repodir/.taskconfig - repo only
* Taskfile - local (to the Taskfile) as additions schema docs ... (git does this too, repo local settings), you should be more open to that ... it would enable some interesting mechanisms.
* CLI options
* environment variables - always the highest priority (since CLI options might be baked into a container image)

You can take it a long way ... that is the problem with configs.

trulede:
> > If we decide to proceed, I strongly suggest `.taskrc.yml`
> 
> Not against this given the industry standard, though I always found this precedent weird since `rc` is supposed to be short for `runcom` or `run commands` which a settings file does not do 🤷

Correct. `rc` is not a great choice.

pd93:
> If you put an "experiment" into the Task executable, it's not an experiment, it's a feature. 

I don't understand what the alternative to having experiments in the executable would be. Where else would they go? Experiments are experimental _features_. They are things that we want to get feedback on before they are added without us having to make backwards compatible changes.

> I understand the intention, but in doing so its created a config issue, and a difficult user experience.

I don't believe there is a config issue. The way it works today is fine and has been since it was introduced. We're simply introducing an alternative to `.env` files so that conflicts do not occur. You still haven't clearly articulated what you think the problem with this approach is.

> A plugin mechanism might be better, and open up Task to external developers.

Plugins are complex to develop in a cross-platform way. Also, plugins tend to "plug in" to a specific part of a system. Experiments need to be able to modify any aspect of the code. Plugins are not a suitable solution here.

> Does Task really need a config? IMO nothing is being achieved with the current mechanism. If I design a task which uses an "experimental" feature, then I want to use that feature. Having some kind of enabling mechanism is just complicating the use of Task. Nothing else.

Experiments aren't always a case of "use it or don't". They are allowed (and usually do) _change_ behaviour. This means two users with the same task can see different behaviour if they have different experiment flags enabled. Its not complicating anything - It is necessary to stop us from breaking users projects.

> In any case, I suggest following the traditional patterns for config files (git does config this way). Start with something general (`taskconfig`) and then consider where to put it. No suggesting to implement it, just as a way of thinking about the problem.

Yes, if we have a proper settings config, then maybe one day it will be hierarchical like you suggested, but this is a long way out of scope for this issue/PR. We are not implementing settings here. We are implementing an alternative way of enabling experiments which _may_ use the same file as a potential future settings file.

> Correct. rc is not a great choice.

As discussed in this thread `rc` is perfectly acceptable as a "runtime configuration" file. Nothing wrong with it.

pd93 (commented):
> I've updated the code based on what we discussed. Currently, the parsing is done in experiment, but we'll move it to a more global scope if we decide to use this file for other settings.

Looks good! I left a couple of minor comments
 
> I've also created a JSON Schema hosted directly in our website.

Once the PR is merged we should submit this to https://github.com/SchemaStore/schemastore

Reference to the [main Taskfile schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/taskfile.json) (It just refers to the one we host on https://taskfile.dev) and the [catalogue entry](https://github.com/SchemaStore/schemastore/blob/24d1bb2bfcd3dba804d43770139815085e2213c2/src/api/json/catalog.json#L6032)

> For now, I did not add version but it can be challenged

I personally lean towards adding it now so that we don't have to maintain support for no version when we inevitably decide to add it later. However, lets get everyone aligned on this before you change anything.

trulede:
> > If you put an "experiment" into the Task executable, it's not an experiment, it's a feature.
> 
> I don't understand what the alternative to having experiments in the executable would be. Where else would they go? Experiments are experimental _features_. They are things that we want to get feedback on before they are added without us having to make backwards compatible changes.
> 
> > I understand the intention, but in doing so its created a config issue, and a difficult user experience.
> 
> I don't believe there is a config issue. The way it works today is fine and has been since it was introduced. We're simply introducing an alternative to `.env` files so that conflicts do not occur. You still haven't clearly articulated what you think the problem with this approach is.
> 

The problem with `.env` as a file name is already clear I would think.

It works, yes, but operationally the use of experiments is a nuisance. I understand the intention, but here is how it goes in a development cycle:

1/ Decide to use a remote task file. Actually, we use versioned Taskfiles (git tags) and associated containers, from remote repos - so this is a very powerful usage pattern.
2/ Implement that the easiest way, which currently is by using the remote taskfile feature.
3/ Overcome the operational complexity of Task - all related to the "experimental" behaviour. We can do this several ways, but for now have a shell script which automates Task (yeah, really ...) and sets environment variables.
4/ Document how to overcome the operational complexity of Task, for other developers and users.
5/ Hope that people only use the shell script, because no one ever reads the docs.
6/ Accept that the behaviour of remote taskfile feature will change in the future, and we will have to adapt to that.

Point 6 is important. We accept the behaviour will change, or disappear. 
Point 3 is illustrative as to how we deal with [Task operational difficulties](https://github.com/boschglobal/dse.sdp/blob/f9b13ede8568b112ebc2b6c4fce80c62837e99f2/examples/openloop/openloop.sh#L14). Note, the Taskfile is not in the root of the repo, and we avoided the complexity of a separate config file, or even thinking about it! It's just a mess to even consider - Task can be frustrating like that.

Now you get the feedback, I guess. Feature is great, operation is very un-task-like.

> 
> > Correct. rc is not a great choice.
> 
> As discussed in this thread `rc` is perfectly acceptable as a "runtime configuration" file. Nothing wrong with it.

It's not a great choice. 'rc' historically stands for "run commands" which is semantically different to configuration.

Also, are you sure about versioning and a schema for a config file? Take a look at how other config files work; git or docker; that is not normally how config files are implemented. Of course, nothing wrong with it either.


It's much easier if switches for behaviour are represented in some kind of normal way. Environment variables, at the root of the taskfile schema, would be ideal. Simple, versioned schema already exists, can simply parse the taskfile early in the process and extract the necessary content. No practical difference to the current `.env` file parsing, really.


Only seems to be adding complexity to me. But as I explained above, I've avoided the config file, so I have no skin in the game.

website/docs/experiments/experiments.mdx
```diff
@@ -39,23 +42,41 @@ Which method you use depends on how you intend to use the experiment:
 1. Prefixing your task commands with the relevant environment variable(s). For
    example, `TASK_X_{FEATURE}=1 task {my-task}`. This is intended for one-off
    invocations of Task to test out experimental features.
-1. Adding the relevant environment variable(s) in your "dotfiles" (e.g.
+2. Adding the relevant environment variable(s) in your "dotfiles" (e.g.
```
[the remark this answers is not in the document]
vmaerten: Yeah I know
As you said, it does not make a difference here

internal/experiments/experiments.go
```diff
@@ -17,6 +19,15 @@ import (
 
 const envPrefix = "TASK_X_"
 
+var defaultConfigFilenames = []string{
+	".taskrc.yml",
+	".taskrc.yaml",
+}
+
+type ExperimentConfigFile struct {
```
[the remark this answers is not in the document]
vmaerten: You are totally right! I always forgot to check exported vs non exported (I am doing too much typescript at work atm :joy: )

Done !

internal/experiments/experiments.go
```diff
@@ -17,6 +19,15 @@ import (
...
+	Experiments map[string]string `yaml:"experiments"`
```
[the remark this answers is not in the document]
vmaerten: @pd93 Good question!  

Since experiment values are typed as strings (because environment variables are strings), I considered two options:  

1. Define `Experiments map[string]string` with `yaml:"experiments"` as `map[string]string`, allowing the parser to handle the conversion. This approach lets me keep the existing code.  
2. Define `Experiments` as `map[string]int` and modify the code to convert all environment variables to integers.  

I choose the first one to keep the existing code but you may be right, we could / should define that experiment should be only int

vmaerten:
>Once the PR is merged we should submit this to [SchemaStore/schemastore](https://github.com/SchemaStore/schemastore?rgh-link-date=2025-01-02T21%3A12%3A18Z)

> Reference to the [main Taskfile schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/taskfile.json?rgh-link-date=2025-01-02T21%3A12%3A18Z) (It just refers to the one we host on [taskfile.dev](https://taskfile.dev/)) and the [catalogue entry](https://github.com/SchemaStore/schemastore/blob/24d1bb2bfcd3dba804d43770139815085e2213c2/src/api/json/catalog.json#L6032)

Noted I'll do it :) 

> I personally lean towards adding it now so that we don't have to maintain support for no version when we inevitably decide to add it later. However, lets get everyone aligned on this before you change anything.

Actually both are fine to me, but yeah let's wait a bit to get everyone aligned

vmaerten:
I rebased this with the latest experiment modifications.
As we discussed, I've converted everything to integers.
Regarding the version, I have a proposal: put the version in the schema but do not validate it yet, as we do not need it for experiments.

AFAIK, we just need to be on the same page about the version to move forward with this.

Any thought ? /cc @pd93 & @andreynering

pd93:
> I rebased this with the latest experiment modifications.
 
Will need another rebase to pass the required 1.24 tests now. Sorry 😛 

> As we discussed, I've converted everything to integers.

Nice, alongside the other changes, this is much tidier now IMO.

> Regarding the version, I have a proposal: put the version in the schema but do not validate it yet, as we do not need it for experiments.
> 
> AFAIK, we just need to be on the same page about the version to move forward with this.
> 
> Any thought ? /cc @pd93 & @andreynering

I think let's leave it for now. Rereading this thread made me agree with your earlier point:

> Haha, for now, we'll only have experiments in it. Maybe we can delay the decision until we add other things?

Experiments are fundamentally not stable, so I don't think we need to make a decision about a schema version until we add something else to this file and I don't want to hold up the PR unnecessarily.

internal/experiments/experiments.go
```diff
@@ -6,13 +6,26 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
+	"gopkg.in/yaml.v3"
+
 	"github.com/joho/godotenv"
 	"github.com/spf13/pflag"
```
pd93, line 14: Nit: Can we remove the newlines so that `gofumpt` sorts these?

pd93 (approved):
Awesome. Thanks for doing this 🚀

Issue #935, opened 2022-11-16
Publish `task` as a Python package


sisp:
I'd love to use `task` as the task runner in a Python project without requiring developers to install `task` manually. In my opinion, it is much better to use `task` over the plethora of language-specific task runners because it tends to be more feature-rich than many alternatives (as far as I can tell) and can be used for projects independent of their programming languages across all major operating systems without requiring a runtime (like the Python interpreter) thanks to the compiled Go binaries. I'd hope that `task` would gain further adoption with additional language-native packages and development efforts would focus on this project rather than being scattered across the language-specific alternatives.

I believe a Python package could be created quite easily with some inspiration from the following projects:
* https://github.com/shellcheck-py/shellcheck-py
* https://github.com/editorconfig-checker/editorconfig-checker.python

Would you be open to accepting a contribution that adds support for publishing a Python package?

Related to #870.

andreynering:
Hi @sisp,

I have zero experience with the Python world, so to even consider doing this I need some guidance from the community.

npm was easy because I'm more familiar with the JS world and the process was relative straightforward.

For anyone that would use this, please add your 👍 so we know how many would really use this.

retr0h:
hi @andreynering @sisp I could help with this if interested.  Would likely utilize [poetry](https://python-poetry.org/) as the mechanism for publishing the package.

sisp:
@retr0h I do like Poetry for Python projects, but I don't think it's useful or even beneficial in this case. The projects referenced build packages that essentially download the binary for the respective platform during installation of the package. This can be done nicely in `setup.py` but I don't quite see how it would be done with Poetry, or how Poetry would be better than using `setup.py` and [`twine`](https://github.com/pypa/twine) (the standard publishing utility for Python) for publishing the package.

sisp:
@retr0h Old tools but widely adopted. There's nothing wrong with that. Anyway, how would you implement the code that downloads the binaries for the given platform and architecture during package installation with Poetry?

Undertone0809:
I see a implement https://pypi.org/project/go-task-bin/#description

pd93:
This is one of the most upvoted issues, but I can't tell if there is actually any reason to keep it open? I'm not a heavy Python user, so haven't verified if https://pypi.org/project/go-task-bin is a good resolution.

If we're happy with it, then it can be added to our [docs](https://taskfile.dev/installation/) as a community installation method and this can be closed.

andreynering:
[...]

PR #1989, opened 2025-01-02
docs: updated installation guides

pd93:
Standardises the layout for all installation methods, adds labels and links to sources where possible. Also reorders the list with the official methods at the top and community owned ones at the bottom.

Also closes the longstanding #935.

### DIFF
--- a/website/docs/installation.mdx
+++ b/website/docs/installation.mdx
@@ -1,6 +1,7 @@
 ---
 slug: /installation/
 sidebar_position: 2
+toc_max_heading_level: 4
 ---
 
 import Tabs from '@theme/Tabs';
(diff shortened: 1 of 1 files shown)


andreynering (approved):
Great, this page needed an overhaul 👏 

The only small thing I noticed that would be nice to fix if easy, is that link anchors for the titles that have badges now have extra dashes:

image

This means that existing links for these anchors will be broken, and new links will look slightly uglier.

pd93:
> link anchors for the titles that have badges now have extra dashes:

Yeah, I noticed this just as you commented! Thankfully Docusaurus allows you to [manually set the anchor ids](https://docusaurus.io/docs/next/markdown-features/toc#heading-ids), so I have updated them all to match the previous values now.

I have also added the `pypi` package as discussed in #935.

PR #1993, opened 2025-01-05
feat: add global precondition

### LINKED ISSUE #294: Add global preconditions
It would be great to be able to set global preconditions to prevent running any tasks in a directory/repo if the precondition is not met. This would be a simple way to add a precondition to all tasks without having to add the same precondition to every task individually.

The use case for this is for example to enforce that a repository has been initialized somehow before allowing to run any other tasks, like setting up release targets for goreleaser or similar. In this example I would have the .goreleaser.yml contain a string like "INIT_PACKAGE" which should be replaced by the package name of the repository, and until that's done all tasks should fail (precondition greps for "INIT_PACKAGE"), and return the same message "This repository hasn't been initialized, please run 'task init'".

Obviously to be able to run the initial `task init` there needs to be a property to disable the global precondition for that specific task. So the Taskfile would end up something like this:

```
version: '2'

preconditions:
    - grep INIT_PACKAGE .goreleaser.yml
      msg: "this repository needs to be initialized, run 'task init' to continue"

tasks:
  build:
    desc: build go project
    cmds:
      - go build .
  
  init:
    desc: initalize repository
    cmds:
      - <do something>
    ignore_global_preconditions: yes
```

vmaerten:
Fixes #294 

It's a fairly popular feature


vmaerten:
I was wondering how it should behave if an included Taskfile defines global preconditions:  

1. Should it produce an error (similar to dotenv)?  
2. Should all preconditions be merged and applied together?  
3. Should the global preconditions in the included Taskfile be ignored?

I started with the 2 in mind, but I am not sure anymore

Any thought ? cc @pd93  / @andreynering

pd93:
As a user, I would expect 2. This is also the safer option if we maintain the principle of "its safer to not run a task that was intended to run than it is to run one that wasn't.

vmaerten:
I followed the same pattern we have with vars.
I am not a huge fan of naming `preconditions` inside `preconditions`,  maybe `values` ? 
I you have some naming ideas

taskfile/ast/precondition.go
```diff
@@ -2,16 +2,56 @@ package ast
...
+type (
+	Preconditions struct {
+		Preconditions []*Precondition
+		mutex         sync.RWMutex
+	}
+
+	Precondition struct {
+		Sh  string
+		Msg string
+	}
```
pd93, line 23: I've moved `Vars` and `Var` into separate files in another PR. Would be good to follow this pattern here.

precondition.go
```diff
@@ -14,7 +14,7 @@ import (
 var ErrPreconditionFailed = errors.New("task: precondition not met")
 
 func (e *Executor) areTaskPreconditionsMet(ctx context.Context, t *ast.Task) (bool, error) {
-	for _, p := range t.Preconditions {
+	for _, p := range append(t.Preconditions, e.Taskfile.Preconditions.Preconditions...) {
```
pd93, line 17: I would expect the global `preconditions` to run first. For example:

```yaml
version: '3'

preconditions:
  - test -f foo.txt

tasks:
  default:
    preconditions:
      - test -f bar.txt
    cmds:
      - echo "default task"
```

Assuming that neither file exists, I would expect to see

```
task: `test -f foo.txt` failed
task: precondition not met
```

Also prefer `slices.Concat` over `append` and `...` now that its in the std lib.

```suggestion
	for _, p := range slices.Concat(e.Taskfile.Preconditions.Preconditions, t.Preconditions) {
```

taskfile/ast/taskfile.go
```diff
@@ -59,28 +60,33 @@ func (t1 *Taskfile) Merge(t2 *Taskfile, include *Include) error {
 	if t1.Tasks == nil {
 		t1.Tasks = NewTasks()
 	}
+	if t1.Preconditions == nil {
+		t1.Preconditions = NewPreconditions()
+	}
 	t1.Vars.Merge(t2.Vars, include)
 	t1.Env.Merge(t2.Env, include)
+	t1.Preconditions.Merge(t2.Preconditions)
```
pd93, line 68: Because we don't have scoping, this is going to lead to issues where global preconditions in included Taskfiles are going to impact tasks in parent files. I think this is going to be pretty unintuitive to users at the moment. For example:

```yml
# Taskfile.yml
version: '3'

includes:
  include: include.yml

tasks:
  default:
    cmds:
      - echo "default task"
```

```yml
# include.yml
version: '3'

preconditions:
  - test -f foo.txt
```

```
task: `test -f foo.txt` failed
task: precondition not met
```

We could:

1. Leave as it is, `vars` already behaves like this. We'll fix it later 🤷‍♂️ 
2. Not allow global preconditions in included taskfiles for now
3. Not merge this functionality until scoping is ready

I think 1 is pretty reckless. We're essentially intentionally adding behaviour that we know is confusing and will raise questions.

3 is very conservative while 2 is a good compromise. It depends how urgently we want to roll out this functionality. If we don't think its urgent then maybe we wait. The question is how long will scoping actually take? I've already taken a stab at it and I can tell you - it's significant piece of work.

pd93 (commented):
Thanks for working on this! It's definitely a key feature. A couple of bits that need discussion though.

pd93:
I completely missed the previous comments when reviewing this 😆 Including my own. 

> I followed the same pattern we have with vars. I am not a huge fan of naming `preconditions` inside `preconditions`, maybe `values` ? I you have some naming ideas

Yeah, `Preconditions.Preconditions` does stutter. I think `Values` is fine, This is what we do for `Vars`.

> > I was wondering how it should behave if an included Taskfile defines global preconditions:
> > 
> > 1. Should it produce an error (similar to dotenv)?
> > 2. Should all preconditions be merged and applied together?
> > 3. Should the global preconditions in the included Taskfile be ignored?
> > 
> > I started with the 2 in mind, but I am not sure anymore
> > 
> > Any thought ? cc @pd93 / @andreynering
>
> As a user, I would expect 2. This is also the safer option if we maintain the principle of "its safer to not run a task that was intended to run than it is to run one that wasn't."

I didn't think about the scoping issues when writing this comment originally. I think option 2 is still the long-term goal here, but we might have to think about doing something like 1 or 3 in the meantime. See https://github.com/go-task/task/pull/1993#discussion_r1960465733

---

P.s 10 years on this platform and I've only just noticed that if you highlight text in a comment before clicking "Quote Reply", it will only quote the selected text 👀 Amazing.

precondition.go
vmaerten: I agree
thanks for the suggestion

taskfile/ast/precondition.go
vmaerten: Yeah
I've splitted in two files

vmaerten:
> I didn't think about the scoping issues when writing this comment originally. I think option 2 is still the long-term goal here, but we might have to think about doing something like 1 or 3 in the meantime. See [#1993 (comment)](https://github.com/go-task/task/pull/1993#discussion_r1960465733)

I think scoping can take some time to roll out, and we could benefit from this feature soon. I agree with you that option 2 is the long-term goal.

I've chosen to produce an error if a precondition is in an included Taskfile. Why? Because I prefer explicit over implicit.
Of course, it can be challenged.

We could merge this feature now and improve it once the scoping feature is ready.

---

P.S : I did not know too, thanks !!

PR #2002, opened 2025-01-11
docs: fix some examples to use spaces instead of tabs

jonathanagustin:
Because if you don't use spaces, it looks like this:

image


andreynering:
Thank you!

PR #2006, opened 2025-01-19
chore: migrate the template engine to sprout

### LINKED ISSUE #1638: Migrate from `sprig` (`slim-sprig`) to `sprout`?
[sprig](https://github.com/Masterminds/sprig/issues/396) is the library that we use for template functions. We forked it on [slim-sprig](https://github.com/go-task/slim-sprig) because the original imports to many packages and also had many functions that do not make sense for a template (crypto functions, etc).

Today, I learned that there's an attempt fork sprig given it hasn't had any activity for over 2 years:

- https://github.com/Masterminds/sprig/issues/396
- https://github.com/42atomys/sprout/issues/1
- https://github.com/42atomys/sprout

It's far from ready, but I decided to open this issue so we don't forget to take a look on how's the fork going in the near future.

42atomys:
This pull request includes the change discussed on issue #1638, primarily focused on updating dependencies, migrating from `slim-sprig` to `sprout`, and refactoring the templater functions.

### Changes
- Updated the Go version from `1.22.0` to `1.23.3` to ensure compatibility with Sprout.
- Replaced `github.com/go-task/slim-sprig/v3` with `github.com/go-sprout/sprout`.
- Removed `slim-sprig` imports and added `sprout` and its registries.
- Implemented backward compatibility for original `sprig` methods.
- Moved templater functions to a new `taskfunc` registry to follow [Sprout standard](https://docs.atom.codes/sprout/advanced/how-to-create-a-registry).
- Added comprehensive tests for the new `taskfunc` package to ensure template functions are correctly tested.
- Update the documentation to match sprout documentation.

Resolve #1638


42atomys:
Hi @andreynering, it's been a long time. I preferred to wait for the v1.0.0 release instead of implementing an unstable version, and now here we are.

I've reflected on all the changes and discovered a few points that need discussion 🙏

1. Regarding the Go Task Functions, you've implemented a `merge` function that behaves like `mergeOverride` in Sprig. This implementation currently overrides the original `merge` behavior, making `merge` and `mergeOverride` functionally the same, although they're based on different libraries. To maintain compatibility, I added an alias to ensure the same behavior is retained, but this seems a bit strange to me: https://github.com/go-task/task/blob/8f2a0519d3cd9d01970f1b05d7ddaaf66d655d69/internal/templater/taskfunc/task_func.go#L56  
What would you prefer? Should we keep this alias and drop the packaged `merge` function? Or should we add a deprecation notice to clarify the difference and guide users to update their templates?

2. I’ve included all functions from `slim-sprig` to define which registries to include. As discussed, the `crypto` package is excluded 😉. Additionally, `semver` and `uuid` are not included either. Let me know if you'd like fewer or more functions.

3. I’ve followed your comments in the code to mark deprecated functions and added other deprecation rules to align with naming conventions:  
https://github.com/go-task/task/blob/8f2a0519d3cd9d01970f1b05d7ddaaf66d655d69/internal/templater/taskfunc/task_func.go#L66-L74  
> [!TIP]  
> Deprecation notices are non-breaking changes! They inform users about updates they need to make to their templates before old functions are removed. This ensures no breaking changes between versions. [Sprout Notice Documentation](https://docs.atom.codes/sprout/features/function-notices#deprecated)

4. While developing Sprout, I discovered several bugs in functions, including a few panics and memory errors. I’ve fixed them, so some functions might now behave differently (e.g., they returned a panic before but now return `nil`). This is just an informational note.

5. To follow Go template conventions, all functions should be errorful. In Sprig, some functions have an error signature `(any, error)` while others don’t. Similarly, some have a `Must` version, and some don’t. In Sprout, we decided to enforce error signatures for consistency, meaning all functions return an error when something goes wrong.  
Sprout also provides the option for [safe function versions](https://docs.atom.codes/sprout/features/safe-functions) to have a no-error signature (disabled by default). Let me know if you’d like to include this feature.

Thanks for your patience and for taking the time to answer me!

PR #2007, opened 2025-01-20
fix: `HTTPNode.Location` when building graph

jeongukjae:
After this [commit](https://github.com/go-task/task/commit/c77c8a419b0c4b6905882c61bc663df5aaef6289), building internal graph can raise error because HTTPNode's Location value can be changed while building graph. (See line 65 (`node.URL = url`))

So if taskfile imports remote file with multiple depth (e.g. A --> remote B --> remote C) without filename, it raises

```
source vertex https://{PATH_TO_REMOTE}/Taskfile.yaml: vertex not found
```

So fixing `HTTPNode.Location` not to be changed over time.

(I couldn't find any related issues, so please let me know if there's issue or I have to create it)

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1239,6 +1239,10 @@ func TestIncludesRemote(t *testing.T) {
 			firstRemote:  srv.URL + "/first/Taskfile.yml",
 			secondRemote: "./second/Taskfile.yml",
 		},
+		{
+			firstRemote:  srv.URL + "/first/",
+			secondRemote: srv.URL + "/first/second/",
+		},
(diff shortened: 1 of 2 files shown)


vmaerten (approved):
lgtm! Thanks

Indeed, when no filename is provided, the node.URL is inconsistent, which causes issues with the graph.

jeongukjae:
@vmaerten Hi! thanks for reviewing! could you merge this one please?

pd93:
@jeongukjae Thanks! 🚀

PR #2011, opened 2025-01-22
feat: make Taskfile initialization less verbose by default

### LINKED ISSUE #2009: Taskfile initialization is too verbose
### Summary
When initializing a new Taskfile with `--init`, it prints the whole file contents to the terminal, which I find unnecessarily verbose.
Also, the `-s, --silent` flags don't seem to affect it at all.

### Suggestion
Both:
- make `--init` take `--silent` into account and don't print non-error messages if it is present, and
- print just the filename instead of the file contents by default: `Taskfile.yml created in the current directory`
  - additionally, hook up `-v, --verbose` to print the file contents if the user **does** want it

HeCorr:
Fixes #2009.

When using the `--init` flag to create a new Taskfile, it currently prints the whole contents of the file to the terminal, which is unnecessarily verbose (and honestly seems to be unintentional).

This PR fixes it by only printing the filename by default and taking the `--silent` and `--verbose` flags into account to control the behavior:

```
$ task --init
Taskfile.yml created in the current directory
```

```
$ task --init --silent
```

```
$ task --init --verbose
# https://taskfile.dev

version: '3'

vars:
  GREETING: Hello, World!

tasks:
  default:
    cmds:
      - echo "{{.GREETING}}"
    silent: true

Taskfile.yml created in the current directory
```

Please note that I purposefully did not update the description of the `--silent` and `--verbose` flags so please let me know if you want me to and what should I change them to.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1022,7 +1022,9 @@ func TestInit(t *testing.T) {
 		t.Errorf("Taskfile.yml should not exist")
 	}
 
-	if err := task.InitTaskfile(io.Discard, dir, 0); err != nil {
+	l := &logger.Logger{Verbose: true}
+
+	if err := task.InitTaskfile(io.Discard, dir, 0, l); err != nil {
 		t.Error(err)
 	}
 
--- a/task_test.go
+++ b/task_test.go
@@ -1022,7 +1022,7 @@ func TestInit(t *testing.T) {
 		t.Errorf("Taskfile.yml should not exist")
 	}
 
-	if err := task.InitTaskfile(io.Discard, dir); err != nil {
+	if err := task.InitTaskfile(io.Discard, dir, 0); err != nil {
 		t.Error(err)
 	}
 
--- a/task_test.go
+++ b/task_test.go
@@ -1011,27 +1011,6 @@ func TestCmdsVariables(t *testing.T) {
 	assert.Contains(t, buff.String(), tf)
 }
 
-func TestInit(t *testing.T) {
-	t.Parallel()
-
-	const dir = "testdata/init"
-	file := filepathext.SmartJoin(dir, "Taskfile.yml")
-
-	_ = os.Remove(file)
-	if _, err := os.Stat(file); err == nil {
-		t.Errorf("Taskfile.yml should not exist")
-	}
-
-	if err := task.InitTaskfile(io.Discard, dir); err != nil {
-		t.Error(err)
-	}
-
-	if _, err := os.Stat(file); err != nil {
-		t.Errorf("Taskfile.yml should exist")
-	}
-	_ = os.Remove(file)
-}
-
 func TestCyclicDep(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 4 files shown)


HeCorr:
I'm considering making the `Taskfile.yml created in the current directory` text green to differentiate it from the file content output if it is present. Thoughts?

pd93 (approved):
I really like this :) Changes seem fine to me. I don't think there is a huge need to update the flag descriptions for this. It should be fairly intuitive.

I do think we could update the `Usage > Getting Started` section to talk about `--init` though. We don't mention it anywhere on that page, despite being an intended starting point.

Not fussed if we do this here or in a separate PR.

pd93:
No problem. I'm happy to add this later. Waiting for a second maintainer to review before we merge.

vmaerten (approved):
Nice addition! :)
Thanks!

HeCorr:
Before this gets merged, should I make the success text green as suggested [here](https://github.com/go-task/task/pull/2011#issuecomment-2607552535)?

pd93:
@HeCorr fine with the green as long as it uses the existing code to colourise the string and respects `NO_COLOR` etc.

HeCorr:
> as long as it uses the existing code to colourise the string and respects `NO_COLOR` etc.

It should, let me commit that change in a little bit and you let me know if it seems correct.

HeCorr:
Oops, I completely missed the test call. No clue how to get a logger in there though.

pd93:
> Oops, I completely missed the test call. No clue how to get a logger in there though.

This is a good point, we should avoid passing the logger into the package API where possible (I know we don't always follow this right now). Ideally, the package API should only do the file creation and the logging should be moved to the CLI. This also means that the verbosity doesn't need to be passed in and we won't break the package API for users.

task_test.go
pd93, line 1: I know this test wasn't originally created by you, but can we move `TestInit` into a new `init_test.go` file? I think we should keep the `task_test.go` for executor tests, not function tests.

HeCorr: I'm going to bed right now but I'll be happy to do so tomorrow.

HeCorr:
> and we won't break the package API for users

I had failed to realize what you meant by that yesterday, but after some very needed sleep I understand it now.

HeCorr:
I'm working on moving the printing into `cmd/task/task.go` but I am facing naming conflicts because the logger is defined as:
```go
logger := &logger.Logger{
    Stdout:  os.Stdout,
    Stderr:  os.Stderr,
    Verbose: flags.Verbose,
    Color:   flags.Color,
}
```

Which means that I can't access `logger.Green` because I am no longer referencing the package, but the struct.
Should I rename the variable to `l`, `log` or something else?

task_test.go
HeCorr: Done in 53ab1e87e8b111aae429ed5f5a78e96e6facc323

pd93 (approved):

pd93:
Thanks @HeCorr! 🚀

PR #2018, opened 2025-01-25
feat: specify --init filename/path

### LINKED ISSUE #2008: Specify --init filename/extension
### Summary
I personally much prefer the `.yaml` extension over `.yml`, but it doesn't seem like I'm able to specify which one to use when initializing a Taskfile with `--init`, so I'm always forced to rename the file afterwards.

### Suggestion
Either:
- add the `-e, --extension` flags that let me specify which file extension to use (either `.yaml` or `.yml`), or
- repurpose the existing `-t, --taskfile` flags to specify the target filename if `--init` is also present
  - maybe also support a wildcard for the filename? i.e. `task -i -t *.yaml` ➡️ `Taskfile.yaml`

HeCorr:
Closes #2008.

Task currently does not provide a way to specify which filename to use when initializing a file (hardcoded as `Taskfile.yml`), forcing users to rename the file/extension after creation.

This PR fixes that by taking the target filename from the first non-flag argument:
```
$ task --init tasks.yml
Taskfile created: tasks.yml
```

If a path with no name (extension only) is provided, the default name "Taskfile" is prepended to it:
```
$ task --init .yaml
Taskfile created: Taskfile.yaml
```

Different directories are also supported:
```
$ task --init sub/NotMakefile.yml
Taskfile created: sub/NotMakefile.yml
```
```
$ task --init sub/.yaml
Taskfile created: sub/Taskfile.yaml
```

This PR also includes the following updates to the docs:

- Fixed the `Flags` header being inside the Tip admonition in [CLI Reference](https://taskfile.dev/reference/cli) page:

image image

- Updated the `--taskfile` flag information:

image
image


website/docs/reference/cli.mdx
```diff
@@ -45,11 +45,30 @@ If `--` is given, all remaining arguments will be assigned to a special
...
+- `taskfile.dist.yml`
+- `Taskfile.dist.yaml`
+- `taskfile.dist.yaml`
+
+:::note
+
+You can use the `--taskfile` flag to run files of any name.
+
+:::
+
```
pd93, line 71: Not convinced that this section is needed. It is already in our `usage.mdx` doc and is not specific to the CLI (i.e. the same defaults apply to the package API too).

taskfile/taskfile.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/go-task/task/v3/internal/sysinfo"
 )
 
+const DefaultTaskInitFilename = "Taskfile.yml"
```
pd93, line 19: We're in the process of stabilising/reducing the package API surface. I don't think this needs to be here. It's more convenient to keep this with the code that produces the file IMO.

init.go
```diff
@@ -23,17 +22,13 @@ tasks:
...
 `
 
-const defaultTaskfileName = "Taskfile.yml"
-
 // InitTaskfile Taskfile creates a new Taskfile
-func InitTaskfile(w io.Writer, dir string) error {
-	f := filepathext.SmartJoin(dir, defaultTaskfileName)
-
-	if _, err := os.Stat(f); err == nil {
+func InitTaskfile(w io.Writer, filepath string) error {
```
pd93, line 26: I prefer to avoid `filepath` as a variable name as it conflicts with the stdlib package of the same name. We're not currently importing it here, but it can lead to confusing error down the line. IMO the std package name was an awful choice, but here we are.

`filePath` is fine if you want to keep the name though. Or use the same name (`path`) as you used in the caller.

cmd/task/task.go
```diff
@@ -77,7 +78,13 @@ func run() error {
 		if err != nil {
 			return err
 		}
-		if err := task.InitTaskfile(os.Stdout, wd); err != nil {
+		name := taskfile.DefaultTaskInitFilename
+		if len(flags.Entrypoint) > 0 {
+			// Replace `*` with `Taskfile` so `*.yaml` results in `Taskfile.yaml`
+			name = strings.Replace(flags.Entrypoint, "*", "Taskfile", 1)
+		}
```
pd93, line 85: The syntax of this isn't intuitive to me. A `*` usually denotes a wildcard. i.e. A character that can be any value, not a placeholder for a specific value. A better syntax would follow the standards set by various print formatters (`%t`) or go templating (`{{.DefaultTaskfileName}}`). 

The latter is just worse way of writing `Taskfile` and while I could be persuaded on % formatting strings, we don't use this anywhere else in Task, and I'm not convinced this functionality is actually that useful.

Second opinions from @andreynering and @vmaerten welcome.

pd93 (commented):
Thanks for this! Generally supportive of this improvement. Just a couple of discussion points on the implementation/docs.

website/docs/reference/cli.mdx
HeCorr: oh, I totally missed that, my bad. I'll remove it.

taskfile/taskfile.go
HeCorr: > keep this with the code that produces the file

Do you mean near [`InitTaskfile()`](https://github.com/go-task/task/blob/0ea6e0086c561f06e74a63f4591c4abe6265f6b6/init.go#L24)? Or where [it is used](https://github.com/go-task/task/blob/97e2a2d6e3e2c83f8de8eeda15956c6987e28770/cmd/task/task.go#L81)?

I've defined it as a const above `defaultTaskfiles` so it's easier to find in case it ever needs to be changed, instead of mixed in with runtime logic as a literal string.

Also, I didn't define it at [`init.go`](https://github.com/go-task/task/blob/0ea6e0086c561f06e74a63f4591c4abe6265f6b6/init.go) because IIRC I couldn't reference it from `task.go`.

I'm open to suggestions on what to do instead.

cmd/task/task.go
HeCorr: My motive was making it as short (and fast) as possible to init a `Taskfile.yaml` file: `task -it *.yaml` (so I don't have to type in "Taskfile").

That said, I totally understand your points.

Another solution could be to detect just `yaml` and prepend `Taskfile.` to that, or maybe drop the `*` and detect `.yaml` instead.

taskfile/taskfile.go
pd93: > Do you mean near [`InitTaskfile()`](https://github.com/go-task/task/blob/0ea6e0086c561f06e74a63f4591c4abe6265f6b6/init.go#L24)? Or where [it is used](https://github.com/go-task/task/blob/97e2a2d6e3e2c83f8de8eeda15956c6987e28770/cmd/task/task.go#L81)?

I mean near `InitTaskfile`. I missed that the value was being passed in from `cmd/task/main.go`.

How about this:

- `InitTaskfile()` accepts a `path`. Could be a directory OR a file.
- In `cmd/task/task.go`, we get the `wd` as before and _only if_ an entrypoint is defined, we `filepathext.SmartJoin(wd, name)`.
  - This means, by default we pass in the `wd` unless `--taskfile` is given
- In `InitTaskfile()`, we replace the stat check with this:

```go
	if fi, err := os.Stat(f); err == nil && !fi.IsDir() { ... }
```

- This means that we only error if a _file_ already exists
- Below this, we can add a line that will add the default filename if `fi.IsDir()` is true.

The nice thing about this is that it doesn't break the package API. Anyone calling `InitTaskfile` with a directory will get the same result as before, but anyone passing a specific file name will now get that specific file. It also means that we're not using the `defaultTaskfile` const outside of `init.go`, so it can stay where it was and that all the logic for handling default files is encapsulated in the package API, not the CLI.

Hope this makes sense and I haven't missed anything. I've not actually tried any of what I've suggested 😆

HeCorr: yeah that sounds good, I'll give it a try.

cmd/task/task.go
trulede: If you want to save typing then you could use a command alias. Having all kinds of behaviour behind such a flag is confusing for what seems like a fairly simple command.

task --init
task --init Taskfile.yaml
task --init some/folder
task --init some/folder/Taskfile.ymal

Those all seem OK. Similar pattern to git init <directory>.

HeCorr: > task --init Taskfile.yaml

That actually makes much more sense. I feel stupid for not considering this.

pd93: Also fine with this if you want to change it.

HeCorr: So the problem with using a string flag which needs to have three states (unset, set, set but empty) is having to use a pointer, which is not used at the moment in any other flag.

If you guys are okay with that and how it would have to be checked, then I can proceed.
```go
if flags.Init != nil {
```

HeCorr: Oh, interesting.. it seems I can't just do:
```go
var Init *string
pflag.StringVarP(Init, "init", "i", "Taskfile.yml", "...")
```
...as it causes a segmentation fault.

This does seem to work:
```go
var Init *string
Init = pflag.StringP("init", "i", "Taskfile.yml", "...")
```
...but the flag's argument is required:
```
$ task --init
flag needs an argument: --init
exit status 2
```
I'll try to use positional arguments instead but I'm not sure if it'll work well. I'm open to ideas.

HeCorr:
@pd93 not sure why `TestInit` is failing now. It didn't fail in #2011, did it?

vmaerten:
> @pd93 not sure why `TestInit` is failing now. It didn't fail in #2011, did it?

It didn't. It fails because, previously in the test, we passed the directory containing the Taskfile as the second parameter, and we concatenated the filename in the `InitTaskfile` method.
Now, the concatenation has been removed, `os.stat` is performed on the directory itself, and it already exists
It prevents the `InitTaskfile` to override existing file.

 (see : https://github.com/go-task/task/blob/23df1f0c6163ea9df1ac4dfaff85bff640df5b59/init_test.go#L23 and https://github.com/go-task/task/pull/2018/files#diff-09d6aefefdc39400d949bac4606f951988cbe618740e96ecfee50236859da58bR26-R32)

HeCorr:
oh, oops. I must have messed up the merge then. thanks.

taskfile/taskfile.go
HeCorr: this is turning out to be much more complicated than I thought. I can't just check `fi.IsDir()` because `fi` might be nil if the path doesn't exist, so I'm having to figure out some `filepath` trickery to differentiate between a dir and a file path, plus updating the "Taskfile created" log to include the relative path it was created at is going to be interesting.

HeCorr: Having some logic inside `InitTaskfile` is proving to be a bit problematic, because I need to replicate that logic at `task.go` to log the correct final path. Not sure what to do at the moment.

HeCorr:
So, this is the problematic code, specifically the highlighted part:
image

I had added the `fp.Dir(path)` part to fix other edge-cases, but it created others.

Could we simply change the printing format altogether to `Taskfile created at %s`? It would greatly simplify this code.

HeCorr:
Refactor complete, manual tests passing, ready to remove Draft status.

Tested commands:
| Command | Result |
|--------|--------|
| `task --init` | `Taskfile created: Taskfile.yml` |
| `task --init tasks.yml` | `Taskfile created: tasks.yml` |
| `task --init .yaml` | `Taskfile created: Taskfile.yaml` |
| `task --init sub/` | `Taskfile created: sub/Taskfile.yml` |
| `task --init sub/tasks.yml` | `Taskfile created: sub/tasks.yml` |
| `task --init sub/.yaml` | `Taskfile created: sub/Taskfile.yaml` |

taskfile/taskfile.go
pd93: > I need to replicate that logic at `task.go` to log the correct final path

This goes against what I've said previously about breaking the API, but I think it would be nice if `InitTaskfile` returned the path of the generated file anyway. This would save some of the pain.

We can still make breaking changes to the package API if needed, but there is an effort to stabilise it for the future, so we need to make good decisions now to make it as usable as possible in the future. If we are struggling with this, then end users will too.

HeCorr: @pd93 so what's the plan? is your suggested API change going to be discussed and implemented in this PR or could we merge this one as-is for now and open another one to discuss the improvement?

pd93: If you're happy to add the changes to the PR now, I'm happy to accept them. Adding in a follow-up PR is fine too if you would prefer. @vmaerten and @andreynering welcome to add their thoughts too, but I don't think this will be controversial.

HeCorr: Nothing against doing it here, I just didn't want the API change to delay the feature itself in case it took a long time to be approved.

HeCorr:
Modified `InitTaskfile()` to return the final calculated file path [as suggested](https://github.com/go-task/task/pull/2018#discussion_r1938519229), which was honestly the right call as that may also help with error handling in case initialization fails.

Only a small quirk with the current implementation which prints the path relative to the current dir is that this may be printed depending on the input path:
```
$ task --init ~/Desktop
Taskfile created: ../../../../../../Desktop/Taskfile.yml
```

...but it is generally much shorter:
```
$ task --init
Taskfile created: Taskfile.yml
```

HeCorr:
Updated the docs one last time as well as the original PR message.

pd93 (approved):

pd93:
@HeCorr Thanks for the hard work a perseverance on this! I've made a couple of small adjustments in ff8c913.

PR #2031, opened 2025-01-30
fix(checker): use only one checker at the same time to improve perf

### LINKED ISSUE #2029: 3.41.0 Regression in sources behavior cause all files to be read even with method: timestamp
### Description

- Running tasks with sources method timestamp causes all files to be fully read 

This purple block in this graph is the data in `sources` being read 
The yellow/green block in the graph is the actual task being run 

Image

I noticed this behavior when upgrading to 3.41.0. Reverting to 3.40.1 resolves this issue completely. 

Even with `method: timestamp` in the task definition it still reads all files. 

If I had to guess it was regressed by https://github.com/go-task/task/pull/1872

It appears to set up both checkers regardless of method 

https://github.com/go-task/task/blob/f5121de4681c708e6a01ee5837b09d923f968ca6/variables.go#L131-L147

which ultimately calls 

https://github.com/go-task/task/blob/f5121de4681c708e6a01ee5837b09d923f968ca6/internal/fingerprint/sources_checksum.go#L90-L115

I believe this behavior to be a bug as when declaring a `method: timestamp` that seems to me that the user expelictly does not want an expensive checksum done 

To fix this I suppose the first referenced code block needs to check the `method` of the task and use this to set up the correct checker. 

### Version

3.41.0

### Operating system

Alpine Linux (task is runing in a docker container)

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: "3"

  cache_sync:
    method: timestamp
    sources:
      - "{{ .CACHE_PATH }}/media/**/*"
    cmds:
      - |
        docker run \
          -v "$CACHE_PATH/media:/src" \
          -v "$BACKING_DATA_PATH/media:/dest" \
          -e "PUID=${PUID}" \
          -e "PGID=${PGID}" \
          --rm \
          --pull=never \
          --name=cache_sync \
          rsync
```

vmaerten:
fixes #2029 

@pd93 Before, both `{{.TIMESTAMP}}` and `{{.CHECKSUM}}` were available in `status`, but according to our documentation, only one should be available depending on the chosen method.

cf : https://taskfile.dev/usage/#using-programmatic-checks-to-indicate-a-task-is-up-to-date

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -965,9 +965,12 @@ func TestStatusVariables(t *testing.T) {
 		Verbose: true,
 	}
 	require.NoError(t, e.Setup())
-	require.NoError(t, e.Run(context.Background(), &ast.Call{Task: "build"}))
+	require.NoError(t, e.Run(context.Background(), &ast.Call{Task: "build-checksum"}))
 
 	assert.Contains(t, buff.String(), "3e464c4b03f4b65d740e1e130d4d108a")
+
+	buff.Reset()
+	require.NoError(t, e.Run(context.Background(), &ast.Call{Task: "build-ts"}))
 
 	inf, err := os.Stat(filepathext.SmartJoin(dir, "source.txt"))
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


andreynering (approved):
Good catch!

pd93 (approved):
👍 Nice one

Issue #2036, opened 2025-02-03
Proposal: Merge `vars` and `env`


pd93:
### Description

One of the things I see a lot, is confusion about the differences between variables and environment variables. They essentially serve the same purpose (holding data), but they have a few differences:

- Both can be parsed into Task via the CLI as strings (though they have different syntaxes)
  - `TASK_VAR=foo task ...` (environment variable)
  - `task ... VAR=foo` (variable)
- Both can be defined in the Taskfile (using different keyword) at a global or task level.
  - `vars` can also be set when calling another task, but `env` cannot.
- `vars` can be of any type. `env` only supports strings (though _sort of_ works with other scalar values).
- `vars` are output using templating syntax and an `env` is output using shell variable syntax.
  - `$FOO` (environment variable)
  - `{{.FOO}}` (variable)
- They have slightly different inheritance/scoping rules.
  - At this point, I'm not even sure what the differences are without investigating properly.

I'm not convinced that any of these distinctions are necessary and I feel like they add an additional layer of confusion for both new users and people wanting to contribute to the project.

Therefore, I propose removing the `env` key entirely. `vars` would then inherit from the system environment and be overridable in Task. If a variable is not a string, it would be output using the stringer version of its value (`%v`). All of the above features below would then apply to `vars`:

- Variables are only passed into Task via the shell environment (whether that be in an .rc file of a shell command).
  - No more `task ... VAR=foo` syntax.
- Variables can be defined at a global or task level and passed when calling dependant tasks.
- Variables can be of any type, though they are still limited to strings when passed via the environment variables.
- Variables are output using templating syntax OR shell variable syntax depending on the user's needs.
- Variables have a consistent inheritance/scoping model as defined in #2035.

Before:

```yaml
version: '3'

env:
  FOO: 'foo'
vars:
  BAR: 'bar'

tasks:
  default:
    cmds:
      - echo '$FOO'
      - echo '{{.FOO}}'
      - echo '$BAR'
      - echo '{{.BAR}}'
```

```
foo


bar
```

After:

```yaml
version: '3'

vars:
  FOO: 'foo'
  BAR: 'bar'

tasks:
  default:
    cmds:
      - echo '$FOO'
      - echo '{{.FOO}}'
      - echo '$BAR'
      - echo '{{.BAR}}'
```

```
foo
foo
bar
bar
```

[...]

Issue #2035, opened 2025-02-03
Proposal: Variable scope and inheritance


pd93:
### Description

This is another area that has caused a lot of confusion. The [`ENV_PRECEDENCE` experiment][1] has mitigated some of this, but does not solve all of the issues. The main cause of our problems is the merging that is done when parsing Taskfiles. In this process, we lose all information about the priority of variables and where they were set.

This leads to a situation where variables are not scoped correctly and leak from a child Taskfile into a parent.

I propose that variables should be set with the following precedence, starting with the lowest priority and finishing with the highest. Each item in the list inherits from the previous items and overrides any existing variables.

1. **Environment Scope**
   1. Set in the host shell
   1. Set in the command line when calling task
1. **Entrypoint Scope**
   1. Global `dotenv` files in the entrypoint Taskfile
   1. Global `vars` in the entrypoint Taskfile
1. **Include Scope(s)** (in the order in which they are included)
   1. Variables specified in the `include` statement
   1. ~~Global `dotenv` files in each included taskfile~~ ([not currently supported](https://github.com/go-task/task/issues/1075))
   1. Global `vars` in each included taskfile
1. **Task Scope**
   1. Variables passed into a task via another task
   1. Task-level `dotenv` files
   1. Task-level `vars`

Note the term "scope". Scopes are used to sandbox variables so that they don't leak into other tasks. Variables in a higher priority scope cannot be accessed from a scope of lower priority or a sibling scope. For example, a variable set in one task, cannot be accessed in another task because that would be a sibling scope. Likewise, it cannot be accessed in a parent Taskfile since this is a lower priority scope.

This kind of precedence and scoping can be achieved by utilizing the new Taskfile DAG and removing merging entirely. This allows us to traverse the graph when searching for a Task and resolve variables as we go. This also has the distinct advantage that we only need to resolve variables that are relevant to the task(s) being called, rather than resolving all variables in all Taskfiles.

[1]: https://github.com/go-task/task/issues/1038

trulede:
How is that different to the existing? Aside from the implied combination of the last two items (in the list below).

Variables can be set in many places in a Taskfile. When executing [templates](https://taskfile.dev/reference/templating/), Task will look for variables in the order listed below (most important first):

* Variables declared in the task definition
* Variables given while calling a task from another (See [Calling another task](https://taskfile.dev/usage/#calling-another-task) above)
[...]

PR #2038, opened 2025-02-04
feat: variable inheritance tests

pd93:
Adds some tests for how Task handles inheritance. This will be useful for stopping regressions and reviewing any potential future changes such as #2035 and #2036.

I have enabled the `ENV_PRECEDENCE` experiment for these tests.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -2,6 +2,7 @@ package task_test
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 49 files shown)


task_test.go
```diff
@@ -3207,6 +3207,112 @@ func TestReference(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var buff bytes.Buffer
+			os.Setenv("VAR", "shell")
+			os.Setenv("ENV", "shell")
+			e := task.Executor{
+				Dir:    fmt.Sprintf("testdata/var_inheritance/%s", test.name),
```
andreynering, line 3302: I wonder if it would make sense to scope by `v2`. We could soon just create a `v3` dir and keep both tests for a while.

```suggestion
				Dir:    fmt.Sprintf("testdata/var_inheritance/v2/%s", test.name),
```

andreynering (approved):
Thanks for adding these tests!

task_test.go
pd93: I'm not sure I'm following. These are testing v3 functionality are they not?

andreynering: I'm sorry. I meant v3 and v4. 😅 

It was just an idea, though.

pd93: Ah understood. I don't have an issue with this. I will update and merge

PR #2042, opened 2025-02-05
refactor: unify how executor tests are written

pd93:
Currently we have a couple of different ways of writing tests in `task_test.go`. Some use the the `fileContentTest` template, some directly craft and run an `Executor`.

This is a draft PR for a new test type called `TaskTest`. The idea is that this is a "do everything" testing suite for the Executor. Instead of relying on handcrafted outputs written inside the go file or creating special taskfiles that output content to a file, it uses `goldie` to create test "golden files". The API uses functional options and post-processing functions to setup tests.

As there are _a lot_ of tests in here, I wanted to gather some feedback on the approach before I continue. I'm also interested in splitting this file up as its currently at over 3400 lines long, but looking for some opinions on the best way to do this. I had considered renaming the existing `testdata` folder to `tests` and moving these tests there alongside their taskfiles, but this doesn't feel very "go-like"?


vmaerten:
Hey @pd93 
First of all thanks for doing this !! I had to write / modify some tests recently and it's painful

>  I'm also interested in splitting this file up as its currently at over 3400 lines long, but looking for some opinions on the best way to do this. I had considered renaming the existing testdata folder to tests and moving these tests there alongside their taskfiles, but this doesn't feel very "go-like"?

I totally agree with you about splitting this file. It may not be very "Go-like," but it makes everything more readable. Having everything in one place is convenient. Every time I work on tests, I find it somewhat difficult to switch between the test file and the testdata directory.

Also, having the golden file named this way makes it easy to identify.

Last but not least, the diff is amazing : 
image


At work, we use [cupaloy](https://github.com/bradleyjkemp/cupaloy), which is also a great tool for handling snapshots.

I am totally in!

pd93:
Just rebased onto the 1.24 changes. I'm going to mark this as ready for review. I would prefer to do this in chunks rather than rewrite every test in one go.

vmaerten (approved):
Once this PR is merged, I'll help you to re-write everything !
Good job mate :heart:

pd93:
Quick update. I've migrated a few more tests. Still lots to do though!

I discovered a [really cool pattern in connectrpc](https://github.com/connectrpc/connect-go/blob/main/option.go) (a project I use a lot). Basically, instead of your regular boring functional option types:

```go
type ThingOption func(*ThingToMutate)
```

you can implement them using an interface type instead:

```go
type ThingOption interface {
    applyToThing(*ThingToMutate)
}
```

The advantage of this is that you can create multiple types that accept _the same_ functional options. This was always something that bothered me about the pattern, but is completely solved by this. The boilerplate is a little bit more verbose, but it makes the API very nice to use.

The main reason I want to use this is that part of my long-term plan for the `Executor` is to start separating the bits of the code that do not execute tasks out of it. For example, the `Reader` and `Formatter` (to be created). This will give us more usable/testable components and reduce the API surface of the `Executor` which is currently doing _a lot of things_.

Most of the tests in `task_test.go` are currently `Executor` tests. However, some of them test the output of other things, such as the `ListTasks` or `Status` functions. If we separate these things out of the `Executor`, we will still need a good way to test them and I want to build this into the test suite. So allowing functional options like `WithName` to be passed into _both_ `ExecutorTest` _and_ `FormatterTest` will stop code duplication and keep the API consistent between test types.

Separating a `Formatter` type out of the `Executor` will take a fair bit of work and I plan to combine this with the work in my [List Experiment Proposal](https://github.com/go-task/task/issues/2091) if we decide to proceed with it. However, splitting these tests now will lay the groundwork for good tests later.

Hope this all makes sense! I've started to move the tests that have been migrated into `executor_test.go` and `formatter_test.go` so its clear which ones need to be done (the ones still in `task_test.go`. We can think about moving them to another location later.

andreynering (approved):

PR #2044, opened 2025-02-06
fix: schema does not allow ignore_error in a for loop

gokeefe-atb:
Visual Studio Code shows a problem when adding `ignore_error` to a `for` loop (`for_cmds_call`), however in practice it appears to work perfectly fine and I can't find anything in the task code that would indicate it shouldn't be allowed.

This change simply adds `ignore_error` as an allowed property of a `for` loop.

## Before

001731 Taskfile yml - projects-ai (Workspace)  SSH_ lxms-d-intops01 atb com  - Visual S

001732 Taskfile yml - projects-ai (Workspace)  SSH_ lxms-d-intops01 atb com  - Visual S

## After

001733 Taskfile yml - projects-ai (Workspace)  SSH_ lxms-d-intops01 atb com  - Visual S

### DIFF
--- a/website/src/public/schema.json
+++ b/website/src/public/schema.json
@@ -442,6 +442,10 @@
           "description": "Silent mode disables echoing of command before Task runs it",
           "type": "boolean"
         },
+        "ignore_error": {
+          "description": "Prevent command from aborting the execution of task even after receiving a status code of 1",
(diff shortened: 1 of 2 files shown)


vmaerten (approved):
Thanks!

PR #2045, opened 2025-02-06
Remote Taskfiles: redact credentials of remote URLs in prompts and logs

iwittkau:
This is to improve https://github.com/go-task/task/issues/1317.

I've implemented redaction support for the `taskfile.Node` interface, specifically for the `taskfile.GitNode` implementation which was logged with the clear-text user access tokens before.


taskfile/node_git.go
```diff
@@ -34,37 +34,43 @@ func NewGitNode(
...
 	opts ...NodeOption,
 ) (*GitNode, error) {
-	base := NewBaseNode(dir, opts...)
-	u, err := giturls.Parse(entrypoint)
+	gitURL, err := giturls.Parse(entrypoint)
 	if err != nil {
 		return nil, err
 	}
+	if gitURL.Scheme == "http" && !insecure {
+		return nil, &errors.TaskfileNotSecureError{URI: entrypoint}
```
iwittkau, line 42: I've moved this up following the early-return principle.

taskfile/node_git.go
```diff
@@ -22,10 +21,11 @@ import (
 // An GitNode is a node that reads a Taskfile from a remote location via Git.
 type GitNode struct {
 	*BaseNode
-	URL    *url.URL
```
iwittkau, line 25: I saw no point in keeping this full URL type since it was only used to retrieve the base URL.

taskfile/node_git.go
```diff
@@ -34,37 +34,43 @@ func NewGitNode(
...
 	}, nil
 }
 
 func (node *GitNode) Location() string {
-	return node.rawUrl
+	return node.rawURL
+}
+
+func (node *GitNode) RedactedLocation() string {
+	return node.redactedRawURL
```
iwittkau, line 73: Using `node.URL.Redacted()` didn't work because it contained only the base URL without the Taskfile file path.

taskfile/node.go
```diff
@@ -18,6 +18,7 @@ type Node interface {
 	Read(ctx context.Context) ([]byte, error)
 	Parent() Node
 	Location() string
+	RedactedLocation() string
```
pd93, line 21: Just had a quick look on my phone, not a full review.

Is the raw (non-redacted) location actually used outside of the git node implementation? You could make the argument that the credentials should never be exposed via the interface and that `Location` should just return the redacted version. This would save us having to add an interface method that only one of the implementations needs.

vmaerten: I agree with you @pd93 !
AFAIK, the `Location` is used only to reference it in the `graph` so, imho no objection to return the `redacted` version.
Also, the http node implementation could / should have the same behavior, as we can use `http://USER:PASS@URL` pattern

iwittkau: That would make the whole change a lot simpler! I saw that the `node.Location` is used to provide the cache key.

I think ignoring the username and password combination for the cache should be fine since different combinations shouldn't yield different Taskfile contents.

Adding the HTTP node redaction should be easy too.

taskfile/node.go
```diff
@@ -22,7 +22,7 @@ type Node interface {
 	Remote() bool
 	ResolveEntrypoint(entrypoint string) (string, error)
 	ResolveDir(dir string) (string, error)
-	FilenameAndLastDir() (string, string)
+	FilenameAndLastDir() (lastDir string, file string) // TODO the return order is implemented opposite to the naming
```
iwittkau, line 25: It looks like this caused some bugs? https://github.com/go-task/task/issues/1317#issuecomment-2620027809

iwittkau:
I've turned this into a draft for now.

taskfile/node_git.go
```diff
@@ -122,5 +121,5 @@ func (node *GitNode) ResolveDir(dir string) (string, error) {
 }
 
 func (node *GitNode) FilenameAndLastDir() (string, string) {
-	return filepath.Base(node.path), filepath.Base(filepath.Dir(node.path))
+	return filepath.Base(filepath.Dir(node.filepath)), filepath.Base(node.filepath)
```
iwittkau, line 124: `taskfile/cache.go` uses it like this:

```go
lastDir, filename := node.FilenameAndLastDir()
```

iwittkau:
Compiling it and running this as a development version actually breaks my own git-based Remote Taskfiles setup 😅 

Back to draft, I guess.

taskfile/node_git.go
```diff
@@ -34,37 +33,37 @@ func NewGitNode(
...
 
-	rawUrl := u.String()
+	ref := gitURL.Query().Get("ref")
+	fullURL := gitURL.Redacted()
 
-	u.RawQuery = ""
-	u.Path = basePath
+	gitURL.RawQuery = ""
+	gitURL.Path = urlPath
+	baseURL := gitURL.Redacted()
```
iwittkau, line 54: We actually have to use the unredacted URL here.

taskfile/node.go
iwittkau: Probably not, it was just always wrong and I can't see why it would cause what was described in the comment.

pd93:
Hi @iwittkau. Sorry for that. As you said, the code is changing a lot at the moment so things often go out of date very quickly. Thanks for your time though. Hopefully the conversation and changes here are still a useful reference to anyone implementing this in the future.

PR #2047, opened 2025-02-07
feat: add TASK_SOURCES special variable

### LINKED ISSUE #948: Use sources as variable

fhofherr:
This commit adds the `TASK_SOURCES` variable to tasks. If a task defines the `sources` key, `TASK_SOURCES` contains the list of paths referenced by `sources`. Globs are expanded.

This was initially implemented by @hernandanielg in #1146 to fix #948. The PR seemed to have stalled since, which is why I'm now opening this new PR. In contrast to #1146 my changes set `TASK_SOURCES` to a list, which allows users to use template functions to transform the value of `TASK_SOURCES`. Additionally, I decided to name the variable `TASK_SOURCES` instead of `SOURCES`, as I wanted to make clear that this variable only exists for tasks.

I'm aware of this [comment](https://github.com/go-task/task/issues/948#issuecomment-2146718626) by @callegar which wondered if `sources` should be changed to accept the value of variables. However, this seemed like a bigger change to me (I may be wrong since I don't know the code in detail). Therefore I opted to stick with the original proposal.

The downside of my change is, that globs get expanded whenever the `TASK_SOURCES` variable is set. I'm not yet sure under which circumstances `Compiler.FastGetVariables` and `Compiler.GetVariables` are called. Both of those methods call the `getVariables` method which in turn calls the `getSpecialVariables` method I modified. Would it be enough to just pass `evaluateShVars` to `getSpecialVariables` (maybe renaming it to `fast` in the process) and skip glob expansion if it is `true`?


internal/compiler/compiler.go
```diff
@@ -184,19 +185,27 @@ func (c *Compiler) ResetCache() {
...
 		"USER_WORKING_DIR": c.UserWorkingDir,
 		"TASK_VERSION":     version.GetVersion(),
 	}
 	if t != nil {
+		dir := filepathext.SmartJoin(c.Dir, t.Dir)
+
+		taskSources, err := fingerprint.Globs(dir, t.Sources)
+		if err != nil {
+			return allVars, fmt.Errorf("expand globs: %v", err)
+		}
```
vergenzt: Should this be moved to the end of this block so that `TASK`, `TASK_DIR`, etc. all still get assigned even if there's an error on glob expansion?

vergenzt:
I would love for this to get merged!

Cheekie25:
Hello,
Could we know why this was closed? I believed it could have been merged quickly once @vergenzt comment would have been taken into account.
Really looking towards this feature !

fhofherr:
I no longer use task. I thus don't really have a chance to test this properly. I can address the comment in the next few days. I've re-opened the PR.

fhofherr:
@vergenzt I've updated the PR and addressed your review comment.

Cheekie25:
> I would love for this to get merged!

@vergenzt I believe this PR is ready.
Anything to add or you would be willing to merge it ?
Thanks, looking forward to use Task once this is available.

vergenzt:
> > I would love for this to get merged!
> 
> 
> 
> @vergenzt I believe this PR is ready.
> 
> Anything to add or you would be willing to merge it ?
> 
> Thanks, looking forward to use Task once this is available.

I'm not a maintainer unfortunately. 😅 Just a drive by reviewer

Cheekie25:
Hello @vmaerten 
Would you mind having a look at this PR and hopefully merge it if everything if fine for you ?
Would greatly appreciate !

compiler.go
```diff
@@ -197,19 +198,27 @@ func (c *Compiler) ResetCache() {
...
 	if t != nil {
+		taskDir := filepathext.SmartJoin(c.Dir, t.Dir)
+
 		allVars["TASK"] = t.Task
-		allVars["TASK_DIR"] = filepathext.SmartJoin(c.Dir, t.Dir)
+		allVars["TASK_DIR"] = taskDir
 		allVars["TASKFILE"] = t.Location.Taskfile
 		allVars["TASKFILE_DIR"] = filepath.Dir(t.Location.Taskfile)
+
+		taskSources, err := fingerprint.Globs(taskDir, t.Sources)
```
trulede, line 217: I think you need to use the logic in the following code, perhaps extracting to a function, so that you get consistent behaviour.
https://github.com/go-task/task/blob/90bcbe9ba5b4e0c80f8f8ea3dbd2dffcaf8bc47b/variables.go#L352

Then, having done that, how many times are the sources getting globed when a task runs? That might be worthwhile checking too ... once should be enough.

Also, it you will do this for TASK_SOURCES then you need to do it for TASK_GENERATES too.

testdata/special_vars/Taskfile.yml
```diff
@@ -22,3 +22,14 @@ tasks:
   print-task-dir:
     dir: 'foo'
     cmd: echo {{.TASK_DIR}}
+
+  print-task-sources:
+    sources:
+      - sources-as-var/**/*.txt
```
trulede, line 28: You probably want to also include globs that don't match a file. So its clear how that case is handled - I think non-matches would not be in the list. One non-match with a wildcard and one without.

executor_test.go
```diff
@@ -382,6 +384,7 @@ func TestSpecialVars(t *testing.T) {
 					task.WithDir(dir),
 					task.WithSilent(true),
 					task.WithVersionCheck(true),
+					task.WithForce(true),
```
trulede, line 387: This line impacts on all the existing tests too. Is that really what you want?

compiler.go
trulede, line 227: Do you need to set a "nil" value here too?

trulede (changes requested):

fhofherr:
Hi @trulede,

Thanks for the review. My workflow has shifted and I’m no longer using Task, so I won’t be able to address these comments or continue with the PR.

If anyone else wants to take this over or build on it, they are more than welcome to do so!

trulede:
@SergioChan in case you are interested to try another issue, this PR is abandoned.

trulede:
Will close. Someone can bring it back to life via #948.

Issue #1508, opened 2024-02-14
Use fsnotify instead of watcher


sirenkovladd:
I have more than 700 files in my project and that's a lot of files to recheck for update (timestamp or hash)

[fsnotify](https://github.com/fsnotify/fsnotify) should be faster and use less CPU than [watcher](https://github.com/radovskyb/watcher)

pd93:
Related to #1179

pdf:
Specificallyy https://github.com/go-task/task/issues/1179#issuecomment-1752978694

andreynering:
Hi @sirenkovladd and @pdf. I've done that on https://github.com/go-task/task/pull/2048.

andreynering:
Implemented on https://github.com/go-task/task/pull/2048

Issue #1179, opened 2023-05-22
Task is not getting re-executed after addition/deletion of files when used with --watch argument. 


PaulisMatrix:
> Please describe the bug you're facing. Consider pasting example

Task is not getting re-executed after addition/deletion of files when used with `--watch` argument. 
Below is the sample taskfile for it. 

```yaml
version: '3'

tasks:
  go-runs:
    desc: Task for building a new binary
    cmds:
      - go build -o main
      - go install 
      - echo 'new binary available'
    sources:
      - '**/*.go
```
Works fine without `--watch` arg when the command `task go-runs` is ran but have to do that manually every time which can be avoided by using `--watch` arg but its not working for it as explained above. 

- Task version: 3.0 
- Operating System: MacOS 12.4

It seems to be an issue from the watcher pkg itself. 
https://github.com/radovskyb/watcher/issues/120

radcool:
@PaulisMatrix, can you please run the latest released version of **task** (3.29.1) with the sample taskfile you provided and with the verbose (`-v`) flag?

Then, please confirm the following:

- "touching" a new file matching the glob pattern in `sources` will trigger a `watching new file` event, but will not run the task itself
- "touching" the same file again (or modifying its content in any way) _will_ trigger running of the task

radcool:
@PaulisMatrix, I've observed the same behavior as you. I also expected tasks being watched to execute anytime a file matching those tasks' sources' glob pattern would be added, but noticed that they weren't.

I took at the code of [watch.go](https://github.com/go-task/task/blob/main/watch.go) and realized this is what is happening:

1. When `task <task_name>... -w` is run, for every task a goroutine is started to listen to [watcher]("github.com/radovskyb/watcher) events.

2. At the same time, a second go routine is started to scan for files matching the sources' glob pattern. Any files that match get added to the watcher. The goroutine sleeps the `watchInterval` duration, and then scans for files matching the sources' glob pattern all over again. Any file the watcher isn't already watching gets added to the watcher. This is how files created _after_ the task program has started get added.

3. The watcher watches any file added (whether initially or subsequently and sends an event on a channel to the first goroutinene every time one of the files changes (size, content, permissions, etc.). Whenever such a file change event is received, the first goroutine runs the task again.

[...]

Issue #985, opened 2023-01-17
Extremely high CPU usage with `-w`


max-sixty:
- Task version: v3.19.1 (h1:2KMJk6mDBacSPuFxPFvlvvHJwGZtU/hN2ENZpaFqR5s=)
- Operating System: MacOS, M1

I'm running a task with `-w` and seeing extremely high CPU usage every few seconds. This is the Vitals chart:

<img width="451" alt="image" src="https://user-images.githubusercontent.com/5635139/212815561-f1df9efd-45bc-48cf-a8a4-8139a7a7d937.png">

I thought it might be the underlying task, but I replaced the command with `echo hello`, and I'm seeing the same behavior. And the task is not re-running during this time (i.e. it's just waiting).

If anyone would like to fully reproduce, it's here; and then `task -w test-rust`
https://github.com/max-sixty/prql/tree/task-repro. Here's the link to the task to view it: https://github.com/max-sixty/prql/blob/701c4141441430f94932a780b32ce8de40b599d8/Taskfile.yml#L181-L208

I'd previously changed our docs to suggest `task` over `watchexec`, as that has some problems with ignoring files in https://github.com/PRQL/prql/pull/1380. Zero stress if the repo is too big / doesn't work well with a large tree as `cargo` though.

andreynering:
Hi @max-sixty,

You mentioned you are using `v3.19.1`. Some performances improvements were made on #982 by @pd93 that I believe may help you.

Can you let me know if you see the same issue on [v3.20.0](https://github.com/go-task/task/releases/tag/v3.20.0)?

max-sixty:
This does still have the same issue with 3.20.0.

Interestingly it only happens on `task -w test-rust` and not `task -w test-rust-fast`. The former uses (almost) the full path, which contains a huge number of files:

```
fd -uu . | wc -l
  208708
```

vs for `test-rust-fast`, which uses a subpath:

```
fd -uu . prql-compiler/ | wc -l
     123
```

---

Is `task` stat-ing the whole path on every pass rather than subscribing to changes, by any chance? If so, that seems very reasonable for small paths, less good for large paths.

pdf:
On this repository, with `Task version: v3.39.2 (h1:Zt7KXHmMNq5xWZ1ihphDb+n2zYLCo4BdRe09AnMMIgA=)`:

https://github.com/delaneyj/datastar

Running `task -w templ`:

```yaml
  templ:
    env:
[...]

Issue #160, opened 2019-01-08
watch: interrupt ongoing task


stelcheck:
Use case: I have a task that starts a server in development mode. I would like to use `watch` to interrupt the ongoing build task before starting the new one. 

I assume this would be the desired behavior for long-running build tasks as well.

andreynering:
Hi @stelcheck,

So, Task actually kill the previous long running command before running it again, BUT it doesn't wait the previous one to finish before it runs the new one. I think this may be causing the issue you're having.

This is unfortunate, but it's due limitation of the Go `context` package (which is asynchronously). I'll have to study how to improve that.

stelcheck:
That would indeed cause problems, at the very least intermittently.

The issue I am seeing though seems to be a bit different. Either:

1. The subprocess is killed with SIGKILL or another more exotic signal
2. Logs stop being outputted while the child process is terminating

I'd be happy to see if I can fix this myself if you can give me a pointer as to where to look in the code.

stelcheck:
Actually, I am now convinced the task is not killed: once I send a signal to the task command, **all** instances of the application receive it at once.

```plaintext
 task run -w
task: Started watching for tasks: run
INFO[0000] Logger system set                             forceColors=true format=text
DEBU[0000]                         ___
DEBU[0000]     o__        o__     |   |\
DEBU[0000]    /|          /\      |   |X\
DEBU[0000]    / > o        <\     |   |XX\
INFO[0000] Setting up metrics system                     subpath=/metrics
INFO[0000] Setting up discovery system                   address="127.0.0.1:8500" scheme=http
INFO[0000] Setting up HTTP Server system                 address="127.0.0.1:8080" prefix=/
INFO[0000] Goal server is up and running
INFO[0000] Logger system set                             forceColors=true format=text
DEBU[0000]                         ___
DEBU[0000]     o__        o__     |   |\
DEBU[0000]    /|          /\      |   |X\
DEBU[0000]    / > o        <\     |   |XX\
INFO[0000] Setting up metrics system                     subpath=/metrics
INFO[0000] Setting up discovery system                   address="127.0.0.1:8500" scheme=http
INFO[0000] Setting up HTTP Server system                 address="127.0.0.1:8080" prefix=/
INFO[0000] Goal server is up and running
INFO[0000] Logger system set                             forceColors=true format=text
DEBU[0000]                         ___
DEBU[0000]     o__        o__     |   |\
DEBU[0000]    /|          /\      |   |X\
[...]

PR #2048, opened 2025-02-08
feat(watch): migrate to fsnotify

andreynering:
We're migrating from [watcher](https://github.com/radovskyb/watcher) to [fsnotify](https://github.com/fsnotify/fsnotify), which is more performant and will avoid issues with high CPU usage.

This will close the following issues:

- https://github.com/go-task/task/issues/1508
- https://github.com/go-task/task/issues/985
- https://github.com/go-task/task/issues/1179

TODO:

- `fsnotify` recommends us to watch directory instead of individual files.
  - [x] Add a function to `internal/fingerprint` to return only the directories and watch for those.
  - [x] When an event comes, we'll need to double check if the file is included in `sources`.
- [ ] ~Mark the `--interval` flag and `interval:` setting as deprecated. We should probably emit a warning if set, as from now on these will be ignored.~
  - I actually reconsidered this and I'm reusing the interval setting as the time we wait for suplicated events. Changed default from 5 seconds to 100ms.
- [x] [Dedup duplicated events](https://github.com/fsnotify/fsnotify/blob/main/cmd/fsnotify/dedup.go)

Possible post-launch improvements:

- https://github.com/go-task/task/issues/1179#issuecomment-2675822570
- https://github.com/go-task/task/issues/160


watch.go
```diff
@@ -71,109 +87,114 @@ func (e *Executor) watchTasks(calls ...*ast.Call) error {
...
+						files, err := fingerprint.Globs(t.Dir, t.Sources)
+						if err != nil {
+							e.Logger.Errf(logger.Red, "%v\n", err)
+							return
+						}
+						relPath, _ := filepath.Rel(e.Dir, event.Name)
+						if !slices.Contains(files, relPath) {
+							e.Logger.VerboseErrf(logger.Magenta, "task: skipped for file not in sources: %s\n", relPath)
+							return
+						}
```
pd93, line 104: What's the expected behaviour when running this with the `--dir` flag? The result seems inconsistent.

Repro (run task commands from project root):

1. Create `./testdata/watcher_interval/src` folder in Task repo
2. `task -wvd ./testdata/watcher_interval`
3. Create file in `./testdata/watcher_interval/src/foo.txt`
    ```
    task: received watch event: WRITE         ".../testdata/watcher_interval/src/foo.txt"
    task: skipped for file not in sources: src/foo.txt
    ```
4. Create `./src` folder in Task repo (keep existing files)
5. Create file in `./src/foo.txt`
6. `task -wvd ./testdata/watcher_interval`
7. Update `./testdata/watcher_interval/src/foo.txt`
    ```
    task: received watch event: CREATE        "/home/pete/dev/github.com/go-task/task/testdata/watcher_interval/src/foo.txt" ← "/home/pete/dev/github.com/go-task/task/testdata/watcher_interval/src/arst"
    task: "default" started
    task: [default] echo "Hello, World!"
    Hello, World!
    task: "default" finished
    task: task "default" finished running
    ```
  
Basically, it seems as though listening to `/testdata/watcher_interval/src` only works when a corresponding file exists in `/src`. Guessing something is off with the paths.

Should running `-w` with `-d` listen to the paths at the given directory or the caller's working directory?

watch.go
pd93, line 1: I'm having issues where if a directory is empty when I start the watcher and then create files, nothing will be detected. If I create any file before starting the watcher, then other new files are detected.

I'm also not able to get deletes to work (on Linux). When I remove a file, I just get something like this:

```
task: received watch event: WRITE         ".../testdata/watcher_interval/src/foo.txt"
task: skipped for file not in sources: src/foo.txt
```

But it does not start the task.

radcool: > I'm having issues where if a directory is empty when I start the watcher and then create files, nothing will be detected. If I create any file before starting the watcher, then other new files are detected.

I've observed the same thing.

watch.go
andreynering: Thanks for testing this scenario and reporting! This issue is now fixed.

watch.go
andreynering: I'm still going to work on this issue another day.

pd93 (approved):
This is looking good now! (At least on Linux). I can't see any unexpected behaviour and the previously mentioned issues seem to be fixed. Performance also seems much better than before.

I can't speak to the remaining points around ulimits on MacOS (I don't own any Apple hardware), but it seems like regardless of that issue, this is a big step forwards and those things could be addressed separately if required. I think for most users, the default ulimit will be enough and `CGO` seems like a line we can't cross in Task anyway.

andreynering:
I'm merging this as this is an significant improvement on it already. We can still do further improvements with time. 🙂

PR #2049, opened 2025-02-08
feat: experiments logging improvements

### LINKED ISSUE #1979: Exit with error when an experiment is specified

pd93:
Closes #1979

Refactors the experiments package slightly so that experiments only need to be specified in one place now. Also moved the print functions to the logger to avoid importing anything in the experiments package. This stops circular imports.

This way of doing things allows us to easily iterate over all experiments which means I was able to easily add a function that validates all experiments. If an experiment is inactive (has no enabled values) and a user tries to specify a value, a warning will now be output.

Finally, I've consolidated how we read any Task-namespaced environment variables into the new `env.GetTaskVar` function that will always prefix `TASK_` to any variables read. The logger now only reads these variables once instead of every time a call to print is made.

### DIFF
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -45,6 +45,6 @@ func isTypeAllowed(v any) bool {
 	}
 }
 
-func GetTaskVar(key string) string {
+func GetTaskEnv(key string) string {
 	return os.Getenv(taskVarPrefix + key)
 }
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -25,7 +25,7 @@ func GetFromVars(env *ast.Vars) []string {
 		if !isTypeAllowed(v) {
 			continue
 		}
-		if !experiments.EnvPrecedence.Enabled {
+		if !experiments.EnvPrecedence.Enabled() {
 			if _, alreadySet := os.LookupEnv(k); alreadySet {
 				continue
 			}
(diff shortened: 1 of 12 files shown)


internal/env/env.go
```diff
@@ -42,3 +44,7 @@ func isTypeAllowed(v any) bool {
 		return false
 	}
 }
+
+func GetTaskVar(key string) string {
```
andreynering, line 48: This could be called `GetTaskEnv`?

andreynering (approved):
👏

pd93:
@andreynering Have added your suggestion and also added some tests for the experiments and did a little bit more refactoring. We'll now get warnings in two cases:

1. A user attempted to enable an experiment with no valid values

   ```
   ❯ TASK_X_ANY_VARIABLES=1 taskd
   task: Experiment "ANY_VARIABLES" is inactive and cannot be enabled
   ```

2. A user attempted to enable an experiment with a value that is not in the allowed values

   ```
   ❯ TASK_X_MAP_VARIABLES=99 taskd
   task: Experiment "MAP_VARIABLES" has an invalid value "99" (allowed values: 1, 2)
   ```

PR #2050, opened 2025-02-08
feat: stop task test installing task

pd93:
@andreynering Maybe I'm missing something. This has been here since long before I came across Task. Is there a reason that `task test` works the way it does? In most of my projects, I just have `task test` setup to run `go test ./...`, but in our own Taskfile it will install the binary first. This isn't usually what I want to do during development.

Just opening this as a suggestion. Thanks :)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -98,21 +98,17 @@ tasks:
   test:
     desc: Runs test suite
     aliases: [t]
-    deps: [install]
+    sources:
+      - "**/*.go"
+      - "testdata/**/*"
     cmds:
-      - go test {{catLines .GO_PACKAGES}}
-    vars:
-      GO_PACKAGES:
-        sh: go list ./...
+      - go test ./...
 
   test:all:
     desc: Runs test suite with signals and watch tests included
-    deps: [install, sleepit:build]
+    deps: [sleepit:build]
     cmds:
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -98,21 +98,16 @@ tasks:
   test:
     desc: Runs test suite
     aliases: [t]
-    deps: [install]
+    sources:
+      - "**.*.go"
```
andreynering, line 102: Looks like a typo, no?

```suggestion
      - "**/*.go"
```

Taskfile.yml
```diff
@@ -98,21 +98,16 @@ tasks:
...
     cmds:
-      - go test {{catLines .GO_PACKAGES}}
-    vars:
-      GO_PACKAGES:
-        sh: go list ./...
+      - go test ./...
 
   test:all:
     desc: Runs test suite with signals and watch tests included
-    deps: [install, sleepit:build]
```
andreynering, line 110: In this case, we really need `sleepit:build` as a dependency, but it is a binary that we actually call inside the test.

andreynering (approved):
This was as is for a long time. I'm fine with the change, and thanks for doing it!

Taskfile.yml
pd93: Yep, thanks 😅

PR #2052, opened 2025-02-09
feat: display allowed values when vars are not provided

### LINKED ISSUE #2027: Show allowed values when variable with set enum is missing
### Description

With a task like the following:

```yaml
version: '3'
tasks:
  deploy:
    cmds:
      - echo "deploying to {{.ENV}}"
    requires:
      vars:
        - name: ENV
          enum: [dev, beta, prod]
```

The following error will be shown when the required variable is missing:

`task: Task "deploy" cancelled because it is missing required variables: ENV`

When an `enum` is set like here, it would be great if the allowed values would be shown in the error. This way the user doesn't have to look into the Taskfile to see what values are allowed in situations where this isn't obvious.

vmaerten:
- Fixes #2027 

Output : 
image


errors/errors_task.go
```diff
@@ -142,17 +142,33 @@ func (err *TaskCancelledNoTerminalError) Code() int {
 }
 
 // TaskMissingRequiredVars is returned when a task is missing required variables.
+
+type MissingVar struct {
+	Name          string
+	AllowedValues []string
+}
 type TaskMissingRequiredVars struct {
```
pd93, line 150: Can we add the `Error` suffix to this type as per the [go naming conventions for errors]( https://go.dev/wiki/Errors#naming). Same with `TaskNotAllowedVars` below.

vmaerten: Yeah good catch
Only for `TaskMissingRequiredVars` right ? MissingVar is not the error but the underlying details

pd93: Yeah exactly :) Thanks!

pd93 (approved):
LGTM

Issue #1322, opened 2023-09-01
`--list-all` spends time globbing for sources.


StefanBRas:
`task --list-all` is computing something even though it seems like it shouldn't to show the list of tasks.

It's an problem because autocompletion relies on `--list-all` so right now it's not usable for me.

It takes ~2 seconds (user time, 12 system) to run `--list-all`.

Not directly related, but it seems to be some interaction with globbing for sources and using `for` in commands that makes it take a long time to run in general.

Given this task file
```yml
version: '3'
tasks:
  default:
    sources:
    - dev/*.json
    cmds:
      - for: sources
        cmd: echo {{ dir .ITEM }}
```
It takes again 2 seconds ish, but
```yml
version: '3'
tasks:
  default:
    sources:
    - "dev/*.json"
    cmds: 
    - "echo hey"
```
is instant.

It's difficult to recreate, because it's only for some folders it takes a long time. I think it's related to #225.

- Task version:  v3.28.0 (h1:PGYGwevlGQdYrqhO6lLCYylC7YuGoQLlVwHkO42gf0I=)
- Operating system: macOS ventura 13.5
- Experiments enabled: No

krystian-panek-wttech:
also `--dry` spends time evaluating dynamic variables; I am not sure if it is always desired; it's more irritating to me than helping

mishamsk:
Same issue here. Started noticing recently. Happens both for `task --list` and `task --list-all`. `Task version: 3.38.0 ()`

Checked usage of `for`, doesn't seem like a culprit. I have two checkouts of the same repo with a Taskfile. autocomplete works reasonably fast in one folder, but not the other. The only difference that I've see is the size of the folders (e.g. one has node_modules populated and the other don't). I do have tasks that defines `sources` keys, I also have tasks with preconditions that run `find`.

I haven't yet triaged the issue to narrow it down to a specific task. I can, assuming anyone from maintainers have capacity to look into this.

Also, I may be missing something, but it looks like list output contains only task names and descriptions. These are all static, so theoretically there is no need to execute anything at all. Just parse yaml. But I may not thinking about some feature of Taskfile of course.

In general - awesome tool, great work and kudos to the team!

PR #2053, opened 2025-02-09
refactor: VeryFastCompile for Task list

vmaerten:
I  was looking at #1322 when I realized that we compile every task (even with Fast Variable) just for `--list` or `--list-all`.
`--list-all` is used for autocomplete, so it should be as fast as possible.

I created a new method to compile specifically for this listing. This way, we do not add complexity to the current one.
I’m aware that it could also be used for the `Editor`, but it seems fine.


notnmeyer:
this would be great to see. we added `**/*.ts` to sources on a task and its impact on `task -l` and especially tab completion was pretty brutal.

PR #2059, opened 2025-02-11
chore: bump minimum version to 1.23

pd93:
Bumps the minimum Go version to v1.23. This allows us to start using 1.23 features in the next version of Task! Tests will now run on 1.23 and 1.24.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.60.1
+          version: v1.64.2
 
   lint-jsonschema:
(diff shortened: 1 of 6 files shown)


pd93:
Waiting for https://github.com/golangci/golangci-lint/pull/5224 for the lint checks to pass.

Also seems that they have [changed the way the versioning information works](https://tip.golang.org/doc/go1.24#tools):

> The go build command now sets the [main module’s version](https://tip.golang.org/pkg/runtime/debug#BuildInfo.Main) in the compiled binary based on the version control system tag and/or commit. A +dirty suffix will be appended if there are uncommitted changes. Use the -buildvcs=false flag to omit version control information from the binary.

This actually appears to have broken one of our tests, but should be simple to fix.

PR #2062, opened 2025-02-13
fix: add defer's `silent` property to schema

### LINKED ISSUE #2061: YAML schema does not recognize `silent` attribute for `defer` blocks
### Description

I use the [Taskfile YAML schema with VS Code](https://taskfile.dev/integrations/#visual-studio-code) and found that it incorrectly reports `silent` as an invalid parameter for `defer` items (screenshot below).  This is definitely [valid behavior that was recently implemented](https://github.com/go-task/task/pull/1879/files).

Image

### Version

3.41.0

### Operating system

Ubuntu 22.04.4

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'

tasks:
  stoaferp-defer-things:
    cmds:
      - defer: echo "this will always execute"
        silent: true
      - exit 1
      - echo "this will not execute"
```

StoaferP:
This smol fix brings the YAML schema in line with the feature implemented in #1879

Fixes #2061

### DIFF
--- a/website/static/schema.json
+++ b/website/static/schema.json
@@ -363,6 +363,10 @@
               "$ref": "#/definitions/task_call"
             }
           ]
+        },
+        "silent": {
+          "description": "Hides task name and command from output. The command's output will still be redirected to `STDOUT` and `STDERR`.",
+          "type": "boolean"
         }
       },
       "additionalProperties": false,


pd93:
Thanks for opening a PR for this. This was completed in #2068 in a slightly different way so I'm going to close this one.

Issue #2065, opened 2025-02-17
Allow to use vars in matrix


DjinNO:
### Description

It is not possible to use variables in a matrix task now

```
version: '3'

vars:
  OS_VAR: ["windows", "linux", "darwin"]
  ARCH_VAR: ["amd64", "arm64"]

tasks:
  default:
    silent: true
    cmds:
      - for:
          matrix:
            OS: {{.OS_VAR}}
            ARCH: {{.ARCH_VAR}}
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
```

```
err:  invalid keys in command
file: /Users/djinn/GIT/gitlab/gitops/installations/generator/Taskfile.yml:12:9
  10 |     silent: true
  11 |     cmds:
> 12 |       - for:
     |         ^
  13 |           matrix:
  14 |             OS: {{.OS_VAR}}
```

Correct me if I'm wrong

pd93:
@DjinNO You are correct that it is currently not possible to pass variables into the matrix. You would need to define the variables statically:

```yaml
version: '3'

tasks:
  default:
    silent: true
    cmds:
      - for:
          matrix:
            OS: ["windows", "linux", "darwin"]
            ARCH: ["amd64", "arm64"]
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
```

The error message you're getting is actually a little misleading and I've opened #2068 to address this. With the changes in that PR, you should see something like `err:  cannot unmarshal !!map into []interface {}`. This indicates that the values in a matrix need to be arrays and hence you wouldn't be able to using templating syntax to do what you want (templates always output strings).

If we support this going forwards, I imagine it would look something like this instead:

```yaml
version: '3'

vars:
  OS_VAR: ["windows", "linux", "darwin"]
  ARCH_VAR: ["amd64", "arm64"]

tasks:
  default:
    silent: true
    cmds:
      - for:
          matrix:
            OS:
              ref: OS_VAR
            ARCH:
              ref: ARCH_VAR
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
```

However, this won't work currently.

illia-weisser:
```
vars:
  ENVIRONMENTS_YML:
    sh: yq e '.environments' environments.yaml
  ENVIRONMENTS:
    ref: "keys ((fromYaml .ENVIRONMENTS_YML))"
  SERVICES:
[...]

PR #2068, opened 2025-02-17
feat: decoding improvements

pd93:
These changes came from investigating #2065. This PR does not fix that issue (I will do that in a follow up PR), but it did uncover a few problems which I've addressed here:

- Refactored how snippets work
  - Moved from `errors` package to `taskfile` package (now part of the package API)
  - Functional options
  - Support for out-of-bounds lines/columns
  - Ability to hide the line/column indicators
  - A bunch of tests
- Added a debug function for `TaskfileDecodeError`. This was just helpful to track down the next issue
- Fixed the `ast.Cmd` Unmarshaler so that errors are correctly propagated to the user.
  - This is what is causing the unhelpful error in #2065.
  - It should now return something like: `err:  cannot unmarshal !!map into []interface {}` which much more accurately describes the problem.
- Fixed the Taskfile schema for defer statements when used in the shorthand commands array syntax


vmaerten (approved):
Way cleaner
I like the fact you splitted `defer` into another file!

PR #2075, opened 2025-02-19
fix: sources brace expansion

### LINKED ISSUE #2073: Using glob curly braces expansion in `sources` prevents change detection for some files
### Description

Hello,

First thanks a lot for all your work on this project, I use it often for personal and professional contexts and saves me a lot of trouble 👍

Using :
```yaml
sources:
  - src/{api,ingest-worker}/**/*
```
triggers correctly the job if I change source files in `src/api` folder but never trigger the job if I change files inside `src/ingest-worker`.
Instead the following :

```yaml
sources:
  - src/api/**/*
  - src/ingest-worker/**/*
```
triggers the job if I change files inside both `src/api` and `src/ingest-worker` folders.

To be sure, I checked my syntax with zglob and seems to be OK :
```bash
❯ /bin/find dist/ src/
dist/
dist/ingest-worker
dist/ingest-worker/ingest-worker.out
dist/api
dist/api/api.out
src/
src/ingest-worker
src/ingest-worker/ingest-worker.source
src/foo
src/foo/foo.source
src/api
src/api/api.source

❯ ./zglob -d 'src/{api,ingest-worker}/**/*'
src/ingest-worker/ingest-worker.source
src/api/api.source
```

Maybe I misunderstood something but both methods should not match same files here ?

### Version

Task version: v3.41.0 (h1:giUddhe0XZLbEWIQ/MuTPipR9ek+teulIA5xf/2IHXg=)

### Operating system

Linux

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'

tasks:
  build:
    cmds:
      - mkdir -p dist/{api,ingest-worker}
      - cp src/api/api.source dist/api/api.out
      - cp src/ingest-worker/ingest-worker.source dist/ingest-worker/ingest-worker.out
    sources:
      - src/{api,ingest-worker}/**/*
```

pd93:
Fixes #2073 

Removes `zglob` as a dependency and instead relies on mvdan's `expand.Fields` with globbing turned on for `sources` globs.

We also use `expand.Fields` in multiple other places to resolve things like `~` and environment variables. However, most of the time, we don't also want brace expansion or globbing, so I have separated this into two functions. One wraps `expand.Literal` which returns a single string with just the variables resolved, and the other wraps `expand.Fields` with braces and globbing enabled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/go-task/slim-sprig/v3 v3.0.0
 	github.com/go-task/template v0.1.0
 	github.com/joho/godotenv v1.5.1
-	github.com/mattn/go-zglob v0.0.6
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/otiai10/copy v1.14.1
 	github.com/puzpuzpuz/xsync/v3 v3.5.1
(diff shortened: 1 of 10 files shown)


andreynering (approved):
👏

PR #2082, opened 2025-02-20
feat: remove logger from taskfile package

pd93:
Our package API should not rely on our internal `logger` package. Logging is a CLI-specific responsibility and users that want to use Task as a package may want to implement their own logging. As such, this PR removes the logger from the `taskfile` package entirely.

Normally this would mean we lose the ability to insert debug logs into our code. However, we can pass a generic function into the `Reader` which will process logs instead. This allows all of Task's current logging to remain in place while a 3rd party user can either omit the logging function when creating their reader (and discard the logs) or implement their own custom logging.

I have also done the same thing for the prompt function so that package API users can implement their own prompt functionality.

Finally, the reader now uses functional options in its constructor. This allows us to set sensible defaults instead of the caller having to specify everything each time (including the new logging functions).

### DIFF
--- a/taskfile/reader.go
+++ b/taskfile/reader.go
@@ -27,13 +27,18 @@ Continue?`
 Continue?`
 )
 
-// A Reader will recursively read Taskfiles from a given source using a directed
-// acyclic graph (DAG).
 type (
-	ReaderOption     func(*Reader)
-	ReaderDebugFunc  func(string)
+	// ReaderDebugFunc is a function that is called when the reader wants to
+	// log debug messages
+	ReaderDebugFunc func(string)
+	// ReaderPromptFunc is a function that is called when the reader wants to
+	// prompt the user in some way
 	ReaderPromptFunc func(string) error
-	Reader           struct {
+	// ReaderOption is a function that configures a Reader.
+	ReaderOption func(*Reader)
+	// A Reader will recursively read Taskfiles from a given source using a directed
+	// acyclic graph (DAG).
+	Reader struct {
 		graph       *ast.TaskfileGraph
 		node        Node
 		insecure    bool
@@ -47,6 +52,7 @@ type (
 	}
 )
 
+// NewReader constructs a new Taskfile Reader using the given Node and options.
 func NewReader(
 	node Node,
 	opts ...ReaderOption,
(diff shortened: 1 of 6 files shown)


taskfile/reader.go
```diff
@@ -28,40 +27,115 @@ Continue?`
...
 		promptMutex: sync.Mutex{},
 	}
+	for _, opt := range opts {
+		opt(reader)
+	}
+	return reader
+}
+
+// WithInsecure enables insecure connections when reading remote taskfiles. By
+// default, secure connections are rejected.
```
vmaerten, line 79: I think there is a typo here
I think you want to write either : 
`by default, only secure connections are allowed` or `by default, insecure connections are rejected`

taskfile/taskfile.go
```diff
@@ -117,7 +115,6 @@ func Exists(l *logger.Logger, path string) (string, error) {
 	for _, taskfile := range defaultTaskfiles {
 		alt := filepathext.SmartJoin(path, taskfile)
 		if _, err := os.Stat(alt); err == nil {
-			l.VerboseOutf(logger.Magenta, "task: [%s] Not found - Using alternative (%s)\n", path, taskfile)
```
vmaerten, line 120: I’m OK with removing it because, even though we could use it for troubleshooting, I don’t think it’s used at all. I just wanted to make sure it wasn’t an oversight.

taskfile/taskfile.go
```diff
@@ -89,7 +88,6 @@ func RemoteExists(ctx context.Context, l *logger.Logger, u *url.URL, timeout tim
 
 		// If the request was successful, return the URL
 		if resp.StatusCode == http.StatusOK {
-			l.VerboseOutf(logger.Magenta, "task: [%s] Not found - Using alternative (%s)\n", alt.String(), taskfile)
```
vmaerten, line 92: I’m OK with removing it because, even though we could use it for troubleshooting, I don’t think it’s used at all. I just wanted to make sure it wasn’t an oversight.

taskfile/taskfile.go
pd93: Yeah, its difficult to log here without passing a function in. If we really want the logs, we could compare the input/output in the caller and log there instead, but I decided it wasn't worth it.

taskfile/reader.go
pd93: Lol whoops. Good spot!

taskfile/taskfile.go
vmaerten: I agree!

vmaerten (approved):
:heart:

PR #2085, opened 2025-02-22
refactor: executor functional options

pd93:
This PR migrates the `Executor` to use the functional options pattern. This follows other similar changes to `taskfile.Reader` and `taskfile.Snippet`. This allows package API users to easily construct an `Executor` with sensible defaults and only supply a list of things they want to override.

Our internal CLI implementation has a special functional option in `internal/flags` called `WithFlags` which will gather all the CLI flags and pass them in as functional options.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -16,9 +16,10 @@ import (
 )
 
 type (
-	// An ExecutorOption is a functional option for an Executor
+	// An ExecutorOption is a functional option for an [Executor].
 	ExecutorOption func(*Executor)
-	// An Executor is a type that is used for processing and executing Taskfiles
+	// An Executor is used for processing Taskfile(s) and executing the task(s)
+	// within them.
 	Executor struct {
 		// Flags
 		Dir         string
@@ -71,6 +72,8 @@ type (
 	}
 )
 
+// NewExecutor creates a new [Executor] and applies the given functional options
+// to it.
 func NewExecutor(opts ...ExecutorOption) *Executor {
 	e := &Executor{
 		Timeout:              time.Second * 10,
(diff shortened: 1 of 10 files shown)


vmaerten (approved):
Good refactor as always!

executor.go
```diff
@@ -0,0 +1,253 @@
...
+		executionHashes      map[string]context.Context
+		executionHashesMutex sync.Mutex
+	}
+	TempDir struct {
+		Remote      string
+		Fingerprint string
+	}
+)
+
+func NewExecutor(opts ...ExecutorOption) *Executor {
```
andreynering, line 75: Each of the functions in this file should have a documentation comment, do make them look pretty at https://pkg.go.dev/github.com/go-task/task/v3.

executor.go
```diff
@@ -0,0 +1,253 @@
...
+	return func(e *Executor) {
+		e.Stdout = stdout
+	}
+}
+
+func WithStderr(stderr io.Writer) ExecutorOption {
+	return func(e *Executor) {
+		e.Stderr = stderr
+	}
+}
```
andreynering, line 247: In theory we could have an additional function that accepts all three, for shortness. Optional suggestion, though.

```go
func WithIO(stdin io.Reader, stdout, stderr io.Writer) ExecutorOption {
	return func(e *Executor) {
		e.Stdin = stdin
		e.Stdout = stdout
		e.Stderr = stderr
	}
}
```

andreynering (approved):
Great work!

Claps

executor.go
pd93: To me, it makes more sense to keep these separate if you're passing different values into each of them. Sometimes you might only want to pass in 1 or 2 of them and leave the others blank. However, I think if you're setting them all to the same value, we could have something like this:

```go
func WithIO(rw io.ReadWriter) ExecutorOption {
	return func(e *Executor) {
		e.Stdin = rw
		e.Stdout = rw
		e.Stderr = rw
	}
}
```

This would be super handy in the new tests I've been writing where we want to write everything to a single buffer.

pd93:
Couple of minor changes:

- Added the missing `ExecutorWithVersionCheck` functional option in `cmd/task`
- Added prefixes to the functional options like in `taskfile.Snippet`
  - This helps with any potential naming conflicts in the future.
- A bunch of docstrings for everything.

PR #2086, opened 2025-02-23
docs: getting started

pd93:
As mentioned in https://github.com/go-task/task/pull/2011#pullrequestreview-2573000290, the getting started docs could do with a bit of an overhaul. The usage page has slowly grown and become a bit daunting to a new user. It also doesn't mention `--init` in any capacity.

This PR solves both of these by adding a new Getting Started page to the website. This page is written more as a tutorial than a reference which is friendlier to new users. It instructs the user how to create a new Taskfile using `--init`, how to call it and how to write a basic build task.

### DIFF
--- a/website/docs/community.mdx
+++ b/website/docs/community.mdx
@@ -1,6 +1,6 @@
 ---
 slug: /community/
-sidebar_position: 9
+sidebar_position: 10
 ---
 
 # Community
--- a/website/docs/contributing.mdx
+++ b/website/docs/contributing.mdx
@@ -1,6 +1,6 @@
 ---
 slug: /contributing/
-sidebar_position: 11
+sidebar_position: 12
 ---
 
 # Contributing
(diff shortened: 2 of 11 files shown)


vmaerten (approved):

andreynering (approved):
That's a great idea!

In theory we could consider adding this page to the top bar, but I don't want to pollute it with too many items.

Screenshot 2025-02-23 at 14 06 45

---

I propose that we at least add a mention on the home, near where we already link the usage page.

Screenshot 2025-02-23 at 14 09 00

pd93:
> That's a great idea!
> 
> In theory we could consider adding this page to the top bar, but I don't want to pollute it with too many items.

Yeah, I think that most users will find it from the sidebar once they're in the docs section anyway.

> I propose that we at least add a mention on the home, near where we already link the usage page.

I hadn't looked at the home page at all, but I agree. I would actually suggest replacing the example Taskfile we have there as this sort of duplicates some of the getting started. Something like this (pushed):

image

andreynering (approved):
Looks good!

PR #2092, opened 2025-02-26
fix: include with dynamic vars

vmaerten:
A regression was introduced in this [PR](https://github.com/go-task/task/pull/1797).

When a dynamic variable is provided, even if it is not used, all other variables become unavailable in the templating system.


andreynering (approved):

pd93 (approved):
Thanks for catching this! ❤️

PR #2094, opened 2025-02-28
Relocate CLI special variable handling

### LINKED ISSUE #2090: Global vars can't use .CLI_ARGS special var as default value
### Description

I wanted to use a global `vars:` key to default to the .CLI_ARGS special var.

```yaml
vars:
  ENV: '{{.ENV | default .CLI_ARGS}}'
```

And then use it in a task, so both of these examples work:

```shell
task debug -- dev  # Should use 'dev' as value for `ENV`
task debug ENV=dev # Should behave similarly as `-- dev`
```

The above behavior works if `vars:` is used directly in the task definition:

```yaml
tasks:
  debug:
    vars:
      ENV: '{{.ENV | default .CLI_ARGS}}'
    cmds:
      - echo '{{.ENV}}'

```

Only when using `vars:` globally `.CLI_ARGS` comes empty and doesn't work.

### Version

v3.41.0 (h1:giUddhe0XZLbEWIQ/MuTPipR9ek+teulIA5xf/2IHXg=)

### Operating system

macOS Sequioa 15.3.1 (24D70)

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'

vars:
  ENV: '{{.ENV | default .CLI_ARGS}}'

tasks:
  debug:
    cmds:
      - echo '{{.ENV}}'
```

trulede:
fixes #2090

This relocates the CLI special vars, with an 
indirection via flags for CLI_ARGS which is done to preserve the associated error handling; since it could not be located in the flags package init() function.

I'm not certain if the handling of Bools is correct with respect  to the overall mechanism (i.e. converting to a string).

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -6,12 +6,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/go-task/task/v3/internal/env"
 	"github.com/go-task/task/v3/internal/execext"
 	"github.com/go-task/task/v3/internal/filepathext"
+	"github.com/go-task/task/v3/internal/flags"
 	"github.com/go-task/task/v3/internal/logger"
 	"github.com/go-task/task/v3/internal/templater"
 	"github.com/go-task/task/v3/internal/version"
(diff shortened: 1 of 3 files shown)


compiler.go
```diff
@@ -203,6 +205,11 @@ func (c *Compiler) getSpecialVars(t *ast.Task, call *Call) (map[string]string, e
 		"ROOT_DIR":         c.Dir,
 		"USER_WORKING_DIR": c.UserWorkingDir,
 		"TASK_VERSION":     version.GetVersion(),
+		"CLI_ARGS":         flags.CliArgs,
+		"CLI_FORCE":        strconv.FormatBool(flags.Force || flags.ForceAll),
+		"CLI_SILENT":       strconv.FormatBool(flags.Silent),
+		"CLI_VERBOSE":      strconv.FormatBool(flags.Verbose),
+		"CLI_OFFLINE":      strconv.FormatBool(flags.Offline),
```
pd93, line 212: We should not be using the `flags` package directly in the main `task` package. The CLI must be decoupled from the `Executor` so that they can be used independently.

These variables should also not be converted to strings. They should stay as booleans or we will break people's Taskfiles.

trulede:
OK.

PR #2098, opened 2025-03-04
Improve handling of TaskfileInvalidError.

### LINKED ISSUE #2067: Improve error message if Taskfile provided via stdin is invalid. Also, skip if empty?
### Description

I just migrated a pre-push git hook to Taskfile.

```
➤ cat .git/hooks/pre-push
#!/bin/sh
set -e

task test-all
```

This fails with a mysterious

```
➤ git push
task: Failed to parse __stdin__:
yaml: line 1: cannot unmarshal !!str into taskfile
error: failed to push some refs to 'github.com:gregtatum/indie-web.git'
```

Running the following works:

```
task test-all
```

I realized that the hook was piping stdin in to the command. I worked around it with:

```
➤ cat .git/hooks/pre-push
#!/bin/sh
set -e

cat > /dev/null | task test-all
```

However, it would be nice to not have a cryptic error when stdin is being written to.

### Version

Task version: v3.35.1 (h1:zjQ3tLv+LIStDDTzOQx8F97NE/8FSTanjZuwgy/hwro=)

### Operating system

macOS

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

trulede:
Change TaskfileInvalidError error formatting to match the prevent error message style. 
Change name of StdinNode.

$ go run ./cmd/task -t help.go
task: Taskfile "help.go" is not valid (failed parsing). yaml: line 32: mapping values are not allowed in this context exit status 109

fixes #2067

### DIFF
--- a/errors/errors_taskfile.go
+++ b/errors/errors_taskfile.go
@@ -47,7 +47,7 @@ type TaskfileInvalidError struct {
 }
 
 func (err TaskfileInvalidError) Error() string {
-	return fmt.Sprintf("task: Failed to parse %s:\n%v", err.URI, err.Err)
+	return fmt.Sprintf("task: Taskfile %q is not valid (failed parsing).\n%v", err.URI, err.Err)
 }
 
 func (err TaskfileInvalidError) Code() int {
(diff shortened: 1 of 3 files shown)


trulede:
These are UI issues.

Adding underscores does nothing to differentiate `stdin` from a filepath, it just looks clumsy.
The format of the message was not consistent with other messages in the same file, so it's reasonable to update it.

PR #2103, opened 2025-03-07
Added TASK_WORKING_DIR variable

### LINKED ISSUE #2102: Add special variable for task working directory
### Description

There are various [special variables](https://taskfile.dev/next/reference/templating/#special-variables) available but none of them reflects the working directory of a respective task. The ones that come closest are: 
- `USER_WORKING_DIR` which contains the directory where _task_ has been called from, but does not honor the `-d <dir>` command line parameter
and
- `TASK_DIR` which merely contains what has been configured under `dir:` in a task of the `Taskfile.yml` file and does not even resolve template expressions.

I see two different solutions to this problem: 

1. ~~Either change the behavior of `USER_WORKING_DIR` so that it takes the `-d` parameter into account, because I see only very little benefit in having a variable which contains the directory where _task_ has been called from when a different working directory has been explicitly set with `-d`~~
2. Introduce a new variable such as `TASK_WORKING_DIR` which contains the desired dir and does not break the current behavior of `USER_WORKING_DIR`.

The current use case which I have is as follows: I have a root directory containing a `Taskfile.yml` file and some subdirectories which doesnt have `Taskfiles.yml` files. I now want to reuse the logic of the `Taskfile.yml` from the root directory in the subdirectories as described [here](https://taskfile.dev/usage/#running-a-taskfile-from-a-subdirectory). This works well if I navigate to a subdirectory and call the respective task from there, however, when I call _task_ from the root directory and specify the subdirectory with the `-d` parameter it doesnt work as `USER_WORKING_DIR` still contains the root directory as outlined above.

jaynis:
This PR adds the `TASK_WORKING_DIR` special variable which contains the absolute path of the directory where _task_ runs on. This is essentially the `TASK_USER_DIR` variable additionally taking into account the `-d` command line parameter.
The documentation of the similar, yet different `TASK_DIR` variable has been updated as well to avoid confusion.

fixes #2102


andreynering:
Closing in favor of #2186. See https://github.com/go-task/task/issues/2102#issuecomment-2816433405.

trulede:
See #2703 for an alternative solution which addresses what is probably the root cause of this issue.

PR #2105, opened 2025-03-09
build: honor version set by ldflag

chenrui333:
Seeing some version report issue in https://github.com/Homebrew/homebrew-core/pull/210289, update to honor the ldflag set.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -11,13 +11,14 @@ var (
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pd93:
Thanks for opening this @chenrui333. We've gone with a [slightly different solution](https://github.com/go-task/task/pull/2131) that should be a bit more robust for everyone.

PR #2112, opened 2025-03-11
refactor: embed the default Taskfile instead of defining it in code

pd93:
Moves the default Taskfile used when calling `--init` to an embed rather than storing it in the code.


andreynering (approved):
Good idea

PR #2113, opened 2025-03-11
feat: process variables in include vars

### LINKED ISSUE #2108: Variables from included Taskfiles cannot be overwritten
### Description

The [documentation](https://taskfile.dev/usage/#namespace-aliases) states that in order to overwrite a variable from an included Taskfile, one has to use the syntax
```
MY_VAR: '{{.MY_VAR | default "my-default-value"}}'
```

That does not work, at least not for 'global' variables defined in the top-level `vars` field of the Taskfile.

#### Taskfile.yaml
```yaml
version: 3

vars:
  FOOBAR: foo

includes:
  bar: Taskfile_bar.yaml

tasks:
  foobar:
    cmds:
    - cmd: echo '{{.FOOBAR}}'
      silent: true
```

#### Taskfile_bar.yaml
```yaml
version: 3

vars:
  FOOBAR: '{{.FOOBAR | default "bar"}}'

tasks:
  foobar:
    cmds:
    - cmd: echo '{{.FOOBAR}}'
      silent: true
```

#### The Problem

Running either `task foobar` or `task bar:foobar` results in `bar`, but both should print `foo` instead.

---

I have to admit that I find the behavior weird in general, I would assume that in most use-cases, variables from the imported Taskfile (often generic/library) should be overwritten by the ones from the importing Taskfile (specialized for project). Having it work the other way around is somewhat confusing. I can work with that though, but if there is no way to overwrite the variables in the included Taskfile from the including one, that is a problem.

---

### Version

v3.41.0

### Operating system

darwin/arm64

### Experiments Enabled

_None_

pd93:
Fixes #2108

Including a Taskfile with a static variable works:

```yaml
version: 3

includes:
  bar:
    taskfile: included.yaml
    vars:
      FOOBAR: foo
```

However, if you want to use an existing variable when including a Taskfile:

```yaml
version: 3

vars:
  FOOBAR: foo

includes:
  bar:
    taskfile: included.yaml
    vars:
      FOOBAR: "{{.FOOBAR}}"
```

you're out of luck, since Task doesn't resolve variables in the include `vars` section. This PR adds that functionality.


vmaerten (approved):
Pretty straitforward

andreynering (approved):

PR #2114, opened 2025-03-11
feat: execution trace gantt chart

viktorasm:
When debugging highly parallelized task executions, it's not immediately obvious where the bottlenecks are.

This PR allows dumping timings of the execution as Gantt chart (Mermaid syntax). Here's a sample of of a real-word invocation:

image

Having this makes it so much easier to understand, how well things parallelize, are tasks unnecessarily blocked, what are the major bottlenecks, etc.

Usage:
* if user provides `--execution-trace-output` flag, then `task` invocation writes contents of the gantt chart to it.


internal/tracing/tracing.go
```diff
@@ -0,0 +1,83 @@
+package tracing
```
trulede, line 1: Here perhaps an interface, and then implementation for the Mermaid tracer.

```
type Tracer interface {
    Start(task)
    Stop()
    ...
}
```

internal/flags/flags.go
```diff
@@ -221,5 +225,9 @@ func WithExecutorOptions() task.ExecutorOption {
 			task.ExecutorWithTaskSorter(sorter),
 			task.ExecutorWithVersionCheck(true),
 		)
+
```
trulede, line 228: I think here, the same pattern as other executors should be used (above calls).

internal/flags/flags.go
```diff
@@ -73,6 +73,8 @@ var (
 	Offline     bool
 	ClearCache  bool
 	Timeout     time.Duration
+
+	ExecutionTraceOutput string
```
trulede, line 77: Might it be possible to consider a more generic operation:

```
# Select one tracer or the other, with default options.
task default --tracer=mermaid
task default --tracer=csv

# Select a tracer and also provide some options, without adding additional CLI parameters.
task default --tracer="mermaid;filename=foo.out"
```

trulede (commented):
This could be interesting a general technique. 

For instance, a tracer could dump each tasks variables as tasks execute, making it _much_ easier to debug Taskfiles.

trulede:
PR #2617 introduces some logging features which could be a more generalised alternative to what this PR is doing. Then it would be a matter of processing the log into the format you need with a simple script.

PR #2115, opened 2025-03-11
docs: specify that install command parameters are order specific

iainvm:
Found out that the install script parameters seem to be order specific so worth documenting it

e.g.
Will ignore the `-b ~/temp` and install `v3.30.0` into `./bin`
```shell
sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d v3.30.0 -b ~/temp
```


vmaerten (approved):

vmaerten:
Thanks !

PR #2121, opened 2025-03-13
feat: allow wildcards to match multiple tasks

### LINKED ISSUE #2072: Multiple matches with wildcards
> 1. I think that adding this restriction is fine for now. It keeps things simple. However, I can definitely see a world where someone with a complex setup asks for the current behaviour as it would allow you to do something like this:
> 
>     ```yml
>     version: 3
> 
>     tasks:
>       foo-*-*:
>         ..
>       foo-*:
>         ..
>     ```
> 
>     This allows Task to do something different depending on if the user does `task foo-bar` vs `task foo-bar-baz`. This would not be possible if we errored when multiple tasks match since `task-bar-baz` matches both tasks in the file.
> 
>     Happy to add the error for now and revisit this later though.

_Originally posted by @pd93 in https://github.com/go-task/task/pull/1489#discussion_r1499811332_

pd93:
Fixes #2072 

As discussed in #2072 and https://github.com/go-task/task/pull/1489#discussion_r1499811332, this PR allows calls to match multiple tasks when using wildcards. For example, given the following Taskfile:

```yaml
version: 3

tasks:
  start:*:*:
    vars:
      SERVICE: "{{index .MATCH 0}}"
      REPLICAS: "{{index .MATCH 1}}"
    cmds:
      - echo "Starting {{.SERVICE}} with {{.REPLICAS}} replicas"

  start:*:
    vars:
      SERVICE: "{{index .MATCH 0}}"
    cmds:
      - echo "Starting {{.SERVICE}}" 
```

Currently, if you run `task start:foo:1`, you will get an error:

```
task: Found multiple tasks (start:*:*, start:*) that match "start:foo:1"
```

However, after this change you will see

```
Starting foo with 1 replicas
```

and if you run `task start:foo`, you will get

```
Starting foo
```

Worth noting that the order is important. If you swap the tasks around, then you will be unable to call `start:*:*`. This is because if multiple tasks match, Task will use the task that is listed first in the Taskfile. In general, it is recommended to write more specific tasks (i.e. ones with more parameters) first.

### DIFF
--- a/task.go
+++ b/task.go
@@ -412,7 +412,6 @@ func (e *Executor) FindMatchingTasks(call *Call) []*MatchingTask {
 		return matchingTasks
 	}
 	// Attempt a wildcard match
-	// For now, we can just nil check the task before each loop
 	for _, value := range e.Taskfile.Tasks.All(nil) {
 		if match, wildcards := value.WildcardMatch(call.Task); match {
 			matchingTasks = append(matchingTasks, &MatchingTask{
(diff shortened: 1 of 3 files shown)


vmaerten (approved):
Good job!

PR #2123, opened 2025-03-14
Clarify docs for --dir option

jberkenbilt:
See #2102. I think the actual behavior of `--dir` is useful and described correctly in the Getting Started section of the documentation, but the help and CLI reference don't explain it correctly.

The docs and help describe it as setting the working directory, but what it actually seems to do is to set the directory from which to start searching for a Taskfile. This is useful for running utility-like Taskfiles and seems like a useful feature that is distinct from setting the work directory. This PR just changes the docs to match the behavior. I'm not sure whether the behavior is wrong, but changing behavior would be non-compatible. Looking through the tests, I think the tool is behaving as intended, but the docs are misleading.

I just started playing with `task` today. If you agree with this but want to word it differently or update more docs or change tests, etc., you can discard this and do it a different way.


andreynering (approved):
Hi @jberkenbilt!

Thanks for contributing.

I made a change to the wording because Task will _also_ execute in that directory, as well as look for Taskfiles there.

jberkenbilt:
Thanks! I guess I didn't quite accurately figure out what it was doing, but the new doc is clear.

Issue #1947, opened 2024-12-11
`--version` doesn't work


yonas:
Running `task --version` outputs `Task version: (devel) ()`.

Tested using source for the latest release: https://github.com/go-task/task/archive/refs/tags/v3.40.1.tar.gz

pd93:
@yonas Are you building the source yourself or installing a precompiled binary? If the former, then this is expected. We get the version information at build-time from Git. If you're building yourself, you need to either set this by using `ldflags` or by building inside our Git repo.

Generally, I would recommend that you install Task via one of the methods listed in our docs unless there is a specific reason why you can't.

yonas:
@pd93 Building from source. I get the same problem if it's build from a git checkout with `HEAD` being the `v3.40.1` tag:

```sh
32fa3a0 - (grafted, HEAD, tag: v3.40.1) v3.40.1 (5 days ago) <Andrey Nering>
```

QinZhanlong:
1.  git clone https://github.com/go-task/task.git
2. git checkout v3.40.1
3. download goreleaser binary [https://goreleaser.com/install/]
4. goreleaser build --clean --verbose
@yonas

yonas:
@gedw99 @QinZhanlong Thanks, I'll try these both. :thumbs:

pd93:
@yonas This should be fixed in the next version of Task. [We have changed how we package the version number](https://github.com/go-task/task/pull/2131) and it should now be available without needing `-ldflags` or a Git checkout. Please let us know if you have any further issues.

Issue #1889, opened 2024-10-28
go-task --version unknown


YurBoiRene:
```console
$ go-task --version
Task version: unknown ()
```

- Task version: 3.39.2 (from `dnf info go-task`)
- Operating system: Fedora 40
- Experiments enabled: None

vmaerten:
This installation method is maintained by the community (cf : https://taskfile.dev/installation/#fedora). We don't know how they build it.
You need to ask them.

YurBoiRene:
Opened an [issue](https://bugzilla.redhat.com/show_bug.cgi?id=2322459) in Fedora/Red Hat's issue tracker/

YurBoiRene:
Fedora maintainer fixed it for Fedora 42 :)

Issue #1709, opened 2024-07-03
Version set via ldflags is ignored when building from source tarball


stefanb:
- Task version: 3.38.0
- Operating system: All
- Experiments enabled: likely irrelevant

In
* https://github.com/Homebrew/homebrew-core/pull/176003

It was noted, that the `--version` test failed with `Expected /Task\ version:\ 3\.38\.0/ to match "Task version: (devel) ()\n`

Meaning that the test was expecting `Task version: 3.38.0` but `task --version` returned `Task version: (devel) ()\n`

Even if the correct version was passed to the compiler in ldflags.

Bug introduced with 
* https://github.com/go-task/task/pull/1663

Where it was assumed that the version would be empty by default:
https://github.com/go-task/task/blob/88b095020e14defcce4645e72f503eb5d880f78d/internal/version/version.go#L13-L21

Where in fact `Version` defaults "(devel)" if Go compiler cannot read git info during build, which is the case when building from the downloaded source tarball.
You can try that in playground, where there is also no git info present: https://go.dev/play/p/MCLXd4QaUru

Because of this the version set via Go `ldflags` is always overrwritten and never used/returned by `GetVersion*()` functions.

This is the reason for
* https://github.com/go-task/task/issues/1707

Issue #1207, opened 2023-06-06
Help setting version in Arch Linux package


carlsmedstad:
I'm struggling to have the binary output the correct version for the AUR package.

See the the `build` function in the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=go-task):

```sh
build() {
  cd "$_archive"

  export CGO_CPPFLAGS="${CPPFLAGS}"
  export CGO_CFLAGS="${CFLAGS}"
  export CGO_CXXFLAGS="${CXXFLAGS}"
  export CGO_LDFLAGS="${LDFLAGS}"
  export GOFLAGS="-buildmode=pie -trimpath -ldflags=-linkmode=external -mod=readonly -modcacherw"

  go build -v -o bin ./cmd/task
}
````

I've tried setting multiple ld flags but I've been unable to get any other output than:

```sh
$ go-task --version
Task version: (devel)
```

For example `-X github.com/go-task/task/version=$pkgver` which I found here: https://github.com/go-task/task/issues/1050.

Assistance very much appreciated.

andreynering:
Hi @carlsmedstad,

This is the flag you need to add to `go build` or `go install`:

```bash
-ldflags="-X github.com/go-task/task/v3/internal/version.version=v9.9.9"
```

Let me know if that works for you or not.

carlsmedstad:
Works like a charm, thank you @andreynering!

PR #2131, opened 2025-03-20
feat: versioning

pd93:
There have been many issues raised over the content of `version.go` - The file where we get the Task version used for the `--version` flag and `.TASK_VERSION` variable.

- #1889
- #1709
- #1207
- #1947
- #2105
- #2150

Currently, we use a mixture of `debug.ReadBuildInfo` and `-ldflags` to set the version variable. The issue with this is that build info is only available when built with a Git checkout and `-ldflags` requires the maintainer of the release mechanism to maintain the version string/checksum correctly. Thins leads to inconsistencies between various builds.

This PR overhauls the whole thing to make some permanent improvements:

- When we run the release script, a `version.txt` file will be generated containing the version number
- The `version.txt` file is then embedded into the binary at build time using `//go:embed`
- This can still be overridden by `-ldflags` so we aren't breaking any existing overrides
- `debug.ReadBuildInfo` is queried and the commit/dirty status appended to the version if the information is available and is a development build (not tagged)
- Removed the `Task version:` prefix so that the output is easier to parse
- Removed the checksum

Examples:

```shell
# Any tagged build or when built outside a Git checkout
3.42.1
# Untagged commit in a Git checkout
3.42.1-2110245
# Uncommitted changes on an untagged commit in a Git checkout
3.42.1-2110245-dirty
```


vmaerten (approved):
Nice ! 
I've added Copilot as reviewer to see if he has something good to say

andreynering (approved):

PR #2134, opened 2025-03-21
feat(completion): let fish complete global tasks if -g or --global is passed

atusy:
In the current implementation, fish completion suggests local tasks regardless of `-g` or `--global` options.
This PR fixes the problem by adding a feature to detect these options and then suggesting global tasks if likely.

I hope this PR can be merged although I do not implement the feature for bash and zsh.
I am not familiar with them...

### DIFF
--- a/completion/fish/task.fish
+++ b/completion/fish/task.fish
@@ -6,10 +6,10 @@ function __task_get_tasks --description "Prints all available tasks with their d
   set -l cmd_args
   eval "set cmd_args $(commandline --current-process)" # split commandline by arguments considering quotes and escapes
   for arg in $cmd_args
-    if test _$arg = _"--"
+    if test "_$arg" = "_--"
       break # ignore arguments to be passed to the task
     end
-    if test _$arg = _--global -o _$arg = _-g
+    if test "_$arg" = "_--global" -o "_$arg" = "_-g"
       set global_task true
       break
     end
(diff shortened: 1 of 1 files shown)


completion/fish/task.fish
```diff
@@ -1,8 +1,26 @@
-set GO_TASK_PROGNAME task
+set -l GO_TASK_PROGNAME task
 
 function __task_get_tasks --description "Prints all available tasks with their description" --inherit-variable GO_TASK_PROGNAME
+  # Check if the global task is requested
+  set -l global_task false
+  set -l cmd_args
+  eval "set cmd_args $(commandline -b)" # split arguments by spaces while taking into account of quotes and escapes
```
atusy, line 7: `eval` is dangerous.

Let's say commandline is

```
$ foo --global; task
```

then `eval` evaluates

```
set cmd_args foo --global; task
```

which in turn 

1. evaluates `task default`...
2. shows global tasks althoug `-g` and `--global` are not specified to `task`

atusy: Okay, I found the problem can be fixed by selecting only a part of commandline instead of whole.

<https://github.com/go-task/task/pull/2134/commits/5f53cf0df7ec509bfcbd8bc742e7984ce5eba78b>

If this is still unlikely, I would avoid `eval` and instead simply split arguments by spaces.

```
set -l cmd_args "$commandline --current-process | string split ' '"
```

The caveat is that string split will complete global tasks in some unexpected cases (e.g., `task "foo -g"`).

atusy: Now I find a way to avoid `eval`!
`read --tokenize` is what we need.

andreynering:
Thank you @atusy!

atusy:
My pleasure!!

PR #2140, opened 2025-03-26
feat: `CLI_ARGS_LIST` global variable

### LINKED ISSUE #2138: Using `for` with `CLI_ARGS` unexpectedly splits arguments containing spaces
### Description

```sh
mkdir inputs
echo 'hello' > inputs/hello.txt
echo 'world' > inputs/world.txt
echo 'hello world' > inputs/"hello world".txt
```

```
$ ls -l inputs 
total 24
-rw-r--r--  1 chrislarose  staff  12 Mar 24 17:22 hello world.txt
-rw-r--r--  1 chrislarose  staff   6 Mar 24 17:22 hello.txt
-rw-r--r--  1 chrislarose  staff   6 Mar 24 17:22 world.txt
```

`Taskfile.yml`

```YAML
# https://taskfile.dev

version: '3'

tasks:
  default:
    cmds:
      - for: { var: CLI_ARGS, as: FILENAME }
        cmd: echo "{{.FILENAME}}"
```

Then

```
task default -- inputs/hello.txt inputs/world.txt inputs/"hello world".txt
```

produces unexpectedly

```
task: [default] echo "inputs/hello.txt"
inputs/hello.txt
task: [default] echo "inputs/world.txt"
inputs/world.txt
task: [default] echo "'inputs/hello"
'inputs/hello
task: [default] echo "world.txt'"
world.txt'
```

The command was executed 4 times, even though only 3 arguments were sent to `task` after `--`. I personally ran into this when trying to invoke `task` using `xargs`

```
 $ find inputs -type f | xargs task default --
task: [default] echo "inputs/world.txt"
inputs/world.txt
task: [default] echo "inputs/hello"
inputs/hello
task: [default] echo "world.txt"
world.txt
task: [default] echo "inputs/hello.txt"
inputs/hello.txt
```

### Version

Task version: v3.42.1 (h1:HOaFbZGLOrAy2V/dLsX2rGJZVG2Qx6268KUIAIXdNE4=)

### Operating system

macOS Sonoma 14.1.2

### Experiments Enabled

_No response_

### LINKED ISSUE #2139: Add a `CLI_ARGS_LIST` global variable
### Description

When the `CLI_ARGS` variable was first introduced, Task only had string variables, so the variable was populated with a single string with shell quoted values. Since then [we have introduced "any type" variables](https://taskfile.dev/blog/any-variables) which means that arrays are now available.

~~Instead of~~ As well as joining the args together, we should make them available in an array. This makes the individual arguments easier to access and loop through without the user having to do additional processing. ~~If the user wants a single string, they are still able to use a string joining function. However,~~ this also adds the flexibility for users to to process arguments first or choose a separator of their own.

~~Since this would be a breaking change, we should make this available as an [experiment](https://taskfile.dev/experiments/).~~

Related issues:

- #2138

Edit: Changed this issue from a proposed experiment to edit an _existing_ variable to a feature request for a _new_ variable. This means it is no longer a breaking change and no longer needs to be an experiment.

pd93:
Fixes #2139 
Fixes #2138

Adds a new `CLI_ARGS_LIST` global variable which contains the pre-dash arguments in an array.

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -13,24 +13,14 @@ import (
 // Get fetches the remaining arguments after CLI parsing and splits them into
 // two groups: the arguments before the double dash (--) and the arguments after
 // the double dash.
-func Get() ([]string, string, error) {
+func Get() ([]string, []string, error) {
 	args := pflag.Args()
 	doubleDashPos := pflag.CommandLine.ArgsLenAtDash()
 
 	if doubleDashPos == -1 {
-		return args, "", nil
+		return args, nil, nil
 	}
-
-	var quotedCliArgs []string
-	for _, arg := range args[doubleDashPos:] {
-		quotedCliArg, err := syntax.Quote(arg, syntax.LangBash)
-		if err != nil {
-			return nil, "", err
-		}
-		quotedCliArgs = append(quotedCliArgs, quotedCliArg)
-	}
-
-	return args[:doubleDashPos], strings.Join(quotedCliArgs, " "), nil
+	return args[:doubleDashPos], args[doubleDashPos:], nil
 }
 
 // Parse parses command line argument: tasks and global variables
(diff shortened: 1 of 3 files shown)


vmaerten (approved):
Nice!
Maybe worth it to wait #2189 and add it :D

pd93:
@vmaerten Just gunna merge this now and we can circle back on the current release items if #2189 gets merged.

PR #2169, opened 2025-04-07
docs: package api docs

pd93:
This PR contains some new package API docs that I have been working on and a very small refactor to the `taskfile.Reader` type that removes the `Node` field in favor of it being passed in as an argument to the `Reader.Read` method.

I have further sections for this doc in progress (Such as `Executing tasks`), but want to wait until I have submitted a few other changes before they are published.


andreynering (approved):
Good stuff! 👏

PR #2184, opened 2025-04-16
docs: Mention version checks introduced in v3.34.0

jubr:
Hi guys, I just ran into my installed version v3.33.0 not reporting it is not new enough as compared to `Taskfile.yaml`:

```
version: '3.41.0'
```

So I thought to provide a suggestion to update the docs for this.


jubr:
Hmm.. thinking about this some more after I clicked "Submit"...

I think there was another fix on this, bumping the effective version this works from. Meaning we might need to update the version in my edit. Not sure what that version is exactly...

pd93:
@jubr You need to select "Next" from the versions drop-down when previewing unreleased changes: https://deploy-preview-2184--taskfile.netlify.app/next/taskfile-versions/

jubr:
Ah yes, missed it! Tnx! (rushing this inbetween my daily job tasks)

So, @pd93 , what do you think, makes sense to add this?

website/docs/taskfile_versions.mdx
```diff
@@ -60,6 +60,13 @@ Now when someone tries to run the Taskfile with an older version of Task, they
 will receive an error prompting them to upgrade their version of Task to
 `v3.17.0` or greater.
 
+:::info
+
+Note that this functionality was introduced in [v3.34.0](https://github.com/go-task/task/releases/tag/v3.34.0),
+so older versions are simply not yet able to give you this warning.
```
pd93, line 66:
```suggestion
so older versions are not able to give you this warning.
```

Nit: Avoid using terms like "simply" in docs. This is a habit I sometimes fall into too.

Also, "not yet" implies they might one day be able to which they won't.

jubr:
And in regard to the first version that is able to detect this properly? What's the smart choice there?

jubr:
Crap^1, IRL happened, sorry for late reply.
Crap^2, can't find it anymore, v3.34.0 should be fine as a starting version.

trulede:
This was introduced in v3.34.0 (e2b85c6).

website/src/docs/taskfile-versions.md
```diff
@@ -60,6 +60,13 @@ Now when someone tries to run the Taskfile with an older version of Task, they
 will receive an error prompting them to upgrade their version of Task to
 `v3.17.0` or greater.
 
+:::info
+
+Note that this functionality was introduced in [v3.34.0](https://github.com/go-task/task/releases/tag/v3.34.0),
+so older versions are not able to give you this warning.
+
+:::
+
```
trulede, line 73: Shorter hint?

> Semantic version checking was introduced with v3.34.0 of Task.

pd93 (approved):
Sorry this took so long to merge. Changes look fine. Thanks!

jubr:
👍

PR #2189, opened 2025-04-21
docs: add since badge

vmaerten:
It's a proposal to indicate the version from which a feature is available.

I've added to some feature ! I also apply it to versionned_docs to have it right now :)

NOTE: I used custom alias to ensure this will work when doing a release (all paths will remain the same)


andreynering:
I loved the idea! I vote for option 5.

website/src/components/AvailableSince.tsx
pd93, line 1: Might be good to apply a formatter here if its easy to do

pd93:
Agreed. 5 Looks good to me.

website/src/components/AvailableSince.tsx
vmaerten: Yes ! I planned to clean and format this

pd93:
Happy with this. However, I do have one question regarding its usage: How far back are we supporting these badges? Is the intention to go through _all_ the docs and add version stamps? (this seems like a lot of work and might lead to very crowded docs in some places where many changes have been made). If not, how far back do we plan to label things and should things become unlabelled after a given period of time or number of versions (this could probably be automated).

For example, we have labelled that map variables are available since v3.43.1, but actually all variable types (except strings) have only been available since v3.37.0.

Another way to do this is to snapshot our docs on each release and have a drop down. Food for thought.

vmaerten:
I like the idea of versionning our docs. We can use docusaurus to do that!
@andreynering what is your thought about this ?

PR #2191, opened 2025-04-21
fix: CLI_ARGS is a string and not an array

vmaerten:
CLI_ARGS shoud be a string and not an array
Introduced in #2166


andreynering:
@vmaerten CI is failing

vmaerten:
Yep I went too fast.
Fixed now. I don't know why argument was inverted in the call line 79 but not in the underlying method

andreynering (approved):

Issue #2192, opened 2025-04-21
Fuzzy suggestions for tasks no longer work


gregtatum:
### Description

Our task file is getting kind of long (not in a bad way). I typed in a task incorrectly and thought it would be quite nice to suggest the closest task when the spelling is slightly off. I've had other tools automatically run the command it thinks you wanted, but just offering a sugestion would improve the ergonomics here.

For instance, something like this:

```
➤ task line-pyright
task: Available tasks for this project:
[...]
* lint:                        Run all available linting tools.
* lint-black:                  Checks the styling of the Python code with Black.
* lint-black-fix:              Fixes the styling of the Python code with Black.
* lint-eslint:                 Checks the styling of the JS code with eslint.
* lint-eslint-fix:             Fixes the styling of the JS code with eslint.
* lint-fix:                    Fix all automatically fixable errors. This is useful to run before pushing.
* lint-pyright:                Runs type checking on python files.
* lint-ruff:                   Lints the Python code with the ruff linter.
* lint-ruff-fix:               Fixes Python code lint errors with the ruff linter.
[...]
task: Task "line-pyright" does not exist.
Did you mean `task lint-pyright`?
```

I searched through the issues and didn't see this feature request anywhere.

pd93:
@gregtatum Thanks for the report. This request is actually a duplicate of #867 which was implemented in #880 and released in [v3.17.0](https://github.com/go-task/task/releases/tag/v3.17.0) in Oct 2022.

However. It looks like it regressed immediately in v3.18.0 and no-one ever reported it. Strange that I never noticed as I remember really liking the feature at the time. I'm going to reclassify this as a bug as the code already exists. Just something not working correctly.

PR #2200, opened 2025-04-23
fix: fuzzy model was not instanciated

vmaerten:
Linked to #2192

Tested locally : 
image

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -937,3 +937,24 @@ func TestVarInheritance(t *testing.T) {
 		)
 	}
 }
+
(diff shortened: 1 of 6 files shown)


pd93 (commented):
Ha, love a one character fix. Maybe a quick test to ensure this doesn't regress again? I'm amazed it took 2.5 years for someone to notice 😆

vmaerten:
You are totaly right!
I've added it to `executor_test`

I just wanted to say, the way you refactored all tests is amazing. Way simpler to use, thanks a lot for this

pd93 (approved):

PR #2211, opened 2025-04-25
chore: update schemas only when a release is done

vmaerten:
Our schemas are updated as soon as changes are merged into the main branch, but this can be confusing since some features may not be released yet.

To solve this, I followed the same pattern we use for the documentation. The files next-schema.json and next-schema-taskrc.json will be copied to schema.json and schema-taskrc.json during the release process.


pd93 (approved):
Nice! This is a good idea. One small suggestion: We could update the CI step that we have for detecting changes to the wrong docs folder so that it detects changes to the wrong schema files too.

vmaerten:
Done ! 
I've made the choice to create another job in the workflow. This way we can have a proper error message

PR #2220, opened 2025-04-28
feat: redact credentials in remote urls

### LINKED ISSUE #2100: Remote taskfile could potentially expose secrets
### Description

I understand this is an experimental feature but the log/warning messages to console should not expose secrets or tokens as plain text. If this is execute by scripts or in a CI environment it could potentially expose the secret tokens or end up in historical logs.

when running with:

```sh
export TASK_X_REMOTE_TASKFILES=1
task -y my-task
```

the following output is echoed to terminal:

```sh
The task you are attempting to run depends on the remote Taskfile at "https://ghp_<secret>@raw.githubusercontent.com/username/taskfile/main/Taskfile.yml".
--- Make sure you trust the source of this Taskfile before continuing ---
Continue? [assuming yes]
```

Ideally, sensitive values would be masked in the output. At minimum there needs to be a way to suppress the warning with an environment variable. The `-s` (`--silence`) flag doesn't hide this message either; i.e. `task -y -s my-task`. An environment variable seems the most logical IMO to ensure the flag isn't missed in scripts and would handle exposure at the repo/system level.

### Version

3.41.0

### Operating system

MacOS

### Experiments Enabled

Remote Taskfiles

### Example Taskfile

```YAML
version: "3"

includes:
  git: https://{{.GH_TOKEN}}@raw.githubusercontent.com/username/taskfile/main/Taskfile.yml

tasks:
   my-task:
     cmds:
        - tasks: git
```

pd93:
Fixes #2100 and replaces #2045

Ensures that any credentials embedded in the URL are redacted by using the built in `url.Redacted()` function. The cache keys have also been altered slightly, so users may have to redownload some files that they already have cached.


iwittkau:
I think this test should be added an pass:

```go
func TestGitNode_redaction(t *testing.T) {
	t.Parallel()

	node, err := NewGitNode("https://git:token@github.com/foo/bar.git//directory/Taskfile.yml?ref=main", "", false)
	assert.NoError(t, err)
	assert.Equal(t, "main", node.ref)
	assert.Equal(t, "directory/Taskfile.yml", node.path)
	assert.Equal(t, "https://git:xxxxx@github.com/foo/bar.git//directory/Taskfile.yml?ref=main", node.Location())
	assert.Equal(t, "https://git:xxxxx@github.com/foo/bar.git", node.URL.String())
	entrypoint, err := node.ResolveEntrypoint("common.yml")
	assert.NoError(t, err)
	assert.Equal(t, "https://git:xxxxx@github.com/foo/bar.git//directory/common.yml?ref=main", entrypoint)
}
```

taskfile/node_git.go
```diff
@@ -40,13 +40,10 @@ func NewGitNode(
...
 
-	basePath, path := func() (string, string) {
-		x := strings.Split(u.Path, "//")
-		return x[0], x[1]
-	}()
+	basePath, path := splitPath(u)
 	ref := u.Query().Get("ref")
 
-	rawUrl := u.String()
+	rawUrl := u.Redacted()
```
iwittkau, line 46: This redacted version should be used as the node.URL and for the `errors.TaskfileNotSecureError.URI`,

taskfile/node_git.go
```diff
@@ -127,11 +124,23 @@ func (node *GitNode) ResolveDir(dir string) (string, error) {
...
+	prefix := filepath.Base(node.path)
 	// Means it's not "", nor "." nor "/", so it's a valid directory
 	if len(lastDir) > 1 {
-		prefix = fmt.Sprintf("%s-%s", lastDir, prefix)
+		prefix = fmt.Sprintf("%s.%s", lastDir, prefix)
+	}
+	return fmt.Sprintf("git.%s.%s.%s", node.URL.Host, prefix, checksum)
+}
+
+func splitPath(u *url.URL) (string, string) {
```
iwittkau, line 136:
```suggestion
func splitGitURLPath(u *url.URL) (string, string) {
```

That's maybe more descriptive.

iwittkau (commented):
Nice! 🎉 

I'll give it a try with my work setup tomorrow.

iwittkau:
I think you could keep GitNode.URL unexported to make sure its `String` and `Redacted` funcs are called correctly.

taskfile/node_git.go
pd93: Thanks, I missed the `errors.TaskfileNotSecureError.URI`

taskfile/node_git.go
pd93: This could potentially be used elsewhere in the future (in which case I imagine it would be moved), so I opted to go for a name that describes what the function does rather than how it is used. I've updated it to `splitURLOnDoubleSlash` for now.

pd93:
> I think you could keep GitNode.URL unexported to make sure its `String` and `Redacted` funcs are called correctly.

Yeah, we don't really need any of the properties on `Node` implementations to be exported. They have constructors and they're quite bespoke. Package API users can just roll their own `Node`s if they need custom behavior. I have unexported all `Node` fields.

iwittkau:
@pd93 I just noticed that this branch needs a `go mod tidy`.

taskfile/node_http.go
```diff
@@ -16,9 +16,8 @@ import (
 
 // An HTTPNode is a node that reads a Taskfile from a remote location via HTTP.
 type HTTPNode struct {
-	*BaseNode
-	URL        *url.URL // stores url pointing actual remote file. (e.g. with Taskfile.yml)
```
vmaerten, line 20: question / nitpick : you renamed `URL` to `url` in the `node_git` but not for the `node_http`, should we do the same here ? or this field has to be public ?

vmaerten (commented):
The is cache naming is way better like this!

taskfile/node_http.go
pd93: Honestly just missed it. Thanks!

vmaerten (approved):
lgtm ! gj :)

iwittkau:
I've been using this for a while now and it works nicely 👍🏻

PR #2227, opened 2025-04-29
Correct the version number in env_precedence experiment

antonio-caceres:
I simply updated the version number in the example files for the Environment Precedence experiment, so it's clearly communicated to users the experiment is only available after release v3.39.0.


antonio-caceres:
I was confused on what the version value meant.

PR #2233, opened 2025-05-02
docs: mention that method key is allowed at root level

### LINKED ISSUE #975: Change the method to timestamp for all the tasks
I'd like to change the method of checking for the state of a task to `timestamp` for all the tasks because the `checksum` method is slow. Is there a way to do this?

vmaerten:
# Summary

The docs does not mention that `method` can be used at root level.

Fixes #975


andreynering (approved):

bijela-gora:
@vmaerten thank you!

PR #2235, opened 2025-05-03
feat: add experiments to taskrc.yml schema

vmaerten:
# Summary

Added all experiments to the schema, clarifying which keys and values are expected. The schema will be updated during releases, allowing new experiments to be released without modifying the current schema


vmaerten:
Conflict resolved

pd93 (approved):

PR #2246, opened 2025-05-06
chore: publish nightly

vmaerten:
Now that we have GoReleaser Pro, it might be a good idea to publish nightly pre-releases.
Some other projects have it. I took inspiration from GoReleaser itself.

I wasn’t able to test it yet because I don’t have the KEY.

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -1,0 +1,31 @@
+name: Realease nightly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 0 * * *
+
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.x
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: release --clean --nightly
+        env:
+          GITHUB_TOKEN: ${{secrets.GH_PAT}}
+          GORELEASER_KEY: ${{secrets.GORELEASER_KEY}}
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -4,8 +4,6 @@ on:
   workflow_dispatch:
   schedule:
     - cron: 0 0 * * *
-
-
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -51,12 +51,12 @@ release:
 
 git:
   ignore_tags:
-    - "{{ if not .IsNightly }}nightly{{ end }}"
+    - "{{if not .IsNightly}}nightly{{end}}"
 
 nightly:
   publish_release: true
   keep_single_release: true
-  version_template: "{{ incminor .Version }}-nightly"
+  version_template: "{{incminor .Version}}-nightly"
 
 snapshot:
   version_template: "{{.Version}}"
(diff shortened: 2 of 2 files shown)


pd93:
I'm not against this if others want it, but I've never seen anyone ask for this and its not the sort of thing I would personally use either. I'm concerned that it's going to pollute our [releases page](https://github.com/go-task/task/releases) with a ton of releases that largely don't contain any changes (we tend to only merge in clumps 2-3 times a month - certainly not daily like bigger projects). Users that really want the latest changes should easily be able to build from source themselves IMO.

If we really wanted to press ahead with this, then I might suggest having a `go-task/task-nightlies` repo which builds the binaries for those who can't do it themselves (or can't be bothered). This would avoid the pollution issue.

vmaerten:
> I'm not against this if others want it, but I've never seen anyone ask for this and its not the sort of thing I would personally use either. I'm concerned that it's going to pollute our [releases page](https://github.com/go-task/task/releases) with a ton of releases that largely don't contain any changes (we tend to only merge in clumps 2-3 times a month - certainly not daily like bigger projects). Users that really want the latest changes should easily be able to build from source themselves IMO.
> 
> If we really wanted to press ahead with this, then I might suggest having a `go-task/task-nightlies` repo which builds the binaries for those who can't do it themselves (or can't be bothered). This would avoid the pollution issue.

Hey @pd93, thanks for your comment.

The main reason I proposed this is that I was looking for a way to gather feedback before a release—whether for a feature or a bug fix. I assumed that not everyone wants to or can build Task locally to test it, and that having a pre-release available might help with that.

Of course, I could be wrong and maybe no one would actually use it!

To address your concern about pollution: only one pre-release would be shown at a time. The same pre-release would simply be updated nightly.
For example, here’s how it looks on [Goreleaser’s releases page](https://github.com/goreleaser/goreleaser/releases):
image


That said, I was just exploring ways to increase adoption and feedback before final releases. If you (or we collectively) feel it wouldn’t be useful, I’m fine with dropping the idea.

pd93:
If only the latest pre-release is shown then I have no concerns about pollution. Does the pre-release disappear once a normal release is made? If so, I'm actually pretty happy with it.

- Regarding `--version`: It is currently only bumped before each release. Usually a nightly will set the version to the next minor release plus whatever build info. e.g. currently it would be `3.43.3+fb68a5f`, but we might want `3.44.0-nightly+fb68a5f`?
- Changelogs: If pre-releases "override" one another in the list, then the changelog should be cumulative IMO. i.e. The entire "Unreleased" section should be included in the nightly changelog, not just the diff since the last nightly (since this would likely be empty most nights). This is probably also easier to implement.

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,31 @@
+name: Realease nightly
+
+on:
+  workflow_dispatch:
```
andreynering, line 4: @vmaerten @pd93 Good to have in mind the with `workflow_dispatch` we can trigger this manually after doing an important fix.

Useful when to ask users to try nightly without waiting for the next day.

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,31 @@
...
+  schedule:
+    - cron: 0 0 * * *
+
+
```
andreynering:
```suggestion

```

.goreleaser.yml
```diff
@@ -48,6 +48,16 @@ archives:
...
 
+
+git:
+  ignore_tags:
+    - "{{ if not .IsNightly }}nightly{{ end }}"
+
+nightly:
+  publish_release: true
+  keep_single_release: true
+  version_template: "{{ incminor .Version }}-nightly"
```
andreynering: Just standardizing Go template a bit:

```suggestion
git:
  ignore_tags:
    - "{{if not .IsNightly}}nightly{{end}}"

nightly:
  publish_release: true
  keep_single_release: true
  version_template: "{{incminor .Version}}-nightly"
```

PR #2256, opened 2025-05-11
feat: add task name to the json output

aleksandersh:
Hi, nice to meet you, and thanks for developing the task tool!

## Rationale

I'm implementing [the tui client](https://github.com/aleksandersh/task-tui) for the task. I have decided that the best way to integrate a user interface is to use the JSON output feature (`task --list --json`).

All works fine except [the labels](https://taskfile.dev/usage/#overriding-task-name). If some tasks have the label field, the JSON output will contain it instead of the real name. Since I can't request the execution by the task label, the client is stuck here.

https://github.com/go-task/task/blob/da4ce5b0a58c328f83c5813806ab5e6265e93fdc/taskfile/ast/task.go#L53

I found the same problem in the vscode plugin (https://github.com/go-task/vscode-task/issues/167) since it is integrated with the task in the same way.

## Changes

I assume that it's not right to change the behavior of this field, because there are some tools which can rely on this. Therefore, I decided to add a separate field (`task`) with the real task name, which can be used exactly for the purpose of executing the task.

The new JSON output will look like below

```json
{
  "tasks": [
    {
      "name": "foobar",
      "task": "foo",
      "desc": "task description",
      "summary": "",
      "aliases": [],
      "up_to_date": false,
      "location": {
        "line": 0,
        "column": 0,
        "taskfile": ""
      }
    }
  ],
  "location": ""
}
```

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -42,9 +42,10 @@ type (
 		FormatterTestOption
 	}
 	TaskTest struct {
-		name           string
-		experiments    map[*experiments.Experiment]int
-		postProcessFns []PostProcessFn
+		name                string
+		experiments         map[*experiments.Experiment]int
+		postProcessFns      []PostProcessFn
+		fixtureTemplateData any
 	}
 )
 
@@ -79,7 +80,11 @@ func (tt *TaskTest) writeFixture(
 	if goldenFileSuffix != "" {
 		goldenFileName += "-" + goldenFileSuffix
 	}
-	g.Assert(t, goldenFileName, b)
+	if tt.fixtureTemplateData != nil {
+		g.AssertWithTemplate(t, goldenFileName, tt.fixtureTemplateData, b)
+	} else {
+		g.Assert(t, goldenFileName, b)
+	}
 }
 
 // writeFixtureBuffer is a wrapper for writing the main output of the task to a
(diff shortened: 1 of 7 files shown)


task_test.go
```diff
@@ -234,6 +239,26 @@ func (opt *setupErrorTestOption) applyToFormatterTest(t *FormatterTest) {
...
+}
+
+func (opt *fixtureTemplateDataTestOption) applyToExecutorTest(t *ExecutorTest) {
+	t.fixtureTemplateData = opt.data
+}
+
+func (opt *fixtureTemplateDataTestOption) applyToFormatterTest(t *FormatterTest) {
+	t.fixtureTemplateData = opt.data
+}
+
```
pd93, line 261: Niceee! I really like this idea. In other tests we use a post-process function to remove absolute paths from the fixtures, but I was never really happy with that solution. This is much cleaner!

I wonder if we should provide a standardised set of variables to all tests for this purpose. No need to do this in this PR though.

aleksandersh: Thank you, Pete! I'm glad you found these changes interesting

pd93 (approved):
🚀 LGTM. Pending a second review

PR #2265, opened 2025-05-20
chore: replace PPRemoveAbsolutePaths with generic fixture template data

pd93:
Followup to the style of test that was added in #2256. Instead of expecting a struct of type `any`, I have converted it to use a `map[string]any`. This largely works the same way, but with the benefit that defining variables is a bit nicer syntactically and we can utilise functions like `maps.Copy()` to do key merging. This allows us to define a predefined set of variables for use in all tests and then merge additional values if required.

Variables being made available for now are:

- `TEST_NAME` - Contains the name of the test
- `TEST_DIR` - Contains the absolute path to the test directory

I have removed `PPRemoveAbsolutePaths` in favor of this new technique.

One small hiccup is that since #2256 was merged, running `task gen:fixtures` will remove the template values from the golden files which then causes the tests to fail. This can be resolved manually, but is a bit of a pain. This is reported in https://github.com/sebdah/goldie/issues/14 and I ~am considering an upstream feature request if @sebdah is interested.~ have [opened a PR](https://github.com/sebdah/goldie/pull/44) to address it.


sebdah:
This is a good idea. I'd be happy to merge this into goldie if you'd want to submit a PR.

pd93:
Goldie v2.7.1 has been released with both my PRs (https://github.com/sebdah/goldie/pull/44, https://github.com/sebdah/goldie/pull/45) merged. I've updated this PR and we should be good to merge now.

Thanks @sebdah for your help getting this done!

andreynering (approved):

Issue #2273, opened 2025-05-28
Idea: Add a global `defer:` attribute to cleanup stuff before exiting `task`?


andreynering:
Keep in mind that the original request asks for running a cleanup _before exiting Task_, and not _after every task ran_.

Not sure if `defer:` if the right keywork for this. `cleanup:` could be clearer on the intention, but `defer:` is more consistent to what we already have.

---

### Discussed in https://github.com/go-task/task/discussions/2272

<div type='discussions-op-text'>

<sup>Originally posted by @ssbarnea May 28, 2025</sup>

#### How can I make task run some commands always before exiting regardless which task was called?

I want to add some special cleanup logic at the end of execution of `task` command but one that would run regardless which task I called. 

For maintenance reasons I cannot really add an extra defer task of each task being added, so I need to rely of a feature that would allow me to define this special type of implicit "task", one that would be called only before exiting.

It would be key that this task is not called before task executable exits, as we don't want to be called accidentally in the middle of task executions.

I am aware that that there is a small risk of this being called if someone is calling task recursively using shell commands, but that is a small price to pay (in theory nesting can be detected via env vars).
</div>

hugo-ricateau:
The proposal #2040 might address such use-cases as well, while being more flexible (e.g. cleanup only occurs if some designated tasks get executed); in this regards, `init:` (resp. `finally:`) is just a matter of declaring `deps:` (resp. `defer:`) on every available task.

@andreynering, what do you think of such an implementation?

PR #2277, opened 2025-05-30
Support init and exit tasks.

trulede:
Outline suggestion (only) for supporting `init` and `exit` tasks using the task.run field. Further testing and refactoring likely required.

General operation; init tasks (run = "init") run before all other tasks, exit tasks (run = "exit") run _after_ all other tasks. 

As a result there is some amount of lifecycle management for starting services, containers, or authentication.

Relates to #2273

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -177,5 +177,20 @@ func run() error {
 		return e.Status(ctx, calls...)
 	}
 
-	return e.Run(ctx, calls...)
+	getRunCalls := func(run string) ([]*task.Call) {
+		var calls = []*task.Call{}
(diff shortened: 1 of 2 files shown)


andreynering:
I think I lean towards having `init` and `finally` keywords at the top, like this:

```yaml
version: '3'

init:
  - echo "init something"
  - task: init-something

finally:
  - echo "finish something"
  - task: finish-something

tasks:
  # ...
```

trulede:
As I think about it, we normally used `run: once` to manage login to remote services (docker registry), so would that be a synonym to `init`? Perhaps ... aside from that particular use case, it does illustrate how we use `run` to manage lifecycle issues.

PR #2280, opened 2025-06-01
Execution Graph to detect cyclic tasks

### LINKED ISSUE #820: "probably an cyclic dep or infinite loop" on huge graphs
If the graph is large, you can get the following error:

```
task: maximum task call exceeded (100) for task: probably an cyclic dep or infinite loop
```

The reasons are as follows:
https://github.com/go-task/task/blob/bf9cd7625bdc9510a309b09a3b47b0fd4b23ac8e/task.go#L110-L112
So, if we try to execute the task >= 100 times, then we crash.

The issue is that `RunTask` is executed once from each of the dependencies:
https://github.com/go-task/task/blob/bf9cd7625bdc9510a309b09a3b47b0fd4b23ac8e/task.go#L200-L206
but we increment the counter before checking if the task is already running or finished. Such check is done below, in `startExecution`:
https://github.com/go-task/task/blob/bf9cd7625bdc9510a309b09a3b47b0fd4b23ac8e/task.go#L300-L327

So, if there is a task on which 100 other tasks depend, then the graph will fail with an error.

Moreover, if you read `startExecution` carefully, then it would appear that the error has nothing to do with cyclic deps. The simplified algorithm is as follows:
```go
RunTask(task) {
  if (hashes.contains(task)) {
    // wait for task
    return;
  }
  hashes.insert(task);
  // iterate over deps
  // run commands
}
```

So, even for cyclic dependency, the task will be executed once (though, it will eventually stuck waiting for itself).

trulede:
Execution graph to detect cyclic tasks, replacing the TaskCalledTooManyTimesError mechanism.

Vertices in the graph are added for each unique call (to a task) and the `PreventCycles` feature of the graph is used to detect cyclic execution. The unique identifying properties are:

- The task name, unique from the Taskfile AST/DAG mechanism.
- The task hash, unique for an "instance" of a task call (e.g. set of vars).
- Call "index" of the task, either from 'deps' or 'cmds'.
- The sources method value, propagated as a call "token". This addresses a special case where a task call is cyclic, but the sources change between each cycle, and thus the task call should continue.

Although this method has worked remarkably well, it could be beneficial to have a mechanism to disable the detection: Environment variable (of course) and perhaps attached to an existing CLI parameter (--force).

fixes #820


trulede:
The graph seems to work, and the theory behind it is becoming more robust. I think the sources for a task will need to be added to the hash, but this needs to be proven with a test.

Also, worthwhile covering some of the previous reported cases: #820 

And, I think it could be useful to be able to disable this with an envar, since ... sometimes people know what they are doing ...

PR #2285, opened 2025-06-07
Task packaging as container image.

### LINKED ISSUE #1801: Distribute as Docker Image

trulede:
This is a dockerfile and workflow for distributing Task in a scratch based image. Images would be pushed to GHCR, which is significantly less a pain in the arse than pushing images to Docker Hub, but you can do that too ... I doubt its worth the effort!

Workflow integration is tedious, the example I have provided should work out-of-the-box.

fixes #1801

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,60 @@
+name: Docker
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - v*
+
+permissions:
+  contents: read
+  pull-requests: read
+  packages: read
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository_owner }}/task
+  # SEMVER_VALUE: ",value=v0.0.4" # use this when testing
+  SEMVER_VALUE: ""
+jobs:
+  docker:
+    name: Docker Container Images
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


trulede:
This is an alternative to #1875 (and #1868?) and just simply distributes task in a container. People can use this image in a multi-stage build to directly install the task binary and avoid compilation delay of manually installing task. And also to pin versions.

Generally, running task in a container provided by go-task is not itself useful, since: we (royal "we") are not going to trust that image (at all!); we put a lot of effort into our own images; and in any case we need to package additional tools which task itself is then automating. Also, providing docker images for broad distribution is a _lot_ of stuffing around ...

Therefore this alternative is provided.

trulede:
Better to use goreleaser for this project.

PR #2287, opened 2025-06-08
Execution Graph to detect cyclic tasks

### LINKED ISSUE #820: "probably an cyclic dep or infinite loop" on huge graphs

### LINKED ISSUE #2302: undetected dependency loop or deadlock with `run: when_changed`
### Description

In the example Taskfile.yml below, running `task` "hangs" forever. Okay, I didn't actually wait forever. But it hangs for long enough for you to think it's in an infinite loop or deadlock. If you remove `run: when_changed`, you get what you would expect:
```
task: Maximum task call exceeded (1000) for task "default": probably an cyclic dep or infinite loop
```
but the presence of `run: when_changed` or `run: once` is preventing this check from triggering.

I'm not really sure what the right behavior should be. My first thought was that it there should be proper loop detection rather than reliance on call depth, though I think limiting call depth is still important. But then I thought, well, if `x` only runs once, I might think the sequence should be `default -> x -> default`: `default` tries to run `x`, which tries to run `default`, which doesn't run `x` because it's `run: once`. But of course `x` didn't actually ever run. It only got invoked. So...you could make `run: once` or `run: when_changed` consider the task to have been run once it was invoked even if it hasn't finished, but I feel like maybe proper loop and deadlock detection may be best. I haven't looked at the code, so this is all speculation. Take it or leave it. You know the code. :-)

### Version

3.44.0

### Operating system

multiple

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'

tasks:
  default:
    deps:
      - x

  x:
    run: when_changed
    deps:
      - default
```

trulede:
Execution graph to detect cyclic tasks, replacing the TaskCalledTooManyTimesError mechanism.

Vertices in the graph are added for each unique call (to a task) and the PreventCycles feature of the graph is used to detect cyclic execution. The unique identifying properties are:

* The task name, unique from the Taskfile AST/DAG mechanism.
* The task hash, unique for an "instance" of a task call (e.g. set of vars).
* Call "index" of the task, either from 'deps' or 'cmds'.
* The sources method value, propagated as a call "token". This addresses a special case where a task call is cyclic, but the sources change between each cycle, and thus the task call should continue.

Although this method has worked remarkably well, it could be beneficial to have a mechanism to disable the detection: Environment variable (of course) and perhaps attached to an existing CLI parameter (--force).

fixes https://github.com/go-task/task/issues/820
fixes #2302


trulede:
@korverdev Would you like to try this for you use case mentioned in comments of #820?

Clone the PR and then run task like this:
`go run ./cmd/task -d <your directory> <your task>`

jberkenbilt:
I can confirm that this PR also fixes #2302

errors/errors.go
```diff
@@ -35,10 +35,11 @@ const (
 	CodeTaskRunError
 	CodeTaskInternal
 	CodeTaskNameConflict
-	CodeTaskCalledTooManyTimes
+	CodeTaskCalledTooManyTimes // Depreciated: replaced by CodeTaskCyclicExecutionDetected.
```
mjpieters: Spelling:

```suggestion
	CodeTaskCalledTooManyTimes // Deprecated: replaced by CodeTaskCyclicExecutionDetected.
```

errors/errors_task.go
```diff
@@ -93,23 +93,30 @@ func (err *TaskNameFlattenConflictError) Code() int {
...
-func (err *TaskCalledTooManyTimesError) Error() string {
-	return fmt.Sprintf(
-		`task: Maximum task call exceeded (%d) for task %q: probably an cyclic dep or infinite loop`,
-		err.MaximumTaskCall,
-		err.TaskName,
-	)
+func (err *TaskCyclicExecutionDetectedError) Error() string {
+	if len(err.CallingTaskName) > 0 {
+		return fmt.Sprintf(
+			`task: Cyclic task call execution detected for task %q (calling task %q)`,
```
mjpieters, line 112: I'd simplify the wording a little: `task: Call cycle detected for task %q (calling task %q)`

errors/errors_task.go
```diff
@@ -93,23 +93,30 @@ func (err *TaskNameFlattenConflictError) Code() int {
...
+			err.TaskName,
+			err.CallingTaskName,
+		)
+	} else {
+		return fmt.Sprintf(
+			`task: Cyclic task call execution detected for task %q`,
```
mjpieters, line 118: Ditto: `task: Call cycle detected for task %q`.

errors/errors.go
```diff
@@ -35,10 +35,11 @@ const (
...
 	CodeTaskCancelled
 	CodeTaskMissingRequiredVars
 	CodeTaskNotAllowedVars
+	CodeTaskCyclicExecutionDetected
```
mjpieters, line 42: Might `CodeTaskCallCycleDetected` be a better name?

errors/errors_task.go
```diff
@@ -93,23 +93,30 @@ func (err *TaskNameFlattenConflictError) Code() int {
 	return CodeTaskNameConflict
 }
 
-// TaskCalledTooManyTimesError is returned when the maximum task call limit is
-// exceeded. This is to prevent infinite loops and cyclic dependencies.
-type TaskCalledTooManyTimesError struct {
+// TaskCyclicExecutionDetectedError is returned when the Execution Graph detects
+// a cyclic execution condition.
+type TaskCyclicExecutionDetectedError struct {
```
mjpieters, line 104: Suggestion: I'd use `CallCycle` instead of `CyclicExecution` everywhere:

```suggestion
// TaskCallCycleDetectedError is returned when the Execution Graph detects
// a call cycle.
type TaskCallCycleDetectedError struct {
```

mjpieters:
This PR also fixes #2546

korverdev:
@trulede Sorry about the late follow up. Last year got really crazy for me.

I'm afraid I can't test this on my example. We had to pull the trigger on replacing task with another tool due to this limitation.

Is there anything I can do to get this over the line?

andreynering:
@trulede Mind rebasing this PR?

trulede:
@andreynering I have not thought too much about this PR recently. It seems to work, and a few people have tried it with good results. Possibly I would add a config/env pair to make it possible to disable this, just in case. 

And, I would just take a second look at how the has is generated because some fields of an ast.Task might be ignored, or not as the tests still work.

@korverdev no, I think its fine. The only thing I did not test myself was lots and lots of tasks. I think some other people may have covered that with their testing of this PR.

PR #2291, opened 2025-06-09
Fix install script for armv5/6/7.

### LINKED ISSUE #1516: Install script fails on Raspberry Pi with `go-task/task crit platform linux/arm is not supported`
Trying to use the install script on my Raspberry Pi 3 Model B fails:

```console
admin@rpi:~ $ sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b ~/.local/bin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9634  100  9634    0     0  35814      0 --:--:-- --:--:-- --:--:-- 35814
go-task/task crit platform linux/arm is not supported.  Make sure this script is up-to-date and file request at https://github.com/go-task/task/issues/new
```

---

- Task version: Install script
- Operating system:
   ```console
    $ uname -s
    Linux
    $ uname -m 
    armv7l
    $ lsb_release -a
    No LSB modules are available.
    Distributor ID: Raspbian
    Description:    Raspbian GNU/Linux 11 (bullseye)
    Release:        11
    Codename:       bullseye
   ```
- Experiments enabled: n/a

---

Possibly related to #842

trulede:
The install-task.sh script was converting "armv7*" to "arm" in function `uname_arch()`, checking that "arm" value in `uname_arch_check()`, and then finally using a case condition based on "armv7" in function `get_binaries()`. So of course the download would fail.

[GO Arch only includes linux/arm](https://pkg.go.dev/internal/platform#pkg-variables) variant.

Solution is to fix the `get_binaries()` function.

fixes #1516 

Download tested on PI aarch64, script modified with hardcoded arch as "arm7l". Expected version of task was downloaded.

### DIFF
--- a/install-task.sh
+++ b/install-task.sh
@@ -64,21 +64,15 @@ get_binaries() {
   case "$PLATFORM" in
     darwin/amd64) BINARIES="task" ;;
     darwin/arm64) BINARIES="task" ;;
-    darwin/armv5) BINARIES="task" ;;
-    darwin/armv6) BINARIES="task" ;;
-    darwin/armv7) BINARIES="task" ;;
+    darwin/arm) BINARIES="task" ;;
     linux/386) BINARIES="task" ;;
     linux/amd64) BINARIES="task" ;;
     linux/arm64) BINARIES="task" ;;
-    linux/armv5) BINARIES="task" ;;
-    linux/armv6) BINARIES="task" ;;
-    linux/armv7) BINARIES="task" ;;
+    linux/arm) BINARIES="task" ;;
     windows/386) BINARIES="task" ;;
     windows/amd64) BINARIES="task" ;;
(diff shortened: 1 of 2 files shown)


andreynering (commented):
Thank you!

FYI: We originally used https://github.com/goreleaser/godownloader to generate these scripts, but the tool got deprecated, so I think we needed to update the file manually for fixes.

@trulede Can you replicate these changes on `website/static/install.sh`? We have two copies because originally people fetched from the repo, but nowadays we instruct users to fetch from the website (`http://taskfile.dev/install.sh`).

trulede:
@andreynering yes, I've pushed that change. FWIW you don't need godownloader (or similar), the script is good enough in its current state.

Issue #1246, opened 2023-06-30
When a parallel task fails allow the other parallel tasks to finish before exiting with failure


scheuk:
Currently when a parallel task fails, Taskfile will kill the other running process. 
If you are running a long-running task like terraform, this will break processes and might leave artifacts arround. 

We should have a config or flag to allow the parallel tasks to complete before quitting with error from the failed task. 

this is similar to: 
https://github.com/go-task/task/issues/1226

supersmile2009:
This is definitely a must have. I'm currently considering an alternative to Just (a competing tool) because it doesn't support parallelization. And it's great the Task does support it, however the implementation without waiting for the remaining tasks to complete is not acceptable to me.

I'd also like to mention, that the best approach IMO would be to default to waiting for all the tasks to finish and only kill the remaining tasks (fail fast) when explicitly requested by user (e.g. via a CLI flag or a task definition option). At the end of the day Task is not a build tool like Make, but a task runner. Killing tasks mid-execution can have detrimental effects as already mentioned in the original post (e.g. you likely don't want to kill `terraform apply`), this should not be the default behavior.

This is also how various test frameworks capable of test parallelization usually work, waiting for all the tests to complete before existing and presenting you with the aggregated report unless instructed otherwise. So it's not an entirely new concept. In fact, most of the tools doing some sort of parallelization, in my experience fail fast only when you explicitly ask them to do so.

KuttKatrea:
I agree with @supersmile2009 that the expected behavior is to finish running the deps/parallel tasks.
Fail-fast could be included as other options (like watch) as both a configuration:

```yaml
task-d:
  deps: [ task-a, task-b, task-c]
  failfast: true
```

and a command line option:

```
task --failfast task-d
```

andrewimeson:
My ugly workaround is to touch a file if there's an error, and then at the end exit nonzero if that file exists. That means I have to clean up that file before starting another run, and it's pretty ugly, but it's a better experience than seeing a stacktrace generated from a program that got sigkilled because another task failed.

Here's a simplified example:

```yaml
---
version: 3
env:
  VENV: .venv
  VENV_ACTIVATE: '{{ .VENV }}/bin/activate'
  ANS_ENV: .ansible
  # TODO: Remove this once mvdan's Go sh interpreter auto-sets OSTYPE, which
  # later versions of Python depend upon being set in the activate script, which
  # breaks with `set -u` enforced.
  #   https://github.com/mvdan/sh/pull/912
[...]

PR #2307, opened 2025-06-24
feat: add failfast support for parallel tasks

elvism-lullabot:
Implement new failfast functionality to allow all parallel child tasks to terminate before failing if any fail instead of aborting immediately.

Fixes https://github.com/go-task/task/issues/1246

Example

With this configuration, task-a, task-b, and task-c will all run in parallel. Even though task-a fails early, the others will continue running. The parent will return a non-zero exit only after all child tasks finish.

   ```yaml
   version: '3'

   tasks:
     task-a:
       cmds: [ "bash -c 'echo A; sleep 1; exit 1'" ]
     task-b:
       cmds: [ "bash -c 'echo B; sleep 2; exit 0'" ]
     task-c:
       cmds: [ "bash -c 'echo C; sleep 3; exit 1'" ]

     parent:
       deps: [task-a, task-b, task-c]
       failfast: true
   ```
You can also enable this behavior via the CLI with the new --failfast flag:

```bash
task --failfast parent
```

This flag overrides the YAML config and forces failfast: true for all tasks.


deviantintegral:
Looking forward to seeing how tests go here!

@elvism-lullabot is this defining "fail fast" backwards? To me, `failfast: true` matches the current behaviour of task. If we set it to false, it would mean we'd "fail slow" and let existing tasks run to completion.

anthony-hayes:
Really excited for this. I'm not a maintainer, but I wonder if it would make sense to make `failFast` default to `true`, to preserve the existing behaviour, so this could be shipped quicker (wouldn't be a breaking change)?

rabbitlair:
I have tested these changes and they seem to work as described. Excited for this one to be merged!

vmaerten:
Fixed in https://github.com/go-task/task/pull/2525

Issue #2301, opened 2025-06-18
star globs do not work in generates attribute


superg:
### Description

I am working on setting up a task that generates sqlc source code.
According to documentation in https://taskfile.dev/reference/schema/#task, generates attribute supports star globs when using timestamp method, however this is not what I observe.
First generation with `task sqlc` creates output files in output. After deleting any file that matches `output/*.sql.go`, for example output/asdf.sql.go and rerunning `task sqlc`, the file is not getting regenerated. When instead of star glob I enumerate all output files manually, `task sqlc` works as expected but it's impractical for my case where a have a lot of files.

Expected: 
```
task sqlc
task: [sqlc] sqlc generate -f sqlc.yaml
```

Happened:
```
task sqlc
task: Task "sqlc" is up to date
```

### Version

v3.39.2 (h1:Zt7KXHmMNq5xWZ1ihphDb+n2zYLCo4BdRe09AnMMIgA=)

### Operating system

Ubuntu 24.04

### Example Taskfile

```YAML
version: '3'

tasks:
  sqlc:
    method: timestamp
    sources:
      - sqlc.yaml
      - sqlc/*.sql
    generates:
      - output/*.sql.go
    cmds:
      - sqlc generate -f sqlc.yaml
```

superg:
Just to clarify, it also happens for the latest version:
```
task --version
3.44.0
```

deicon:
Looks like it considers only the presence of the timestamp file but does not compare it with any other timestamp.

So after deleting every file within output folder of your example, it still thinks its up to date. 



`go
	timestampFile := checker.timestampFilePath(t)

	// If the file exists, add the file path to the generates.
	// If the generate file is old, the task will be executed.
	_, err = os.Stat(timestampFile)
	if err == nil {
		generates = append(generates, timestampFile)
	} else {
		// Create the timestamp file for the next execution when the file does not exist.
		if !checker.dry {
			if err := os.MkdirAll(filepath.Dir(timestampFile), 0o755); err != nil {
				return false, err
			}
			f, err := os.Create(timestampFile)
			if err != nil {
				return false, err
			}
			f.Close()
		}
	}

`

deicon:
To Reproduce

```
version: '3'

tasks:
  default:
    method: timestamp
    sources:
      - sqlc.yaml
      - sqlc/*.sql
    generates:
      - output/sqlc/*.sql
    cmds:
      - for: sources
[...]

Issue #2181, opened 2025-04-15
misleading feature: "generates" and checksum checker


viktorasm:
### Description

It's very misleading how `task.generates` feature works.

One would assume that the task will run if combined checksum of sources and outputs does not match sources/results of previous execution, e.g. having such definition

```yaml
sources:
   - openapi.yaml
generates:
   - client/go/**/*.go
```

.. One would expect that if some generated files are missing, or are unintentionally changed, this will cause task to rerun.

How it actually works with checksum checker:
* Checksum is calculated and checked for `sources`;
* For `generates` list, it's only verified that glob matches any files. 

This means that "generates" is only somewhat usable with absolute file names (globbing is basically useless).

Probably the best fix would be to actually include generated files into checksum calculation.

### Version

main branch

### Operating system

all

### Example Taskfile

```YAML

```

ameergituser:
This is really one of those features that is REALLY required.  For us to live up to the "simpler than Make" statement. The basic feature of only executing when required needs to be supported.

My previous issue: #1741 

My last statement: https://github.com/go-task/task/issues/1741#issuecomment-2437013900

Other related issues: #1945

Related PR: #1816

hanoii:
Also I think is useful if one can simple tell generates about a directory, i.e:

```yml
tasks:
  build:
    dir: web/theme
    cmds:
      - n auto
      - npm i -g yarn
      - yarn install
      - npx mix --production
    sources:
      - '*.lock'
    generates:
      - dist # or - dist/

```

PR #2313, opened 2025-07-02
include generated files in the checksum method

superg:
This PR includes generated files into the calculated checksum to trigger task only when needed.

Namely, that implements:
https://github.com/go-task/task/issues/2181
and closes:
https://github.com/go-task/task/issues/2301 (original bug report for timestamp method but this PR will allow to use default checksum method and preserves original timestamp method behavior)

Example Taskfile.yml:
```
version: '3'

tasks:
  default:
    sources:
      - input/*.in
    generates:
      - output/*.out
    cmds:
      - mkdir -p output
      - cp input/one.in output/one.out
      - cp input/two.in output/two.out
```

- initial task run will generate one.out and two.out
- subsequent task run will report task is up to date
- rename of any *.out file and running task will generate the files again
- removal of any *.out file and running task will generate the files again

### DIFF
--- a/internal/fingerprint/sources_checksum.go
+++ b/internal/fingerprint/sources_checksum.go
@@ -15,7 +15,7 @@ import (
 )
 
 // ChecksumChecker validates if a task is up to date by calculating its source
-// files checksum
+// and destination (generates) files checksum
 type ChecksumChecker struct {
 	tempDir string
 	dry     bool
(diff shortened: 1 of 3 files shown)


superg:
Closing this, I lost any interest and moved on.
It's hard to imagine a build system that is comparing itself to GNU Make and at the same time doesn't implement a core functionality that is make: define rules that specify how to create target files from source files.

andreynering:
Hi @superg.

I know we took a long time to review, but we're volunteers doing this on free time. This is not our full time job. 🙂 

With regard to the feature itself, The existing behavior is somewhat by design. We still check if the `generates` files were all deleted and run again in that case.

I'll add this to my TODO list to maybe review soon.

Issue #2309, opened 2025-06-26
Fuzzy suggestions causes performance degredation


maxmzkrcensys:
### Description

We are seeing a large performance difference with the fuzzy suggestions commit on a large project. We have 1006 tasks in our project right now and are seeing a 17.554s increase in runtime

```
$ gco a33544101a32fc6cd9dc23189077fa7aa76ae97f~1          
Previous HEAD position was a3354410 fix: fuzzy model was not instanciated (#2200)
HEAD is now at 1c35358f v3.43.3
$ time go run ./cmd/task -l -d ~/pipeline | wc -l
1006
go run ./cmd/task -l -d ~/pipeline  13.59s user 1.52s system 1327% cpu 1.138 total
wc -l  0.00s user 0.00s system 0% cpu 1.137 total
$ gco a33544101a32fc6cd9dc23189077fa7aa76ae97f
HEAD is now at a3354410 fix: fuzzy model was not instanciated (#2200)
$ time go run ./cmd/task -l -d ~/pipeline | wc -l 
1006
go run ./cmd/task -l -d ~/pipeline  83.61s user 19.30s system 550% cpu 18.692 total
wc -l  0.00s user 0.01s system 0% cpu 18.691 total
```

```
diff --git a/cmd/task/task.go b/cmd/task/task.go
index 31538afc..b97b8a5a 100644
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -5,6 +5,7 @@ import (
        "fmt"
        "os"
        "path/filepath"
+       "runtime/pprof"
 
        "github.com/spf13/pflag"
 
@@ -42,6 +43,15 @@ func main() {
 }
 
 func run() error {
+       // write profile to a file
+       f, err := os.Create("/home/max/task/cpu.prof")
+       if err != nil {
+               return fmt.Errorf("failed to create CPU profile file: %w", err)
+       }
+       defer f.Close()
+       pprof.StartCPUProfile(f)
+       defer pprof.StopCPUProfile()
+
        log := &logger.Logger{
                Stdout:  os.Stdout,
                Stderr:  os.Stderr,
```

yields

![Image](https://github.com/user-attachments/assets/346d859f-4c9b-4fd8-81f9-903865d86aec)

### Version

v3.44.0

### Operating system

Linux

### Example Taskfile

```YAML

```

trulede:
[...]

PR #2323, opened 2025-07-09
fix: don't suggest internal tasks

maxmzkrcensys:
fixes https://github.com/go-task/task/issues/2309

Internal tasks cannot be ran from the command line, so it doesn't make sense to suggest them when fixing typos on the command line

Without the changes, the added test fails with

```
    executor_test.go:974:
                Error Trace:    /home/max/task/executor_test.go:205
                Error:          Error message not equal:
                                expected: "task: Task \"intern\" does not exist"
                                actual  : "task: Task \"intern\" does not exist. Did you mean \"internal\"?"
                Test:           TestFuzzyModel/intern
```

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -957,6 +957,15 @@ func TestFuzzyModel(t *testing.T) {
 		),
 		WithTask("install"),
 	)
+
+	NewExecutorTest(t,
+		WithName("intern"),
+		WithExecutorOptions(
+			task.WithDir("testdata/fuzzy"),
+		),
+		WithTask("intern"),
+		WithRunError(),
+	)
 }
 
 func TestIncludeChecksum(t *testing.T) {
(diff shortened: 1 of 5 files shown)


executor_test.go
```diff
@@ -957,6 +970,15 @@ func TestFuzzyModel(t *testing.T) {
...
 		WithTask("install"),
 	)
+
+	NewExecutorTest(t,
+		WithName("intern"),
+		WithExecutorOptions(
+			task.WithDir("testdata/fuzzy"),
+		),
+		WithTask("intern"),
+		WithRunErrorMessage("task: Task \"intern\" does not exist"),
```
pd93: I don't think we need this new option. `WithRunError` should already create a fixture containing the error message. However, generating the fixtures is a bit annoying until after #2265 is merged. I suggest updating as per the suggestion below and then running `task gen:fixtures` on the project. You can then just discard the change to `TestJsonListFormat.golden` and commit everything else.

```suggestion
		WithRunError(),
```

Once #2265 is merged, running `task gen:fixtures` will no longer create the change in `TestJsonListFormat.golden`.

maxmzkrcensys: Ah! Easy peasy. Thanks! Updated

PR #2338, opened 2025-07-26
docs: option to add taskfile via install-release

Rishang:
I have added a extra option of installing taskfile via [install-release](https://github.com/Rishang/install-release) which is a package manager created in order to install any single binary tools directly from their github release based on the linux/macos distro with support to upgrade/hold or remove installed package.

Repo doc: https://github.com/Rishang/install-release?tab=readme-ov-file#table-of-contents-


andreynering:
Hi @Rishang,

Thanks for you interest in contributing, but looks like this package manager is not too popular yet, so I don't think we should list it yet. Maybe in the future if it gets more widely used.

PR #2344, opened 2025-07-31
Implement Starlark parsing for Taskfiles

rrdelaney:


andreynering:
Hey @rrdelaney, thanks for your love about the project!

Our previous experiences having extra formats weren't too good. In the beginning, I accepted contributions to add support for JSON and TOML, but I soon regretted it once I noticed the maintanence overhead that it'd add to the project in the long run. I then decided to remove that support sooner than later until too many people started relying on that.

So, I'm also going to close this sooner than later before you expend more time on it.

If you really want to use Starlark, I think a possible path is a build an external tool that generates Taskfiles.

Or... [you can also import Task as a library](https://taskfile.dev/reference/package) if you want.

Issue #2345, opened 2025-07-31
Switch statements


dakom:
### Description

There is already an open issue for `if` statements at https://github.com/go-task/task/issues/608. This can be considered an alternative proposal since it would cover that use-case as well, but it really speaks to a different use-case of branching on multiple variants of a single test. That's a bit wordy, let's look at an example :)

Imagine we have `if` and we want to run different tasks depending on a `DEPLOY_ENV` var. We could model it like this:

```
tasks:
  deploy:
    deps: [deploy-prod, deploy-staging, deploy-dev]

  deploy-prod:
    if:
      - sh: '[ "{{.DEPLOY_ENV}}" = "prod" ]'
    cmds:
      - echo "Deploying to production"
    
  deploy-staging:
    if:
      - sh: '[ "{{.DEPLOY_ENV}}" = "staging" ]'
    cmds:
      - echo "Deploying to staging"

  deploy-dev:
    if:
      - sh: '[ "{{.DEPLOY_ENV}}" = "dev" ]'
    cmds:
      - echo "Deploying to development"
```

However, there's a few problems:

1. The overall logic is spread among a bunch of tasks, it's hard to see what's happening as we add more
2. It's too easy for the subtasks to add their own additional logic, which can drastically impact the ability to reason about what's going on
3. There's no exhaustiveness checks.

In other words, `if` works well enough for one or two small "branches", but it gets too unwieldy at scale.

What if instead, we had `switch`, and it looked like this:

```
tasks:
  requires:
      vars: [DEPLOY_ENV]
  deploy:
    - switch "{{.DEPLOY_ENV}}"
      - case "prod"
        - cmds:
          - echo "Deploying to prod"
      - case "staging"
        - cmds:
          - echo "Deploying to staging"
      - case "dev"
        - cmds:
          - echo "Deploying to dev"
      - default
        - cmds:
          - echo "Unknown deploy target {{.DEPLOY_ENV}}"
```

Solves all those problems:
1. It's all located together
2. The case matching is done on only the specified switch, no room to introduce hidden logic
3. Exhaustiveness checks with `default`

Much cleaner imho!

(btw, just for the sake of completeness, this _exact_ problem could be solved by just having subtasks and running `task deploy-{{.DEPLOY_ENV}}` but it's just an example, real-world problems aren't always so trivial)

trulede:
What about:

```
tasks:
  requires:
      vars: [DEPLOY_ENV]
  deploy:
    tasks:
      deploy-prod:
[...]

PR #2346, opened 2025-08-01
Defer tasks until a later named point.

### LINKED ISSUE #1082: Feature Request: Dependency Cleanup
Hello, and thanks for this great tool!

I was writing a task today and wished for a better way to express it:
- I want to execute a command to generate documentation
- This commands depends on a background service to be running. (e.g., `docker-compose up ...`)
- I want the service to stop once it is no longer necessary (e.g., `docker-compose down ...`)

(For context, the service in question takes a lot of RAM, so I do want to stop it when done)

What I have right now is something like this:

```
version: '3'

tasks:

  start-docker-service:
    cmds:
      - docker compose up

  stop-docker-service:
    cmds:
      - docker compose down

  default:
    cmds:
      - task: start-docker-service
      - defer: {task: stop-docker-service}
      - echo "Doing stuff"

```

This works well enough, but the "docker-service" would be better expressed as a dependency. It should also be possible to have multiple concurrent tasks that depend on this.

I think it would be great if we could do something like this:

```
version: '3'

tasks:

  with-docker-service:
    cmds:
      - docker compose up ...
      - defer-dep: docker compose down ...  # similar to `defer`, but only executes after all tasks that depend on it are complete

  default:
    deps:
      - with-docker-service
    cmds:
      - echo "Doing stuff"
```

`defer` is pretty close to what I want, but it only waits for the next commands, not the dependents tasks. I guess we need a new keyword (Or modifying the semantics of defer, which sounds more problematic)

### LINKED ISSUE #2040: Proposal: `defer:` availability at task level
### Description

I have similar use-cases as reported in this [feature request](https://github.com/go-task/task/issues/1082) (running post-processing tasks at the very end of the overall workflow), but I would like to propose a different approach for handling such case; as it may have a slightly different purpose / implementation, I created a distinct report, but close it if it seems like a duplicate.

The proposal would be to add support for the `defer:` keyword at task level, with the exact same behaviour as for commands; defining a symmetric to the `deps:` feature. For instance,
```yaml
version: 3

tasks:
  pre:
    internal: true
    run: once
    defer:
      - post
    cmds:
      - pre-cmd

  a:
    deps:
      - pre
    cmds:
      - a-cmd

  b:
    deps:
      - pre
    cmds:
      - b-cmd

  post:
    internal: true
    run: once
    cmds:
      - post-cmd
```
when invoked via `task a b`, would result in running `pre` then `a` then `b` and finally `post`.

For now, the proposed syntax has no effect.

### LINKED ISSUE #2273: Idea: Add a global `defer:` attribute to cleanup stuff before exiting `task`?

trulede:
This PR makes it possible to defer a command to a "named point" such as `exit` (i.e. when task itself is exiting) using a new `when` keyword. For example:

```yaml
version: '3'
vars:
  GREETING: Hello, World!
tasks:
  default:
    cmds:
      - defer:
          task: when-message
          when: exit
      - defer:
          task: defer-message
      - task: message
  message:
    cmds:
      - echo "MESSAGE - {{.GREETING}}"
  defer-message:
    cmds:
      - echo "DEFER - {{.GREETING}}"
  when-message:
    run: once
    cmds:
      - echo "WHEN - {{.GREETING}}"
```

which produces this output (note the order of the messages):

```bash
$ go run ../cmd/task/task.go -s
MESSAGE - Hello, World!
DEFER - Hello, World!
WHEN - Hello, World!
```

The "when" mechanism operates in a method similar to a "pub/sub". Defered tasks a pushed to a named when list, and then pulled at some later point. Only the `exit` tasks are pulled and executed at the moment, however it would be possible to trigger the execution of tasks contained in a named list - such a mechanism only needs to be defined (e.g. a tempate function).

fixes #2273 
fixes #2040 
fixes #1082 
#141 - partial, would need addition, for instance defer/if (see https://github.com/go-task/task/issues/2345#issuecomment-3145600773)

### DIFF
--- a/call.go
+++ b/call.go
@@ -8,4 +8,5 @@ type Call struct {
 	Vars     *ast.Vars
 	Silent   bool
 	Indirect bool // True if the task was called by another task
+	When     string
 }
--- a/executor.go
+++ b/executor.go
@@ -71,10 +71,16 @@ type (
 		executionHashes      map[string]context.Context
 		executionHashesMutex sync.Mutex
 		watchedDirs          *xsync.MapOf[string, bool]
+		whenTasks            map[string][]WhenTaskCall
 	}
 	TempDir struct {
 		Remote      string
 		Fingerprint string
+	}
+	WhenTaskCall struct {
+		Task  *ast.Task
+		Call  *Call
+		Index int
 	}
 )
 
@@ -98,6 +104,7 @@ func NewExecutor(opts ...ExecutorOption) *Executor {
 		mkdirMutexMap:        map[string]*sync.Mutex{},
 		executionHashes:      map[string]context.Context{},
 		executionHashesMutex: sync.Mutex{},
+		whenTasks:            map[string][]WhenTaskCall{},
 	}
 	e.Options(opts...)
 	return e
--- a/task.go
+++ b/task.go
@@ -231,8 +231,6 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
 				continue
 			}
 
-			// TODO push to exit defer list
-
 			if err := e.runCommand(ctx, t, call, i); err != nil {
 				if err2 := e.statusOnError(t); err2 != nil {
 					e.Logger.VerboseErrf(logger.Yellow, "task: error cleaning status on error: %v\n", err2)
--- a/task.go
+++ b/task.go
@@ -97,6 +97,15 @@ func (e *Executor) Run(ctx context.Context, calls ...*Call) error {
 		return e.watchTasks(watchCalls...)
 	}
 
+	if whenList, ok := e.whenTasks["exit"]; ok {
+		for _, w := range whenList {
+			if err := e.runCommand(ctx, w.Task, w.Call, w.Index); err != nil {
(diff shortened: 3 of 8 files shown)


trulede:
If this PR is interesting, I can develop further (tests, doc ... extensions for #141 ... etc).

Issue #2091, opened 2025-02-26
List experiment


pd93:
## Context

This is a proposal for a new API for the `--list` flag. It aims to unify the API for outputting tasks in a more consistent and intuitive way.

Obviously, we cannot break the API for the existing `--list` flag in v3, so I am proposing that the new API go behind an experimental (i.e. `TASK_X_LIST`). This will allow us to iterate and gather feedback on the design without worrying about breaking anything for existing users. The flag should come with a warning to users that it does not have a backward-compatible guarantee and should not be relied upon for scripts/CI. If the experiment is successful, then we could replace the existing `--list` and `--list-all` flags in a new major version release.

## Design

This proposal has 3 main aims:

- Consolidate the output of tasks into a single `--list` flag.
  - Any other flags related to the output should be implemented as "subflags" that supplement the `--list` flag.
- Make the design more intuitive and consistent with other well-known CLIs.
- Tidy up the code responsible for output and decouple it from the executor. This make it easier to:
  - Add new filters and formats
  - Write functional unit tests.

The following subsections describe the proposed API design in more detail:

### Tags

`tags` is a new field that can be added to a task. It allows the user to attach some arbitrary metadata to a task that can then be used later for filtering and grouping. Multiple tags can be specified per task:

```yaml
version: 3

tasks:
  foo:
    desc: 'Task foo'
    tags: [foo, bar]
  bar:
    desc: 'Task bar'
    tags: [foo]
```

### Filtering (`--filter`)

Filtering allows the user to filter the list of tasks using task attributes. We can do this by supporting some basic wildcard functionality (or maybe regex?). The flag can be supplied multiple times to add more filters. These filters are applied in the order specified. `!` can be prefixed to a filter to negate its effect (i.e. it becomes an exclusion). Initial supported filtering attributes are:

- `name`
- `desc`
- `tags`

Example usage:

- `task --list --filter name=foo*` - Only lists tasks that start with "foo".
  - "foo1" and "foo2" will be included.
- `task --list --filter name={foo,bar}*` - Only lists tasks whose names start with "foo" or "bar".
  - "foo1" and "foo2" will be included.
  - "bar1" and "bar2" will be included.
- `task --list --filter !desc=` - Removes tasks with no description (matches the current `--list` behavior).
  - We could also allow for the shorthand `--filter !desc` for this.
- `task --list --filter tags=foo` - Only lists tasks with the tag "foo".

### Sorting (`--sort`/`--sort-by`)

[...]

Issue #931, opened 2022-11-12
Add `--debug` (or similar) flag to output debugging metadata about the Taskfile


andreynering:
Sometimes users need to debug something on their Taskfiles, and we could have a flag to output information that could help them:

- Computed/expanded variables
- Computed/expanded env
- Computed/expanded sources and envs (#924)

There may be some intersection between this and #764 / #884, so this will probably be an extension of `--json` with extra data that would be useful for debugging but not for the editor. Because of this, it may be more intelligent to implement `--json` before this one (`--debug`).

trulede:
The PR #2617 might be a good way to realise this, especially the TASK log level that is intended to provide structured information about task itself  - for integration with other tools, more advanced use cases.

trulede:
Additionally, PR #2739 introduces variable debugging.

ssbarnea:
A more generic partern would be to allow multiple `-v` options for increasing the verbosity. This would allow fine tuning the verbosity.

Usually a `-vv` would translate to debug mode and `-vvv` would translate to trace mode (extremly verbose).

PR #2348, opened 2025-08-02
Add templating support to CLI flags list/list-all

### LINKED ISSUE #1695: Generate documentation based on your taskfile
We have option for displaying summary, also we configure input variables. I we use multiple task modules and sharing between teams, document generation tool can really helpout to have better visibility on how to use the task module one created.

I created my own python code to generate docs: https://github.com/Rishang/task-templates/blob/main/generate_docs.py

And here is how the docs look: https://github.com/Rishang/task-templates/blob/main/docs/_aws.md

### LINKED ISSUE #2261: Option to disable output of commands in summary
### Description

Hi, I read the issue #1350 and PR #1656. You have chosen summary to output multi-line description of a task, along with the commands inside the task.

My use case is to document tasks which have parameters. Currently `--summary` produces way more output than I wish it was exposed to someone trying to understand (or to remember) how to use my taskfiles.

For example, I have this field:

```yaml
  terragrunt_apply:
    desc: |
      Executes the terraform apply command.

      CI: executes terraform in non-interactive way. Default: false
      PLAN: executes terraform plan before apply. Default: false
```

Which renders this way with `task --list`:

```text
$ task
* terragrunt_apply:         Executes the terraform apply command.  CI: executes terraform in non-interactive way. Default: false PLAN: executes terraform plan before apply. Default: false 
```

`--list` is interesting because it list all tasks, but it collapses multi-line. `--summary` respect multi-line, but it only works one command at a time, and I can't control its output to only show the summary field's content, it will show also the task commands.

I tried to manipulate `--list` output, adding `\n` to the text inside description field and throwing the result inside `printf`, something like this:

```yaml
tasks:
  default:
    cmds:
      - printf "$(task --list | sed 's/\\n/\\n\\t/g')"
```
Which produced:

```text
* terragrunt_apply:         Executes the terraform apply command.
          CI: executes terraform in non-interactive way. Default: false
         PLAN: executes terraform plan before apply. Default: false
```

But I don't want to go this path, I would rather rely on task to make this.. So can you help this use case?

Thanks in advance!

timrulebosch:
This PR is adding template support for CLI commands --list / --list-all. The implementation adds an optional flag (--template) which is used to specify a template file(s). When specified, the template files(s) are loaded and rendered. The templating mechanism makes use of the existing slim-sprig functions available in Task.

Primary motivation for the PR is to generate documentation for a library of Taskfiles. Some of the documentation can be extracted from the existing Taskfile schema, and other elements of the documentation are embedded in the Taskfiles. This later embedded documentation ([example here](https://github.com/boschglobal/dse.fmi/blob/main/Taskfile.yml)) would be made available to the templating mechanism as well (PR in development).

As it currently stands, the PR replaces the `--list` code (a for loop) with the templating mechanism.

### mddoc.template
```
# Taskfile
Available tasks for this project:
{{ range . }}
## {{.Task}}
{{.Desc}}
{{if len .Aliases}}
### Aliases
{{range .Aliases}}- {{.}}
{{end}}{{end}}{{end}}
```

### Console Output
```bash
$ task --list-all --template mddoc.template
# Taskfile
Available tasks for this project:

## clean
Cleans temp files and folders

### Aliases
- clear

## default


## format
Runs golangci-lint and formats any Go files

### Aliases
- fmt
- f

## generate
Runs all generate tasks

### Aliases
- gen
- g
```

## Notes:

fixes #1695
fixes #2261
related #1916
related #931 (perhaps)
related #2091

Outstanding items needed to complete PR:
* Refinement of operation - e.g. relationship between list list-all and template CLI options.
* Documentation with examples.
* Application to the --summary command?
* Support template definition in the taskrc file.?

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -128,6 +128,7 @@ func run() error {
 		flags.ListAll,
 		flags.ListJson,
 		flags.NoStatus,
+		flags.Template,
 	)
 	if listOptions.ShouldListTasks() {
 		if flags.Silent {
--- a/help.go
+++ b/help.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"html/template"
 	"io"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/Ladicle/tabwriter"
 	"golang.org/x/sync/errgroup"
+
+	sprig "github.com/go-task/slim-sprig/v3"
 
 	"github.com/go-task/task/v3/internal/editors"
 	"github.com/go-task/task/v3/internal/fingerprint"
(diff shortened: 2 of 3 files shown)


trulede:
It occurs to me that the `listDefaultTemplate` could also be optionally loaded from the taskrc file which would allow for easier customisation of `--list`/`--list-all`/`--summary` behaviour, such as mentioned in #2486.

Probably something like:

```yaml

templates:
  cli:
    list: >
      task: Available tasks for this project:{{ range . }}
      {{color "Yellow"}}* {{color "Green"}}{{ .Task }}{{color "Reset"}}:{{"\t"}} 
      {{ .Desc | replace "\n" " " }}{{if len .Aliases}}{{color "Cyan"}}{{"\t"}}(aliases: {{.Aliases | join ", "}})
      {{color "Reset"}}{{end}}{{end}}
    list-all: >
      task ....
    summary: >
      task ...
```

adriens:
That would indeed help implrement things like that:
- https://github.com/go-task/task/issues/2503

adriens:
You inspired me @trulede 

- https://github.com/adriens/task-gomplates

PR #2350, opened 2025-08-03
Fix concurrent writes to stdout in tests.

### LINKED ISSUE #1208: Grouped output is mixed with Task output
Hello, when using grouped output for the tasks, it seems like the Tasks' logs are mixed within the output, in the example below you can see that `task: Task "one" is up to date` is being printed within the task group and the output is different with every run, sometimes this statement is printed is before the group statement, sometimes it is after the closing group statement. I would expect this to be consistent across the runs or atleast outside of the group output.

Please run the Taskfile below multiple times to reproduce.

```
::group::two
task: Task "one" is up to date
2
::endgroup::
```

- Task version: 3.25.0
- Operating system: MacOS
- Experiments enabled: 0

```yaml
version: '3'

output:
  group:
    begin: '::group::{{.TASK}}'
    end: '::endgroup::'

tasks:
  default:
    deps: [one,two]
    cmds:
    - task: one
    - task: two
    - task: one

  one:
    sources: [Taskfile.yml]
    cmds: [sleep 1]
  two: echo 2
```

```
~/sandbox/task-test ❯ task
task: [two] echo 2
::group::two
task: Task "one" is up to date
2
::endgroup::
task: Task "one" is up to date
task: [two] echo 2
::group::two
2
::endgroup::
task: Task "one" is up to date

~/sandbox/task-test ❯ task
task: [two] echo 2
::group::two
2
::endgroup::
task: Task "one" is up to date
task: Task "one" is up to date
task: [two] echo 2
::group::two
2
::endgroup::
task: Task "one" is up to date

~/sandbox/task-test ❯ task
task: [two] echo 2
task: Task "one" is up to date
::group::two
2
::endgroup::
task: Task "one" is up to date
task: [two] echo 2
::group::two
2
::endgroup::
task: Task "one" is up to date
```

### LINKED ISSUE #2349: Occasional flaky tests with grouped output.
### Description

There is an occasional flaky test result. It seems like the output of task is captured with split lines, that is to say, a partial line is written as one line, and then the rest of that line is written as a subsequent line. This output, when sorted, the causes the goldie check to fail.

For instance, in the following example, the single line "d3\n" seems to be "d3" and "\n". As a result the check fails.

```
--- FAIL: TestDeps (0.14s)
      executor_test.go:401: Result did not match the golden fixture. Diff is below:
          
          --- Expected
          +++ Actual
          @@ -1,2 +1,2 @@
          -d1
          +
           d11
          @@ -8,3 +8,2 @@
           d23
          -d3
           d31
          @@ -12,2 +11,3 @@
           d33
          +d3d1
           
          
  FAIL
  FAIL	github.com/go-task/task/v3	6.995s
```

Its possible that the function https://github.com/go-task/task/blob/e72c35f79ffd32856efe3226ff9aadcd78d0804c/task_test.go#L304 is not processing the output correctly. It may first need to join lines without a trailing '\n' to the following line, and then apply the sorting algorithm. However, this is not proven, and the output "d3d1\n" does seem odd. In theory, with grouped output, such a combination should not be possible, right? Therefore I wonder if the post processing (`PPSortedLines()`) causes the issue.

I observe this problem most often with the "deps" test: https://github.com/go-task/task/blob/e72c35f79ffd32856efe3226ff9aadcd78d0804c/testdata/deps/Taskfile.yml#L1

### Version

v3.44.1

### Operating system

Ubuntu

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

trulede:
Writes to stdout from `deps` are occasionally "mixed up" because of stdout buffering. To prevent this, the _entire_ grouped output is written with _one_ call to `gw.buffer.WriteString()` rather than the previous two calls.

Additionally, to reduce flaky tests, use a SyncBuffer which prevents interleaving of stdout/err (typically from `deps` which run in parallel).

fixes #2349
fixes #1208

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -144,12 +144,12 @@ func (tt *ExecutorTest) run(t *testing.T) {
 	t.Helper()
 	f := func(t *testing.T) {
 		t.Helper()
-		var buf bytes.Buffer
+		var buffer SyncBuffer
 
 		opts := append(
 			tt.executorOpts,
-			task.WithStdout(&buf),
-			task.WithStderr(&buf),
+			task.WithStdout(&buffer),
+			task.WithStderr(&buffer),
 		)
 
 		// If the test has input, create a reader for it and add it to the
@@ -172,7 +172,7 @@ func (tt *ExecutorTest) run(t *testing.T) {
 		if err := e.Setup(); tt.wantSetupError {
 			require.Error(t, err)
 			tt.writeFixtureErrSetup(t, g, err)
-			tt.writeFixtureBuffer(t, g, buf)
+			tt.writeFixtureBuffer(t, g, buffer.buf)
 			return
 		} else {
 			require.NoError(t, err)
@@ -193,7 +193,7 @@ func (tt *ExecutorTest) run(t *testing.T) {
 		if err := e.Run(ctx, call); tt.wantRunError {
 			require.Error(t, err)
 			tt.writeFixtureErrRun(t, g, err)
-			tt.writeFixtureBuffer(t, g, buf)
+			tt.writeFixtureBuffer(t, g, buffer.buf)
 			return
 		} else {
 			require.NoError(t, err)
@@ -206,7 +206,7 @@ func (tt *ExecutorTest) run(t *testing.T) {
 			}
 		}
 
-		tt.writeFixtureBuffer(t, g, buf)
+		tt.writeFixtureBuffer(t, g, buffer.buf)
 	}
 
 	// Run the test (with a name if it has one)
--- a/internal/output/group.go
+++ b/internal/output/group.go
@@ -44,10 +44,15 @@ func (gw *groupWriter) close() error {
 		// don't print begin/end messages if there's no buffered entries
 		return nil
 	}
-	if _, err := io.WriteString(gw.writer, gw.begin); err != nil {
-		return err
+	if len(gw.begin) > 0 {
+		// Rewrite the gw.buff with the beginning text.
+		s := gw.buff.String()
+		gw.buff.Reset()
+		gw.buff.WriteString(gw.begin)
+		gw.buff.WriteString(s)
 	}
 	gw.buff.WriteString(gw.end)
+	// Return the entire gw.buff to ensure the group is written atomically to stdout.
 	_, err := io.Copy(gw.writer, &gw.buff)
 	return err
 }


task.go
```diff
@@ -353,9 +357,17 @@ func (e *Executor) runCommand(ctx context.Context, t *ast.Task, call *Call, i in
 			Stdout:    stdOut,
```
andreynering, line 353: Hey @trulede,

I think this won't work as you expect because you forgot to replace here, right?

```suggestion
			Stdout:    _stdOut,
```

That said, the whole change will actually change the Task behavior in an unintended way.

By default, Task assumed `output: interleaved`, which means that if you have multiple long-running command being run, you'll see mixed lines being printed. There's `output: group` for those who want to see the full output of each command only once it finished running.

This change will essentially break `interleaved` because the output will only be printed once the command finished, but that's not what we want. For a HTTP server for example, we want to see logs as soon as the app prints it.

So the solution for the flacky test will be something completely different in my view.

trulede: @andreynering The problem itself is a simple data race, so I think there will be an equally simple solution, probably using a mutex. It (the problem) is especially prevalent in the code area addressed by the PR because the grouped output is written to _stdout_ without protection of a mutex and the effect (i.e.... the data race) is particularly observable when tasks run in parallel (e.g. deps).

I think you are right about interleaved output. I guess there is no unit test covering that, right?

Let me take a second look, I have an idea that we can put mutex protection in a Writer class, and then "install" that so it wraps whatever _stdout_ really is. Then we get the solution for all use cases (interleaved, grouped ...).

trulede: @andreynering It turned out there was already a SyncBuffer in the testing framework which can be used. Therefore I reverted the change and switched the Executor tests to use the SyncBuffer _always_. Why _always_? This is because the data race there was really bad in some cases - I wrote a test for interleaved output which simply failed _all_ the time. 

I've started to have problems with another test. I'm not sure why, so I have pushed the latest code ... mostly to see how the CI responds. I still want to run more test cycles to confirm the improved behaviour.

EDIT: OK, the CI is better but one of the new tests was not reliable. I have to look a bit more on that.

@pd93 Do you have some comments about the using SyncBuffer this way?

trulede: @andreynering summary:

- Use SyncBuffer in tests, fixes flaky tests.
- Fix the issue causing #1208.
- Write a test for `output` to capture the effects of each setting.

I've run 100's of the new tests without failure, and those that were flaky before. When I introduce the original problems, it does not take more than a few test cycles for errors to occur. However only more tests on more systems can really prove the changes are durable.

I have no access to a `darwin` system to figure out why the new tests failed there. So I skip the test in that case.

It seems like the task writes to os.Stdout are "thread-safe" (100's of tests run). When I check these things I used a modified pattern in the ping/pong scripts which makes observing the issue easier:

```
::endgroup::
task: [bar] bash pong.sh
task: [foo] bash ping.sh
::group::bar
---------------------------- 1
---------------------------- 2
---------------------------- 3
---------------------------- 4
---------------------------- 5
---------------------------- 6
---------------------------- 7
---------------------------- 8
---------------------------- 9
---------------------------- 10
::endgroup::
::group::foo
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 1
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 2
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 3
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 4
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 5
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 6
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 7
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 8
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 9
XXXXXXXXXXXXXXXXXXXXXXXXXXXX 10
::endgroup::
task: [bar] bash pong.sh
task: [foo] bash ping.sh
::group::bar
```

trulede:
I will have to tag the tests with "signals" (or delete them), its too difficult to get them running reliably in the CI.

andreynering:
Thank you @trulede!

I pushed a follow-up commit to tidy up the code a bit, but your diagnose was correct!

* 43074c20f205f084650540ed9711e069ee655507

PR #2359, opened 2025-08-07
docs: migrate website to vitepress

vmaerten:


Taskfile.yml
```diff
@@ -5,6 +5,10 @@ includes:
     aliases: [w, docs, d]
     taskfile: ./website
     dir: ./website
+  website_vitepress:
+    aliases: [ wp ]
+    taskfile: ./website_vitepress
+    dir: ./website_vitepress
```
pd93: This taskfile doesn't appear to exist?

website/TODO.md
pd93, line 1: We could put this in the PR description so it doesn't get accidentally merged

website/src/team.md
```diff
@@ -0,0 +1,58 @@
...
+    ]
+  },
+  {
+    avatar: 'https://www.github.com/pd93.png',
+    name: 'Pete Davison',
+    title: 'Maintainer',
+    sponsor: 'https://github.com/sponsors/pd93',
+    links: [
+      { icon: 'github', link: 'https://github.com/pd93' },
+      { icon: 'x', link: 'https://x.com/youyuxi' }
```
pd93: No need to keep my Twitter, I will likely never use it.

```suggestion
      { icon: 'bluesky', link: 'https://bsky.app/profile/pd93.uk' }
```

website/src/team.md
```diff
@@ -0,0 +1,58 @@
...
+
+const members = [
+  {
+    avatar: 'https://www.github.com/andreynering.png',
+    name: 'Andrey Nering',
+    title: 'Creator & Maintainer',
+    sponsor: 'https://github.com/sponsors/andreynering',
+    links: [
+      { icon: 'github', link: 'https://github.com/andreynering' },
+      { icon: 'x', link: 'https://x.com/andreynering' }
```
pd93: I assume @andreynering will want these updating too.

website/src/team.md
```diff
@@ -0,0 +1,58 @@
...
+</script>
+
+<VPTeamPage>
+  <VPTeamPageTitle>
+    <template #title>
+      Our Team
+    </template>
+    <template #lead>
+      The development of VitePress is guided by an international
+      team, some of whom have chosen to be featured below.
```
pd93: Something related to Task instead of VitePress. Probably along the lines of us being amazing for donating our time 😉

website/src/index.md
```diff
@@ -0,0 +1,39 @@
...
+      text: Schema
+      link: /docs/reference/schema
+    - theme: alt
+      text: CLI
+      link: /docs/reference/cli
+
+
+features:
+  - title: 30-Second Setup
+    details: Single binary download, zero dependencies. Works with Homebrew, Snapcraft, Scoop.
```
pd93:
```suggestion
    details: Single binary download, zero dependencies. Works with Homebrew, Snapcraft, Scoop etc.
```

or maybe

```suggestion
    details: Single binary download, zero dependencies. Works with Homebrew, Snapcraft, Scoop and more.
```

just so it doesn't look like an exhaustive list.

website/.vitepress/config.ts
pd93, line 1: Image

Not a fan of the `.html`. Can we add `cleanUrls: true` here? See https://vitepress.dev/reference/site-config#cleanurls

website/src/docs/reference/cli.md
pd93, line 1: Love this!

Image

website/src/docs/reference/schema.md
```diff
@@ -0,0 +1,836 @@
...
+- **Options**: `checksum`, `timestamp`, `none`
+- **Description**: Default method for checking if tasks are up-to-date
+
+```yaml
+method: timestamp
+```
+
+### `includes`
+
+- **Type**: `map[string]Include`
```
pd93, line 61: Would be great if this could be a link to the `Include` section like we have in the current docs. Same goes for other references (e.g. `Variable`, `Task`). Doesn't have to work the same way as our current docs, but some kind of quick way to get to the inner type would be nice.

website/src/docs/reference/templating.md
```diff
@@ -0,0 +1,912 @@
...
+1: 1
+2: 1
+3: 1
+```
+
+## Special Variables
+
+Task provides special variables that are always available in templates. These override any user-defined variables with the same name.
+
+### CLI Context
```
pd93: I don't think all these titles need to have the word "Context"

website/src/docs/reference/templating.md
```diff
@@ -0,0 +1,912 @@
...
+### Output Functions
+
+#### `print`, `printf`, `println`
+Formatted output functions
+```yaml
+cmds:
+  - echo "{{printf "Version: %s.%d" .VERSION .BUILD}}"
+```
+
+## Slim-Sprig Functions
```
pd93: I actually don't think most of our users know (or care) what the difference between built-in, slim-sprig and task-specific functions are? I wonder if its worth just merging them all for simplicity.

It makes more sense to group them logically then by where they are defined in some cases. E.g. `randInt` and `randIntN` are in entirely different places despite being logically very similar because they're defined in different places.

website/src/blog/index.md
pd93, line 1: It's a shame that Vitepress doesn't have a way of generating this automatically. Seems a bit annoying to have the same text in multiple places.

pd93, line 1: Image

The social icons are not rendering correctly for me.

website/src/blog/task-in-2023.md
```diff
@@ -0,0 +1,83 @@
...
+outline: deep
+---
+
+# Introducing Experiments
+
+<AuthorCard :author="$frontmatter.author" />
+
+Lately, Task has been growing extremely quickly and I've found myself thinking a lot about the future of the project and how we continue to evolve and grow. I'm not much of a writer, but I think one of the things we could do better is to communicate these kinds of thoughts to the community. So, with that in mind, this is the first (hopefully of many) blog posts talking about Task and what we're up to.
+
+## 📅 So, what have we been up to?
```
pd93: Was there any particular reason to stop using shortcodes here? It seems like [they are supported](https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/full.mjs). I'm generally not a fan of having emojis in code as they often render strangely and break the monospacing.

Same applies to other files.

website/src/docs/changelog.md
```diff
@@ -852,7 +851,511 @@ it a go and let us know what you think via a
...
   #475, #626).
-- Remove long deprecated and undocumented `$` variable prefix and `^` command
+- Remove long deprecated and undocumented `---
+  title: Changelog
+  description: Historique complet des versions et changements de Task, de la v3.44.1 jusqu'à la v1.0.0
+  permalink: /changelog/
+  outline: deep
+---
+
+# Changelog
```
pd93: Looks like a copy paste error?

website/src/docs/contributing.md
```diff
@@ -81,16 +80,16 @@ by using `task website` (requires `nodejs` & `yarn`). All content is written in
...
 (enforced by Prettier).
 
-When making a change, consider whether a change to the [Usage Guide](/usage) is
+When making a change, consider whether a change to the [Usage Guide](/usage.md) is
 necessary. This document contains descriptions and examples of how to use Task
 features. If you're adding a new feature, try to find an appropriate place to
 add a new section. If you're updating an existing feature, ensure that the
-documentation and any examples are up-to-date. Ensure that any examples follow
-the [Taskfile Styleguide](/styleguide).
+documentation and any examples are up-to-date. Ensure that any examples follow[contributing.md](/contributing.md)
```
pd93: Don't think this is supposed to be here

```suggestion
documentation and any examples are up-to-date. Ensure that any examples follow
```

website/src/docs/contributing.md
```diff
@@ -81,16 +80,16 @@ by using `task website` (requires `nodejs` & `yarn`). All content is written in
...
+the [Taskfile Styleguide](/styleguide.md).
```
pd93: Inconsistent with using `/path` or `/path.md`. It doesn't appear that the `.md` is needed? (See the link to `/experiments` further up in this file) and in fact, it is converted to `.html` in the link which I already suggested we move away from.

This applies to all links

website/prettier.config.js
pd93, line 1: I would actually like to keep this. One of the nice things about using `.md` instead of `.mdx` is that prettier actually supports it properly. It will mean that we get our automatic 80 line limit and removal of erroneous new lines back.

More stuff we don't have to complain about in reviews.

website/.vitepress/config.ts
```diff
@@ -0,0 +1,285 @@
...
+        {
+          text: 'Community',
+          link: '/community'
+        },
+        {
+          text: 'Style Guide',
+          link: '/styleguide'
+        },
+        {
+          text: 'Contribution',
```
pd93:
```suggestion
          text: 'Contributing',
```

website/.vitepress/meta.ts
```diff
@@ -0,0 +1,5 @@
+export const taskName = 'Task';
+export const taskDescription =
+  'A fast, cross-platform build tool inspired by Make, designed for modern workflows.';
```
pd93, line 3: Is this officially our new tagline? If so, we should use this everywhere (GitHub/socials etc). Otherwise, this should stay the same as it is today.

pd93 (commented):
Fantastic start!

I still have a lot of reviewing to do, but this felt like enough for now. I'll keep looking over the next few days.

website/src/docs/contributing.md
pd93, line 1: This document makes references to `yarn`, but we appear to have moved to `pnpm` with this change. We might want to correct this.

I'm not a frontend dev, so don't generally ponder these things. Is any particular reason for the move away from yarn?

website/Taskfile.yml
```diff
@@ -33,20 +33,16 @@ tasks:
...
   clean:
     desc: Clean temp directories
     cmds:
-      - rm -rf ./build
+      - rm -rf ./vitepress/dist
 
   deploy:
-    desc: Build and deploy Docusaurus
-    summary: Requires GIT_USER and GIT_PASS envs to be previous set
+    deps: [install]
```
pd93: Does this need to run the `build` task too?

website/src/index.md
```diff
@@ -0,0 +1,39 @@
...
+    - theme: brand
+      text: Get Started
+      link: /docs/getting-started
+    - theme: alt
+      text: Schema
+      link: /docs/reference/schema
+    - theme: alt
+      text: CLI
+      link: /docs/reference/cli
+
```
pd93: These point to `/docs/*` but the pages are not under `/docs`. Somehow the link still works though?

Either way, I think these _should_ be hosted under `/docs` like they are today. It means we don't have to set up any redirect rules and keeps doc pages logically grouped under a URL path.

website/src/donate.md
```diff
@@ -0,0 +1,74 @@
...
+
+[Donate via PayPal](https://www.paypal.com/cgi-bin/webscr?cmd=_donations&business=GSVDU63RKG45A&currency_code=USD&source=url)
+
+---
+
+## 🇧🇷 PIX (Brazil only)
+
+If you're in Brazil, you can also support @andreynering via PIX:
+
+![PIX QR Code](/img/pix.png)
```
pd93: Maybe decrease the size of the QR code (Sorry @andreynering 😆). It's quite large atm.

website/src/donate.md
```diff
@@ -0,0 +1,74 @@
...
+outline: false
+---
+
+# 🙏 Support Taskfile
+
+If you find this project useful, consider supporting its ongoing development.
+
+> This is just a way to say **“thank you”** — donations won’t provide priority support or special privileges.
+
+---
```
pd93: These `---` separators aren't needed as the headings already create them. Having both means we get double separators:

Image

website/src/docs/deprecations/version-2-schema.md
```diff
@@ -1,17 +1,17 @@
 ---
-slug: /deprecations/version-2-schema/
+title: "Version 2 Schema (#1197)"
+description: Dépréciation du schéma version 2 des Taskfiles et migration vers la version 3
```
pd93: en anglais s'il vous plaît.

Same in:

- `/deprecations/completion-scripts.md`
- `/deprecations/index.md`
- `/deprecations/template-functions.md`
- `/deprecations/template.md`
- `/deprecations/version-2-schema.md`
- `/experiments/env-precedence.md`
- `/experiments/gentle-force.md`
- `/experiments/index.md`
- `/experiments/remote-taskfiles.md`

website/src/docs/experiments/env-precedence.md
```diff
@@ -1,31 +1,27 @@
...
-
-:::info
-
+::: info
 To enable this experiment, set the environment variable:
 `TASK_X_ENV_PRECEDENCE=1`. Check out [our guide to enabling
-experiments][enabling-experiments] for more information.
-
+experiments](/experiments/#enabling-experiments) for more information.
 :::
```
pd93: I mentioned this on Discord already and its not that important, but if its possible, it would be nice to re-add the icons we had on these admonitions on the old site. They look a bit plain without them.

website/src/docs/deprecations/template.md
pd93, line 1: We have this template for the deprecations pages, but the one for experiments is missing.

Taskfile.yml
vmaerten: Removed thanks !

website/TODO.md
vmaerten: Should have been deleted ! thanks

website/.vitepress/config.ts
vmaerten: Done

website/src/docs/reference/cli.md
vmaerten: Me too ! I find this way more clear than our old table (I took example from vitest and vitepress)

website/src/blog/index.md
vmaerten: I agree
The only way to do it, is to develop it ... but as we do not blog a lot, I assumed that it's something we can improve later

website/src/index.md
vmaerten: Oopsy the URL is not correct. it should be `/reference/cli`

I'd love have docs but in our current setup it's not : 
- https://taskfile.dev/reference/cli
- https://taskfile.dev/usage/

website/src/donate.md
vmaerten: I've put 200px x 200px :)

website/src/blog/task-in-2023.md
vmaerten: You're totally right! I did not pay attention but somehow the AI modify this
I'll change it !

website/src/docs/changelog.md
vmaerten: hmm I do not get it ^^"

website/prettier.config.js
vmaerten: Very good point !

vmaerten: I've ran prettier on everything : *.vue, *.md and *.json

website/src/docs/deprecations/version-2-schema.md
vmaerten: - `/deprecations/completion-scripts.md` :heavy_check_mark: 
- `/deprecations/index.md` :heavy_check_mark: 
- `/deprecations/template-functions.md` :heavy_check_mark: 
- `/deprecations/template.md` :heavy_check_mark: 
- `/deprecations/version-2-schema.md` :heavy_check_mark: 
- `/experiments/env-precedence.md`  :heavy_check_mark: 
- `/experiments/gentle-force.md`  :heavy_check_mark: 
- `/experiments/index.md`  :heavy_check_mark: 
- `/experiments/remote-taskfiles.md`  :heavy_check_mark:

website/src/index.md
pd93: Oh my mistake! I really thought we had `/docs` in the URL for some reason. Disregard the latter part of this comment then 😆

website/src/blog/index.md
vmaerten: Somehow, I does not work locally with dev server but it works when built and run with real server.
I'll take a look later

website/src/docs/changelog.md
pd93: Sorry, comment wasn't very clear. The file starts at v3.44.1 and goes all the way back to v3.10.0, then on line 861, it looks like a copy of the file has been inserted again all the way until line 1364-ish. It's much clearer if you look at the full diff on GitHub.

website/prettier.config.js
vmaerten: I have an issue with custom containers (called "Custom container" in VitePress, similar to admonitions).

Before prettier : 

```md
::: info
When using `-x/--exit-code`, failed command exit codes are passed through instead of the above codes.
:::
```

After : 

```md
::: info When using `-x/--exit-code`, failed command exit codes are passed
through instead of the above codes. :::
```

This breaks the rendering...

I’ve thought about two possible solutions, but I don’t really like either:

Option 1: Add a blank line before and after the content:

```md
::: tip

When using `-x/--exit-code`, failed command exit codes are passed
through instead of the above codes. 

:::
```

Option 2: Use Prettier ignore comments:

```md
<!-- prettier-ignore-start -->
::: info
When using `-x/--exit-code`, failed command exit codes are passed through instead of the above codes.
:::
<!-- prettier-ignore-end -->
```

vmaerten: What do you think ? @pd93  + @andreynering

pd93: This is a [known problem](https://github.com/prettier/prettier/issues/8640) with Prettier and we had the same issue before we switched to `.mdx` on the old site.

Back then we used the former syntax (Option 1). Its ironically not the prettiest, but I think its easier than adding the comments everywhere and it won't stop prettier from making other formatting changes inside the admonition block.

vmaerten: Ok I'll apply it everywhere soo
Thanks!

website/.vitepress/meta.ts
vmaerten: I haven't told you, I tried something more engaging
But I am perfectly fine to stay with the current's one (that's why it's stored in the variable, easy to change)

What are your thoughts on it @andreynering and @pd93  ?

website/src/docs/contributing.md
vmaerten: I've updated it.
I'm also a JavaScript developer at work (both frontend and backend), and I work with Go and Rust too 😄.
There are several reasons for the switch, but mainly: it's faster, uses less space for node_modules, gives more control over hosting, and is actively maintained.
Since we're still on Yarn 1, which is now deprecated, I decided to migrate to pnpm instead of moving to Yarn 4, for the reasons mentioned above.

website/src/docs/experiments/env-precedence.md
vmaerten: image


Dark mode : 
image

Disclaimer, I am not a designer (and I have really bad taster in termes of design) so feel free to comment it

pd93: These are perfect! Nice one

website/src/donate.md
```diff
@@ -0,0 +1,64 @@
...
+title: Donate
+layout: doc
+outline: false
+---
+
+# 🙏 Support Taskfile
```
pd93: More emojis in here

website/src/docs/experiments/remote-taskfiles.md
```diff
@@ -34,19 +36,35 @@ when including Taskfiles.
...
+
+```text [Git over HTTP]
+https://github.com/go-task/task.git//website/static/Taskfile.yml?ref=main
+```
+
+```text [Git over SSH]
+git@github.com/go-task/task.git//website/static/Taskfile.yml?ref=main
+```
+
+:::
```
pd93, line 53: Kind of a shame that we lose the id/anchor features from Docusaurus here.

i.e. https://taskfile.dev/experiments/remote-taskfiles/?method=git-ssh will take you to the page with the tab set to the given method and changing the method updates every set of tabs on the page.

Maybe this is a feature request we can make and re-add the functionality later

website/src/docs/contributing.md
pd93, line 1: Just noticed the documentation section will need updating to mention VitePress instead of Docusaurus, Yarn and MDX.

website/src/docs/contributing.md
pd93: Sorry, you're going to hate me 😆 I just checked the branch out again and found that the absolute links (`/styleguide` etc) mean that navigating locally in a git checkout doesn't work.

If it was a relative link including file extension (`./styleguide.md`), then these would work. The editor will even bring up the file browser when you start typing `./` and suggest the files you can pick from and the links in the markdown previewer will work too.

None of this seems to break the actual rendered site either!

website/src/docs/experiments/remote-taskfiles.md
vmaerten: Should I sleep in the night or develop this feature for vitepress ? do not tempt me :joy:

website/src/docs/contributing.md
vmaerten: Oh I see
Just to be sure, I can replace all link to relative link with extension ? 
Claude agree with us : 

image

pd93: > Just to be sure, I can replace all link to relative link with extension ?

Yeah, I think replacing all weblinks with relative paths and file extensions is good!

website/src/docs/experiments/remote-taskfiles.md
pd93: See related issue: https://github.com/vuejs/vitepress/issues/2954

Sleep in the night! No need to destroy your schedule for this 😆

website/src/docs/reference/templating.md
vmaerten: I agree ! I pushed new version !

website/src/docs/changelog.md
vmaerten: Fixed !

website/Taskfile.yml
```diff
@@ -1,29 +1,35 @@
-version: "3"
+version: '3'
 
 tasks:
-  yarn:install:
+  install:
     desc: Setup Docusaurus locally
```
andreynering:
```suggestion
    desc: Setup VitePress locally
```

website/Taskfile.yml
```diff
@@ -33,20 +39,15 @@ tasks:
...
-    cmds:
-      - npx docusaurus deploy
-
-  upgrade:
-    desc: Upgrade Docusaurus
+    deps: [build]
+    desc: Build and deploy Website to Netlify
     cmds:
-      - yarn upgrade @docusaurus/core@latest @docusaurus/preset-classic@latest @docusaurus/module-type-aliases@latest @docusaurus/tsconfig@latest @docusaurus/types@latest
+      - pnpm netlify deploy --prod --site=054cd897-66d2-46fa-9b57-4831a8d61b37
```
andreynering: @vmaerten Which site ID is this? I suppose it's a test one you created.

We're probably going to remove this because we'll have two sites now: `taskfile.dev` and `next.taskfile.dev`. We should set this via ENV.

Basically we'll need to:

* Create a new website on Netlify to be `next.taskfile.dev` or change the domain of the current one to be `next.taskfile.dev`. IIRC it's Netlify that receives webhooks for pushes to `main` and builds the site.
* Set up the website that will be the new `taskfile.dev` to deploy from GitHub Action on release as you did.

I can work on this if you want.

website/src/docs/usage.md
```diff
@@ -39,11 +35,12 @@ the file tree until it finds one (similar to how `git` works). When running Task
 from a subdirectory like this, it will behave as if you ran it from the
 directory containing the Taskfile.
 
-You can use this functionality along with the special `{{.USER_WORKING_DIR}}`
-variable to create some very useful reusable tasks. For example, if you have a
-monorepo with directories for each microservice, you can `cd` into a
-microservice directory and run a task command to bring it up without having to
-create multiple tasks or Taskfiles with identical content. For example:
+You can use this functionality along with the special
+<span v-pre>`{{.USER_WORKING_DIR}}`</span> variable to create some very useful
```
andreynering: Looks like this was a search & replace that gone wrong.

website/src/docs/usage.md
```diff
@@ -65,18 +62,18 @@ will be brought up.
 
 If you call Task with the `--global` (alias `-g`) flag, it will look for your
 home directory instead of your working directory. In short, Task will look for a
-Taskfile that matches `$HOME/{T,t}askfile.{yml,yaml}` .
+Taskfile that matches `$HOME/`<code>{T,t}askfile.{yml,yaml}</code> .
```
andreynering: This one too.

They still render okay but we should still revert those changes.

And there are other like those.

website/src/docs/usage.md
```diff
@@ -256,11 +257,11 @@ version: '3'
 
 includes:
   docs:
-    taskfile: ./docs/Taskfile.yml
+    taskfile: ./Taskfile.yml
     dir: ./docs
```
andreynering: The path here was intentional. We should revert this change.

andreynering (commented):
Looking good!

I didn't do a full pass yet but added a couple of comments in the meantime.

website/Taskfile.yml
vmaerten: Exactly, I planned to change it as soon we agree on the PR and netlify has been updated

We don't need to variabilise it because only `taskfile.dev` will be deployed with this Task
`next.taskfile.dev` will be deployed "from Github" like we have today (pushing on main)
So this siteId should be the `taskfile.dev`'s one (deploy in the release with the netlify CLI)

> I can work on this if you want.

As you wish, if you have time, I am fine with it :)

website/src/docs/usage.md
vmaerten: Are you talking about `<span v-pre>` ?
If yes, it's because `{{}}` is interpreted as vue 

image

website/src/docs/usage.md
vmaerten: I've removed it!

website/src/docs/usage.md
andreynering: Got it

website/.vitepress/meta.ts
andreynering: I think it's time for a new headline. It's the perfect time to change as we're migrating the website.

This one looks good, we can still change if we find something even better.

andreynering:
FYI: I just created a new website on Netlify to be https://next.taskfile.dev/ and should be live soon. Just need to wait for DNS to propagate. In the meantime it's available on https://next-taskfile.netlify.app/

I configured it to build on this branch (`migrate-to-vitepress`) for now, so we can preview it as we iterate. We'll just change it to `main` once merged.

Issue #2355, opened 2025-08-04
Preconditions: environment variable substitution stopped working


christophberger:
### Description

A Taskfile that worked fine for months is now failing when trying to evaluate an environment variable within a precondition.

This Taskfile works:

```yaml
version: "3"

env:
  PORT: 1313

tasks:
  port:
    cmds:
      - echo $PORT
```

Output:

```
> task port
task: [port] echo $PORT
1313
```

This one fails:

```yaml
version: "3"

env:
  PORT: 1313

tasks:
  noserverrunning:
    preconditions:
      - sh: "! nc -z localhost $PORT"
        msg: Port $PORT is not available.
  port:
    deps:
      - noserverrunning
    cmds:
      - echo $PORT
```

Output:

```
> task port
task: Port $PORT is not available.
task: precondition not met
```

Expected (if the precondition isn't met): `Port 1313 is not available.`

The Taskfile has been the same for months, but when I used it today, it started failing. Hence I assume it's a bug. 

### Version

3.44.1

### Operating system

macOS Sequoia 15.5

### Example Taskfile

```YAML
version: "3"

env:
  PORT: 1313

tasks:
  noserverrunning:
    preconditions:
      - sh: "! nc -z localhost $PORT"
        msg: Port $PORT is not available.
  port:
    deps:
      - noserverrunning
    cmds:
      - echo $PORT
```

vmaerten:
Hello ! 
You need to use the go templating system : 

```yml
version: "3"

env:
  PORT: 1313

tasks:
  noserverrunning:
    preconditions:
      - sh: "! nc -z localhost {{.PORT}}"
        msg: Port {{.PORT}} is not available.
  port:
    deps:
      - noserverrunning
    cmds:
      - echo {{.PORT}}
```

christophberger:
Thanks @vmaerten, I'll try this, but can you tell my why my version worked fine until just recently?

christophberger:
Unfortunately, `{{.PORT}}` doesn't seem to work inside the `sh` command string (I tested the string with and without double quotes). I always get "Port 1313 is not available" even though no server listens on this port.

christophberger:
[As per the docs](https://taskfile.dev/usage/#environment-variables), using `$VAR` syntax is valid for environment variables: 

```yaml
version: '3'

env:
  GREETING: Hey, there!

tasks:
  greet:
    cmds:
      - echo $GREETING
```

christophberger:
[...]

PR #2360, opened 2025-08-07
feat: add native core utils to improve compatibility on windows

### LINKED ISSUE #197: Support built-in shell commands for an improved Windows support
For cross-platform usage, some built-in commands will helpful.

andreynering:
TODO:

- [ ] Wait for Go v1.25 launch, because we support the latest two Go major versions and this requires Go 1.24
- [ ] Update documentation to mention the `TASK_USE_GO_CORE_UTILS` env setting

---

This PR makes Task start using some core utils from the [u-root](https://github.com/u-root/u-root) project so we can have a much improved compability on Windows.

Although it should work most of the time, it's important to know that there might be small implementation differences between GNU or macOS core utils and these Go-implemented ones. Because of this, I decided to only enable it by default on Windows for now, but you can still control that:

* If a `TASK_USE_GO_CORE_UTILS` is set to either `0` (disabled) or `1` (enabled), this will be respected no matter the operating system.
* Otherwise, it is enabled by default on Windows but not on the other operating system.
* When not enabled, it simply means it will use the system core utils if available, or fail otherwise.

Available core utils More can be added with time):

* base64
* cat
* chmod
* cp
* find
* gzip
* ls
* mkdir
* mktemp
* mv
* rm
* shasum
* tar
* touch
* xargs

---

* Closes #197


fgimian:
Go 1.25.0 is out now. 🚀

andreynering:
This is now ready to Go.

I made the env shorter, just `TASK_CORE_UTILS`.

I wrote a blog post which can be previewed on [this link](https://deploy-preview-2360--next-taskfile.netlify.app/blog/windows-core-utils).

fgimian:
Super excited about this being merged and released, thank you so much @andreynering!!! 😄

pd93 (approved):
Amazing work! So good to finally see this land 🚀

internal/execext/coreutils.go
```diff
@@ -0,0 +1,20 @@
...
+var useGoCoreUtils bool
+
+func init() {
+	// If TASK_CORE_UTILS is set to either true or false, respect that.
+	// By default, enable on Windows only.
+	if v, err := strconv.ParseBool(env.GetTaskEnv("CORE_UTILS")); err == nil {
+		useGoCoreUtils = v
+	} else {
+		useGoCoreUtils = runtime.GOOS == "windows"
+	}
```
vmaerten, line 19: Should this be configurable via the `taskrc.yml` file? (Similar to what we discussed on Discord)

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	golang.org/x/sync v0.16.0
 	golang.org/x/term v0.34.0
 	gopkg.in/yaml.v3 v3.0.1
+	mvdan.cc/sh/moreinterp v0.0.0-20250807215248-5a1a658912aa
 	mvdan.cc/sh/v3 v3.12.0
```
vmaerten, line 32: I'm not sure what Task's guideline is on referencing unreleased libraries. Do we need to wait until it's officially released (e.g., tagged on GitHub) ?

vmaerten (commented):
Finally 🚀
I’ve left some questions.

go.mod
pd93: I don't see this as a huge issue. We've imported `golang.org/x/exp` in the past when we wanted to try out `slices`/`maps` before they were in the std lib.

We just have to be a bit more careful when updating.

vmaerten: Oh! So we could just use the latest version to fix the issue that’s impacting us?
I am talking about https://github.com/go-task/task/issues/2355 that has been fixed in the upstream but not released yet

pd93: If we're super keen to get a fix out and it passes all our tests, then I'm not against it. Let's get @andreynering's thoughts too though.

internal/execext/coreutils.go
andreynering: Probably yes.

Do you think it's better to on this PR or later on a separate one? I know some changes still weren't merged to `main`.

vmaerten: I think we can do this on a separate one :)

go.mod
andreynering: A bit of history here...

Me and @mvdan had some issues tagging a version on `mvdan.cc/sh/moreinterp` because the project uses a vanity domain (`mvdan.cc`), and some temporary misconfiguration messed with Go's proxy cache. Until we figure out how to address that, we agreed that tagging a commit is fine for now.

When it comes to the interpreter itself (`mvdan.cc/sh/v3`), I'd prefer to pin tags if possible, but if a given bug is more serious nothing stops us from pinning a commit temporarily so we can release the fix for our users.

vmaerten: Thanks a lot for the explaination! 
Fine to me!

PR #2362, opened 2025-08-10
build: publish deb and rpm to cloudsmith

vmaerten:
I’ve created a team, and we’re using the Open Source free tier, which should be sufficient for our needs!
I’ve added the `CLOUDSMITH_TOKEN` as a secret variable in the GitHub Actions workflow.

I haven’t been able to test it yet, but according to Cloudsmith, we can use `any-version` since our binary is statically compiled.

I haven’t updated the documentation yet because the website is being rewritten. Also, I wanted to test how it works first before making any updates.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -135,3 +135,19 @@ winget:
           owner: microsoft
           name: winget-pkgs
           branch: master
+
+cloudsmiths:
+  - organization: "task"
+    repository: "task"
+    formats:
+      - deb
+      - rpm
+    distributions:
+      deb:
+        - "ubuntu/any-version"
+        - "debian/any-version"
+      rpm:
+        - "el/any-version"
+        - "fedora/any-version"
+    component: main
+    republish: true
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -135,3 +135,19 @@ winget:
...
           name: winget-pkgs
           branch: master
+
+cloudsmiths:
+  - organization: "task"
+    repository: "task"
+    formats:
+      - deb
+      - rpm
+    distributions:
```
andreynering, line 146: What about Alpine?

AFAIK `go-task` is published officially there, so maybe not needed, but asking anyway.

In theory would allow users to have a new release sooner.

andreynering (approved):
Thank you for setting this up!

.goreleaser.yml
vmaerten: We could add it indeed!  I’d say the more things we handle ourselves, the better it’ll be.

andreynering: Yeah, I think we can add to test, and if anything we can just not document it.

PR #2247, opened 2025-05-06
feat: add global .taskrc in HOME

vmaerten:

PR #2380, opened 2025-08-14
feat: XDG config

pd93:
Based on #2247

This takes the idea of having a config in `$HOME` and takes it further by:

- Using `$XDG_CONFIG_HOME` instead of `$HOME`
- Adding proper nested config support (including use of `$HOME` if its a parent).
  - See [the updated docs](https://deploy-preview-2380--next-taskfile.netlify.app/docs/reference/config#file-precedence) for a full example

### DIFF
--- a/website/.vitepress/config.ts
+++ b/website/.vitepress/config.ts
@@ -208,12 +208,12 @@ export default defineConfig({
               link: '/docs/reference/schema'
             },
             {
-              text: 'Configuration',
-              link: '/docs/reference/config'
-            },
-            {
               text: 'Environment',
               link: '/docs/reference/environment'
+            },
+            {
+              text: 'Configuration',
+              link: '/docs/reference/config'
             },
             {
               text: 'CLI',
--- a/website/src/docs/reference/cli.md
+++ b/website/src/docs/reference/cli.md
@@ -10,8 +10,8 @@ outline: deep
 Task has multiple ways of being configured. These methods are parsed, in
 sequence, in the following order with the highest priority last:
 
-- [Configuration files](./config.md)
 - [Environment variables](./environment.md)
+- [Configuration files](./config.md)
 - _Command-line flags_
 
 In this document, we will look at the last of the three options, command-line
--- a/website/src/docs/reference/config.md
+++ b/website/src/docs/reference/config.md
@@ -10,11 +10,11 @@ outline: deep
 Task has multiple ways of being configured. These methods are parsed, in
 sequence, in the following order with the highest priority last:
 
-- _Configuration files_
 - [Environment variables](./environment.md)
+- _Configuration files_
 - [Command-line flags](./cli.md)
 
-In this document, we will look at the first of the three options, configuration
+In this document, we will look at the second of the three options, configuration
 files.
 
 ## File Precedence
--- a/website/src/docs/reference/environment.md
+++ b/website/src/docs/reference/environment.md
@@ -9,11 +9,11 @@ outline: deep
 Task has multiple ways of being configured. These methods are parsed, in
 sequence, in the following order with the highest priority last:
 
-- [Configuration files](./config.md)
 - _Environment variables_
+- [Configuration files](./config.md)
 - [Command-line flags](./cli.md)
 
-In this document, we will look at the second of the three options, environment
+In this document, we will look at the first of the three options, environment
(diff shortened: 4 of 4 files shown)


taskrc/taskrc.go
```diff
@@ -1,6 +1,62 @@
...
 }
+
+// GetConfig loads and merges local and global Task configuration files
+func GetConfig(dir string) (*ast.TaskRC, error) {
+	var config *ast.TaskRC
+	reader := NewReader()
+
+	// Read the XDG config file
+	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
+		xdgConfigNode, err := NewNode("", xdgConfigHome)
```
pd93: Middle of the night thought to self. Fix this to be a task subdirectory...

vmaerten: I totally agree!

vmaerten:
Regarding the code, and IMO how it should behave, the docs are wrong.
 > Configuration Reference
> Task has multiple ways of being configured. These methods are parsed, in sequence, in the following order with the > highest priority last:
> 
> 1. Configuration files
> 2. Environment variables
> 3. Command-line flags

The local config file should override env variables, and that’s how it works now.  
Personally, I think the priority should be:  
1. Global config file (HOME / XDG)  
2. Env variables (though this is tricky since we also have local env vars for experiments)  
3. Local config file  
4. CLI flags  

I just don’t want env vars from my `.zshrc` to beat my local `.taskrc.yml`.

vmaerten:
That being said, I totally forgot about XDG config, and I love this! Good job!
Nice refacto tho

pd93:
> The local config file should override env variables, and that’s how it works now. Personally, I think the priority should be:
> 
> 1. Global config file (HOME / XDG)
> 2. Env variables (though this is tricky since we also have local env vars for experiments)
> 3. Local config file
> 4. CLI flags

I don't see the distinction between a "local" file and one in the users home directory config. With a project path of `$HOME/path/to/project`, there could be config files in each of:

- `$HOME/path/to/project`
- `$HOME/path/to`
- `$HOME/path`
- `$HOME`

and they would all override one another with the more specific paths taking priority. I don't see why the one in `$HOME` should be treated differently.

> I just don’t want env vars from my `.zshrc` to beat my local `.taskrc.yml`.

I can understand why the XDG config might be treated differently, but why not just have environment variables as the lowest priority then? I think its confusing to put env vars inbetween the two types of config in the priority order.

Thinking about it, is there any particular reason we need environment variables at all? Why would you apply settings in an rc file when you could just define a config?

vmaerten:
> and they would all override one another with the more specific paths taking priority. I don't see why the one in $HOME should be treated differently.

Because a global env variable would take precedance over a config file located in the project (but See :arrow_down: )

> I can understand why the XDG config might be treated differently, but why not just have environment variables as the lowest priority then? I think its confusing to put env vars inbetween the two types of config in the priority order.

> Thinking about it, is there any particular reason we need environment variables at all? Why would you apply settings in an rc file when you could just define a config?

Here we are! If you ask me, now that we have proper config files, env vars aren’t really needed anymore.  
Aside from `TASK_COLOR` and `TASK_TEMP_DIR`, the rest are tied to experiments and could be removed.  

That would simplify both our codebase **and** the user experience.

pd93:
> Because a global env variable would take precedance over a config file located in the project (but See ⬇️ )

Ok, I think we're on the same page 👍
 
> Here we are! If you ask me, now that we have proper config files, env vars aren’t really needed anymore. Aside from `TASK_COLOR` and `TASK_TEMP_DIR`, the rest are tied to experiments and could be removed.

Since config files are new, I assume we would put `TASK_COLOR` and `TASK_TEMP_DIR` as the lowest priority and nothing should break. We can then deprecate them.

vmaerten (approved):
I've removed my two debug logs!
LGTM

panchoh:
Hi, and sorry I’m late to the party!

Could I suggest (for another PR, but keeping this context) that we strip the initial dot in the `taskcfg.yml` that lives on `$XDG_CONFIG_HOME/task`?

Having a hidden file there only makes it harder to see it, and it is a common pattern. See for instance `git`: `~/.gitconfig` vs. `~/.config/git/config`.

Thanks for Task as well as this PR!!! <3

PR #2388, opened 2025-08-18
feat: added variable override flag --var

sashokbg:
This PR adds a new --var flag that allows to override variables set in the Taskfile


andreynering:
Hi @sashokbg,

Thanks for your PR, but this is probably not the right direction as we already have a proper syntax for setting vars via the CLI:

```bash
task my-task FOO=bar
```

So in theory there's no need for:

```bash
task my-task --var FOO=bar
```

There's the priority thing, but add even one more way of setting vars would just make things even for confusing.

sashokbg:
Hello @andreynering I understand, I missed out that positional args are treated as vars.

Can we please keep the variable overriding at least ?

So that I can still run task my-task FOO=bar but the FOO from cli args will take precedence over the one defined in the template file (like is the case with many other cli tools).

Thank you for your understanding

andreynering:
@sashokbg The current behavior is by design. You should use `default` if you want to make a variable overridable:

https://taskfile.dev/docs/reference/templating#default-values-and-coalescing

sashokbg:
ok, it is all clear now @andreynering, thank you for your explanations.

I worked on the overriding / defaults in relation to https://github.com/go-task/task/pull/2379 can you come check it please ?

The idea is to have variables set by interactive input, but we need to handle the case on non-TTY shells and thus use defaults etc. Any comment and proposals would be really helpful.

Thank you and have a nice day !

Issue #1403, opened 2023-11-17
Feature request: Allow taskfile schema to use options related to remote taskfiles


c-ameron:
As discussed [here](https://github.com/go-task/task/pull/1345#issuecomment-1791296189), I'd like to see options for the remote taskfiles to be included in taskfile schema.

Having them as a key in the .yml would make it easier to manage the options and a less clunky CLI invocation (especially in the context of sourcing/sharing different remote taskfiles).

As an example
``` yaml
includes:
  my-remote-namespace: https://raw.githubusercontent.com/my-org/my-repo/main/Taskfile.yml
  offline: true
  auto-update-ttl: weekly
```

I understand that you don't want to do schema changes while the remote taskfiles are still an experiment, but something to track/think about if it makes it to GA.

Thanks again for your work!

pd93:
@c-ameron I don't think we're going to be adding these to the main Taskfile schema. However, they will likely be added to the new `.taskrc` config file instead so I have changed the title accordingly and we will track those changes here. 

@vmaerten has already started this process with #2210

PR #2389, opened 2025-08-18
feat: add config to taskrc.yml

vmaerten:
- Added  flags to the config file (CLI values take precedence over config).
- Documented the `remote` CLI flag (marked as *experimental* with a badge).
- Documented the `gentle-force` CLI flag (marked as *experimental* with a badge).
- Updated documentation with cross-links between CLI and config pages.

Fixes https://github.com/go-task/task/issues/1403

### DIFF
--- a/website/src/docs/experiments/remote-taskfiles.md
+++ b/website/src/docs/experiments/remote-taskfiles.md
@@ -296,7 +296,6 @@ This experiment adds a new `remote` section to the [configuration file](../refer
 
 - **Type**: `object`
 - **Description**: Remote configuration settings for handling remote Taskfiles
-- **Requires**: [Remote Taskfiles experiment](../experiments/remote-taskfiles.md) to be enabled
 
 ```yaml
 remote:


website/src/docs/reference/cli.md
pd93, line 1: Up for debate, but I'm not convinced that these flags should be documented here. I've always seen the experiments as independent from the rest of Task and fully documented within their experiments page. This means that the main docs are stable and not changing all the time.

The intention would then be to move the experiment docs to the appropriate place when released.

taskrc/ast/taskrc.go
```diff
@@ -24,4 +35,28 @@ func (t *TaskRC) Merge(other *TaskRC) {
...
+		t.Remote.CacheExpiry = other.Remote.CacheExpiry
+	}
+
+	if other.Verbose != nil {
+		t.Verbose = other.Verbose
+	}
+
+	if other.Concurrency != nil {
+		t.Concurrency = other.Concurrency
+	}
```
pd93: `cmp.Or` would tidy this up a lot. Could also be used for `Version`.

Also ordering to match field names.

```suggestion
	t.Remote.Insecure = cmp.Or(other.Remote.Insecure, t.Remote.Insecure)
	t.Remote.Offline = cmp.Or(other.Remote.Offline, t.Remote.Offline)
	t.Remote.Timeout = cmp.Or(other.Remote.Timeout, t.Remote.Timeout)
	t.Remote.CacheExpiry = cmp.Or(other.Remote.CacheExpiry, t.Remote.CacheExpiry)

	t.Verbose = cmp.Or(other.Verbose, t.Verbose)
	t.Concurrency = cmp.Or(other.Concurrency, t.Concurrency)
```

taskrc/ast/taskrc.go
```diff
@@ -2,13 +2,24 @@ package ast
...
 	"github.com/Masterminds/semver/v3"
 )
 
 type TaskRC struct {
 	Version     *semver.Version `yaml:"version"`
 	Experiments map[string]int  `yaml:"experiments"`
+	Remote      Remote          `yaml:"remote"`
+	Verbose     *bool           `yaml:"verbose"`
+	Concurrency *int            `yaml:"concurrency"`
+}
```
pd93: Nit: Just ordering. Makes sense to have globals first and experiments last IMO

```suggestion
type TaskRC struct {
    Version     *semver.Version `yaml:"version"`
	Verbose     *bool           `yaml:"verbose"`
	Concurrency *int            `yaml:"concurrency"`
	Remote      Remote          `yaml:"remote"`
    Experiments map[string]int  `yaml:"experiments"`
}
```

website/src/docs/reference/config.md
```diff
@@ -68,3 +68,118 @@ option_3: foo # Taken from $XDG_CONFIG_HOME/task/.taskrc.yml
...
+remote:
+  insecure: false
+  offline: false
+  timeout: "30s"
+  cache-expiry: "1h"
+
+# Global settings
+verbose: true
+concurrency: 2
+```
```
pd93: Nit: more ordering

```suggestion
```yaml
# Global settings
verbose: true
concurrency: 2

# Remote configuration
remote:
  insecure: false
  offline: false
  timeout: "30s"
  cache-expiry: "1h"

# Enable experimental features
experiments:
  REMOTE_TASKFILES: 1
```
```

File: internal/flags/flags.go [hunk 4]
```diff
@@ -5,17 +5,17 @@ import (
...
 	"github.com/spf13/pflag"
 
 	"github.com/go-task/task/v3"
 	"github.com/go-task/task/v3/errors"
 	"github.com/go-task/task/v3/experiments"
-	"github.com/go-task/task/v3/internal/env"
 	"github.com/go-task/task/v3/internal/sort"
 	"github.com/go-task/task/v3/taskfile/ast"
+	"github.com/go-task/task/v3/taskrc"
+	taskrcAST "github.com/go-task/task/v3/taskrc/ast"
```
andreynering: Package names or alises should be all lowercase.

```suggestion
	taskrcast "github.com/go-task/task/v3/taskrc/ast"
```

website/src/docs/reference/cli.md
andreynering: I agree

vmaerten: I get your point, but it still feels a bit odd to me that it’s in `reference/config.yml` but not in `cli.md`.  
Personally, I think all the flags you see in `task -h` should also show up in `cli.md`.

Remote experiments are used a lot, and it’s not always clear which flags exist or what they actually do just from the experiment page. That’s why I suggested documenting it, but with a note like:

> This flag is linked to an experiment (see XXXXX) and could be removed at any time.

That said, if you both prefer not to include it here, I’m fine with removing it.

pd93: I think for the config reference, I would include a section for the `experiments:` key, but not include the specific experiments. Just say something along the lines of "please refer to the experiments pages for more details" and then describe the name of the key and values over there.

I don't think specific experiments should be described in any pages outside the experiments docs themselves.

website/src/docs/reference/config.md
```diff
@@ -68,3 +68,132 @@ option_3: foo # Taken from $XDG_CONFIG_HOME/task/.taskrc.yml
...
+remote:
+  insecure: false
+  offline: false
+  timeout: "30s"
+  cache-expiry: "1h"
+
+# Global settings
+verbose: true
+concurrency: 2
+```yaml
```
pd93: Think this got accidentally doubled up. Suggested changes are annoying sometimes 😆

vmaerten: :joy:

website/src/docs/reference/cli.md
vmaerten: Not sure what you mean but I've removed all remote and gentle force stuff from CLI.md

vmaerten: Got it! Let me know if this is what you had in mind 🙂

PR #2391, opened 2025-08-19
feat: change XDG taskrc naming

### LINKED ISSUE #2390: `XDG_CONFIG_HOME/task` config should not be a hidden file
Could I suggest that we strip the initial dot in the `taskcfg.yml` that lives on `$XDG_CONFIG_HOME/task`?

Having a hidden file there only makes it harder to see it, and it is a common pattern. See for instance `git`: `~/.gitconfig` vs. `~/.config/git/config`.

_Originally posted by @panchoh in https://github.com/go-task/task/issues/2380#issuecomment-3199204619_

pd93:
Fixes #2390 by changing the name of the `$XDG_CONFIG_HOME/task` config to `taskrc.yml` or `taskrc.yaml` (without the preceding `.`).

@vmaerten @panchoh w.r.t discussion in #2390, I'm not against `config.yml`, but after a bit of thinking I actually feel like ripgrep might have got this right. The naming consistency is valuable IMO. It lets you know that you're editing the same kind of file with the same schema and I think `$XDG_CONFIG_HOME/task/taskrc.yml` is the way to go. Happy to be outvoted on this though!

Once agreed and merged, I will open a PR to the schemastore repo.

---

Also the related discussion on Discord around the inclusion of the `$HOME` directory when running from a directory that does not have `$HOME` as a parent.

See updated docs for a full description.

### DIFF
--- a/taskrc/node.go
+++ b/taskrc/node.go
@@ -11,9 +11,10 @@ type Node struct {
 func NewNode(
 	entrypoint string,
 	dir string,
+	possibleFileNames []string,
 ) (*Node, error) {
 	dir = fsext.DefaultDir(entrypoint, dir)
-	resolvedEntrypoint, err := fsext.SearchPath(dir, defaultTaskRCs)
+	resolvedEntrypoint, err := fsext.SearchPath(dir, possibleFileNames)
 	if err != nil {
 		return nil, err
 	}
--- a/taskrc/taskrc.go
+++ b/taskrc/taskrc.go
@@ -4,15 +4,22 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/go-task/task/v3/internal/fsext"
 	"github.com/go-task/task/v3/taskrc/ast"
 )
 
-var defaultTaskRCs = []string{
-	".taskrc.yml",
-	".taskrc.yaml",
-}
+var (
+	defaultXDGTaskRCs = []string{
+		"taskrc.yml",
+		"taskrc.yaml",
+	}
+	defaultTaskRCs = []string{
+		".taskrc.yml",
+		".taskrc.yaml",
+	}
+)
 
 // GetConfig loads and merges local and global Task configuration files
 func GetConfig(dir string) (*ast.TaskRC, error) {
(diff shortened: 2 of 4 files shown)


vmaerten:
First of all, I agree with `taskrc.yml ` as filename. 
@andreynering suggested here that we could support both (with dot and without) everywhere : https://github.com/go-task/task/issues/2390#issuecomment-3201144603

## Project without a `.taskrc.yml`
- Walk up the directory tree to find a `.taskrc.yml`.
- Read `$HOME/.taskrc.yml`  if is not a parent of the current directory.
- Read `$XDG_CONFIG_HOME/task/taskrc.yml`.

### Project with a `.taskrc.yml`
- Do not walk up the tree, since the config is already defined locally.
- Optionally merge with `$XDG_CONFIG_HOME/task/taskrc.yml` if it exists.


### Windows-specific case
If a user has a local `.taskrc.yml` in their project but no `XDG_CONFIG_HOME`,  
they cannot define a global Task configuration.  
In this scenario, `$HOME/.taskrc.yml` would **not** be merged with the local config (assuming the project is under `$HOME/xxx/xxxx/xxxx`.
You may need to remove the condition `!strings.Contains(home, dir)`

**Question:**  
Am I correct in this interpretation?  
Do we want to support this use case?

pd93:
Happy to debate if it should work differently, but currently it works like this:

- Walk up the directory tree to find _all_ `.taskrc.yml` files.
  - If current dir is a child of `$HOME` then any `.taskrc.yml` is automatically included in this walk.
  - Otherwise, we explicitly add any `.taskrc.yml` in `$HOME`.
- Look for `taskrc.yml` in `XDG_CONFIG_HOME`

We then gather _all_ the files we found merge them in reverse order. Starting with `$XDG_CONFIG_HOME` (as the lowest priority), then `$HOME`, then files in parent directories and finally any files in the current directory.

> If a user has a local `.taskrc.yml` in their project but no `XDG_CONFIG_HOME`,
> they cannot define a global Task configuration.
> In this scenario, `$HOME/.taskrc.yml` would **not** be merged with the local config (assuming the project is under `$HOME/xxx/xxxx/xxxx`.

I think the misunderstanding here is that it is gathering _all_ files in parent directories, not just the first one it finds. Whether this is correct is up for debate. However, as it stands, if `$XDG_CONFIG_HOME` is not set, we will _always_ use `$HOME` if set - Either via the recursive search or by explicitly pulling it in.

Using the example from the updated docs:

```yaml
# $XDG_CONFIG_HOME/task/taskrc.yml
# lowest priority global config
option_1: foo
option_2: foo
option_3: foo
```

```yaml
# $HOME/.taskrc.yml
option_1: bar
option_2: bar
```

```yaml
# $HOME/path/to/project/.taskrc.yml
# highest priority project config
option_1: baz
```

You would end up with the following configuration:

```yaml
option_1: baz # Taken from $HOME/path/to/project/.taskrc.yml
option_2: bar # Taken from $HOME/.taskrc.yml
option_3: foo # Taken from $XDG_CONFIG_HOME/task/.taskrc.yml
```

vmaerten:
You are totally right! I'm dumb :man_facepalming: 
It's good for me then :) 
Sorry!

Issue #1018, opened 2023-02-22
Allow `:` separator to define multiple paths on a single `sources` entry


eitsupi:
I am using the following Taskfile, but there are several tasks that refer to partially overlapping sources, and I felt it would be useful to be able to specify multiple sources at once.

https://github.com/eitsupi/prqlr/blob/cade3aa1a8cd4c5b1a39b3ea38aea4d95d9e1fdc/Taskfile.yml#L42-L67

```yml
  test-source:
    sources:
      - tests/**/*
      - R/*
      - src/Makevars*
      - src/rust/Cargo.toml
      - src/rust/src/**/*.rs

  test-vignettes:
    sources:
      - vignettes/**/*.Rmd
      - R/*
      - src/Makevars*
      - src/rust/Cargo.toml
      - src/rust/src/**/*.rs
```

I would like to define `R/*`, `src/Makevars*`, `src/rust/Cargo.toml`, `src/rust/src/**/*.rs` somewhere and designate it collectively.

_Originally posted by @eitsupi in https://github.com/go-task/task/discussions/1009_

andreynering:
Hi @eitsupi,

Note that you can use variables to share at least single source entries between tasks:

```yaml
version: '3'

vars:
  GLOB: 'src/**/*.js'

tasks:
  task-A:
    sources:
      - '{{.GLOB}}'

  task-AB:
    sources:
      - '{{.GLOB}}'
```

Of course, that doesn't solve the whole problem given you need multiple entries. I'll keep this issue open so we can consider addressing this in the future.

I think a possible solution would be to allow a separator character (probably `:` to be consistent with PATH on Linux) so multiple paths could be set on a single source entry.

```yaml
version: '3'

vars:
  GLOB: 'path/1.js:path/2.js:path/**/*.js'

tasks:
  task-A:
    sources:
      - '{{.GLOB}}'

  task-AB:
    sources:
      - '{{.GLOB}}'
```

PR #2404, opened 2025-08-28
feat: add support for colon-separated glob patterns in sources

lakshya1goel:
## Description

Adds support for colon-separated glob patterns in the `sources` field, allowing multiple patterns to be specified in a single line for more concise Taskfile syntax.

## Problem

Currently, each glob pattern in `sources` must be specified as a separate array item:

```yaml
tasks:
  build:
    sources:
      - "*.go"
      - "internal/**/*.go" 
      - "cmd/**/*.go"
    cmds:
      - go build
```

## Solution

Enable colon-separated patterns within a single string:

```yaml
tasks:
  build:
    sources:
      - "*.go:internal/**/*.go:cmd/**/*.go"
    cmds:
      - go build
```

## Changes

- **Modified** `internal/templater/templater.go`: Enhanced `ReplaceGlobs` function to split colon-separated patterns
- **Added** integration test `TestColonSeparatedSources`
- **Updated** documentation and JSON schema

## Backwards Compatibility

 **Fully backwards compatible** - existing Taskfiles work unchanged.

## Example

```yaml
vars:
  GO_FILES: "*.go:internal/**/*.go:cmd/**/*.go"

tasks:
  test:
    sources:
      - "{{.GO_FILES}}"
    cmds:
      - go test ./...
```

## Testing

-  Integration test added and passing
-  All existing tests pass
-  Manual testing with various glob patterns


lakshya1goel:
Hi @andreynering, please review this PR, whenever you get some time.
Thank you!

pd93:
Tbh, I really don't like the `foo:bar` syntax for sources. Using a colon as a list separator feels like hidden behaviour and unintuitive when other sources are defined using YAML lists. I'm generally against having 2 ways of doing 1 thing.

IMO, now that we have proper variable types in Task, a better solution to #1018 would be to add support for the `ref` keyword to `sources` and pass a real YAML list into the tasks. For example:

before:
```yaml
version: '3'

vars:
  PATHS:
    - 'path/1.js'
    - 'path/2.js'
    - 'path/**/*.js'

tasks:
  task-A:
    sources:
      - ref: .PATHS

  task-AB:
    sources:
      - ref: .PATHS
      - 'path/3.js'
```

lakshya1goel:
Hmm, I got your point, I will modify the implementation to adapt it to `ref` keyword instead of using : separated paths.

andreynering:
Yeah, I also don't think that using colon for this is the right path.

@lakshya1goel I'm going to close this one, and if you implement the suggested `ref` you can open an new PR with a new branch.

PR #2408, opened 2025-09-03
feat: add `joinEnv` and `joinUrl` string functions and 2 new system vars 

### LINKED ISSUE #2406: joinEnv - Just like joinPath but with ; for windows else :
### Description

```
env:
```

solvingj:
Add two new system variables which map directly to platform specific go constants:
  - PATH_LIST_SEPARATOR -> os.PathListSeparator
  - FILE_PATH_SEPARATOR -> os.PathSeparator

Add joinEnv string function which is sibling to joinPath:
- strings.Join(elem, string(os.PathListSeparator))

Add joinUrl string function which is sibling to joinPath:
- path.Join(elem...)

Add all to docs

closes #2406


trulede (commented):
@solvingj there are quite a few functions already in https://go-task.github.io/slim-sprig/ and perhaps its better to use those existing list functions (join, append, prepend) since that gives a better control over how such lists/strings are generated.

The PathSeparator and PathListSeparator are useful however I would suggest following the existing most-predominate naming convention:

* PathSeparator -> osPathSeparator (note also the absence of "file")
* PathListSeparator -> osPathListSeparator 

since the allcaps just looks ODD.


Observation: this could also be done with a .env file, introducing these constants (and others). Perhaps this PR is more effort than value.

solvingj:
Up to @pd93 and @andreynering I suppose.

vmaerten (changes requested):
Hello!
Sorry for the delay, we have been busy lately.

`PATH_LIST_SEPARATOR` / `FILE_PATH_SEPARATOR` does not work in the templating system.

If you take this Taskfile : 
```yml
# https://taskfile.dev

version: '3'

tasks:
  default:
    cmds:
      - echo "{{.PATH_LIST_SEPARATOR}}"
      - echo "{{.FILE_PATH_SEPARATOR}}"

```
 You will get this result: 

Image

For these variables, you need to define them in `getSpecialVars` inside `compiler.go`

solvingj:
> For these variables, you need to define them in `getSpecialVars` inside `compiler.go`

Ah, of course, moved...

vmaerten:
@solvingj the lint is failing. Do you still want to work on it?

solvingj:
Yah I can fix it tomorrow. Just to confirm, are you sure you want to merge it?

solvingj:
@vmaerten i just added this proposed example to the ENV_PRECEDENCE documentation:
https://github.com/go-task/task/commit/d5431e0705457d72bdb07faa10b454ebd83bb403

Just pointing out that the use case would be a touch more elegant with this feature.

internal/templater/funcs.go
```diff
@@ -94,6 +98,14 @@ func IsSH() bool {
 	return true
 }
 
+func joinEnv(elem ...string) string {
```
trulede, line 102: Why is this called `joinEnv`? It has nothing to do with environment.

There is already a `join` function in slim-sprig which takes  a second argument for the separator. Would this not be more suitable, along with the added seperator (in this PR).

internal/templater/funcs.go
```diff
@@ -94,6 +98,14 @@ func IsSH() bool {
...
+	return strings.Join(elem, string(os.PathListSeparator))
+}
+
+func joinUrl(elem ...string) string {
+	return path.Join(elem...)
```
trulede, line 107: There is already a `joinPath` function. Any reason not to use that? It calls `filepath.Join`.

Also, there is an existing `urlJoin` which is actually joining a URL. And a corresponding `urlParse`. Those are both working with URLs, the function presented here is not...

trulede:
@solvingj why not just add `pathJoinList` alongside the existing `pathJoin`? And don't bother will all the additional stuff.

For your problem, its much neater than dealing with the separators. It seems like it would just solve your problem (sorry I don't exactly know the templating syntax to use here, but you will get the point):

`
export PATH={{pathJoinList env PWD/tools/bin env PATH}}
`


https://github.com/SergioChan/task/blob/54bdcba369357b47e19066b57badfb216a4c8d95/internal/templater/funcs.go#L35

internal/templater/funcs.go
solvingj: > Why is this called `joinEnv`? It has nothing to do with environment.

`os.PathListSeparator` is abstract rune which resolves to the character ":" when compiled on posix operating systems, and ";" on windows. It represents the special value used by these operating systems when parsing 1 or more filesystem paths from an environment variable, such as in the special case of `PATH`. 

This function returns a string, which is the combination of input strings, but with the appropriate platform-specific character placed in-between each. In this way, it returns a string which is valid to be assigned to an environment variable which is designated to hold 1 or more filesystem paths.

Thus, this function is related to environment variables, and that is why it was called `joinEnv`.  

I apologize for not explaining this more clearly in the original description.

internal/templater/funcs.go
solvingj: > There is already a `joinPath` function. Any reason not to use that? It calls `filepath.Join`.

`filepath.Join` is abstract method which takes 1 or more strings understood to be filesystem paths, and combines them by putting the platform-specific separator in-between each element.  When compiled on posix operating systems, the separator used is "/", when compiled on windows, the separator is "\\". 

When joining path elements for a URL, the inputs must be combined together, with the character "/" placed in-between each element.  The `path.Join` function does this. The use of `joinPath` on posix operating-systems returns a valid URL with the "/" character, however when called on windows with the same inputs, the function returns a string with the separator "\\" placed in-between each element.  This result is not a valid URL.  

This is one of two key reason why this function is implemented, and why it used the `path.Join` function instead of `filepath.Join`.

> Also, there is an existing `urlJoin` which is actually joining a URL. And a corresponding `urlParse`. Those are both working with URLs, the function presented here is not...

`urlJoin` is a function which takes an input of type `map` and combines them into a URL based on a number of optional inputs, according to [RFC 3986](https://www.ietf.org/rfc/rfc2718.txt).  It looks for the following keys (with example values provided):
```
scheme:   'http'
host:     'server.com:8080'
path:     '/api'
query:    'list=false'
opaque:   nil
fragment: 'anchor'
userinfo: 'admin:secret'
```

The `path` key supports a single string value.  If the `path` of the url has multiple path elements, such as `api`,`foo`, and `bar`, they must be combined into a single string prior to calling this function. To be valid, they must be combined with the "/" character in-between each element, such as `/api/foo/bar`. 

This is the second of the two key reason why this function is implemented. It provides a first-class way to prepare a `path` string for use with URLs, including for use with the existing `urlJoin` function.

I apologize for not explaining these things more clearly in the original description.

solvingj:
> @solvingj why not just add `pathJoinList` alongside the existing `pathJoin`? And don't bother will all the additional stuff.
> 
> For your problem, its much neater than dealing with the separators. It seems like it would just solve your problem (sorry I don't exactly know the templating syntax to use here, but you will get the point):
> 
> `export PATH={{pathJoinList env PWD/tools/bin env PATH}}`
> 
> https://github.com/SergioChan/task/blob/54bdcba369357b47e19066b57badfb216a4c8d95/internal/templater/funcs.go#L35

This suggested implementation does not result in the desired behavior (it does not pass the included unit tests).  To try to help understand why, I have provided some responses to your other comments and questions in this PR, with some key details about how the functions you mentioned work.  Your confusion is understandable, as the details of these functions are somewhat esoteric, and cross-platform nuances are often not well-understood.  I apologize I did not give more detailed explanations in my original PR description.

andreynering:
Hi @solvingj,

Looks like this build is broken. This needs a rebase.

solvingj:
@andreynering done. All tests pass locally now.

solvingj:
@andreynering any additional blockers?

andreynering (approved):
Thank you for your patience!

PR #2418, opened 2025-09-12
Pass task vars to deferred tasks.

### LINKED ISSUE #2244: Variable definitions are re-evaluated in deferred cmds
### Description

It seems that every time a `defer:` command is run, variable templates are re-evaluated. This is problematic when using nondeterministic variable definitions that include things like the `randInt` builtin. In addition, this breaks from the behavior of normal commands where all executions seem to share the same variable values.

For example, I would expect the example task to print the same path value 4 times, but instead it prints 2 of the same path then 2 more different paths (a total of 3 unique paths).

### Version

3.43.2

### Operating system

Windows

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
tasks:
  test:
    vars:
      DIR: '{{.ROOT_DIR}}/temp-{{randInt 1000000 9999999}}'
    cmds:
      - echo "{{.DIR}}"
      - defer: echo "{{.DIR}}"
      - defer: echo "{{.DIR}}"
      - echo "{{.DIR}}"
```

trulede:
A deferred task was recalculating vars which results in re-evaluation of _templated_ and _sh_ vars. This might not be desired in the case where values change with each evaluation; e.g. template function `randInt`.

This PR passes the task Vars to the deferred task, and removes the code which previously re-evaluated the vars. This works and I do not notice any point in the code where the task Vars would be mutilated (changed) so I did not use the DeepCopy() function of the Vars object when calling `runDeferred()`. I think this will be OK, but I would still review that part again.

fixes #2244

### DIFF
--- a/task.go
+++ b/task.go
@@ -211,7 +211,7 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
 
 		for i := range t.Cmds {
 			if t.Cmds[i].Defer {
-				defer e.runDeferred(t, call, i, &deferredExitCode)
+				defer e.runDeferred(t, call, i, t.Vars, &deferredExitCode)
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


andreynering (approved):
This simply makes total sense. Thank you @trulede!

PR #2430, opened 2025-09-20
docs: Update Remote Taskfiles default values

merusso:
This change updates the documentation for [Remote Taskfiles > Configuration](https://taskfile.dev/docs/experiments/remote-taskfiles#configuration) for `timeout` and `cache-expiry` to match the defaults in code.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -240,7 +240,7 @@ func (o *timeoutOption) ApplyToExecutor(e *Executor) {
 }
 
 // WithCacheExpiryDuration sets the duration after which the cache is considered
-// expired. By default, the cache is considered expired after 24 hours.
+// expired. By default, the cache is 0 (disabled).
 func WithCacheExpiryDuration(duration time.Duration) ExecutorOption {
 	return &cacheExpiryDurationOption{duration: duration}
 }


merusso:
I couldn't tell if this page was generated or not. I did also see this other reference to remote configuration settings, wasn't sure if this needed to be updated also: https://github.com/go-task/task/blob/main/website/src/public/schema-taskrc.json

website/src/docs/experiments/remote-taskfiles.md
```diff
@@ -342,7 +342,7 @@ remote:
 #### `cache-expiry`
 
 - **Type**: `string`
-- **Default**: Not specified
+- **Default**: 24h
```
pd93: The cache expiry defaults to `0s` (i.e. the cache expires immediately and so files are redownloaded every time).

merusso: I was confused by this one. Looking [here](https://github.com/go-task/task/blob/8716ab81bebd84e4905607914584cadf416200a3/internal/flags/flags.go#L157), the `cache-expiry` default is `0`, but looking [here](https://github.com/go-task/task/blob/8716ab81bebd84e4905607914584cadf416200a3/executor.go#L242-L246), the comment suggests a default of `24h`. I couldn't find any actual code setting the default to `24h` though.

Is the comment in executor.go wrong? Should I update it in this PR?

pd93: Ah yeah, the comment is wrong. This used to work differently before #2176 and I guess the comment was never updated. Feel free to fix in this PR :)

merusso:
```suggestion
- **Default**: 0s (no cache)
```

merusso: Okay, thanks, I updated the comment

PR #2432, opened 2025-09-21
chore: delete unused file + fix lint

andreynering:


cmd/tmp/main.go
andreynering, line 1: @pd93 Looks like this file is not needed for anything useful, right? Let me know if I'm wrong.

pd93: Looks like I was just testing something. Who knows what 😆 Probably didn't mean to commit this.

vmaerten (approved):

Issue #2012, opened 2025-01-22
Idea: Make `--init` more powerful by having Taskfile templates


andreynering:
From previous discussion in https://github.com/go-task/task/discussions/2010 with @HeCorr. This is an idea for discussion, and we can discard if we think it'd not be worth it.

Our default Taskfile generated from `--init` is pretty basic. It's basically a skeleton so the user can start adding their own tasks.

In theory we could have some templates for each kind of app: Go, Rust, Ruby on Rails, npm, Yarn, Deno, Docker, etc. IMO they should still be basic, but would help the user by having a starting point.

Implementation wise, we could have a directory in our repository called `templates` with each template, and have that embed in the binary with `go embed`. Once `--init=rails` is called, it'll init with the Rails template, for example (`templates/rails/Taskfile.yaml`).

Additionally, we could allow the user to override these templates with a directory of his own. Not sure how it'd be more practical, if a CLI flag or an ENV like `TASK_EXTRA_TEMPLATES_DIR`, but we would merge those with our default templates.

officel:
Hi!

The other day, go-task was upgraded and the location for the global configuration file changed to XDG Base.
I was hoping that meant the --global taskfile could now be placed in ~/.config/task/Taskfile.yml, but sadly that's not the case.

While pondering this, it occurred to me: maybe I could just place my own base template taskfile in the global tasks directory so it gets loaded?

like this.
https://github.com/officel/config_task/blob/main/Taskfile.yml#L60

If you're waiting for a feature that prepares multiple templates and deploys them during init, I figured it would be faster to just build it into the task itself.

trulede:
Could we add a template to the taskrc file perhaps? Or even a list of named templates, from which one might be selected.

```
task --init  # Use the default template from taskrc, if defined, otherwise use the task default.
task --init=foo  # Use the named template.
task --init=  # Lists all named templates from taskrc, same if the named template did not exist, so helpful fail.
```

B13rg:
I tried adding a flag `--init-template`, but it seemed a little long to type and 
Creating the various templates should be a separate  task. 


Add an optional env var like `TASK_INIT_FILE` to override the template file created with the plain `--init` command.
Maybe also support setting a directory and use same filename search logic as elsewhere. -> #2814 



IF there's effort to create some community Taskfile templates 
[...]

PR #2433, opened 2025-09-21
fix: a couple of fixes and improvements on `task --init`

andreynering:
* Fixed check for an existing Taskfile: look for all possibilities, and not only `Taskfile.yml` specifically.
* Added a description (`desc`) to the `default` task. Important to at lest `task --list` work by default (a core feature).
* ~Changed default extension from `.yml` to `.yaml`.~
* ~Added `yaml-language-server` comment.~

### DIFF
--- a/taskfile/templates/default.yml
+++ b/taskfile/templates/default.yml
@@ -1,12 +1,13 @@
-# https://taskfile.dev
+# yaml-language-server: $schema=https://taskfile.dev/schema.json
 
 version: '3'
 
 vars:
-  GREETING: Hello, World!
+  GREETING: Hello, world!
 
 tasks:
   default:
+    desc: Print a greeting message
     cmds:
       - echo "{{.GREETING}}"
     silent: true
--- a/init.go
+++ b/init.go
@@ -6,9 +6,10 @@ import (
 
 	"github.com/go-task/task/v3/errors"
 	"github.com/go-task/task/v3/internal/filepathext"
+	"github.com/go-task/task/v3/taskfile"
 )
 
-const defaultTaskFilename = "Taskfile.yml"
+const defaultFilename = "Taskfile.yml"
 
 //go:embed taskfile/templates/default.yml
 var DefaultTaskfile string
@@ -20,22 +21,30 @@ var DefaultTaskfile string
 //
 // The final file path is always returned and may be different from the input path.
 func InitTaskfile(path string) (string, error) {
-	fi, err := os.Stat(path)
-	if err == nil && !fi.IsDir() {
+	info, err := os.Stat(path)
+	if err == nil && !info.IsDir() {
 		return path, errors.TaskfileAlreadyExistsError{}
 	}
 
-	if fi != nil && fi.IsDir() {
-		path = filepathext.SmartJoin(path, defaultTaskFilename)
-		// path was a directory, so check if Taskfile.yml exists in it
-		if _, err := os.Stat(path); err == nil {
+	if info != nil && info.IsDir() {
+		// path was a directory, check if there is a Taskfile already
+		if hasDefaultTaskfile(path) {
 			return path, errors.TaskfileAlreadyExistsError{}
 		}
+		path = filepathext.SmartJoin(path, defaultFilename)
 	}
 
 	if err := os.WriteFile(path, []byte(DefaultTaskfile), 0o644); err != nil {
 		return path, err
 	}
-
 	return path, nil
+}
+
+func hasDefaultTaskfile(dir string) bool {
+	for _, name := range taskfile.DefaultTaskfiles {
+		if _, err := os.Stat(filepathext.SmartJoin(dir, name)); err == nil {
+			return true
+		}
+	}
+	return false
 }
--- a/taskfile/node_file.go
+++ b/taskfile/node_file.go
@@ -19,7 +19,7 @@ type FileNode struct {
 
 func NewFileNode(entrypoint, dir string, opts ...NodeOption) (*FileNode, error) {
 	// Find the entrypoint file
-	resolvedEntrypoint, err := fsext.Search(entrypoint, dir, defaultTaskfiles)
+	resolvedEntrypoint, err := fsext.Search(entrypoint, dir, DefaultTaskfiles)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
 			return nil, errors.TaskfileNotFoundError{URI: entrypoint, Walk: false}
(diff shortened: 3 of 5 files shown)


taskfile/templates/default.yaml
```diff
@@ -0,0 +1,15 @@
+# https://taskfile.dev
+#
+# yaml-language-server: $schema=https://taskfile.dev/schema.json
```
andreynering, line 1: I'm a bit unsure about this one. It's not actually needed for `--init` because `Taskfile.yaml` is recognized by default. Not having it makes it clean.

On the other hand, it's useful to inform users that they can do this to keep the YAML LSP working for non-standard names. This would be useful for non-standard names used by `task -t foo.yaml` or Taskfiles added to `includes`.

andreynering:
@pd93 @vmaerten This is a bit oppinated, so I want your opinions on this one. Feel free to disagree or suggest changes.

This is the default Taskfile for years. Needed some love for a while. We can eventually consider changing it completely, but in that case we'll need to update [the documentation](https://taskfile.dev/docs/getting-started) as well. Perhaps the right moment for it would be on:

* https://github.com/go-task/task/issues/2012

pd93:
I'm fine with most of this. I'm not convinced about `yaml` over `yml` though. We've had the `yml` convention for a long time and most Taskfiles I've seen use this convention - including our own repo. Most YAML config files for other tools also typically prefer `yml` too (Mockery, GoReleaser, GolangCI etc). What would be the reason to change this now?

taskfile/templates/default.yml
```diff
@@ -0,0 +1,15 @@
+# https://taskfile.dev
+#
+# yaml-language-server: $schema=https://taskfile.dev/schema.json
+
+version: '3'
```
pd93, line 3: We could consider setting this to the current version of the binary creating the file. This ensures that the same features will be available for the person initialising the file and the people reading it.

It would also encourage the use of this feature which I think is underused. Most people just put `3` despite requiring a much more recent version.

taskfile/templates/default.yaml
trulede: Its 3 more lines to delete each time I use the --init command option ...

If someone wanted that, they could setup a template. Would be useful to document.

andreynering:
OK, I reverted the default extension change and the language server comment.

Still kept the addition of `desc` and the fix to look for all default names on `--init`.

taskfile/templates/default.yml
andreynering: This is an interesting idea.

Would we want to add as `3.45.4` or `3.45`?

taskfile/templates/default.yaml
pd93: Taskfiles are recognized by default, but only if they have one of the default Taskfile names. I often include files with other names like `foo.yml` and the directives at the top are useful for instructing the YAML server. I don't think the link to `taskfile.dev` is necessary, but I'd leave the YAML server instruction personally.

It's much easier for a user who knows what they're doing to remove a line than it is for a new user who has no idea what a YAML directive/schema is to add it.

taskfile/templates/default.yml
pd93: `3.45` is probably enough since we only add features on minor revisions anyway. The chances are that anyone getting a message to upgrade will immediately install the latest version instead of the minimum version anyway.

taskfile/templates/default.yaml
vmaerten: My two cents
I agree with @pd93 
I think `taskfile.dev` is not necessary, I'd even say it adds more noise than value. I’m also in favor of keeping the YAML server instruction.
Two days ago, I gave a talk about Task, I created the file with `task --init` and I had to serch in our website to find the YAML server instruction

trulede:
@andreynering here is an idea, perhaps you want an ENVAR/taskrc-file to allow setting of default taskfile names, which if present overrides this list:
https://github.com/go-task/task/blob/4e84c6bb766e10a6a9b4b08bfe75519a240ceff8/taskfile/taskfile.go#L15 
Then, to get the taskfile name to use with the `--init` option, take the first item from that list.

```
$ TASK_TASKFILE_DEFAULT_NAMES = Taskfile.yaml;Taskfile.yml
$ task --init
$ ls
Taskfile.yaml
```

taskfile/templates/default.yml
trulede: A different perspective: unless a taskfile _needs_ a specific version, then doing this will become an annoyance when people are working/sharing/using different versions of task/taskfiles. We don't use the feature because its not very useful for us ... the version of Task in use is already managed, which is sufficient.

pd93: The version key is not intended to dictate a _specific_ version. It is a _minimum_ version required for the features in the Taskfile. The vast majority of Taskfiles will require a minimum version since they'll use features introduced after v3.0.0. For instance, any Taskfile that uses aliases would need to have a minimum version of 3.17.

This is extremely useful as instead of receiving a confusing error when they use a version of Task that is too old, they'll receive a message that tells them to update.

I think the main reason this is underused is because it didn't work this way until v3.34 and is underdocumented. None of our examples/tests use it and the tool itself doesn't advertise the feature. Adding this to the `--init` flag is analogous to `go mod init` adding the current version of `Go` to the file.

pd93:
@trulede You can already change the name of the created file using the `-t`/`--taskfile` flag. e.g.

```shell
task --init --taskfile foo.yml
```

vmaerten (commented):
> OK, I reverted the default extension change and the language server comment.
> 
> Still kept the addition of `desc` and the fix to look for all default names on `--init`.

@andreynering could we reconsider the language server comment?

To quote both Pete and me:

> but I'd leave the YAML server instruction personally.
> 
> It's much easier for a user who knows what they're doing to remove a line than it is for a new user who has no idea what a YAML directive/schema is to add it.

> Two days ago, I gave a talk about Task, I created the file with task --init with a custom name and I had to serch in our website to find the YAML server instruction


Otherwise, I'm fine with the other changes and looking forward to seeing this get merged :fire:

taskfile/templates/default.yaml
andreynering: OK, I removed the `taskfile.dev` comment and added the YAML language server back.

taskfile/templates/default.yml
andreynering: I'll keep as is for now so we can move with the other fixes.

We can still do that in a follow-up PR, though.

Issue #2441, opened 2025-09-24
Unhelpful error with bad include


maxmzkrcensys:
### Description

When you have a taskfile that includes a directory that exists, but no taskfile exists in the directory, you get the error
```
file does not exist
```

Before c5afffb55114f7e0a68d383bcd01159abfaf1ae5, the error was
```
task: No Taskfile found at "/empty/directory"
```

### Version

3.45.4

### Operating system

ubuntu

### Example Taskfile

```YAML

```

PR #2442, opened 2025-09-24
fix: improve missing taskfile error message

maxmzkrcensys:
fixes https://github.com/go-task/task/issues/2441

Leaving some inline comments about things I'm not thrilled about and happy to fix if they are issues!

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1030,8 +1030,7 @@ func TestIncludesOptionalImplicitFalse(t *testing.T) {
 	const dir = "testdata/includes_optional_implicit_false"
 	wd, _ := os.Getwd()
 
-	message := "stat %s/%s/TaskfileOptional.yml: no such file or directory"
-	expected := fmt.Sprintf(message, wd, dir)
+	expected := fmt.Sprintf("task: No Taskfile found at %q", filepath.Join(wd, dir))
 
 	e := task.NewExecutor(
 		task.WithDir(dir),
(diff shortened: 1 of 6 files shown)


taskfile/node_file.go
```diff
@@ -20,8 +21,23 @@ type FileNode struct {
 func NewFileNode(entrypoint, dir string, opts ...NodeOption) (*FileNode, error) {
 	var err error
 	base := NewBaseNode(dir, opts...)
+	originalDir := base.dir
 	entrypoint, base.dir, err = fsext.Search(entrypoint, base.dir, defaultTaskfiles)
 	if err != nil {
+		// Check if this is a directory that exists but has no taskfile
+		if os.IsNotExist(err) {
```
maxmzkrcensys, line 28: tbh, I don't really understand what entrypoint and dir are. I followed the logic that search does. If there is a better thing to use, then I can use that

testdata/includes_empty_directory/empty_dir/.gitignore
```diff
@@ -0,0 +1,2 @@
+# This file exists to make the directory committable
+# The directory should remain empty of taskfiles for the test
```
maxmzkrcensys, line 2: let me know if there is a preferred way to have an empty directory

task_test.go
```diff
@@ -1030,8 +1030,7 @@ func TestIncludesOptionalImplicitFalse(t *testing.T) {
 	const dir = "testdata/includes_optional_implicit_false"
 	wd, _ := os.Getwd()
 
-	message := "stat %s/%s/TaskfileOptional.yml: no such file or directory"
-	expected := fmt.Sprintf(message, wd, dir)
+	expected := fmt.Sprintf("task: No Taskfile found at %q", filepath.Join(wd, dir))
```
maxmzkrcensys, line 1033: are we okay with this error message changing? I can attempt to change it so that we get no such file or directory errors when the directory doesn't exist. The search function returns the same error if the file doesn't exist and if the directory doesn't exist. So I would need to either change the errors that search returns or inspect upon getting an error

maxmzkrcensys:
Sorry, I'm going to reopen a new pr with a better base branch. And fix merge conflicts

PR #2445, opened 2025-09-27
FIx calculation of task.Dir for tasks in included taskfiles.

### LINKED ISSUE #2443: Absolute path bug when using include and dir
### Description

When setting an absolute directory (ex. /tmp) from a variable and using that in the `dir:` directive of an included taskfile, it is treated as a relative path.  However, it is dependent on the syntax of the include.  Minimal working example included below. 

When included using this syntax, the variable is treated as a relative path:
```
vars:
  TARGET_DIR: '/tmp'
includes:
  broken: 
    taskfile: ./Taskfile.broken.yml
```

When used like this:
```
tasks:
  broken:
    dir: "{{.TARGET_DIR}}"
    cmds:
      - |
          echo "PWD DIR: $(pwd)" 
```

### Version

3.45.4

### Operating system

OSX 26

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
# Taskfile.yml
version: '3'

vars:
  TARGET_DIR: '/tmp'

includes:
  working: ./Taskfile.working.yml
  broken: 
    taskfile: ./Taskfile.broken.yml
    # this also doesn't fix it
    #vars:
    #  TARGET_DIR: "{{.TARGET_DIR}}"
tasks:
  default:
    cmds:
      - task: working:working
      - task: broken:broken
```

```YAML
# Taskfile.working.yml
version: '3'

tasks:
  working:
    dir: "{{.TARGET_DIR}}"
    cmds:
      - |
          echo "TARGET_DIR: {{.TARGET_DIR}}"
          echo "PWD DIR: $(pwd)"          
          cd {{.TARGET_DIR}}
          echo "PWD AFTER CD: $(pwd)"
```

```YAML
# Taskfile.broken.yml
version: '3'

tasks:
  broken:
    dir: "{{.TARGET_DIR}}"
    cmds:
      - |
          echo "TARGET_DIR: {{.TARGET_DIR}}"
          echo "PWD DIR: $(pwd)"          
          cd {{.TARGET_DIR}}
          echo "PWD AFTER CD: $(pwd)"
```

trulede:
Relocate the calculation of an included tasks task.Dir until the templating is available. 

Previously, task.DIr would be calculated earlier which resulted in SmartJoin() incorrectly evaluating the second parameter, which had not been expanded (i.e. {{.FOO}}. To get correct behaviour, SmartJoin() needed to be called later, during compilation of the imported task, so that the second parameter could first be expanded, and then used correctly with SmartJoin().

fixes #2443

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -91,17 +91,6 @@ func (c *Compiler) getVariables(t *ast.Task, call *Call, evaluateShVars bool) (*
 	rangeFunc := getRangeFunc(c.Dir)
 
 	var taskRangeFunc func(k string, v ast.Var) error
-	if t != nil {
-		// NOTE(@andreynering): We're manually joining these paths here because
-		// this is the raw task, not the compiled one.
-		cache := &templater.Cache{Vars: result}
-		dir := templater.Replace(t.Dir, cache)
-		if err := cache.Err(); err != nil {
-			return nil, err
-		}
-		dir = filepathext.SmartJoin(c.Dir, dir)
-		taskRangeFunc = getRangeFunc(dir)
-	}
 
 	for k, v := range c.TaskfileEnv.All() {
 		if err := rangeFunc(k, v); err != nil {
@@ -119,6 +108,22 @@ func (c *Compiler) getVariables(t *ast.Task, call *Call, evaluateShVars bool) (*
 				return nil, err
 			}
 		}
+
+		// NOTE: Calculate the task.Dir and taskRangeFunc here, after
+		// IncludeVars have been added to the templater. The call to SmartJoin()
+		// will then work as expected (i.e. when `dir` becomes an absolute path
+		// after the templater.Replace() call).
+		// NOTE: taskRangeFunc() will be available to subsequent calls
+		// in this function despite the convoluted code structure here.
+		cache := &templater.Cache{Vars: result}
+		dir := templater.Replace(t.Dir, cache)
+		t.Dir = filepathext.SmartJoin(t.IncludeDir, dir)
+		if err := cache.Err(); err != nil {
+			return nil, err
+		}
+		dir = filepathext.SmartJoin(c.Dir, dir)
+		taskRangeFunc = getRangeFunc(dir)
+
 		for k, v := range t.IncludedTaskfileVars.All() {
 			if err := taskRangeFunc(k, v); err != nil {
 				return nil, err
--- a/taskfile/ast/task.go
+++ b/taskfile/ast/task.go
@@ -44,6 +44,7 @@ type Task struct {
 	Location      *Location
 	// Populated during merging
 	Namespace            string
+	IncludeDir           string
 	IncludeVars          *Vars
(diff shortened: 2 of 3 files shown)


trulede:
I put this as a draft until there is an indication that a fix is wanted.

trulede:
After working some more on this PR, all but one test are passing.

```
=== FAIL: . TestIncludedTaskfileVarMerging/foo (0.03s)
    task_test.go:1306: 
        	Error Trace:	/Users/runner/work/task/task/task_test.go:1306
        	Error:      	"foo\n/Users/runner/work/task/task/testdata/included_taskfile_var_merging/bar\n" does not contain "included_taskfile_var_merging/foo\n"
        	Test:       	TestIncludedTaskfileVarMerging/foo
```

This is an unexpected failure, as var merging was not directly altered. Some investigation will be required to figure out if the test is correct/false, or the behaviour somehow changed.

trulede:
@andreynering @vmaerten In this PR I've relocated the point where an imported task.Dir is calculated, and as a result it is possible to use template variables in the `include.dir`.  This is certainly helpful as many users of the Import feature encouter this limitation as some point and it is confusing as to why it does not work.

However, some tests (2) are proving to be remarkably flakey. These are sometimes passing, sometimes failing.
* TestIncludedTaskfileVarMerging/foo ([taskfile](https://github.com/go-task/task/blob/main/testdata/included_taskfile_var_merging/foo/Taskfile.yaml))
* TestIncludedTaskfileVarMerging/bar ([taskfile](https://github.com/go-task/task/blob/main/testdata/included_taskfile_var_merging/bar/Taskfile.yaml))

These test cases have two imported Taskfiles, each with the same global var `DIR`, and each setting that var to a different value. I wonder if there is a data-race (existing or introduced).

You might have ideas why that is the case? It seems unrelated to the effects of #2350, and somehow related to this PR. In either case I would likely add some additinal testcases to cover the change in functionality added by this PR.

andreynering:
Hey @trulede!

First of all, thanks for your patience! I know if have many PRs opened and we are overdue on reviewing them.

I debugged this flaky test a little bit, but couldn't quickly find the problem, and want to focus on other PRs for today.

But to add context, this was originally fixed by @pd93 on the following PR, and your changes probably made that fix insufficient for some reason.

* https://github.com/go-task/task/pull/1649

trulede:
@andreynering OK, thanks. I will take a look.

trulede:
@andreynering I've sorted out the flakiness and added a test covering the specific scenario (of the PR, not the flake).

If the var being referenced is "dynamic", the outcome might be unexpected (possibly similar to the current behaviour). Its not clear if that could be resolved since dynamic vars are resolved on later passes (of the var stack) and they need a working directory to be resolved (which to use ....).

That limitation seems reasonable with the current implementation. It might be resolved with refactoring of the import mechanism.

trulede:
I will revisit this after #2666 is resolved. I believe it can be fixed in a novel way, in the same code area.

PR #2452, opened 2025-10-02
fix: make list tasks dry to avoid override checksum

johnvox:
When using `--json` option `--list-all` commands update checksum which is not desirable

### DIFF
--- a/help.go
+++ b/help.go
@@ -60,6 +60,7 @@ func (o ListOptions) Filters() []FilterFunc {
 // The function returns a boolean indicating whether tasks were found
 // and an error if one was encountered while preparing the output.
 func (e *Executor) ListTasks(o ListOptions) (bool, error) {
+	e.Dry = true
 	tasks, err := e.GetTaskList(o.Filters()...)
 	if err != nil {
 		return false, err


vmaerten:
Hello!
Thanks for your contribution, I've merged #2053 which should also solve this issue

vyivel:
https://github.com/go-task/task/pull/2053 didn't fix the problem.

In `ListTasks()`, `e.ToEditorOutput` is called:

```go
	if o.FormatTaskListAsJSON {
		output, err := e.ToEditorOutput(tasks, o.NoStatus, o.Nested)
		if err != nil {
			return false, err
		}
		// ...
	}
```

…which in turn checks whether the tasks are up to date with `e.Dry` set to `false`.

```go
	// Look over each task in parallel and turn it into an editor task
	for i := range tasks {
		g.Go(func() error {
			editorTask := editors.NewTask(tasks[i])
			// ...
			upToDate, err := fingerprint.IsTaskUpToDate(context.Background(), tasks[i],
				fingerprint.WithMethod(method),
				fingerprint.WithTempDir(e.TempDir.Fingerprint),
				fingerprint.WithDry(e.Dry),
				fingerprint.WithLogger(e.Logger),
			)
			if err != nil {
				return err
			}
			// ...
		})
	}
```

This results in https://github.com/go-task/vscode-task/issues/200, which I have fixed by applying the patch locally for now. As such, please consider reopening this PR.

PR #2454, opened 2025-10-05
docs: update Arch and Nix community links

panchoh:
I’ve updated links to Arch/AUR and Nix/NixOS.

See the individual commits for context on each one.

First PR, yay!

Thanks for this project, which I use every single day <3


vmaerten:
Thanks!

PR #2474, opened 2025-10-24
feat: Command timeouts

### LINKED ISSUE #1569: Timeout for running tasks
Be able to specify a run timeout, for a task and/or a command in the task.

Eg i'm using task to run various commands in a pipeline. In certain cases there are issues that are not properly handled by the underlying command, keeping the command (and thus task) hanging. By providing a timeout, we can bail out with an error after eg 10 min.

lox:
## Summary

Add command-level timeout support to prevent commands from hanging indefinitely in task pipelines.

Closes #1569

## Usage

```yaml
version: '3'

tasks:
  deploy:
    cmds:
      - cmd: npm run build
        timeout: 5m

      - cmd: ./deploy.sh
        timeout: 30m
```

Commands exceeding their timeout are terminated:

```
task: Failed to run task "deploy": task: [deploy] command timeout exceeded (5m): context deadline exceeded
```

Implementation Details

- Timeout applies per command, not per task
- Works with both shell commands and task calls
- Deferred commands use separate context and run even if parent times out
- ignore_error flag does not suppress timeout errors
- Zero/unspecified timeout maintains current behavior (no timeout)

### DIFF
--- a/taskfile/ast/cmd.go
+++ b/taskfile/ast/cmd.go
@@ -1,6 +1,8 @@
 package ast
 
 import (
+	"time"
+
 	"gopkg.in/yaml.v3"
 
 	"github.com/go-task/task/v3/errors"
@@ -19,6 +21,7 @@ type Cmd struct {
 	IgnoreError bool
 	Defer       bool
 	Platforms   []*Platform
+	Timeout     time.Duration
 }
 
 func (c *Cmd) DeepCopy() *Cmd {
@@ -36,6 +39,7 @@ func (c *Cmd) DeepCopy() *Cmd {
 		IgnoreError: c.IgnoreError,
 		Defer:       c.Defer,
 		Platforms:   deepcopy.Slice(c.Platforms),
+		Timeout:     c.Timeout,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


taskfile/ast/cmd.go
```diff
@@ -62,10 +66,20 @@ func (c *Cmd) UnmarshalYAML(node *yaml.Node) error {
...
 			return errors.NewTaskfileDecodeError(err, node)
 		}
+
+		// Parse timeout if specified
+		if cmdStruct.Timeout != "" {
+			timeout, err := time.ParseDuration(cmdStruct.Timeout)
+			if err != nil {
+				return errors.NewTaskfileDecodeError(err, node).WithMessage("invalid timeout format")
+			}
+			c.Timeout = timeout
```
trulede, line 81: Should the timeout value (string) also support templating?

timeout: '{{.SOME_TIMEOUT}}'

vmaerten:
Hello !
I think it's a good feature to have, are you wiliing to rebase ?
Sorry for the late review

lox:
Sorry, no, I've moved on to other things.

PR #2475, opened 2025-10-25
chore: pin golangci-lint tool version

lox:
Pinned golangci-lint as a Go tool so my first contribution wouldn’t turn into a hunt for the "right" version instead of actual code, and in the process captured its whole dependency graph to keep lint runs hermetic and reproducible for everyone.

The exact commands are:

```
go get -tool github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.0
go tool golangci-lint run ./...
```

Because it causes so much churn to go.mod, I'd suggest a core contributor verify the changes!

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,10 +22,11 @@ jobs:
 
       - uses: actions/checkout@v5
 
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v8
-        with:
-          version: v2.1.0
+      - name: Download modules
+        run: go mod download
+
+      - name: Run golangci-lint
+        run: go tool golangci-lint run ./...
 
   lint-jsonschema:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


go.mod
andreynering, line 1: golangci-lint is a huge dependency. I wonder if it’s viable to extract tools into a separate Go module.

lox: Would need a wrapper script, at which point might be easier to use https://github.com/cashapp/hermit for `bin/golangci-lint`.

trulede: How hard is it to just install version 2.1.0 of golangci-lint?

`curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.1.0`

Having a task which does that, to setup a dev environment, might be nice. However, users of Task don't want to install all those modules.

lox: It's a huge pain to have to swap your version of golangci-lint for every go project you use. 

Agree the mod changes are heavy.

PR #2489, opened 2025-11-01
docs: add tasks.task.dir

### LINKED ISSUE #2481: Property "tasks.task.dir" missing from reference documentation
### Description

Hi,

It seems that the `dir` property of a task is not documented in the [reference](https://taskfile.dev/docs/reference/schema#task), even though it is mentioned in the [guide](https://taskfile.dev/docs/guide#running-a-taskfile-from-a-subdirectory).

I believe the property should be present in the reference.

There may be other undocumented properties but it is the only one I've noticed so far.

If you confirm this issue is valid, then I'd be happy to provide a PR.

Thanks!

SamK:
This PR documents the `dir` property at the task level.

Maybe should I add a default value? I was thinking about `ROOT_DIR` or `{{.ROOT_DIR}}` or "the location of the root Taskfile" but couldn't find anything suitable. Let me know if you want one.

Closes #2481

### DIFF
--- a/website/src/docs/reference/schema.md
+++ b/website/src/docs/reference/schema.md
@@ -614,6 +614,21 @@ tasks:
       - ./deploy.sh
 ```
 
+### `dir`
+
+- **Type**: `string`
+- **Description**: The directory in which this task should run
+- **Default**: If the task is in the root Taskfile, the default `dir` is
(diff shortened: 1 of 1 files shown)


SamK:
I added a _Default_ section the best I could. Let me know what you think.

Also rebased on top of main.

vmaerten:
Thanks for your contribution!

PR #2491, opened 2025-11-02
feat: add --trust CLI and remote.trust config for remote tasks

### LINKED ISSUE #2473: Remote Taskfiles: add a new trust mechanism
### Description

Hey!

This is a follow-up of my comment here: https://github.com/go-task/task/issues/1317#issuecomment-3386352226

In a [Remote Taskfiles experiment ](https://github.com/go-task/task/issues/1317), I would like to propose adding a new `--trust` CLI flag and `remote.trust` configuration option.

Currently, when using remote taskfiles, one is asked for a confirmation. This new flag/setting will bypass that.

This is different than `--yes`, because:
1. It can be configured in taskrc
2. It doesn't answer `yes` for all prompts

# Example case

1. I have a quite large system, structured in a modular way, with many taskfiles.
2. I have some generic taskfiles placed in a separate git repository, so they can be reused in many places.
3. I trust that generic taskfiles - they are managed by the same team of people.
4. If I was care about preventing supply chain attack, which I am not really, I could use checksum feature
5. I don't want to force people to answer `yes` or add `--yes` to every invocations, especially as there are some potentially destructive tasks protected with additional prompt so `--yes` may be dangerous, when used without thinking
6. Certainly, I don't want to add `--yes` to my CI/CD pipelines (read previous point)

maciej-lech:
This PR implements the feature discussed in #2473.

It introduces a new --trust CLI flag and a corresponding remote.trust configuration option in taskrc. These options are available when the Remote Taskfiles experiment is enabled.

The --trust flag can be specified multiple times to define trusted hosts (optionally including ports). Any remote Taskfiles fetched from these trusted hosts will not prompt for confirmation on their initial download or when their checksums change.

Closes #2473

### DIFF
--- a/website/src/docs/experiments/remote-taskfiles.md
+++ b/website/src/docs/experiments/remote-taskfiles.md
@@ -216,8 +216,8 @@ interactive terminal, so you are not able to accept a prompt. In these cases you
 are able to tell task to accept these prompts automatically by using the `--yes`
 flag or the `--trust` flag. The `--trust` flag allows you to specify trusted
 hosts for remote Taskfiles, while `--yes` applies to all prompts in Task. You
-can also configure trusted hosts in your [taskrc configuration](#trust) using
-`remote.trust`. Before enabling automatic trust, you should:
+can also configure trusted hosts in your [taskrc configuration](#trusted-hosts) using
+`remote.trusted-hosts`. Before enabling automatic trust, you should:
 
 1. Be sure that you trust the source and contents of the remote Taskfile.
 2. Consider using a pinned version of the remote Taskfile (e.g. A link


curtbushko:
Looking forward to this feature!

task_test.go
```diff
@@ -784,6 +785,11 @@ func TestIncludesRemote(t *testing.T) {
 
 			var buff SyncBuffer
 
+			// Extract host from server URL for trust testing
+			parsedURL, err := url.Parse(srv.URL)
+			require.NoError(t, err)
+			trustedHost := parsedURL.Host
```
SamK, line 793: Can't I trust URLs? I trust github.com/myself but I don't trust github.com/shady.

maciej-lech: I was thinking about that, but using URLs requires more assumptions how to compare provided URL with the `trust` config.

1. Exact match, so full URL comparison. In my case it would require configuring every single remote taskfiles (more than dozen now) which is not a big deal, but may not be a best DX.
1. Prefix match. A problem: I want to trust `https://github.com/myself` but not `https://github.com/myselfHackedByShady` - which could be easily solved by setting `https://github.com/myself/` and not `https://github.com/myself`. So maybe this is the best way.
1. Glob-like style: `https://github.com/myself/*` or extended version `https://github.com/myself/**/*`
1. Regex: `https:\/\/github\.com\/myself\/.*`

taskrc/ast/taskrc.go
```diff
@@ -43,6 +44,31 @@ func (t *TaskRC) Merge(other *TaskRC) {
...
+		for _, host := range t.Remote.Trust {
+			if !seen[host] {
+				seen[host] = true
+				merged = append(merged, host)
+			}
+		}
+
+		t.Remote.Trust = merged
+	}
+
```
vmaerten: Since the order doesn’t matter (a host is trusted regardless of its position in the slice), we can simplify it to:


```suggestion
	if len(other.Remote.Trust) > 0 {
		merged := slices.Concat(other.Remote.Trust, t.Remote.Trust)
		slices.Sort(merged)
		t.Remote.Trust = slices.Compact(merged)
	}

```

website/src/docs/experiments/remote-taskfiles.md
```diff
@@ -353,3 +359,38 @@ remote:
...
+
+You can also specify trusted hosts via the command line:
+
+```shell
+# Trust specific host for this execution
+task --trust github.com -t https://github.com/user/repo.git//Taskfile.yml
+
+# Trust multiple hosts
+task --trust github.com --trust gitlab.com -t https://github.com/user/repo.git//Taskfile.yml
+
```
vmaerten:
```suggestion
task --trust github.com --trust gitlab.com -t https://github.com/user/repo.git//Taskfile.yml
task --trust github.com,gitlab.com -t https://github.com/user/repo.git//Taskfile.yml
```

maciej-lech:
> Thanks for your PR! I would prefer using `trusted-hosts` instead of `trust` (both in the CLI and in the config file).
> 
> * It's more explicit, even if it's longer
> * Tools like pip already use this flag (see: https://pip.pypa.io/en/stable/cli/pip/#cmdoption-trusted-host)
> * Using the CLI with comma-separated values (`--trusted-hosts a.com,b.com`) avoids repeating the flag name

Thank you for your feedback.  Yes, I stumbled across PIP when I was searching for similar implementations. Although, `--trusted-hosts` in PIP is more like per-domain `--insecure`, I fully agree it is more explicit. Comma-separated imho is also fine. Let me update this PR

internal/flags/flags.go
```diff
@@ -230,6 +233,16 @@ func (o *flagsOption) ApplyToExecutor(e *task.Executor) {
...
 
+	// Parse comma-separated trusted hosts
+	var trustedHosts []string
+	if TrustedHosts != "" {
+		for host := range strings.SplitSeq(TrustedHosts, ",") {
+			if trimmed := strings.TrimSpace(host); trimmed != "" {
+				trustedHosts = append(trustedHosts, trimmed)
+			}
+		}
+	}
```
vmaerten: You do not need this, pflag already support it with `StringSliceVar` cf: https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar

vmaerten: So, we can simplify and use `StringSliceVar`. This way we support either `--trusted-hosts="a.com,b.com"` `either --trusted-hosts="a.com" --trusted-hosts="b.com"`

maciej-lech: Ohh, I didn't know that. Thanks :)

PR #2493, opened 2025-11-04
Watch: restarts tasks only when at least one watched file changes.

### LINKED ISSUE #2477: Watch hijacking dependency process
### Description

Given the following tasks: 

```yaml
  provision:
    desc: "Bring up some dependencies and write the result to .env file"
  ...

  build:
    desc: "Build the service binaries"
    sources:
      - '**/*.go'
      - 'go.mod'
      - 'go.sum'
    generates:
      - './out/bin/server'
      - './out/bin/client'
    cmds:
      - task: provision
      - go build -o ./out/bin/ ./cmd/...
    method: timestamp

  run:
    desc: "Run the services"
    dotenv: ['.env']
    deps: [run:server, run:client]
    watch: true
    sources:
      # do not include .env as it is auto-generated by the provision task
      - exclude: '.env'
      - '*.sql'
    method: none

  run:server:
    deps: [build]
    cmd: ./out/bin/server

  run:client:
    deps: [build]
    cmd: ./out/bin/client
```

Running `task run -v` seems to yield a deadlock of some kind, where we have: 

```
task: "provision" finished
task: [build] go build -o ./out/bin/ ./cmd/...
task: received watch event: WRITE         "/Users/myself/Projects/myproject/.env"
task: [run:server] ./out/bin/server
task: skipped for file not in sources: .env
```

However, running `task run:server` works just fine. 

It seems that, for some reason, having `watch` turned on makes the `run:server` step not wait for the `build` step any longer, so the build doesn't even get to finish, but it tries to run the binary which has not yet been created. 

Haven't been able to figure out why this behavior happens though. 

### Version

3.45.4

### Operating system

MacOS 15.7.1

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

trulede:
When a task(s) is watched, and a watched event occurs, then; the current context is cancelled (which signals the task and sub-tasks), then the watch conditions are evaluated, and finally if the watch conditions are satsified, then the task(s) are restarted.

This means that if a watch condition is _not_ satisified (i.e. "skipped for file not in sources") then the task is cancelled _and not restarted_.

In this PR the logic is altered as follows: First the watch condition is evaluated, if at least one task(s) satisifies the condition then _all_ tasks are cancelled, and then _all_ tasks restarted. When no task satisified the watch contition, no action is taken. Given the way this code behaves, I believe its an improvement within the scope of the known limitations of the method used (i.e. the Context object).

closes #2477

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -78,41 +78,46 @@ func (e *Executor) watchTasks(calls ...*Call) error {
 				}
 				e.Logger.VerboseErrf(logger.Magenta, "task: received watch event: %v\n", event)
 
-				cancel()
-				ctx, cancel = context.WithCancel(context.Background())
-
-				e.Compiler.ResetCache()
-
+				// Count the tasks that would be restarted by the watch (_before_ cancelling them).
+				watchCount := 0
 				for _, c := range calls {
-					go func() {
-						if ShouldIgnore(event.Name) {
-							e.Logger.VerboseErrf(logger.Magenta, "task: event skipped for being an ignored dir: %s\n", event.Name)
-							return
-						}
-						t, err := e.GetTask(c)
-						if err != nil {
-							e.Logger.Errf(logger.Red, "%v\n", err)
-							return
-						}
-						baseDir := filepathext.SmartJoin(e.Dir, t.Dir)
-						files, err := e.collectSources(calls)
-						if err != nil {
-							e.Logger.Errf(logger.Red, "%v\n", err)
-							return
-						}
-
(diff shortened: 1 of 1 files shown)


andreynering:
This needs a rebase.

trulede:
@andreynering rebased

trulede:
PR #2743 may be slightly neater than this. Set to draft for now.

trulede:
duplicate of #2743

PR #2510, opened 2025-11-14
feat: apply insteadOf rules from global config

robertgzr:
When accessing a remote taskfile, it is unexpected and in some cases
necessary to respect these settings. Since go-git doesn't support this
yet, this workaround fixes it on our end.


vmaerten:
Thanks for your PR but we have migrated to go-getter

robertgzr:
Yes I was following that as well, awesome thanks o/

Issue #2521, opened 2025-11-20
Completions have become significantly slower starting from 3.44.0


vctls:
### Description

I have generated taskfiles with a lot of tasks (several hundreds) that I specifically use for auto-completion.
When switching from v3.43.3 to 3.44.0, completion time has gone up from around 1 second to around 3 or 4 seconds.
I tried with fish and bash, they are both slower.
The problem appeared in 3.44.0 and is still present in 3.45.5.

### Version

3.45.5

### Operating system

Ubuntu 24.4

### Example Taskfile

```YAML

```

trulede:
Is task slow from the CLI with `--silent --list-all` options?

vmaerten:
I suspect the fuzzy matcher. I’ve opened a draft where it’s disabled. Can you test it?

```
go install github.com/go-task/task/v3/cmd/task@fix-disable-fuzzy
```

We may want to consider adding an option to disable it.

trulede:
@vmaerten AFAICS the fuzzy model is only used (to generate a list of suggestions) when the specified task cannot be found. An alternative to your suggestion would be to relocate the code that builds the fuzzy match to [the point where it is used](https://github.com/go-task/task/blob/7901cce831caa573af85aae75c21378390dd2405/task.go#L453).

That removes it, and its associated performance hit, from the "normal" program flow.

vctls:
> I suspect the fuzzy matcher. I’ve opened a draft where it’s disabled. Can you test it?
> 
> ```
> go install github.com/go-task/task/v3/cmd/task@fix-disable-fuzzy
> ```

This is indeed much faster than the latest version.

vmaerten:
It will be available in the next release

PR #2523, opened 2025-11-20
refactor: optimize fuzzy matching with lazy initialization

vmaerten:
## Problem

  The fuzzy model was trained on every Task invocation (including shell autocompletion), causing performance issues with hundreds/thousands of tasks. The model is only needed when a task doesn't exist to show "Did you mean...?" suggestions.

## Solution

  - **Lazy initialization**: Train fuzzy model only when a task is not found (not during `Setup()`) (thanks @trulede for the suggestion)
  - **New option**: `--disable-fuzzy` flag and `disable-fuzzy` taskrc option to completely disable the feature

## Changes

  - Moved fuzzy model training from `Setup()` to lazy init in `GetTask()`
  - Added `DisableFuzzy` option to taskrc, flags, and Executor
  - Updated shell completions (zsh, fish, powershell)
  - Added documentation for new option

## Performance

  - **Before**: Fuzzy model trained every time, even for autocompletion
  - **After**: Zero overhead for normal operations, training only on task-not-found errors

## Compatibility

  Fully backward compatible. Fuzzy matching enabled by default, users can opt-out via `.taskrc.yml` or `--disable-fuzzy`.

##  Related Issues
- #2521

### DIFF
--- a/taskrc/ast/taskrc.go
+++ b/taskrc/ast/taskrc.go
@@ -15,7 +15,7 @@ type TaskRC struct {
 	DisableFuzzy *bool           `yaml:"disable-fuzzy"`
 	Concurrency  *int            `yaml:"concurrency"`
 	Remote       Remote          `yaml:"remote"`
-  Failfast    bool            `yaml:"failfast"`
+	Failfast     bool            `yaml:"failfast"`
 	Experiments  map[string]int  `yaml:"experiments"`
 }
 


website/src/docs/reference/config.md
```diff
@@ -109,6 +120,7 @@ Here's a complete example of a `.taskrc.yml` file with all available options:
 ```yaml
 # Global settings
 verbose: true
+disable-fuzzy: false
```
andreynering, line 134: I just realized that, we are apparently inconsistency when it comes to `_` vs `-` for word separation?

task.go
```diff
@@ -450,8 +450,14 @@ func (e *Executor) GetTask(call *Call) (*ast.Task, error) {
...
-		if e.fuzzyModel != nil {
-			didYouMean = e.fuzzyModel.SpellCheck(call.Task)
+		if !e.DisableFuzzy {
+			// Lazy initialization of fuzzy model
+			if e.fuzzyModel == nil {
+				e.setupFuzzyModel()
+			}
+			if e.fuzzyModel != nil {
+				didYouMean = e.fuzzyModel.SpellCheck(call.Task)
+			}
```
andreynering, line 463: We should probably wrap this into a mutex.

vmaerten: Good catch! I've implemented a Once to be sure this is executed only once

website/src/docs/reference/config.md
vmaerten: I think the default is to use `-`, and we only use `_` for experiments to mirror the environment variable. I’m open to discussing it, tho

PR #2525, opened 2025-11-22
feat: add `--failfast` and `failfast: true` to control dependencies

andreynering:
* Closes https://github.com/go-task/task/issues/1246

This PR introduces a small behavior change on how dependencies work. Task will now wait for all dependencies to finish running by default, even when any of them fail.

To opt for the previous behavior, either:

* Add `failfast: true` to your `.taskrc.yml`.
* Add `failfast: true` to a specific task.
* Use the `--failfast` flag.
* The `--failfast` flag will also be considering when using `--parallel`.


rabbitlair:
It's working for me as expected, thanks a lot!

My `Taskfile.yml`:
```
version: '3'

tasks:
  default:
    deps:
      - dep1
      - dep2
      - dep3
      - dep4

  dep1: sleep 1 && echo 'dep1'
  dep2: sleep 2 && echo 'dep2'
  dep3: sleep 3 && echo 'dep3'
  dep4: exit 1
```

image

andreynering:
@vmaerten Thank you for your review!

I ended up removing `fastfast` as a global option in the Taskfile and added it to `.taskrc.yml`. Thank you for this suggestion, I think it indeed makes sense.

PR #2534, opened 2025-11-27
Package API: Support in-memory taskfile nodes with task.Executor.

trulede:
When using the Package API, make it possible to create taskfile objects from []byte/memory objects (i.e. raw unparsed YAML data).

This use case only exists when using the Package API. Its behaviour/operation is similar to `NodeStdin`.

### DIFF
--- a/taskfile/node_byte.go
+++ b/taskfile/node_byte.go
@@ -1,0 +1,49 @@
+package taskfile
+
+import (
+	"github.com/go-task/task/v3/internal/execext"
+	"github.com/go-task/task/v3/internal/filepathext"
+)
+
+// A ByteNode is a node that reads a taskfile direct from a []byte object.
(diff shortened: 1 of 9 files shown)


taskfile/node_byte.go
```diff
@@ -0,0 +1,49 @@
...
+		data:     data,
+	}, nil
+}
+
+func (node *ByteNode) Location() string {
+	return "__bytes__"
+}
+
+func (node *ByteNode) Remote() bool {
+	return true
```
trulede, line 26: Here, I'm not certain if this should actually be false. It may not matter either way.

andreynering:
I'll delegate this review to @pd93. Also, this needs a rebase.

trulede:
Rebased.

PR #2538, opened 2025-11-30
Add posix and bash opts for status and precondition commands.

### LINKED ISSUE #1901: `set:` options don't apply to dynamically (shell) expanded variables
```yaml
version: "3"

set:
  - pipefail

tasks:
  default:
    vars:
      TESTVAR:
        sh: |-
          false | true
          echo "test"
    cmds:
      - |-
        false | true
        echo "Testing {{.TESTVAR}}"
```

When `pipefail` is enabled, the task command fails, but the `TESTVAR` expansion does not.

```
$ task -v
task: [REDACTED] Not found - Using alternative (Taskfile.yml)
task: dynamic variable: "false | true\necho \"test\"" result: "test"
task: "default" started
task: [default] false | true
echo "Testing test"
task: Failed to run task "default": exit status 1
```

- Task version:
```
$ task --version
Task version: 3.39.2 ()
```
- Operating system: macOS Sequoia 15.1
- Experiments enabled: None

### LINKED ISSUE #2469: Seems like `pipefail` is not applied to `status` commands
### Description

Taskfile:

```yaml
version '3'

set:
  - errexit
  - nounset
  - pipefail
  - xtrace

...
  foo:
    cmd: echo RUNNING FOO
    status:
      - cat non-existent-file | jq '{}'
...
```

Running this:
```shell
$ task foo
task: Task "foo" is up to date
```

As a user, I expected `pipefail` to be applied and used for `status` commands as well, not just for `cmd`/`cmds`.

### Version

3.45.4

### Operating system

macOS

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version '3'

set:
  - errexit
  - nounset
  - pipefail
  - xtrace

tasks:
  foo:
    cmd: echo RUNNING FOO
    status:
      - cat non-existent-file | jq '{}'
```

trulede:
set/shopts are applied to cmd variants:
* status
* precondition
* if
* sh both status and precondition commands

fixes #2469 
fixes #1901

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -81,7 +81,9 @@ func (c *Compiler) getVariables(t *ast.Task, call *Call, evaluateShVars bool) (*
 				return nil
 			}
 			// If the variable is dynamic, we need to resolve it first
-			static, err := c.HandleDynamicVar(newVar, dir, env.GetFromVars(result))
+
+			static, err := c.HandleDynamicVar(newVar, dir, env.GetFromVars(result),
+				[]string{}, []string{})
 			if err != nil {
 				return err
 			}
@@ -145,7 +147,7 @@ func (c *Compiler) getVariables(t *ast.Task, call *Call, evaluateShVars bool) (*
 	return result, nil
 }
 
-func (c *Compiler) HandleDynamicVar(v ast.Var, dir string, e []string) (string, error) {
+func (c *Compiler) HandleDynamicVar(v ast.Var, dir string, e []string, posixOpts []string, bashOpts []string) (string, error) {
 	c.muDynamicCache.Lock()
 	defer c.muDynamicCache.Unlock()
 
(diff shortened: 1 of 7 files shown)


vmaerten:
Hey  Thanks for working on it.
I think we can avoid passing opts everywhere by merging them earlier, in compiledTask:
```go
  // variables.go
  Set:   slicesext.UniqueJoin(e.Taskfile.Set, origTask.Set),
  Shopt: slicesext.UniqueJoin(e.Taskfile.Shopt, origTask.Shopt),
```
  Since compiledTask always runs before status checks, preconditions, if, etc., the compiled task already carries the merged opts and all the call just do `PosixOpts: t.Set` with no extra plumbing. StatusChecker loses its stored fields, WithPosixOpts/WithBashOpts go away, HandleDynamicVar keeps its original  signature

What do you think ?
Also it may worth it to add a test for this :)

trulede:
@vmaerten I took your suggestion, but it did not help much with `HandleDynamicVar` as that function does not have access to the task owning the variable. However, that function is called from only two places ... both have a task variable (of some kind), so its probably possible to change the signature to include the relevant task.

A bigger question is if the existing approach was actually correct. Consider this:

```
err = execext.RunCommand(ctx, &execext.RunCommandOptions{
    ...
    PosixOpts: slicesext.UniqueJoin(e.Taskfile.Set, t.Set, cmd.Set),
    BashOpts:  slicesext.UniqueJoin(e.Taskfile.Shopt, t.Shopt, cmd.Shopt),
})
```

Should these options be a set/merge of all provided options (taskfile, task, cmd)? Or, should it be an overriding selection, so use cmd.opts if defined, else use task.opts if defined, else use taskfile.opts.

I think the latter is more correct, since it makes it possible to disable an earlier/lower setting. For example:

```
-- Taskfile.yml --
version: '3'

set: [pipefail]

tasks:
  default:
    cmds:
      - task: global
      - task: task
      - task: command
  global: echo **/*.go 
  task:
    set: []  # ***** Disable pipefail
    cmds:
      - cmd: echo **/*.go
  command:
    set: [hashall]  # ***** Replace pipefail with hashall
    cmds:
      - cmd: echo **/*.go
        set: [pipefail hashall]   # ***** Replace hashall with hashall and pipefail
```

PR #2564, opened 2025-12-10
feat: add conditional execution for tasks and commands

### LINKED ISSUE #608: Ability to Run Tasks Conditionally
There are some cases where `task:` is called under `cmds:` and the task needs to be conditionally run. Take the following as an example:

```
---
version: '3'

tasks:
  python:requirements:
    deps:
      - :install:software:python
    run: once
    cmds:
      - task: python:requirements:poetry
      - task: :{{if eq .REPOSITORY_TYPE "ansible"}}common:python:requirements:ansible{{else}}donothing{{end}}
    status:
      - '[[ "${container:=}" == "docker" ]]'
```

I'm currently getting around the issue by having a task called `donothing` that gets called when the task should not run. It would be nice if we could conditionally run tasks somehow. Maybe something like this:

```
---
version: '3'

tasks:
  test:
    cmds:
      - task: anothertask
        when:
          sh: '[[ "$CONTAINER" == 'docker' ]]'
```

vmaerten:
* Closes #608

## Summary

- Add `if` attribute for conditional execution of tasks and commands
- When the shell command in `if` exits with non-zero, the task/command is skipped (not failed)
- Supports Go template expressions (e.g., `{{eq .VAR "value"}}`) which evaluate to `true`/`false`
- Works with `for` loops, evaluating the condition for each iteration

Turns out it was not that hard and it can solve most of problems described in the issue

  ## Example

  ```yaml
  tasks:
    deploy:
      if: '[ "$CI" = "true" ]'
      cmds:
        - cmd: echo "Production build"
          if: '{{eq .ENV "production"}}'
        - ./deploy.sh

### DIFF
--- a/testdata/if/testdata/TestIf-task-if-false.golden
+++ b/testdata/if/testdata/TestIf-task-if-false.golden
@@ -1,1 +1,1 @@
-task: "task-if-false" if condition not met - skipped
+task: if condition not met - skipped: "task-if-false"
--- a/testdata/if/testdata/TestIf-task-level-template-false.golden
+++ b/testdata/if/testdata/TestIf-task-level-template-false.golden
@@ -1,1 +1,1 @@
-task: "task-level-template-false" if condition not met - skipped
+task: if condition not met - skipped: "task-level-template-false"


task.go
```diff
@@ -129,6 +129,17 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
 		return nil
 	}
 
+	if t.If != "" {
+		if err := execext.RunCommand(ctx, &execext.RunCommandOptions{
+			Command: t.If,
+			Dir:     t.Dir,
+			Env:     env.Get(t),
```
trulede, line 137: Sometimes people ask to have Posix and Bash opt available on these commands. I don't know if that is a good or a bad idea, never the less please see #2538 where the idea is in draft form.

vmaerten: I think it make sense
I'll take a look at your PR soon ! :)

testdata/if/Taskfile.yml
andreynering, line 1: I liked that you kept all tests in the same file. Looks very clean!

task.go
```diff
@@ -129,6 +129,17 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
 		return nil
 	}
 
+	if t.If != "" {
```
andreynering: `TrimSpace` looks like a good idea here to me, to avoid this triggering if the string happens to have any whitespace.

```suggestion
	if strings.TrimSpace(t.If) != "" {
```

task.go
```diff
@@ -129,6 +129,17 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
...
+		if err := execext.RunCommand(ctx, &execext.RunCommandOptions{
+			Command: t.If,
+			Dir:     t.Dir,
+			Env:     env.Get(t),
+		}); err != nil {
+			e.Logger.VerboseOutf(logger.Yellow, "task: %q if condition not met - skipped\n", call.Task)
```
andreynering: I wonder if moving the command to the end is better? Nitpick, tho.


```suggestion
			e.Logger.VerboseOutf(logger.Yellow, "task: if condition not met - skipped: %q\n", call.Task)
```

task.go
```diff
@@ -305,6 +317,18 @@ func (e *Executor) runDeferred(t *ast.Task, call *Call, i int, vars *ast.Vars, d
 func (e *Executor) runCommand(ctx context.Context, t *ast.Task, call *Call, i int) error {
 	cmd := t.Cmds[i]
 
+	// Check if condition for any command type
+	if cmd.If != "" {
```
```suggestion
	if strings.TrimSpace(cmd.If) != "" {
```

PR #2566, opened 2025-12-11
fix: always run a watch task regardless of run setting.

### LINKED ISSUE #1388: --watch detects change but skips task when "run: when_changed" is set
Running with `--watch` or `watch: true` does not actually run the task again when a change is detected if `run: when_changed` is set. Use the following Taskfile and run with `task -w -vvv test` to reproduce.

```yml
version: '3'
interval: 500ms

# Comment this out to see expected correct behavior
run: when_changed

tasks:
  test:
    sources:
      - content.txt
    cmds:
      - echo "Hello, world!"
```

Editing `content.json` shows
```shell
% task -w -vvv test
task: Started watching for tasks: test
task: Watching for changes every 500ms
task: "test" started
task: watching new file: /Users/joshwilson/code/content.txt
task: Task "test" is up to date
task: received watch event: FILE "content.txt" WRITE [/Users/joshwilson/content.txt]
task: skipping execution of task: test:871025353146782510
```

I usually set `run: when_changed` at the global scope to avoid running diamond dependencies multiple times in a general manner. I was surprised by this interaction with the `--watch` option.

A workaround is to override the global `run` setting with `run: always` in the watched task definition. I would expect that a watched task would behave as if `run: always` was implicitly set. In the meantime I will revise my Taskfiles to default to `run: always` and explicitly set `run: when_changed` on the diamond dependencies.

- Task version: 3.31.0
- Operating system: macOS 13.6.1 (22G313) (Ventura)
- Experiments enabled: None

trulede:
A task will not run (once or when_changed) if its hash is "known". Therefore, a watch task where `run` is configured, will never start.

In this PR, the check of task hash is extended to consider if watch is configured/requested.

fixes #1388

### DIFF
--- a/task.go
+++ b/task.go
@@ -366,7 +366,7 @@ func (e *Executor) startExecution(ctx context.Context, t *ast.Task, execute func
 		return err
 	}
 
-	if h == "" {
+	if h == "" || t.Watch {
 		return execute(ctx)
 	}
 


vmaerten (approved):
Thanks!

PR #2568, opened 2025-12-12
feat: emit error annotations in GitHub Actions

vmaerten:
## Summary

  When a task fails in GitHub Actions, finding the error often means scrolling through hundreds of lines of logs. This PR changes that.

  Task now automatically emits `::error::` annotations when running in GitHub Actions (`GITHUB_ACTIONS=true`). These annotations:

  - **Appear directly in the workflow summary** - no more hunting through logs
  - **Show the task name** - immediately know which task failed
  - **Link to the error** - click to jump to the relevant log line

  ### Before

  Errors are buried in logs, easy to miss when you have dozens of tasks running.

  ### After

  ::error title=Task 'build' failed::exit status 1

  The error appears as a red annotation at the top of the workflow run, visible at a glance.

image

  ## Why this matters

  - **Faster debugging** - See what failed without reading logs
  - **Better DX in CI** - Task becomes a first-class citizen in GitHub Actions
  - **Zero configuration** - Works automatically, no setup required

  ## Test plan

  - [x] `GITHUB_ACTIONS=true task failing-task` emits `::error title=Task 'X' failed::...`
  - [x] Without `GITHUB_ACTIONS=true`, no annotation is emitted
  - [x] Annotation includes task name when available (TaskRunError)
  - [x] Generic errors also emit annotations

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/spf13/pflag"
 
@@ -46,7 +47,7 @@ func main() {
 
 // emitCIErrorAnnotation emits an error annotation for supported CI providers.
 func emitCIErrorAnnotation(err error) {
-	if isGA, _ := strconv.ParsePool(os.Getenv("GITHUB_ACTIONS")); !isGA {
+	if isGA, _ := strconv.ParseBool(os.Getenv("GITHUB_ACTIONS")); !isGA {
 		return
 	}
 	if e, ok := err.(*errors.TaskRunError); ok {


cmd/task/task.go
```diff
@@ -28,19 +28,34 @@ func main() {
...
+		emitCIErrorAnnotation(err)
 		l.Errf(logger.Red, "%v\n", err)
 		os.Exit(errors.CodeUnknown)
 	}
 	os.Exit(errors.CodeOk)
 }
 
+// emitCIErrorAnnotation emits an error annotation for supported CI providers.
+func emitCIErrorAnnotation(err error) {
+	if os.Getenv("GITHUB_ACTIONS") != "true" {
```
andreynering: [`ParseBool`](https://pkg.go.dev/strconv#ParseBool) is usually a good idea in these scenarios. GA might be fixed at `true`, but still...

```suggestion
	if isGA, _ := strconv.ParsePool(os.Getenv("GITHUB_ACTIONS")); !isGA {
```

vmaerten: You are totally right

vmaerten: from @andreynering : 

```suggestion
	if isGA, _ := strconv.ParsePool(os.Getenv("GITHUB_ACTIONS")); !isGA {
```

PR #2570, opened 2025-12-12
Enable ANSI escape code handling on windows consoles.

### LINKED ISSUE #2076: Coloring on windows.
### Description

Issue #343 and related PR #635 were closed unsolved because the issue author did not respond/whatever else.

It is broken on windows terminal on windows 10. I don't know about 11 but terminal and conhost are the same across both.

Image

Going by the previous issues (3 and 5 years ago), it was never fixed. A fix was proposed in #635 but it was deemed to have too many side effects and not done properly.

I think it's safe to say that it is not a windows terminal issue, considering literally every other app has managed to do colors well, especially since wt is much more modern than other terminals. (examples include `clink`, `eza`, `bat`, the whole `charmbracelet/*` bunch i.e. `gum`, `glow`, any `bubbletea` app, etc. - if you need more examples please tell me, I will look)

<details>
<summary>Some examples</summary>
`clink` can be seen in the prompt.

`eza`
Image

`bat` running inside `tv`

Image

</details>

Do you want me to use Debug Tap on windows terminal? This will allow you to see all the actual ascii sequences sent to the terminal.
But since the issue had already been diagnosed to one line in #343 itself, I'm not sure if that was needed.

### Version

v3.41.0

### Operating system

Windows 10

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

trulede:
Enable ANSI escape code handling on windows consoles.

fixes #2076 

This sets the windows terminal so that it handles ANSI codes (esp color). Under some circumstances (not known) a windows console may be in a mode where it does not process ANSI codes, as a result task output (color=true _or_ false) will show escape sequences:

```
C:\Users\trule\Desktop>task
←[31mtask: No Taskfile found at ""
←[0m
```

This PR pushes the terminal into the correct mode (ENABLE_VIRTUAL_TERMINAL_PROCESSING).

### DIFF
--- a/cmd/task/console_win.go
+++ b/cmd/task/console_win.go
@@ -14,9 +14,11 @@ func init() {
 	// Ensure that Windows console handles ANSI escape-codes correctly.
 	if runtime.GOOS == "windows" {
 		stdout := windows.Handle(os.Stdout.Fd())
-		var originalMode uint32
-		if err := windows.GetConsoleMode(stdout, &originalMode); err == nil {
-			windows.SetConsoleMode(stdout, originalMode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+		if stdout != 0 {
+			var originalMode uint32
+			if err := windows.GetConsoleMode(stdout, &originalMode); err == nil {
+				windows.SetConsoleMode(stdout, originalMode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+			}
 		}
 	}
 }


andreynering:
Hey @trulede.

Were you able to reproduce the problem? If so, how? Which terminal app?

trulede:
@andreynering I modified the code in the PR to force the condition (clearing the flag), proving the reported behaviour (and thus the solution).  But would it solve the issue for the originator? Probably.

Still, its not a Task issue. Something in the terminal is wrong of the originators pc. They have indicated as much.

Although I don't think there is much risk with this PR, in case the console is not a terminal, the call to GetConsoleMode() will err and nothing else will happen. At least someone else should verify the PR - I suggest forcing the condition, my efforts to reproduce the broken console took a long time and were not effective.

Almost certainly, something else is the cause.


There could be a task user who does not want this flag set? We could restrict it further, for instance, only setting it if `color` is enabled. Or make the change, and wait for feedback.

trulede:
It seems that faith/color also has this init code, so its likely that some other affect is causing this - and not likely related to Task.

Issue #2479, opened 2025-10-28
`CLI_ASSUME_YES` flag


dbeecham:
### Description

Hi. I wanted to bikeshed my Taskfile.yaml-file I used for my NixOS machines (around 8-12 of them). The UI from Taskfile can get a little messy at times, so I try to keep it clean using [Charm](https://charm.land/) or [Rich](https://rich.readthedocs.io/en/latest/).

I'd love to see either a way to run a command to query for prompt validation, or by exposing a global variable such as `.CLI_ASSUME_YES` we can use to kind of "side-step" the `prompt` when we need additional control.

For a little bit of extra context, also work a lot with Ansible. Ansible also has issues with it's `prompt` functionality, specifically that it doesn't scale very well. Either you could have 200k machines that need an acklowledgment, or you're running in a frontend where a TTY is not available. In these cases, it's neat to be able to hook in your own special solution to ask for confirmation.

trulede:
New envar `CLI_ASSUME_YES` reflecting the CLI option `--yes`.

semihbkgr:
I'd like to take this issue if no one is currently working on it.

PR #2577, opened 2025-12-14
feat: expose .CLI_ASSUME_YES templating variable

semihbkgr:
This PR exposes the state of the `--yes` (or `-y`) CLI flag as a global templating variable named `.CLI_ASSUME_YES`.

This allows users to conditionally execute logic in their Taskfiles based on whether the user has confirmed prompts via the CLI flag.

see: #2479

**Example:**

```yaml
tasks:
  example:
    cmds:
      - >
        {{if .CLI_ASSUME_YES}}
        echo "Assume yes is true"
        {{else}}
        echo "Assume yes is false"
        {{end}}
```

**Changes:**

- Added `.CLI_ASSUME_YES` to the global variables in `cmd/task/task.go`
- Documented the new variable in `website/src/docs/reference/templating.md`

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -182,6 +182,7 @@ func run() error {
 	globals.Set("CLI_SILENT", ast.Var{Value: flags.Silent})
 	globals.Set("CLI_VERBOSE", ast.Var{Value: flags.Verbose})
 	globals.Set("CLI_OFFLINE", ast.Var{Value: flags.Offline})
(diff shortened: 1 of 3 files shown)


semihbkgr:
Got it, that makes more sense now. Thanks for the explanation. I was initially confused by the 'env var' terminology in the comments. I understand now that this should be a global variable exposed in the templating system. I'll update the PR accordingly.

vmaerten (approved):
Thanks!

PR #2578, opened 2025-12-14
chore(ci): add on-demand PR build workflow

vmaerten:
# Summary

  Similar to our nightly builds, but for PRs!

  Currently, testing a PR requires having Go installed locally to build from source. This makes it harder for users to help test fixes or new features before they're merged.

  This adds a workflow that builds binaries on-demand when a maintainer adds the `needs-build` label to a PR. Binaries are uploaded as artifacts and a comment is posted with the download link.

This will uses our `task-bot`.

This [PR](https://github.com/go-task/task/pull/2570) could benefits this feature

> [!NOTE]
> The label is `needs-build` but it can be modified. We can also use somesthing like a comment but it's harder to detect 

## Example: 

image

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,0 +1,69 @@
+name: PR Build
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: write
+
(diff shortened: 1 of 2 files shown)


andreynering:
@vmaerten This is cool, thank you!

PR #2579, opened 2025-12-14
feat(vars): add interactive prompting for required variables

### LINKED ISSUE #2079: Allow users to input variables interactively
### Description

We could probably use [Huh](https://github.com/charmbracelet/huh) for this.

vmaerten:
# Summary

fixes #2079

Add interactive prompting for missing required variables using Bubble Tea UI.

When `interactive: true` is set in `.taskrc.yml` or `--interactive` is passed, Task will prompt users for any missing required variables instead of failing with an error.

> [!NOTE]
> I chose to bind this to requires.vars instead of defining it for each var because required vars already express the semantic of "this must be provided" so prompting is a natural fallback when they're missing.
 Similar to Terraform's approach. If needed, we could add an interactive: false flag per var to opt-out, but I'm not sure it's necessary.

### Variable types supported

- **Text input**: For simple string variables
- **Enum selection**: For variables with predefined allowed values (with vim keybindings j/k)

```yaml
tasks:
  deploy:
    requires:
      vars:
        - API_KEY                    # Text prompt
        - name: ENVIRONMENT          # Enum selector
          enum: [dev, staging, prod]
```

### Bound to `requires`

Interactive prompting only applies to variables declared in `requires.vars`. Regular variables with defaults are not prompted.

### Configuration

Enable in `.taskrc.yml`:

```yaml
interactive: true
```

Or via CLI flag:

```bash
task deploy --interactive
```

### Auto-detection of non-TTY environments

Task automatically detects non-TTY environments (CI, piped input, etc.) and skips prompts, failing with the standard missing variable error.

### Hybrid approach: deps vs cmds

Task handles `deps` and `cmds` differently when it comes to prompting:

#### Dependencies (`deps`) - Upfront collection

Dependencies execute **in parallel**, which means multiple prompts could appear simultaneously:

```
? Enter value for VAR_A: >
? Select value for VAR_B:
❯ option1
? Enter value for VAR_C: >
    option2
```

To avoid this, we **collect all required vars upfront** before any execution begins:

1. Traverse the dependency tree recursively
2. Collect all missing required variables
3. Prompt for everything at once
4. Then execute all tasks with the collected values

```
? Select value for BUILD_MODE: release
? Select value for TEST_SUITE: all

task: [build] Building in release mode...
task: [test] Running all tests...
```

#### Commands (`cmds`) - Just-in-time prompting

Commands that call other tasks (`task: foo`) execute **sequentially**, so there's no risk of interleaved prompts. We prompt **just-in-time** when we encounter a task with missing variables:

```yaml
pipeline:
  cmds:
    - task: setup      # Prompts for PROJECT_NAME when reached
    - task: build      # Prompts for BUILD_MODE when reached
    - task: deploy     # Prompts for ENVIRONMENT when reached
```

This approach is more natural for sequential workflows - users see prompts in context as the pipeline progresses.

### Why this hybrid approach?

| Aspect | Deps (parallel) | Cmds (sequential) |
|--------|-----------------|-------------------|
| Execution | Concurrent | One at a time |
| Prompt timing | Must be upfront | Can be just-in-time |
| UX | All prompts together | Prompts in context |

The hybrid approach gives the best of both worlds:
- **Deps**: No garbled output from parallel prompts
- **Cmds**: Natural flow where you answer questions as you reach each step

### Variable caching

Once a variable is prompted, its value is cached in `e.promptedVars`. If the same variable is required by multiple tasks, the user is only prompted once.

### Handling duplicate task calls with different vars

When the same task is called multiple times with different variables:

```yaml
deps:
  - task: build
    vars: { TARGET: linux }   # TARGET provided
  - task: build
    vars: { MODE: release }   # TARGET missing!
```

The collection phase checks all calls, not just unique task names, ensuring we catch all missing variables.

### Integration with `if:` on tasks

When the `if:` feature on tasks lands, we'll need to ensure that tasks with `if: false` (that would be skipped) are **not** prompted for their variables during the upfront collection phase. This will require evaluating the `if` condition before collecting vars from a task.

## Test plan

- Run `task pipeline` in `testdata/interactive_vars/` with `interactive: true`
- Verify deps prompts appear upfront before execution
- Verify cmds prompts appear just-in-time as tasks are reached
- Test enum selection with arrow keys and j/k vim bindings
- Verify `--interactive=false` disables prompts
- Verify prompts auto-skip in non-TTY (pipe output to file)
- Test task with nested deps requiring different variables
- Test `full-deploy` (deps + cmds combined)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
   @trulede).
 - Task aliases can now contain wildcards and will match accordingly (e.g., `s-*`
   as alias for `start-*`) (#1900, #2234 by @vmaerten).
+- Added conditional execution with the `if` field: skip tasks, commands, or task
+  calls based on shell exit codes or template expressions like
+  `{{ eq .ENV "prod" }}` (#2564, #608 by @vmaerten).
 
 ## v3.46.4 - 2025-12-24
 
--- a/executor_test.go
+++ b/executor_test.go
@@ -1105,3 +1105,65 @@ func TestFailfast(t *testing.T) {
 		)
 	})
 }
+
+func TestIf(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		task    string
+		vars    map[string]any
+		verbose bool
+	}{
+		// Basic command-level if
+		{name: "cmd-if-true", task: "cmd-if-true"},
+		{name: "cmd-if-false", task: "cmd-if-false"},
+
+		// Task-level if
+		{name: "task-if-true", task: "task-if-true"},
+		{name: "task-if-false", task: "task-if-false", verbose: true},
+
+		// Task call with if
+		{name: "task-call-if-true", task: "task-call-if-true"},
+		{name: "task-call-if-false", task: "task-call-if-false", verbose: true},
+
+		// Go template conditions
+		{name: "template-eq-true", task: "template-eq-true"},
+		{name: "template-eq-false", task: "template-eq-false", verbose: true},
+		{name: "template-ne", task: "template-ne"},
+		{name: "template-bool-true", task: "template-bool-true"},
+		{name: "template-bool-false", task: "template-bool-false"},
+		{name: "template-direct-true", task: "template-direct-true"},
+		{name: "template-direct-false", task: "template-direct-false"},
+		{name: "template-and", task: "template-and"},
+		{name: "template-or", task: "template-or"},
+
+		// CLI variable override
+		{name: "template-cli-var", task: "template-cli-var", vars: map[string]any{"MY_VAR": "yes"}},
+
+		// Task-level if with template
+		{name: "task-level-template", task: "task-level-template"},
+		{name: "task-level-template-false", task: "task-level-template-false", verbose: true},
+
+		// For loop with if
+		{name: "if-in-for-loop", task: "if-in-for-loop", verbose: true},
+	}
+
+	for _, test := range tests {
+		opts := []ExecutorTestOption{
+			WithName(test.name),
+			WithExecutorOptions(
+				task.WithDir("testdata/if"),
+				task.WithSilent(true),
+				task.WithVerbose(test.verbose),
+			),
+			WithTask(test.task),
+		}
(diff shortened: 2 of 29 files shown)


internal/input/input.go
```diff
@@ -0,0 +1,223 @@
+package input
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/charmbracelet/bubbles/textinput"
+	tea "github.com/charmbracelet/bubbletea"
```
andreynering: Change to import from `charm.land/bubbletea/v2`.

v2 is in release candidate and will be released soon!

sashokbg:
Hello, @vmaerten great work!
I have a question, can the enum be replaced with a **sh** ? I worked on a similar feature and had implemented sh where it will split the newlines and present them as possible choices.
This way the select options can be fed from a script

internal/input/input.go
vmaerten: I have change import to target v2

vmaerten:
> Hello, @vmaerten great work! I have a question, can the enum be replaced with a **sh** ? I worked on a similar feature and had implemented sh where it will split the newlines and present them as possible choices. This way the select options can be fed from a script

Hello @sashokbg!
For the initial version, I won’t introduce `sh` within interactive variables, but I completely understand the use case.
We could consider introducing dynamic enums using `sh:`. Could you open an issue so we can discuss the need?

vmaerten:
> This looks a lot cooler in practice than I expected! Good job.
> 
> I liked that you added an `interactive` option to avoid this being a breaking change, and also that it works with enums!
> 
> You may also write a blog post for this if you want. (Or perhaps it could be a single post for this and #2564, as they'll likely be released together).
> 
> I'm holding the approve until you upgrade to Bubble Tea v2 RC, but otherwise this looks good!

Thanks a lot!
I’ve upgraded to Bubble Tea v2 RC.

sashokbg:
Hello, I have opened #2647 to propose and discuss the `sh` improvements. I would be happy to help and build on top of your PR @vmaerten 
Have a nice day !

PR #2582, opened 2025-12-15
Fix Taskfile not found error handling

### LINKED ISSUE #2581: Taskfile not found error shows empty path and missing --init suggestion
### Description

When running `task` in a directory without a Taskfile (and no Taskfile in parent directories), the error message displays an empty path and is missing the helpful suggestion to run `task --init`:

```
task: No Taskfile found at ""
```

The error message does not include:
- `(or any of the parent directories)`
- `Run `task --init` to create a new Taskfile.`

This is a regression from previous versions where the error message properly displayed the directory path being searched **and included a suggestion to run `task --init`**.

## Previous Behavior

In previous versions, the error message was informative and helpful:

```
task: No Taskfile found at "/path/to/directory" (or any of the parent directories). Run `task --init` to create a new Taskfile.
```

## Root Cause

The issue is caused by conflicting error handling between `NewFileNode()` and `getRootNode()`:

**In `taskfile/node_file.go`**, `NewFileNode()` was already wrapping `os.ErrNotExist` with `TaskfileNotFoundError`:

```go
    resolvedEntrypoint, err := fsext.Search(entrypoint, dir, DefaultTaskfiles)
    if err != nil {
            return nil, errors.TaskfileNotFoundError{URI: entrypoint, Walk: false}
        }
        return nil, err
    }
    // ...
}
```

**In `setup.go`**, `getRootNode()` was checking for `os.IsNotExist(err)` to wrap the error:

```go
func (e *Executor) getRootNode() (taskfile.Node, error) {
    node, err := taskfile.NewRootNode(e.Entrypoint, e.Dir, e.Insecure, e.Timeout)
    if os.IsNotExist(err) {  // This check fails because err is already TaskfileNotFoundError
        return nil, errors.TaskfileNotFoundError{
            URI:     fsext.DefaultDir(e.Entrypoint, e.Dir),
            Walk:    true,
            AskInit: true,
        }
    }
    // ...
}
```

The problem: `os.IsNotExist(err)` returns `false` when `err` is `TaskfileNotFoundError` (not `os.ErrNotExist`), so the error is returned as-is with an empty or incorrect URI and without the `AskInit` suggestion.

## Steps to Reproduce

1. Navigate to an empty directory (with no Taskfile in parent directories)
2. Run `task`
3. Observe the error message shows an empty path and doesn't suggest `task --init`

### Version

main

### Operating system

macOS

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

semihbkgr:
Fixes #2581

When running `task` in a directory without a Taskfile, the error message was showing an empty path (`task: No Taskfile found at ""`) and missing the helpful `--init` suggestion. This was caused by conflicting error handling where `NewFileNode()` was already wrapping `os.ErrNotExist` as `TaskfileNotFoundError`, but `getRootNode()` was checking for `os.IsNotExist(err)` which failed on the wrapped error.

## Solution

1. **Moved `--init` suggestion logic from error type to caller**: Removed the `AskInit` field from `TaskfileNotFoundError` and moved the suggestion logic to `getRootNode()` in `setup.go`. This separates the concern of "what happened" (file not found) from "what to suggest to the user" (run --init), and ensures the suggestion only appears when appropriate (i.e., when running plain `task` without `-t` or `-d` flags, since `--init` ignores those flags).

2. **Fixed URI construction**: Added `fsext.GetSearchPath()` helper to properly determine the absolute path being searched. This ensures `TaskfileNotFoundError.URI` always contains a meaningful path instead of an empty string.

3. **Simplified error handling**: Changed `getRootNode()` to use `errors.As()` to properly detect `TaskfileNotFoundError` and wrap it with the `--init` suggestion when appropriate, instead of trying to re-wrap `os.ErrNotExist`.

## Result

- `--init` suggestion appears only when running `task` without flags
- Proper "parent directories" message when doing recursive search
- Error messages now show absolute paths

### DIFF
--- a/errors/errors_taskfile.go
+++ b/errors/errors_taskfile.go
@@ -11,18 +11,14 @@ import (
 // TaskfileNotFoundError is returned when no appropriate Taskfile is found when
 // searching the filesystem.
 type TaskfileNotFoundError struct {
-	URI     string
-	Walk    bool
-	AskInit bool
+	URI  string
+	Walk bool
 }
 
 func (err TaskfileNotFoundError) Error() string {
 	var walkText string
 	if err.Walk {
 		walkText = " (or any of the parent directories)."
-	}
-	if err.AskInit {
-		walkText += " Run `task --init` to create a new Taskfile."
 	}
 	return fmt.Sprintf(`task: No Taskfile found at %q%s`, err.URI, walkText)
 }
--- a/internal/fsext/fs.go
+++ b/internal/fsext/fs.go
@@ -49,6 +49,24 @@ func ResolveDir(entrypoint, resolvedEntrypoint, dir string) (string, error) {
 	return filepath.Dir(resolvedEntrypoint), nil
 }
 
+// GetSearchPath returns the absolute path to search for the entrypoint. If the
+// entrypoint is set, it returns the absolute path to the entrypoint. If the dir
+// is set, it returns the absolute path to the dir. If both are empty, it returns
+// the current working directory.
+func GetSearchPath(entrypoint, dir string) (string, error) {
+	if entrypoint != "" {
+		return filepath.Abs(entrypoint)
+	}
+	if dir != "" {
+		return filepath.Abs(dir)
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+	return wd, nil
+}
+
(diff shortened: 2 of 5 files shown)


trulede:
Hi, this got fixed in the meantime by PR #2682, more or less the same outcome.

Issue #2588, opened 2025-12-18
Variables passed via command line are not processed in task 3.46.1


msardara:
# Description

I have a set of script calling task to accomplish some builds.

I pass the variables using the following syntax:

```
task VALUE=value
```

This used to work until version 3.46.1, which broke this behavior.

# Minimal reproducible example

Taskfile.yaml
```yaml
---

version: '3'

vars:
  PROFILE: '{{.PROFILE | default "debug"}}'
  RELEASE:
    sh: '[[ {{.PROFILE}} == "release" ]] && echo "--release" || echo ""'

tasks:
  print:
    desc: "Print release"
    cmds:
      - echo {{.RELEASE}}
```

Working:
```
task --version
3.45.4

task -v print PROFILE=release
task: dynamic variable: "[[ release == \"release\" ]] && echo \"--release\" || echo \"\"" result: "--release"
task: "print" started
task: [print] echo --release
--release
task: "print" finished
```

Not working:
```
task --version
3.46.1

task -v print PROFILE=release
task: dynamic variable: "[[ debug == \"release\" ]] && echo \"--release\" || echo \"\"" result: ""
task: "print" started
task: [print] echo

task: "print" finished
```

### Version

3.46.1

### Operating system

Any

### Example Taskfile

```YAML
---

version: '3'

vars:
  PROFILE: '{{.PROFILE | default "debug"}}'
  RELEASE:
    sh: '[[ {{.PROFILE}} == "release" ]] && echo "--release" || echo ""'

tasks:
  print:
    desc: "Print release"
    cmds:
      - echo {{.RELEASE}}
```

trulede:
@vmaerten if I revert 5889ff6b6525f804e7d301beb9fd70445946a3e2 this works ... ~~which makes me wonder if that underlying issue was inadvertently resolved at some point?~~

```
---
version: '3'

silent: true

vars:
  FOO: '{{.FOO | default "foo"}}'

tasks:
  default:
    cmds:
      - echo FOO={{.FOO}}
```

yields (with 3.46.1)

```
$ task FOO=bar
FOO=foo
```

trulede:
OK, 5889ff6b6525f804e7d301beb9fd70445946a3e2 is fixing this issue in 3.46.1, so it is having an effect.

```
version: '3'

vars:
  ENV: '{{.ENV | default .CLI_ARGS}}'   # use of CLI_ARGS here

tasks:
  debug:
    cmds:
      - echo '{{.ENV}}'
```

[...]

PR #2589, opened 2025-12-18
fix: CLI variables should take priority over Taskfile defaults

vmaerten:
fixes https://github.com/go-task/task/issues/2588

When using `task FOO=bar` with a Taskfile containing `FOO: '{{.FOO | default "foo"}}'`, the CLI value was being overwritten by the Taskfile default.

The ReverseMerge function was incorrectly allowing Taskfile variables to overwrite CLI globals. Now it skips variables that already exist in the CLI globals, ensuring user-provided values take priority while still making CLI globals available for templating.

Fixes regression introduced in #2403.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -172,18 +172,23 @@ func run() error {
 		calls = append(calls, &task.Call{Task: "default"})
 	}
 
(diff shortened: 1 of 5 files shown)


trulede:
I also tested it for both cases, and a few others, seems good.

PR #2607, opened 2025-12-26
feat(config): add environment variable support for all taskrc options

### LINKED ISSUE #1066: add environment variable-based configuration

vmaerten:
# Summary

Fixes #1066 
  Add environment variable support for all taskrc configuration options. Allow users to configure Task via `TASK_*` environment variables (e.g., `TASK_VERBOSE`, `TASK_COLOR`, `TASK_CONCURRENCY`) with priority: CLI flags > env vars > taskrc config > defaults.

  - Add typed env var helpers (`GetTaskEnvBool`, `GetTaskEnvInt`, `GetTaskEnvDuration`, `GetTaskEnvString`, `GetTaskEnvStringSlice`) in `internal/env`
  - Update `getConfig` to check env vars before falling back to taskrc
  - Support all existing taskrc options: `VERBOSE`, `COLOR`, `CONCURRENCY`, `FAILFAST`, `DISABLE_FUZZY`, `REMOTE_*` (insecure, offline, timeout, cache-expiry, cache-dir, trusted-hosts)

### DIFF
--- a/website/src/docs/reference/cli.md
+++ b/website/src/docs/reference/cli.md
@@ -332,6 +332,8 @@ Task automatically detects non-TTY environments (like CI pipelines) and
 skips prompts. This flag can also be set in `.taskrc.yml` to enable prompts
 by default.
 
+- **Environment variable**: [`TASK_INTERACTIVE`](./environment.md#task-interactive)
+
 ```bash
 task deploy --interactive
 ```
--- a/website/src/docs/reference/environment.md
+++ b/website/src/docs/reference/environment.md
@@ -75,6 +75,12 @@ variables. The priority order is: CLI flags > environment variables > config fil
 - **Default**: `false`
 - **Description**: Assume "yes" as answer to all prompts
 
+### `TASK_INTERACTIVE`
+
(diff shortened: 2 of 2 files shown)


trulede:
@vmaerten could we also address #1066 in this PR? I think what you have in the PR would be adaptable.

vmaerten:
> @vmaerten could we also address #1066 in this PR? I think what you have in the PR would be adaptable.

Yes! It's the plan :) 
I need to find the time to finish it!

PR #2615, opened 2025-12-30
[WIP] Access Task `sources`, `generates`, and task maps

CuckooEXE:
**Do not merge, opening for discussion**

This PR was created by taking [fhofherr's PR](https://github.com/go-task/task/compare/main...fhofherr:task:948-uses-sources-as-variable) and feeding it through Cursor's AI. I don't know golang, let alone this codebase, so I was more exploring if what I was looking for would even be possible.

Specifically, I'm looking for two features:

1. Access my own Task mapping as a variable
2. Access other Task's mapping as a variable

What I mean is, when writing a simple `gcc`-based target, you'll find yourself repeating a lot of code. Right now this is how you would do it:

```
myApp:
    sources: ["myApp/*.c"]
    generates: ["bin/myApp"]
    cmd: gcc -o bin/myApp myApp/*.c
```

By implementing the first feature, we could instead write something like this (using this PR):

```
myApp:
    sources: ["myApp/*.c"]
    generates: ["bin/myApp"]
    cmd: gcc -o {{ .TASK_GENERATES | first }} {{ .TASK_SOURCES }}
```

`TASK_GENERATES` works a little funny because the files don't exist yet. So any globs would have to be passed in as raw strings with the glob pattern.

The other feature would allow me to re-use information already in the Taskfile in other targets, for example (and this is *not* implemented correctly in this PR):

```
lint:
    sources: [ myApp.sources ]
    cmd: clang-format {{ .TASK_SOURCES }}

clean:
    cmd: rm -f {{ myApp.TASK_GENERATES }}
```

This is **partially** implemented in this PR by exposing a Task's mapping in a for loop as such:

```
clean:
    cmds:
    - for: [myApp]
      cmd: rm -rf {{ .ITEM.generates | join " " }}
```

I'm very invested in this feature (my team is looking to migrate away from our Makefile-hell), so anything I can do to help out with this, I'm here for!

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -76,6 +76,8 @@ type (
 		executionHashes      map[string]context.Context
 		executionHashesMutex sync.Mutex
 		watchedDirs          *xsync.Map[string, bool]
+		taskMapCache         map[string]map[string]any
+		taskMapCacheMutex    sync.RWMutex
 	}
 	TempDir struct {
 		Remote      string
--- a/task_test.go
+++ b/task_test.go
@@ -2675,29 +2675,6 @@ func TestReferenceSelf(t *testing.T) {
 	assert.Contains(t, buff.String(), `[]`)
 }
 
-func TestReferenceOtherTasks(t *testing.T) {
-	t.Parallel()
-
-	const dir = "testdata/reference_other_tasks"
-
-	var buff bytes.Buffer
-	e := task.NewExecutor(
-		task.WithDir(dir),
-		task.WithStdout(&buff),
-		task.WithStderr(&buff),
-		task.WithForceAll(true),
-	)
-	require.NoError(t, e.Setup())
-	require.NoError(t, e.Run(t.Context(), &task.Call{Task: "default"}))
-
-	assert.Contains(t, buff.String(), `task: [check-sources-empty] echo []`)
-	assert.Contains(t, buff.String(), `[]`)
-	assert.Contains(t, buff.String(), `task: [check-generates] echo [a b *.txt]`)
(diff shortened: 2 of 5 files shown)


trulede:
You can do everything you want with existing features, essentially, a Taskfile that works like a template. Exactly how ... for starters:

```yaml
myApp:
    vars:
      APP_NAME: '{{.USER_WORKING_DIR}}'
    sources: ["*.c"]
    generates: ["{{.ROOT_DIR}}/bin/{{.APP_NAME}}"]
    cmd: gcc -o {{.ROOT_DIR}}/{{.APP_NAME}} *.c
```

1/ replace `myApp` with something derived from `USER_WORKING_DIR`. See https://taskfile.dev/docs/reference/templating
2/ put the Taskfile.yml in the root of your repo.
3/ then - `cd path/to/myApp; task build`.

That's it. More of less ...

I suggest starting a topic in the "Discussions", if you need more help. A PR will not get much attention.

CuckooEXE:
I think that pattern forces codebases to comply with Taskfile's usage (rather, limitations). For that to work I have to change either how I (and my CI/CD) build (by changing directories before building), and it forces me to layout my files in a specific structure. I'm not sure your suggestion actually satisfies the original points in my description.

This PR stemmed from discussion in the discord with @andreynering.

CuckooEXE:
I (read: the AI) significantly changed the code (and the files modified). I (read: this time actually me) wrote the tests and they pass for my updated implementation. I think this makes more sense after studying the codebase, and it more matches the pattern of existing variables.

PR #2619, opened 2026-01-08
Ensure that errors from interop are of type interp.ExitStatus.

### LINKED ISSUE #2466: `ignore_error` not working on new built-in `cp` command
### Description

- The new built-in `cp` command to copy a file that may or may not exist, with the `ignore_error: true` property set on the `cmd:` scope
- Expected `cp` command errors to be suppressed
- `cp` errors were not suppressed and tasks fail 
- No change if `-f` is used or not. 
- Adding `powershell` before the `cp` command works as expected even if the powershell command fails. 

This task is just a best-effort file copy for some testing, and some files may not exist for now, and simply suppressing the error is the simplest solution for now. 

Example error

```
task: Failed to run task "dev-plugins-fetch": GetFileAttributesEx C:\tests\config\Essentials.Accounts.json: The system cannot find the file specified.
```

### Version

3.45.4

### Operating system

Windows

### Experiments Enabled

None

### Example Taskfile

```YAML
version: 3

tasks:

  default:
    cmds:
     - task: dev-build-plugin
       vars:
         ASM_NAME: "TestAssembly"
         OUTPUT_DIR: "{{ .USER_WORKING_DIR }}"

  dev-build-plugin:
    internal: true
    dir: '..'
    requires: { ASM_NAME, OUTPUT_DIR}
    cmds:
     - cmd: mkdir -p '{{ .OUTPUT_DIR }}'

     - cmd: cp -f config/{{ .ASM_NAME }}.json '{{ .OUTPUT_DIR }}/{{ .ASM_NAME }}.json'
       ignore_error: true
```

trulede:
When commands from coreutils fail, the error type is not the expected interop.ExitStatus. The handling of those errors, especially in relation to ignore-error, depends on that error type.

Solution presented here first prints the error (otherwise you have no feedback as to why the command fails), and then replaces the err object with a new interop.ExitStatus error. Location is also chosen because logging is available (for the original error).

Output formatting of the error is similar (in format) to the errors from defaul interop.

```
# Windows
.\task.exe -t .\ISSUES\2466 -v
task: "default" started
task: [default] cp -f foo fuvar/bar
CreateFile \\wsl.localhost\Ubuntu\home\trule\git\task\ISSUES\2466\foo: The system cannot find the file specified.
task: [default] command error ignored: exit status 1
task: "default" finished

# Linux
$ task -t ./ISSUES/2466/ -v
task: "default" started
task: [default] cp -f foo fuvar/bar
cp: cannot stat 'foo': No such file or directory
task: [default] command error ignored: exit status 1
task: "default" finished
```

Fixes #2466

### DIFF
--- a/internal/execext/coreutils.go
+++ b/internal/execext/coreutils.go
@@ -7,14 +7,14 @@ import (
 	"github.com/go-task/task/v3/internal/env"
 )
 
-var useGoCoreUtils bool
+var UseGoCoreUtils bool
 
 func init() {
 	// If TASK_CORE_UTILS is set to either true or false, respect that.
 	// By default, enable on Windows only.
 	if v, err := strconv.ParseBool(env.GetTaskEnv("CORE_UTILS")); err == nil {
-		useGoCoreUtils = v
+		UseGoCoreUtils = v
 	} else {
-		useGoCoreUtils = runtime.GOOS == "windows"
+		UseGoCoreUtils = runtime.GOOS == "windows"
 	}
 }
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -144,7 +144,7 @@ func ExpandFields(s string) ([]string, error) {
 }
 
 func execHandlers() (handlers []func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc) {
-	if useGoCoreUtils {
+	if UseGoCoreUtils {
 		handlers = append(handlers, coreutils.ExecHandler)
 	}
 	return handlers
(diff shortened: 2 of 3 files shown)


andreynering (commented):
Hey @trulede,

I think that having an specific error type in the middleware will work very well and be more reliable. What do you think?

* https://github.com/mvdan/sh/pull/1241

trulede:
@andreynering the problem I encountered was finding a way to print the cause of the error (its in the err object) so that the user has some indication of the failure.

A specific error type in the middleware would also facilitate that. However that solution is a bit more complex, IIRC you need changes in a few places/interfaces, and task also needs to handle _that_ error object. Its probably the _more_ correct approach. Its possibly a change over 3 repos - its quite a complex integration.

vmaerten:
It should be fixed by https://github.com/mvdan/sh/commit/1e6494c74f35978c6c97c42affd208c8f153fa6e
Included in https://github.com/go-task/task/pull/2943
If it does not solve the issue, feel free to re-open it

PR #2640, opened 2026-01-19
fix(includes): propagate silent mode from included Taskfiles to tasks

### LINKED ISSUE #1319: file-global silent:true not respected when in included file
- Task version: Task version: v3.29.1 (h1:q4mqGSR40qTOf9XZp2ySY3cM6enb2d+AqaxI/pEBiLk=)
- Operating system: Ventura 13.5
- Experiments enabled: none I guess

```yaml
# taskfile.yml
version: "3"
includes:
  silent: taskfile.silent.yml
```

```yaml
# taskfile.silent.yml
version: "3"
silent: true # <- look here 
tasks:
  me-not-silent:
    desc: "hello"
    cmds:
      - echo hello
 ```

```sh
$ task silent:me-not-silent
task: [silent:me-not-silent] echo hello # <- not very silent :(
hello
```

When silent set to the included job directly, then it works as expected.

trulede:
When merging a taskfile, retain the taskfile silent value, when set true, by applying that value to all merged tasks.

fixes #1319 

Test case to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Fixed `ROOT_TASKFILE` variable pointing to directory instead of the actual
   Taskfile path when no explicit `-t` flag is provided (#2635, #1706 by
   @trulede).
+- Included Taskfiles with `silent: true` now properly propagate silence to their
+  tasks, while still allowing individual tasks to override with `silent: false`
+  (#2640, #1319 by @trulede).
 
 ## v3.47.0 - 2026-01-24
 


vmaerten:
The issue with this approch is `slient: false` does not work anymore : 
```yml

  hello-not-silent:
    silent: false
    cmds:
      - echo "Hello from include not silent task"
```

I took a look back in time, and figure out it'll be hard to solve this issue without the scoping.
(FYI, I'm going open a new experiment soon about scoped taskfiles)

trulede:
@vmaerten I added a commit which handles that by parsing YAML into a `*bool`. But if you are going to change the merging with scoped taskfiles then its not worth the bother here.

PR #2659, opened 2026-01-25
feat: add SCOPED_TASKFILES experiment

vmaerten:
## Summary

This PR introduces the `TASK_X_SCOPED_TASKFILES` experiment which fundamentally changes how variables work in included Taskfiles. The goal is to eliminate variable conflicts and unexpected leakage between includes.

related to #2035

### What this experiment does

**1. Environment namespace**

Environment variables (OS + `env:` section) move to a dedicated namespace:

```yaml
# Before (legacy)
- echo "{{.HOME}}"
- echo "{{.MY_ENV}}"

# After (scoped)
- echo "{{.env.HOME}}"
- echo "{{.env.MY_ENV}}"
```

Special variables like `{{.TASK}}`, `{{.ROOT_DIR}}` stay at root level.

**2. Variable isolation**

Included Taskfiles can only see:
- Variables from the root Taskfile (inherited)
- Their own variables
- Variables explicitly passed via `includes: name: vars:`

They can NOT see variables from sibling includes anymore.

```yaml
# api/Taskfile.yml
tasks:
  show:
    cmds:
      - echo "{{.ROOT_VAR}}"  # works (inherited from root)
      - echo "{{.API_VAR}}"   # works (own variable)
      - echo "{{.WEB_VAR}}"   # empty (can't see sibling's vars)
```

**3. Clear priority order**

From lowest to highest:
1. Root Taskfile vars
2. Included Taskfile vars
3. Passthrough vars (`includes: foo: vars:`)
4. Task vars
5. Call vars (when calling a task with `vars:`)
6. CLI vars (`task foo VAR=value`)

**4. Escape hatch: `flatten: true`**

For gradual migration, you can use `flatten: true` on specific includes to restore legacy behavior:

```yaml
includes:
  api: ./api           # scoped (default)
  shared:
    taskfile: ./shared
    flatten: true      # legacy merge behavior
```

### Breaking changes

- `{{.HOME}}` → `{{.env.HOME}}` for all env vars
- Sibling includes can't see each other's variables anymore


solvingj:
I'm excited for the "can't see siblings" part, even if nothing else makes it through. 
Does this mean "{{.TASKFILE_DIR}}" will always resolve to the file it's referenced in, even on passthrough cases?
Right now I don't think passthrough includes can't resolve and define vars relative to themselves.  (happy to be wrong if i'm missing something).

trulede:
Regarding this:

```
# Before (legacy)
- echo "{{.HOME}}"
- echo "{{.MY_ENV}}"

# After (scoped)
- echo "{{.env.HOME}}"
- echo "{{.env.MY_ENV}}"
```

Envars can come from several locations: os, dotenv, env; and this is some disagreement on how those should be prioritised ( think dotenv, then os.env, and lastly taskfile myself). One real problem is to override whatever mechanism task is using, which could be done with `coalesce ` if each category of env were separated, which to some degree they are (2 of 3 at least, right now task envs and dotenv get "merged").

The scoping does however look a bit odd, how would it look with a function instead (see #1065 )

```
FOO: '{{ coalesce env "FOO" dotenv "FOO" osenv "FOO" "default" }}'

FOO: '{{ coalesce .env.FOO .dotenv.FOO .osenv.FOO "default" }}'
```

probably the . method is more natural looking.


Also, one useful thing with Task is that all envars become vars, so you can for the most part only work with vars in a taskfile which is useful for setting defaults, and probably other things. I do realise that some task user struggle with that, but it really is a fairly normal situation with makefiles/scripting (that envars becomes vars in the script). I guess that does not happen anymore? 


EDIT: I'm prototyping a Templater with some state (taskfile, task) that can support a syntax like:

```
FOO: '{{ coalesce tfenv "FOO" tdotenv "FOO" env "FOO" "default" }}'
```

The `tfenv`, `tdotenv`, `tenv` would be templater functions that directly reference the stateful vars of the templater.

PR #2666, opened 2026-01-29
Correct import directory resolution for nested imports.

### LINKED ISSUE #903: Working directory of included taskfiles is differs depending on task call location
Given the following Taskfiles:

```yaml
# test/Taskfile.yaml
---
version: 3

includes:
  foo:
    taskfile: lib/foo.yaml
    dir: ./

tasks:
  pwd:
    cmds:
      - echo $PWD
  toplevel:foo-bar-pwd:
    cmds:
      - task: foo:bar:pwd
```

```yaml
# test/lib/foo.yaml
---
version: 3

includes:
  bar:
    taskfile: ./bar.yaml
    dir: ./

tasks:
  pwd:
    cmds:
      - echo $PWD

  foolevel:bar-pwd:
    cmds:
      - task: bar:root
```

```yaml
# test/lib/bar.yaml
---
version: 3
tasks:
  pwd:
    cmds:
      - echo $PWD
```

We get the following behaviour when calling the above tasks:

```shell
🐟 ❯ task pwd
task: [pwd] echo $PWD
/tmp/test

🐟 ❯ task foo:pwd
task: [foo:pwd] echo $PWD
/tmp/test

took 10s
🐟 ❯ task foo:foolevel:bar-pwd
task: [foo:bar:pwd] echo $PWD
/tmp/test/lib

🐟 ❯ task foo:bar:pwd
task: [foo:bar:pwd] echo $PWD
/tmp/test/lib
```

IMO, this is unexpected behaviour and a *bug* - as I'm calling all tasks from `test/Taskfile.yaml`, which specifically declares the exec directory should be `./` for the included taskfiles. 

In otherwords, the correct behaviour should be as follows if running `foo.yaml` in the `lib` directory:

```shell
🐟 ❯ task -t foo.yaml pwd
task: [pwd] echo $PWD
/tmp/test/lib

🐟 ❯ task -t foo.yaml foolevel:bar-pwd
task: [bar:pwd] echo $PWD
/tmp/test/lib

🐟 ❯ task -t foo.yaml bar:pwd
task: [bar:pwd] echo $PWD
/tmp/test/lib
```

tl;dr: It looks like the value of `dir` is dependent on the location of task being called within the taskfiles includes tree, instead of being correctly merged/updated with the parent taskfile's `dir` when recursing into the included files.

## Task Version

```shell
🐟 ❯ task --version
Task version: v3.17.0 (h1:tDkE4X8VUUSNzFMTQNuEEhR/3cfI9hILLuwm365sNCM=)
```

### LINKED ISSUE #2620: Inconsistent path resolution when using different includes syntax
### Description

In the directory `C:\Users\xxx\Desktop\teskfile-err`, running task `web:portal` returns the path `C:\Users\xxx\Desktop\teskfile-err\`.But if I modify the includes block in web/main.yml to specify the taskfile key explicitly like this:
```
yaml
includes:
  portal:
    taskfile: portal.yml
```
The output path will change to ```C:\Users\xxx\Desktop\teskfile-err\web.```

### Version

3.46.4

### Operating system

win11

### Experiments Enabled

_No response_

### Example Taskfile

YAML
taskfile.yml

```version: '3'

includes:
  web: web/main.yml
vars:
  GREETING: Hello, world!

tasks:
  default:
    desc: Print a greeting message
    cmds:
      - echo "{{.GREETING}}"
    silent: true
```

web/main.yml
```version: '3'
includes:
  portal: portal.yml

tasks:
  web:
    desc: Run web-related tasks
    cmds:
      - pwd
      - echo "Running web tasks..."
```

web/portal.yml

```version: '3'

tasks:
  portal:
    desc: Run portal-related tasks
    cmds:
      - pwd
      - echo "Running portal tasks..."
```

trulede:
The importing mechanism was/is determining the import directory based on the path of the importing taskfile. This means that the directory of an imported taskfile is not following the expectation of users, and effectively becomes unpredictable as imports are nested.

```
A imports foo/B imports foo/bar/C (in each case, import dir not specified)
A TASK_DIR = ./
B TASK_DIR = ./
C TASK_DIR = ./foo  *** the logic behind this is cloudy ...
```

The logic seems to be that, C was imported from B, and therefore C should run in the folder of B (./foo above). However, if B is running in the folder of A (./) then having C run in the folder of B, even though B is running in the folder of A, seems to be wrong. C should run in the folder of B, if B runs in the folder of A, then C should _also_ run in the folder of A.

This PR changes that behaviour, so that it matches the user documentation, and imports taskfiles based on the resolved directory of the importing taskfile. As a result, imports are relative to one another, based on their resolved directories.

Essentially; if taskfile A imports taskfile B then it expects taskfile B to have the same working dir as A, unless the importing mechanism is configured by A (via import dir), in which case taskfile B would have that configured working dir.

fixes #903 
fixes #2620

### DIFF
--- a/taskfile/node_file.go
+++ b/taskfile/node_file.go
@@ -18,7 +18,7 @@ type FileNode struct {
 }
 
 func NewFileNode(entrypoint, dir string, opts ...NodeOption) (*FileNode, error) {
-	// Find the entrypoint file
+	// Find the entrypoint file.
 	resolvedEntrypoint, err := fsext.Search(entrypoint, dir, DefaultTaskfiles)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
@@ -27,7 +27,7 @@ func NewFileNode(entrypoint, dir string, opts ...NodeOption) (*FileNode, error) 
 		return nil, err
 	}
 
-	// Resolve the directory
+	// Resolve the directory.
 	resolvedDir, err := fsext.ResolveDir(entrypoint, resolvedEntrypoint, dir)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


Cediddi:
Hi @trulede why did you close this PR? I'm having the same problem and I can't find a workaround it, this PR would've fix my problem.

trulede:
@Cediddi There was exactly no interest (until now). When someone wants the PR they are welcome to use the code here as a starting point.

PR #2671, opened 2026-02-01
Sources accepts variables which represent arrays.

### LINKED ISSUE #88: More powerful sources and generates patterns
I have a task that needs to re-run whenever any tracked file in the entire folder changes (I use git to track what is a "tracked file"). I wanted to use `**/*` as the pattern, but obviously this doesn't work without tweaking, because the task will pick up changes in the build artifacts that the task itself produces.

The simplest approach would just be to allow for:

```
sources:
  sh: <given filenames on stdin, exit or stdout line determines whether it is an input>
```

I could also make do with something like `sources: { include: [...], exclude: [...] }`.

PS: Just wanted to thank you for this amazing tool. I really like that you have sparse, minimal feature set that composes in powerful ways (e.g. recursive invocation).

### LINKED ISSUE #709: sources to support reading an `ignore`, and looking at the opposite of what's ignored
I found myself writing this:

```yaml
  docker:bake:
    deps:
    - docker:create-multiarch-builder
    # this should match what's in .dockerignore
    sources:
    - ./cmd/**/*
    - ./internal/**/*
    - ./proto/**/*
    - ./testdata/**/*
    - ./go.*
```

with a `.dockerignore` file of:

```
*
!cmd/
!internal/
!proto/
!testdata/
!go.*
```

and thought, wouldn't it be nice if...

### LINKED ISSUE #948: Use sources as variable

### LINKED ISSUE #1018: Allow `:` separator to define multiple paths on a single `sources` entry

### LINKED ISSUE #2400: Dynamic variables in `sources` cause glob expansion to hang due to unpopulated placeholders and symlink loops

### LINKED ISSUE #2483: Specify task `sources` from dynamic list of files
### Description

It would be extremely useful to be able to specify the `sources` of a task from a dynamic list of files.

I'm builing a collection of reusable tasks in a monorepo and many of them accept a list of files to operate on. This list could be from the output of `find` or `ripgrep` or something similar. These tasks accept a `FILES` variable which I'm able to manipulate into the proper format for use in the task command. I'm unable, however, to set the `sources` of the task based on this variable.

To be clear, the list of files is something like `['file1.txt', 'file2.txt', 'file3.txt']`. It is not a pattern, and it is not fixed length.

It seems that setting `sources` to a dynamic list of files in any intuitive way is impossible.

Intuitive ways that don't work:

a) Setting `sources` directly equal to a dynamic list:`sources: '{{.MY_ARRAY}}'`.

b) Expanding individual items of `sources` into multiple files, similar to how a pattern is expanded.
b.i) By allowing the items to be arrays instead of just strings, similar to (a) but one level deeper.
b.ii) By allowing multiple patterns per item separated by, e.g., newlines or null-bytes.

Currently I'm resorting to building a massive string like `{file1.txt,file2.txt,file3.txt}` but even with this I'm encountering some annoying corner cases. And honestly at this point things are getting a bit silly.

trulede:
If a source (or generate) glob is a templated variable, and that variable is/was an array, then add that array of globs. The following example demonstrates.

* Fixes #1018
* Fixes #2400 
* Fixes #2483
* Fixes #88
* Fixes #709 
* Fixes #948 

```yaml
version: '3'

method: none

vars:
  GLOB_SINGLE: 'foo.*'
  GLOB_ARRAY: ['*.txt', '*.yml'] 
  GLOB_STRING: '*.txt:*.yml' 
  GLOB_SPLIT: '{{splitList ":" .GLOB_STRING}}'

tasks:
  default:
    cmds:
      - cmd: touch foo.txt
      - cmd: touch bar.txt
      - cmd: touch foo.yml
      - cmd: touch bar.yml
      - task: direct
      - task: single
      - task: array
      - task: string
      - task: split

  direct:
    sources:
      - '*.txt'
    cmds:
      - echo "== {{.TASK}} =="
      - for: sources 
        cmd: echo {{.ITEM}}
  single:
    sources:
      - '{{.GLOB_SINGLE}}'
    cmds:
      - echo "== {{.TASK}} =="
      - for: sources 
        cmd: echo {{.ITEM}}
  array:
    sources:
      - '{{.GLOB_ARRAY}}'
      - exclude: Taskfile.yml
    cmds:
      - echo "== {{.TASK}} =="
      - for: sources 
        cmd: echo {{.ITEM}}
  string:
    sources:
      - '{{splitList ":" .GLOB_STRING}}'
      - exclude: Taskfile.yml
    cmds:
      - echo "== {{.TASK}} =="
      - for: sources 
        cmd: echo {{.ITEM}}
  split:
    sources:
      - '{{.GLOB_SPLIT}}'
      - exclude: Taskfile.yml
    cmds:
      - echo "== {{.TASK}} =="
      - for: sources 
        cmd: echo {{.ITEM}}
```


timrulebosch:
It might be worthwhile combining changes from #2624 including/extending test coverage.

marat2509 (approved):

guettli:
Until this PR is merged, I use a work-around which I explain here: [skip this task unless files in this path changed since the last successful run](https://github.com/guettli/taskfile-is-great?tab=readme-ov-file#missing-feature) (feedback is welcome).

variables.go
```diff
@@ -99,15 +99,23 @@ func (e *Executor) compiledTask(call *Call, evaluateShVars bool) (*ast.Task, err
 	}
 
 	cache := &templater.Cache{Vars: vars}
+	globber := func(globs []*ast.Glob) []*ast.Glob {
+		// Delays globbing until dynamic variables are available.
+		if evaluateShVars {
+			return templater.ReplaceGlobs(globs, cache)
+		} else {
+			return origTask.Sources
```
vmaerten, line 107: I think this should be `globs` because if it's call for `generates` it'll return always `sources`

vmaerten (commented):
Thanks for the PR!  I have some concerns about the string-parsing approach in ReplaceGlobs.

  The detection logic strings.HasPrefix(v, "[") && strings.HasSuffix(v, "]") conflicts with valid glob syntax. [a-z] is a standard glob character class, but this code would interpret it as an array:

```yml
  tasks:
    example:
      sources:
        - '[a-z]'  # valid glob matching single-char files a, b, c... z
      cmds:
        - echo {{.ITEM}}
```

  Currently this works (in main) and matches files like a, b, c. With this PR, [a-z] would be parsed as an array containing a-z.

I know glob are not widely used and it can be edge cases.


I think relying on string representation of arrays is fragile.

  IMHO, Go's string representation of slices ([a b c]) is a debug format, no escaping, no quoting, no delimiter.
`strings.Fields` may also break on elements containing spaces.

  I think we can work on a type-aware resolution

  We already have  `template.Resolve()` which returns typed values (slices, strings, etc.) instead of stringifying them. This is the same mechanism used by `ResolveRef` for matrix variables.

If we try  `Resolve()` before falling back to `Replace()`, we can detect actual arrays at the type level.

I can try to make some suggestion in the code if you want

variables.go
trulede: Regarding the glob array type. I think this is for the case (#88):

```
GLOB_DYNAMIC:
    sh: 'echo "[*.txt *.text]"'
```

where the glob pattern really is a string encoded array (i.e. JSON), not a Go type. By the time this code is hit the globs have already been decoded from arrays to strings, so its only a matter of decoding the string itself ... if it seems like an array.

Of course, this might be better as a real JSON string, perhaps, or dropped from the PR altogether.

timrulebosch: I think reading in globs from a command is interesting, but a JSON encoding might be better and also not collide with the glob expressions?


So, code like this:

```
func appendGlobsFromJSON(v string, g *ast.Glob, new []*ast.Glob) []*ast.Glob {
	var decoded any
	if err := json.Unmarshal([]byte(v), &decoded); err != nil {
		return new
	}

	switch x := decoded.(type) {
	case []any:
		for _, item := range x {
			if glob, ok := item.(string); ok {
				new = append(new, &ast.Glob{Glob: glob, Negate: g.Negate})
			}
		}
	case string:
		return appendGlobsFromJSON(x, g, new)
	}

	return new
}
```

trulede: @vmaerten I've reworked that and it seems good. Additionally I've used JSON encoded strings for a more formal/robust representation of "string arrays".

If that is OK, we might consider documentation too?

trulede:
There is a conflict with recent For additions. Additionally, fingerprint needs the occur after the task is compiled.

Both items need to be checked/addressed.

trulede:
@vmaerten I've removed the support for dynamic vars, which was impossible to support correctly. Anyone wanting that can use the technique in PR #2711.

Your other feedback is addressed in the latest (squashed) commit.

andreynering:
@trulede Can you rebase this PR?

PR #2682, opened 2026-02-07
fix: improve error message when searching for Taskfile.

### LINKED ISSUE #1683: Clearly error when ownership check fails when walking parent directories to find a Taskfile
I ran into a tricky issue where Task wasn't finding an expected Taskfile: https://github.com/orgs/ddev/discussions/6273

What was confusing to me is that https://taskfile.dev/next/usage/#running-a-taskfile-from-a-subdirectory says

> it will walk up the file tree until it finds one (similar to how git works)

I think it's common for this to refer to git not crossing mounted file systems when looking for a `.git` directory. It wasn't clear to me that this would also involve ownership checks, which is done here: https://github.com/go-task/task/blob/40d77156df28e87c7cded7d99f312d98257439fc/taskfile/taskfile.go#L148-L150

I assume this is all to prevent potential security issues like the one in Git https://github.blog/2022-04-12-git-security-vulnerability-announced/.

It'd be great if Task could either:

- Clearly say why it stopped walking parent directories.
- Or, if we don't need to check ownership then don't.

trulede:
When searching for a Taskfile, and the directory ownership changes, provide a message indicating that.

In implementing this change, I noticed that the handling of TaskfileNotFoundError in setup.go would never be hit (bug) and also that the code in fs.go was a little odd - in so much that detecting this condition required returning valid paths _and_ an err, which in many cases the caller would ignore anyway. But I think its OK, just a bit odd.

```
~/git$ task --list-all
task: No Taskfile found at "" (or any of the parent directories until ownership changed). Run `task --init` to create a new Taskfile.

```
Fixes #1683

### DIFF
--- a/errors/errors_taskfile.go
+++ b/errors/errors_taskfile.go
@@ -11,14 +11,17 @@ import (
 // TaskfileNotFoundError is returned when no appropriate Taskfile is found when
 // searching the filesystem.
 type TaskfileNotFoundError struct {
-	URI     string
-	Walk    bool
-	AskInit bool
+	URI         string
+	Walk        bool
+	AskInit     bool
+	OwnerChange bool
 }
 
 func (err TaskfileNotFoundError) Error() string {
 	var walkText string
-	if err.Walk {
+	if err.OwnerChange {
+		walkText = " (or any of the parent directories until ownership changed)."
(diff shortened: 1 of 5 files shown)


vmaerten:
Thanks!

PR #2709, opened 2026-02-26
fix: address "taskfile not found" in some environments like docker

### LINKED ISSUE #2708: Taskfile not found on certain Linux distros
### Description

I tried to run `task` in a folder where a `Taskfile.yml` was located. I got the following message

```
task: No Taskfile found at "" (or any of the parent directories until ownership changed). Run `task --init` to create a new Taskfile.
```

I encountered the problem on Rocky Linux 8.10 and inside an Alpine docker container. It does not happen on Ubuntu 25.10 or Debian 13. 

I am pretty sure the problem came into being in the last few weeks because it did not exist before.

### Version

3.48.0

### Operating system

Rocky Linux 8.10, Alpine 3.23.3

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

trulede:
Revise the order of checks for better interop with Docker containers.

The caller SearchPathRecursively() **now** ignores the error _if_ a path was found _and_ consumes any error (previously it propagated the error).

fixes #2708

### DIFF
--- a/internal/fsext/fs.go
+++ b/internal/fsext/fs.go
@@ -153,13 +153,16 @@ func SearchPath(path string, possibleFilenames []string) (string, error) {
 // also check if the user ID of the directory changes and abort if it does.
 func SearchPathRecursively(path string, possibleFilenames []string) (string, error) {
 	paths, err := SearchNPathRecursively(path, possibleFilenames, 1)
-	if err != nil {
-		return "", err
+	if len(paths) > 0 {
+		// Regardless of the error, return the first possible filename.
+		return paths[0], nil
+	} else {
+		if err != nil {
+			return "", err
+		} else {
+			return "", os.ErrNotExist
+		}
 	}
-	if len(paths) == 0 {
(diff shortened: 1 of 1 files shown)


trulede:
@andreynering I do have a way to test this kind of thing automatically, but its too much effort within the Task repo. I may make an experiment (just for learning) ... but in this case I think the change will be "no worse". I will ask the originator to test it too ...

triole:
Hi, I tried your pull request but unfortunately it did not help. The message still occurs and task does not run anything. Sorry.

trulede:
@andreynering I think I have found and fixed the issue. With my testing its now working.

The problem was caused by f63a63fa6cf654dcdc647d5714d4bf102bbf8611

Confirmed that 3.47.0 and 3.48.0 worked, and pre-release 3.49.0/main was broken.

andreynering (approved):
Looks good, thank you!

triole:
Tested and it works fine. Thanks.

PR #1627, opened 2024-04-28
fix: expand task-level dotenv in brackets

aminya:
Fixes #997

Task level dotenv didn't expand the variables referenced by `{{.var}}` previously This fixes the issue. I added a test for it to prevent the bug from happening again.

PR #2711, opened 2026-03-01
Exchange values between task and deps using dotenv files.

### LINKED ISSUE #1381: dotenv files are loaded before task dependencies are ready

### LINKED ISSUE #1449: Add ability to register output of a task and make it available in templates context
`rash-sh` allows passing arbitrary values from one task from another one via register (see [here](https://rash-sh.github.io/docs/rash/v1.10/tasks.html))

```
tasks:
  sources:
    cmds:
      - sh: find . -type f -name '*.c'
  mytask:
    cmds:
      - task: sources
         register: sources_result
      - echo discovered files {{ .sources.output }}
      - echo stderr {{ .sources.extra.stderr }}
      - echo return code {{ .sources.extra.return_code }}
```

Starting from the above we can go further and allow multiple named returns

```
  sources:
    cmds:
      - sh: find . -type f -name '*.c'
         register: c_sources
      - sh: find . -type f -name '*.cpp'
         register: cpp_sources
      - sh: find . -type f -name '*.h'
         register: h_sources
    register:
         sources: # the result would be flattened (concatenation of lists) 
            - c_sources
            - cpp_sources
         headers: h_sources
      
  mytask:
    cmds:
      - task: sources
         register:
            inputs: sources
            headers: headers
            rc: _rc # assume we always register return code in `_rc` output
            sterr: _stderr # assume we always register return code in `_stderr` output
      - echo discovered files {{ .inputs }}
      - echo header files {{ .headers }}
      - echo stderr {{ .stderr }}
      - echo return code {{ .rc }}
```

This design has sufficient level of indirection to avoid any name collisions. Also it doesn't pollute the global namespace.

### LINKED ISSUE #1575: Lazy Loading Vars
Given the task file below I would love to be able to run *task list-prepared-statements* and have it first run the department of *list-work-groups* before it trys to evaluate the output into the var *WORK_GROUPS*. Either that or better still a way to capture the output of the first task as a var for future use.

Thanks

```yaml
version: 3

dotenv: [".env", "{{.ENV}}/.env.", "{{.HOME}}/.env"]

tasks:
  start-json-file:
    vars:
      FILE_NAME: '{{default "file_name.json" .FILE_NAME}}'
    cmds:
      - echo "[" > {{.FILE_NAME}}
    

  end-json-file:
    vars:
      FILE_NAME: '{{default "file_name.json" .FILE_NAME}}'
    cmds:
      - sed -i 's/}/},/' {{.FILE_NAME}}
      - sed -i '$ s/.$//' {{.FILE_NAME}}
      - echo "]" >> {{.FILE_NAME}}
      - cat {{.FILE_NAME}} | jq '.' > temp.json && mv temp.json {{.FILE_NAME}}

  list-work-groups:
    preconditions:
      - test {{.AWS_REGION}}
      - test {{.AWS_PROFILE}}
    cmds:
      - aws athena list-work-groups | jq -rc '.WorkGroups | .[].Name' > work-groups

  list-prepared-statements:
    preconditions:
      - test {{.AWS_REGION}}
      - test {{.AWS_PROFILE}}
    deps: [list-work-groups]
    vars:
      FILE_NAME: "prepared-statements.json"
      WORK_GROUPS: 
        sh: cat work-groups
    cmds:
      - task: start-json-file
        vars: {FILE_NAME: "{{.FILE_NAME}}"}
      - for: { var: WORK_GROUPS}
        cmd: echo "checking work group {{.ITEM}}"; aws athena list-prepared-statements --work-group $x >> {{.ITEM}}
      - task: end-json-file
        vars: {FILE_NAME: "{{.FILE_NAME}}"}
```

### LINKED ISSUE #2385: tasks outputs that dependent tasks can read from
### Description

# Usecase

Consider a setup where many tasks needs same credentials for different operations.

Therefore, the logical thing is to have a task that reads from some secret manager creds that must not be preserved on disk, and passes them **in-memory** to all its dependents.

# Current state
The only way I could find how to do it now - is have the "reader" task print something that the downstream stasks have to pass to eval - e.g. a step like `- eval $(task creds)`.

```
version: '3'

tasks:
  creds:
    cmds:
    - echo "export USR=me; export PWD=shhhh!!!;"
   
  operation1:
    cmds:
    - >
       eval $(task creds)
       some-cli-tool -u $USR -p $PWD some operation1

  operation2:
    deps: [creds]
    cmds:
    - >
       eval $(task creds)
       some-cli-tool -u $USR -p $PWD some operation2
```

That is less than ideal, first because it is platform-dependent and shell dependent. second - if I want to debug something - I should print it to `>&2` - and that's also a shell specific thing. third - the `eval` must happen on every command - it is not persisted even on the task level... If I want to pass it to env - it gets very clunky:

```
tasks:
  creds:
    cmds:
    - echo '{"USR":"me","PWD":"shhhh!!!;"}'
   
  operation1:
    vars:
       USR:  { sh: task creds | jq.USR }
       PWD:  { sh: task creds | jq.PWD}
    cmds:
    - >
       some-cli-tool-expects-ENV-vars operation3
```
this negates all the reuse beause the `operation<#>.vars` is coppied between all the operations that depend on `creds` :( 

# Proposal

I would propose tasks to have `outputs`, similar to how outputs work in github actions, just less cumbersome...

e.g:
```
version: '3'

tasks:
  creds:
    vars:
      JSON:
        sh: aws sts ....
      USR: 
        sh: echo {{.JSON}} | jq .USR
        exports: true  #makes it available on context of dependent tasks
      PWD: 
        sh: echo {{.JSON}} | jq .PWD
        exports: { as: SVC_API_KEY } # makes it exported as `SVC_API_KEY`

  operation1:
    deps: [creds]
    cmds:
    - some-cli-tool -u $USR some operation1
```
Mind that on `operation1` - it explicitly uses `$USR` and not `{{.USR}}` so that the value is not spat to screen.

It would be nice to support secrets:
```
  creds:
    vars:
      JSON:
        sh: aws sts ....
      USR: 
        sh: echo {{.JSON}} | jq .USR
        exports: true  # make it available on contexts of dependent tasks
      PWD:
         sh: echo {{.JSON}} | jq .PWD
         exports: true
         secret: true  # masks the secret from being print 
```
but that can be dealt in another feature

### LINKED ISSUE #2826: computed vars in dotenv paths
### Description

When a `dotenv` path references a variable computed via `sh`, the template renders incorrectly — the variable appears empty, causing the resolved path to fall back to the task's working directory.

Expected behavior

`{{.DOTENV}}` resolves to the output of `mktemp` and the dotenv file at that path is loaded.

Actual behavior

`{{.DOTENV}}` renders to an empty string. The empty path is resolved against the task's working directory.

```sh
$ task
task: Failed to read env file ...: read /home/user/project: is a directory
```

Related issues

- #434 — variable interpolation in dotenv: paths (implemented for static vars only)
- #1912 — CLI-provided vars not visible in dotenv: paths (includes a sh: mktemp example; closed pointing to PR #2288, which was not merged)
- #1312 — {{.USER_WORKING_DIR}} not available in global dotenv: paths (closed with workarounds only)

Workaround

Avoid `dotenv` entirely and source the file explicitly in each task that needs it:

```yaml
cmds:
  - |
    set -a && source {{.DOTENV}} && set +a
```

### Version

3.48.0

### Operating system

Ubuntu 20.04

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'

vars:
  DOTENV:
    sh: mktemp

tasks:
  default:
    dotenv:
      - '{{.DOTENV}}'
    cmds:
      - echo "AWS_KEY=$AWS_KEY"
```

trulede:
This PR allows the exchange of variables between tasks, and depend tasks, using the dotenv mechanism.

If a task has `deps` and `dotenv` then the `dotenv` files are reloaded which makes them available in the environment of subsequent task commands. 

Using this method the task does not need to be recompiled, which avoids complexities which might occur if templated variables in a task would change.

Fixes #1381 (and many others)
Fixes #1575 
Fixes #2385
Fixes #1449
Fixes #2826

Alternative to #1746 (improves upon)
Related to #1627 (alternative to)

In general, works as you might expect:

```
version: '3'
tasks:
  gen:
    cmds:
      - echo "BAR=bar" > .genenv  
  default:
    dotenv:
      - .genenv
    deps:
      - gen
    cmds:
      - echo "BAR=$BAR"
```


EvansJahja:
This would require writing to a file, feels like it's leaking context to the outside world, just because we don't have a ".RESULT" variable or something.

Also, concurrent tasks would be fighting on the same '.env' file

trulede:
> This would require writing to a file, feels like it's leaking context to the outside world, just because we don't have a ".RESULT" variable or something.
> 
> Also, concurrent tasks would be fighting on the same '.env' file

Its a normal way to do this. Simple to implement and easy to understand (and it already works!).
For more sophisticated use cases there things like Redis which can serve the same purpose without involving the filesystem.

PR #2718, opened 2026-03-05
docs: schema: add `tasks.task.method`

derhuerst:
As documented in https://github.com/go-task/task/blob/90bcbe9ba5b4e0c80f8f8ea3dbd2dffcaf8bc47b/website/src/docs/guide.md?plain=1#L816-L830.


vmaerten:
Thanks !

PR #2730, opened 2026-03-09
fix(fish): honor GO_TASK_PROGNAME for experiments cache

### LINKED ISSUE #2727: Task completion with the fish shell raise a warning when distro use the go-task name instead of task
### Description

Hello,

The task completion with the shell **fish** raised a warning on **manjaro** (and probably on **archlinux** too) because these systems use the name "go-task" instead of"task".

```
go-task fish: Unknown command: taskcom
in command substitution
        called on line 19 of file /usr/share/fish/vendor_completions.d/go-task.fish
in function '__go_task_get_experiments'
        called on line 27 of file /usr/share/fish/vendor_completions.d/go-task.fish
in function '__go_task_is_experiment_enabled' with arguments 'REMOTE_TASKFILES'
in command substitution
/usr/share/fish/vendor_completions.d/go-task.fish (line 19): Unknown command
    set -g __go_task_experiments_cache (task --experiments 2>/dev/null)
                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
in function '__go_task_get_experiments'
        called on line 27 of file /usr/share/fish/vendor_completions.d/go-task.fish
in function '__go_task_is_experiment_enabled' with arguments 'REMOTE_TASKFILES'
in command substitution
```

The completion still works afterwards, but the warning remains.

I've started looking at the code and i saw this : 

https://github.com/go-task/task/blob/54bdcba369357b47e19066b57badfb216a4c8d95/completion/fish/task.fish#L18-L19

If i understand this file correctly it seems like instead of using `task` directy the code should use `$GO_TASK_PROGNAME` .

Would you accept a PR for this ?

Thank you for your time !

### Version

3.48.0

### Operating system

Linux 6.18.12-1-MANJARO

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

SergioChan:
## Summary
- use `$GO_TASK_PROGNAME` instead of hardcoded `task` when refreshing fish completion experiment cache
- fixes fish completion warnings on distros where the binary name is `go-task`
- keeps completion behavior unchanged for default `task` installs

## Testing
- `git diff --check`
- `go test ./...` *(not runnable in this environment: required Go toolchain `go1.25` is unavailable via `GOTOOLCHAIN` download)*

## Related
- Fixes #2727

### DIFF
--- a/completion/fish/task.fish
+++ b/completion/fish/task.fish
@@ -5,7 +5,7 @@ set -g __task_experiments_cache ""
 set -g __task_experiments_cache_time 0
 
 # Helper function to get experiments with 1-second cache
-function __task_get_experiments
+function __task_get_experiments --inherit-variable GO_TASK_PROGNAME
     set -l now (date +%s)
     set -l ttl 1  # Cache for 1 second only
 


trulede (approved):
Fix seems OK. 
I'm not familiar with fish and can't verify.

appenzellergeller:
I did face the same issue today and have already tried the fix as it is proposed in this PR.
This will unfortunately not work in any case.

The are situations where the `__go_task_is_experiment_enabled` is called from another script and then the variable `$GO_TASK_PROGNAME` has not been defined in that scope.

If you add the `--inherit-variable GO_TASK_PROGNAME` to the function then it will work. See below.

```
function __go_task_get_experiments --inherit-variable GO_TASK_PROGNAME
[...]
    # Refresh cache
    set -g __go_task_experiments_cache ($GO_TASK_PROGNAME --experiments 2>/dev/null)
[...]
end
```

SergioChan:
Thanks for the repro details and suggestion — I pushed an update to this PR.

### What changed
- Added `--inherit-variable GO_TASK_PROGNAME` to `__task_get_experiments` in `completion/fish/task.fish`.
- This ensures the helper keeps access to `GO_TASK_PROGNAME` when invoked through `__task_is_experiment_enabled` from other completion contexts.

### Validation
- Attempted: `go test ./...`
- Result: could not run in this environment because the required Go 1.25 toolchain is unavailable.

If you want, I can also add `--inherit-variable GO_TASK_PROGNAME` to `__task_is_experiment_enabled` for extra explicitness, but this change should address the scope issue reported here.

SergioChan:
Thanks for the concrete fish-shell suggestion (`--inherit-variable GO_TASK_PROGNAME`) and for testing this path.

I’ll incorporate that approach and push an update.

PR #2733, opened 2026-03-09
templater: add joinEnv helper for PATH-like vars

### LINKED ISSUE #2406: joinEnv - Just like joinPath but with ; for windows else :

SergioChan:
## Summary
- add a new `joinEnv` template helper that joins path-like values with the OS-specific path list separator (`:` on POSIX, `;` on Windows)
- expose `joinEnv` in the templater function map so Taskfiles can build cross-platform `PATH` values cleanly
- add unit coverage for both direct helper behavior and template function registration

## Testing
- `git diff --check`
- `go test ./internal/templater` *(not runnable in this environment: required Go toolchain `go1.25` is unavailable via `GOTOOLCHAIN` download)*

## Related
Fixes #2406

### DIFF
--- a/internal/templater/funcs.go
+++ b/internal/templater/funcs.go
@@ -1,6 +1,7 @@
 package templater
 
 import (
+	stdos "os"
 	"maps"
 	"math/rand/v2"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/templater/funcs.go
```diff
@@ -89,6 +91,10 @@ func splitArgs(s string) ([]string, error) {
 	return shell.Fields(s, nil)
 }
 
+func joinEnv(elems ...string) string {
```
trulede, line 94: I would suggest to call this function `pathJoinList`.

trulede:
A more succinct solution to duplicate #2408

SergioChan:
Thanks, agree this should be more succinct and we should avoid redundant direction with #2408.

I’ll revisit this with a tighter approach and either update this PR accordingly or close it if superseded.

PR #2734, opened 2026-03-09
feat: add global preconditions at the taskfile level

### LINKED ISSUE #294: Add global preconditions

drichardson:
Closes #294

## Summary

Adds a top-level `preconditions` key to Taskfiles that applies to all tasks in that file before they run.

```yaml
version: '3'

preconditions:
  - sh: test -f .env
    msg: "Missing .env file. Run 'task init' to set up the project."
    inherit: true   # also applies to tasks in included Taskfiles

includes:
  tools: ./tools

tasks:
  init:
    skip_preconditions: true  # skips parent's inherited preconditions
    cmds:
      - cp .env.template .env
```

### Behavior

- Global preconditions are **file-scoped by default** — they apply to tasks in the file they are defined in, but not to tasks in included Taskfiles
- Setting `inherit: true` on a precondition causes it to propagate into included Taskfiles, transitively through the include hierarchy
- A task with `skip_preconditions: true` opts out of inherited preconditions from parent Taskfiles; global preconditions from the task's own file still apply
- `skip_preconditions` has no effect on root-level tasks (they have no inherited preconditions)
- Global preconditions are checked before a task's own preconditions

### Changes

**Core:**
- `taskfile/ast/taskfile.go` — `Preconditions []*Precondition` field on `Taskfile`
- `taskfile/ast/precondition.go` — `Inherit bool` field on `Precondition`
- `taskfile/ast/task.go` — `SkipPreconditions bool` field on `Task`
- `taskfile/ast/tasks.go` — scoped injection in `Tasks.Merge`: own-file tasks always get all preconditions; transitive-include tasks only get `inherit: true` preconditions unless `skip_preconditions`
- `taskfile/ast/graph.go` — same scoped logic applied to root preconditions after full merge

**Tests (20 new cases):**
- Global precondition met/not met for root and included tasks
- Non-inheriting precondition does not affect included tasks
- `inherit: true` blocks included tasks and propagates transitively (root → mid → deep)
- Mid-level `inherit: true` blocks sub-includes but does not affect root tasks
- `skip_preconditions` skips inherited preconditions
- `skip_preconditions` does not skip own-file preconditions
- `skip_preconditions` is a no-op on root tasks

**Docs:**
- `website/src/docs/guide.md` — new "Global preconditions" section with inherit and skip_preconditions usage
- `website/src/docs/reference/schema.md` — new entries in Root Schema, Task, and Precondition sections
- `website/src/public/schema.json` — `preconditions` on root taskfile, `inherit` on precondition object, `skip_preconditions` on task


drichardson:
don't have plans to complete this any time soon

PR #2735, opened 2026-03-10
templater: expose wildcard MATCH in include vars

### LINKED ISSUE #2732: Wildcards don't work with vars specified in task includes
### Description

Given this config file:

```yaml
version: '3'

includes:
  'stack:*':
    taskfile: ./Taskfile.stack.yml
    vars:
      ENV: '{{index .MATCH 0}}'
```

the app raises the following error:
```
template: :1:2: executing "" at <index .MATCH 0>: error calling index: index of untyped nil
```

The matching using `index .MATCH 0` works inside the included `Taskfile.stack.yml`, but since the configuration allows passing vars for the included task files, would it be possible to support `index .MATCH` in the includes vars section as well? That enables a clean approach where included tasks don't have to know how to extract the variables. 

### Version

3.49.1

### Operating system

MacOS 15.7.4

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

SergioChan:
## Summary
- seed call-time vars before include var templating so wildcard `MATCH` is available while evaluating `includes.*.vars`
- preserve existing variable resolution flow; change is scoped to include-var evaluation order
- add a regression fixture and test covering `ENV: '{{index .MATCH 0}}'` in wildcard includes

## Testing
- `go test ./... -run TestWildcardIncludeVarsCanUseMatch -count=1` *(fails locally: repo requires Go 1.25 toolchain, host has Go 1.22.2 and cannot download go1.25 in this environment)*
- Manual verification via regression test logic: before change include vars evaluation hit `index of untyped nil`; after change the test asserts output contains `ENV=prod`

## Related
- Fixes #2732

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -115,6 +115,13 @@ func (c *Compiler) getVariables(t *ast.Task, call *Call, evaluateShVars bool) (*
 		}
 	}
 	if t != nil {
+		if call != nil {
+			for k, v := range call.Vars.All() {
+				if err := rangeFunc(k, v); err != nil {
+					return nil, err
+				}
+			}
+		}
 		for k, v := range t.IncludeVars.All() {
 			if err := rangeFunc(k, v); err != nil {
 				return nil, err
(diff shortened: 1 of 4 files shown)


trulede:
I assume this is not wanted.

PR #2737, opened 2026-03-10
templating: add resolvePath function for absolute path resolution

### LINKED ISSUE #2681: Add a resolvePath or absPath function
### Description

Current function templates do not allow someone to resolve a path like like FOO=/foo/bar/zzz with `..`.

I looked at relPath and the other path related functions but I found none that can resolve paths (ideally even resolving symlinks).

Current workaround is to rely on bash to do it.

SergioChan:
## Summary
- add a new `resolvePath` template function to resolve relative segments into an absolute path
- make `resolvePath` resolve symlinks when available, with a safe fallback to the absolute path when symlink resolution is not possible
- add focused unit tests for path normalization and symlink resolution behavior

## Testing
- Not run in this environment: `go test ./internal/templater`
- Reason: the repository requires Go 1.25 toolchain and this runtime cannot download/provision that toolchain (`toolchain not available`)

## Related
- Fixes #2681

### DIFF
--- a/internal/templater/funcs.go
+++ b/internal/templater/funcs.go
@@ -34,6 +34,7 @@ func init() {
 		"IsSH":         IsSH, // Deprecated
 		"joinPath":     filepath.Join,
 		"relPath":      filepath.Rel,
+		"resolvePath":  resolvePath,
 		"merge":        merge,
 		"spew":         spew.Sdump,
 		"fromYaml":     fromYaml,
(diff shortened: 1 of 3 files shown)


internal/templater/funcs.go
```diff
@@ -34,6 +34,7 @@ func init() {
 		"IsSH":         IsSH, // Deprecated
 		"joinPath":     filepath.Join,
 		"relPath":      filepath.Rel,
+		"resolvePath":  resolvePath,
```
trulede, line 37: Is more granular approach better?

absPath: filepath.Abs
evalSymlinks: filepathEvalSymlinks

SergioChan: Great question! I intentionally kept this as a single `resolvePath` helper because issue #2681 is about resolving template paths to a final usable absolute path, and exposing one focused function keeps the template surface smaller.

Current behavior is effectively:
1) `filepath.Abs(...)`
2) best-effort `filepath.EvalSymlinks(...)`
3) fallback to the absolute path if symlink evaluation fails

If you prefer explicit `absPath` and `evalSymlinks` helpers instead, I can split this PR that way.

trulede: The general approach has been to add individual functions. I think its more useful.

PR #2738, opened 2026-03-10
docs: correct stdin Taskfile command example

### LINKED ISSUE #2610: Faulty documentation for reading Taskfile from stdin
### Description

Hi! :) 

Check out the [documentation for reading the Taskfile from stdin](https://github.com/go-task/task/blob/main/website/src/docs/guide.md#reading-a-taskfile-from-stdin). There is one example that simply does not work:
```bash
task -t - <(cat Taskfile.yaml)
```
I think what you actually mean is one of these:

```bash
# input redirection
task -t - < Taskfile.yaml
# process substitution
task -t <(cat Taskfile.yaml)
```

If you're not familiar with Bash process substitution (I'm unsure if a POSIX shell even supports that), run `echo -t <(cat Taskfile.yaml)`, which will show you what the shell resolves this to. On my system, it gave me `task -t /dev/fd/63`, with `/dev/fd/63` being the a name by which stdout of the `cat Taskfile.yaml` process can be accessed.

BTW: I think the examples implicitly assume Bash or a POSIX shell, which could be written in a more explicit way. That would be a major overhaul though.

Cheers!

Uli

### Version

n/a

### Operating system

Ubuntu 24.04

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

```

SergioChan:
## Summary
- fix the `Reading a Taskfile from stdin` guide example that used `task -t - <(cat ./Taskfile.yml)`
- replace it with `task -t - < ./Taskfile.yml`, which correctly passes stdin to Task when using `-t -`
- keep the existing piped example (`cat ./Taskfile.yml | task -t -`) unchanged

## Testing
- verified the docs snippet now contains the corrected stdin redirection command:
  - `rg -n "task -t - < ./Taskfile.yml|cat ./Taskfile.yml \| task -t -" website/src/docs/guide.md`
- note: `task` CLI is not installed in this runner, so executable command validation was not run locally

## Related
Fixes #2610

### DIFF
--- a/website/src/docs/guide.md
+++ b/website/src/docs/guide.md
@@ -99,7 +99,7 @@ from stdin, you must specify the `-t/--taskfile` flag with the special `-`
 value. You may then pipe into Task as you would any other program:
 
 ```shell
-task -t - <(cat ./Taskfile.yml)
+task -t - < ./Taskfile.yml
 # OR
 cat ./Taskfile.yml | task -t -
 ```


trulede (approved):
Correction seems reasonable.
Doc change only.

PR #2741, opened 2026-03-13
feat: add official pre-commit hook support

juftin:
## Summary

Closes #2562.

Adds official [pre-commit](https://pre-commit.com/) support by adding a `.pre-commit-hooks.yaml` at the repository root. This allows users to reference `repo: https://github.com/go-task/task` directly in their `.pre-commit-config.yaml` rather than maintaining a local hook definition.

## Changes

- **`.pre-commit-hooks.yaml`**: defines a single `task` hook with `language: golang`, `pass_filenames: false`, `require_serial: true`, and `minimum_pre_commit_version: "3.0.0"` (required for golang language support)
- **`precommit_test.go`**: validates the hook file is present, parses as valid YAML, and asserts all fields are correctly set
- **`website/src/docs/integrations.md`**: documents the integration with setup instructions, a configuration reference table, and collapsible examples covering common use cases
- **`website/src/docs/changelog.md`**: adds a changelog entry


precommit_test.go
trulede, line 1: This a static config file added to the task repo. I don't see the point of a test.

juftin: That's completely fair - I'm happy to remove it. The idea there was that a test could monitor that the file doesn't change its format - but agreed that it doesn't do much to guarantee functionality.

juftin: File deleted: https://github.com/go-task/task/pull/2741/commits/085a1c81ee15bc11af84c804ea5743d813feb5d9

trulede (approved):
LGTM.

I don't use hooks myself, but can see how this would be useful.

Issue #2715, opened 2026-03-02
`watch` issue


butuzov:
### Description

Change issued in files elsewhere (not in watch patters) stops watch (issues kill to running command). I do belive this is connected to switch to `v[3.43.1](https://github.com/go-task/task/releases/tag/v3.43.1)` introduced `fsnotify`.

part of the log with -v and -w flags
```
application - 11:06:59  INFO  layer=server/http msg=http setup
application - 11:06:59 DEBUG  layer=server/http/openapi msg=serving openapi base=/docs spec=/docs/openapi.yaml
application - 11:06:59  INFO  msg=[HTTP] server listening on: [::]:8000
task: received watch event: WRITE         "/Users/butuzov/Desktop/api.service/scripts/test_api.py"
application - 11:07:13  INFO  msg=[HTTP] server stopping
task: "build:live:all" finished
task: task "build:live:all" finished running
task: dynamic variable: "git describe --tags --match 'v*' --always --dirty" result: "dae5ced-dirty"
task: dynamic variable: "git rev-parse HEAD" result: "dae5cedf1460ee7fea4e8bff47c4a703b3e4ef43"
task: dynamic variable: "go env -json | jq -r '.GOOS'" result: "darwin"
task: dynamic variable: "go env -json | jq -r '.GOARCH'" result: "arm64"
task: skipped for file not in sources: scripts/test_api.py
```

### Version

3.48, i did track it back to 3.42.1 as last one not killing current execution, with false signals. 

### Operating system

macOS

### Example Taskfile

```YAML

```

trulede:
@butuzov you need to provide a minimal Taskfile (using simple commands like touch, sleep) which demonstrates the problem, and clear instructions on how to reproduce the effect. Thanks.

butuzov:
> [@butuzov](https://github.com/butuzov) you need to provide a minimal Taskfile (using simple commands like touch, sleep) which demonstrates the problem, and clear instructions on how to reproduce the effect. Thanks.

Tbh. I cannot recreate it with another repo... i am using such sources list in both cases

```yaml
x-be:dev:base: &x-be-dev-base
  sources:
  - exclude: ./cmd/service/wire_gen.go
  - ./**/*.proto
  - ./**/*.yaml
  - ./**/*.go
  - ./scripts/build-pipeline.sh
  - ./scripts/build-servies.sh
  method: timestamp
  silent: true
  ignore_error: false
```

in both cases its unpacking  into task via regular yaml syntax

```yaml
[...]

PR #2742, opened 2026-03-15
fix: Selective watching

butuzov:
Attempt of fixing selective watching, that become broken after switch to `fsnotify`.

Issue available at #2715
Alternative Solution (with tests) #2740


trulede:
duplicate of #2743

PR #2743, opened 2026-03-15
Watcher should ignore files not in sources before cancelling any running tasks.

### LINKED ISSUE #2477: Watch hijacking dependency process

### LINKED ISSUE #2715: `watch` issue

trulede:
Fix several issues with the watcher:
* the watcher would cancel running tasks _before_ evaluating the watch condition and, if the skip condition existed, _not even restart_ the cancelled tasks.
* events not being watched would cause running tasks to be cancelled (same root cause as above).
* a watched dir that was deleted and then recreated would no longer be watched.
* a new dir created, which should be watched, may not actually be watched.

Additional tests are added to cover most normal conditions.

* Closes #2477
* Closes #2715
* Duplicate #2493 
* Duplicate #2742

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -15,7 +15,6 @@ import (
 	"github.com/puzpuzpuz/xsync/v4"
 
 	"github.com/go-task/task/v3/errors"
-	"github.com/go-task/task/v3/internal/filepathext"
 	"github.com/go-task/task/v3/internal/fingerprint"
 	"github.com/go-task/task/v3/internal/fsnotifyext"
 	"github.com/go-task/task/v3/internal/logger"
@@ -24,6 +23,8 @@ import (
 )
 
 const defaultWaitTime = 100 * time.Millisecond
+
+var refreshChan = make(chan string)
 
 // watchTasks start watching the given tasks
 func (e *Executor) watchTasks(calls ...*Call) error {
@@ -68,9 +69,24 @@ func (e *Executor) watchTasks(calls ...*Call) error {
 
 	closeOnInterrupt(w)
 
+	watchFiles, err := e.collectSources(calls)
+	if err != nil {
+		cancel()
+		return err
+	}
 	go func() {
 		for {
 			select {
+			case path := <-refreshChan:
+				// If a path is added its necessary to refresh the sources, otherwise the
+				// watcher may not pick up any changes in that new path.
+				_ = path
+				watchFiles, err = e.collectSources(calls)
+				if err != nil {
+					e.Logger.Errf(logger.Red, "%v\n", err)
+					continue
+				}
+
 			case event, ok := <-eventsChan:
 				if !ok {
 					cancel()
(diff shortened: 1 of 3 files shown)


watch.go
```diff
@@ -78,34 +83,43 @@ func (e *Executor) watchTasks(calls ...*Call) error {
 				}
 				e.Logger.VerboseErrf(logger.Magenta, "task: received watch event: %v\n", event)
 
+				// Check if this watch event should be ignored.
+				if ShouldIgnore(event.Name) {
+					e.Logger.VerboseErrf(logger.Magenta, "task: event skipped for being an ignored dir: %s\n", event.Name)
+					continue
+				}
+				if event.Has(fsnotify.Remove) || event.Has(fsnotify.Rename) {
```
mkeeler:
```suggestion
				if event.Has(fsnotify.Remove) || event.Has(fsnotify.Rename) || event.Has(fsnotify.Write) {
```

Without this I think writes to existing files that are not in the sources would still trigger cancellation.

mkeeler:
@trulede Feel free to bring the tests over here if you like. What about the second half to my original PR to fix the task fingerprinting uniqueness issue? Should I refactor my PR to only have that fix within it.

watch.go
trulede: Yes, thanks. I've added that.

trulede:
@mkeeler I've copied over the watch_tests and will get them running and then update this PR.

Yes, refactor your PR to have the fix for the checksums.

mkeeler:
@trulede Is there any reason this cannot be merged (and #2740 too)

trulede:
I think @andreynering plans to merge this one soon.

vmaerten (changes requested):
Thanks for the PR ! 

In some cases the watch does not work as intended.

Considering this Taskfile : 
  ```yml
  version: '3'
  tasks:
    default:
      sources:
        - "**/*.txt"
      cmds:
        - echo "Task running!"
```

And those steps to reproduce : 

```bash
echo initial > a.txt
/tmp/task-repro --watch --verbose > /tmp/repro-watch.log 2>&1 &
sleep 1
echo updated-1 > a.txt
sleep 1
mkdir newdir && echo hello > newdir/new.txt
sleep 2
echo probe > a.txt
sleep 3
```

## Observed output

cat /tmp/repro-watch.log


```text
  task: Started watching for tasks: default
  task: [default] echo "Task running!"
  task: watching new dir: .
  Task running!
  task: task "default" finished running
  task: received watch event: WRITE         "/tmp/repro-watch/a.txt"
  task: [default] echo "Task running!"
  Task running!
  task: task "default" finished running
  task: received watch event: CREATE        "/tmp/repro-watch/newdir"
  task: watching new dir: newdir
  # here you should see an event for the write to a.txt but nothing appear
```

If I run it with main : 
```txt
  task: Started watching for tasks: default
  task: watching new dir: .
  task: "default" started
  task: [default] echo "Task running!"
  Task running!
  task: "default" finished
  task: task "default" finished running
  task: received watch event: WRITE         "/tmp/repro-watch2/a.txt"
  task: "default" started
  task: [default] echo "Task running!"
  Task running!
  task: "default" finished
  task: task "default" finished running
  task: received watch event: CREATE        "/tmp/repro-watch2/newdir"
  task: skipped for file not in sources: newdir
  task: received watch event: WRITE         "/tmp/repro-watch2/a.txt"
  task: "default" started
  task: [default] echo "Task running!"
  Task running!
  task: "default" finished
  task: task "default" finished running
  task: watching new dir: newdir
```

trulede:
That is a race condition in the underlying watcher (I believe). Its been there for a while.

The directory and file are created close together, such that the directory is not added until _after_ the file WRITE event of the file. Therefore it is missed. There is currently a go routine, which runs on an interval, and rebuilds/rescans the watcher and picks up these new directories. So, after a period of time, the watcher should trigger on the new directory (and its contained files).

That code pre-existed this PR. However, this PR does improve the rescan algorithm so that deleted and recreated directories are correctly handled. Previously, a deleted directory handle in the watcher would become stale, and recreated directories would _never_ trigger watch events.


The race condition might be solved in a subsequent PR, after refactoring out of some duplicated code. However its also possible that the existing periodic rescan is actually the only solution. It depends somewhat on the order that watcher events are raised (the order not always what you think).

butuzov:
does excluded files works again?

PR #2751, opened 2026-03-19
Fix dir not creating directory when shell variables are used

### LINKED ISSUE #1001: `dir` does not create a directory in some cases
- Task version:  v3.20.0

`Taskfile.yml`:
```
  test-dir:
    dir: does-not-exist
    cmds:
      - echo {{.TEST}}
    vars:
      TEST:
        sh: cat ../exist.txt
```

```
```

qianchongyang:
## Problem

When a task uses both a `dir` directive and shell variables, the directory is not created before the shell command executes. The task fails with `chdir: no such file or directory` because:

1. Shell variables (defined with `sh:`) are evaluated first
2. The directory creation happens after variable substitution
3. The shell command tries to `chdir` to a non-existent directory

## Solution

Reordered the execution flow to ensure the directory is created before shell variable commands are executed:

1. Create the specified directory before any variable evaluation
2. Then proceed with shell variable substitution
3. Finally execute the commands

This matches the documented behavior where `dir` should create the directory if it does not exist.

## Validation

```bash
# Before fix - fails with "no such file or directory"
task test-dir

# After fix - succeeds, creates directory and runs command
task test-dir
# Directory "does-not-exist" is created
# Command executes successfully
```

Fixes #1001

### DIFF
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -160,13 +160,14 @@ func dirOption(path string) interp.RunnerOption {
 			return nil
 		}
 
-		// If the specified directory doesn't exist, it will be created later.
-		// Therefore, even if `interp.Dir` method returns an error, the
-		// directory path should be set only when the directory cannot be found.
+		// If the specified directory doesn't exist, create it.
 		if absPath, _ := filepath.Abs(path); absPath != "" {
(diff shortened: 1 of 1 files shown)


trulede:
This is not the correct approach for the problem.

PR #2773, opened 2026-04-04
feat: add use_gitignore option to exclude ignored files from sources/generates

vmaerten:
## Summary

Add `use_gitignore: true` option at the Taskfile (global) and task level to automatically exclude files matching `.gitignore` rules from `sources` and `generates` glob resolution

Uses lightweight `sabhiram/go-gitignore` library (zero transitive dependencies) instead of `go-git`


trulede:
sabhiram/go-gitignore is a single file, you might just copy it into the Task repo.

Perhaps a test for negation (!) in a nested `.gitignore` file, or similar.

PR #2782, opened 2026-04-12
ci: add security action with `govulncheck` + enable `gosec` linter

andreynering:
Some explanation:

* [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) is a native Go tool that checks if we're using any vulnarable dependencies, including modules from `go.mod` and also Go itself. When it fails, we need to update that specific dependency or [the minimal Go version](https://github.com/go-task/task/blob/8fa9dc04aca4a372bd4c5a7ba84a2bc3e7eab82a/go.mod#L3) to the latest patch.
* [gosec](https://github.com/securego/gosec) is a tool that checks for security issues on our Go code. We're running it via `golangci-lint`. When not relevant, can be ignored with `//nolint:gosec`.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,19 @@
+name: Security
+
+on:
+  pull_request:
+  push:
+    tags:
+      - v*
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  govulncheck:
+    name: govulncheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{matrix.platform}}
     steps:
       - name: Check out code
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go ${{matrix.go-version}}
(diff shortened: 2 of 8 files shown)


taskfile/node_cache.go
```diff
@@ -63,7 +63,7 @@ func (node *CacheNode) WriteChecksum(checksum string) error {
 	if err := node.CreateCacheDir(); err != nil {
 		return err
 	}
-	return os.WriteFile(node.checksumPath(), []byte(checksum), 0o644)
+	return os.WriteFile(node.checksumPath(), []byte(checksum), 0o600)
```
trulede: I just wonder about this general change: 644 -> 600, there are some circumstances where this might not be a good idea. Especially Docker when a container is run as root, since any created file (by the container) would not be visible to the user (if they would subsequently run task outside of a container).

trulede (commented):
Only a comment on the 644 -> 600 changes.

taskfile/node_cache.go
andreynering: @copilot What’s your opinion? Is this change a good idea or can it break stuff?

vmaerten (approved):

taskfile/node_cache.go
pd93: I think I agree with @trulede here. `644` is only being flagged as a _potential_ security issue, not a detected one. There is some [chatter about it from the gosec team here](https://github.com/securego/gosec/issues/107). For example, this is here so that it catches cases like writing to a credentials file and is intended for the author to think about what permissions they actually need.

I don't believe that any of the cases here are actually security issues and can probably be `nolint`ed.

andreynering: @trulede @pd93 Yeah, I agree with you guys.

Just pushed an update to `.golangci.yml` to ignore those issues for our projects. Kept `//nolint:gosec` for the remaining ones.

pd93 (approved):
🚀

Issue #2784, opened 2026-04-12
go.mod integration as a tool: forward to the declared Task version


dolmen:
### Description

# Summary

When working in a Go module where a version of Task is declared as a [tool](https://go.dev/doc/modules/managing-dependencies#tools) dependency, transfer execution to `go tool task`.

# Rationale

Since Go 1.24 the Go toolchain has special support for versioning tools in `go.mod`: https://go.dev/doc/modules/managing-dependencies#tools
As `Task` is written in Go, it can be installed as a tool (see also #2783).
But for a complete integration, it would be better to ensure that `task` and `go tool task` run the same version.
So I propose that the `task` command running from `$PATH` would delegate to the version of the `task` command declared in `go.mod`.
 
# Proposal

When:
1. `task` detects it works from a Go module (a go.mod is found in a directory up to the directory of the `Taskfile.yml` config)
2. `go env GO111MODULE` is not `off`
3. that go.mod declares a version of task (`github.com/go-task/task/v[1-9][0-9]*/cmd/task`) in a [`tool` directive](https://go.dev/ref/mod#go-mod-file-tool): the `go tool` command lists the declared tools
4. there is a single tool named `task` declared (avoid ambiguous tool)
5. `task` isn't already the exact requested version (matching [`runtime/debug.BuildInfo`](https://pkg.go.dev/runtime/debug#BuildInfo)): avoid useless recursion
6. `task` has been launched as a simple command (`os.Args[0]` is not an absolute path, and doesn't even contain a path separator): avoid infinite recursion, and let the user override this behavior

Then:
* Pass all the received arguments (`os.Args[1:]`) to `go tool task` and exit

# Related issues

* #2783
* #2785
* golang/go#78673

# Implementation hints

Usefull `go` commands:
* `go list tool`: list tools declared in the Go module
* `go list -f '{{.ImportPath}}{{if not .Module.Main}}@{{.Module.Version}}{{end}}' tool`: list tools with their versions
* `go list -f '{{.Module.Version}}' github.com/go-task/task/v3/cmd/task`: list the requested version of Task

pd93:
> it would be better to ensure that `task` and `go tool task` run the same version.

Why? What exactly is the use case/benefit of this?

[...]

PR #2785, opened 2026-04-12
build(deps): declare ourself as a 'go tool' for ourself

dolmen:
In go.mod declare ourself as a [tool](https://go.dev/doc/modules/managing-dependencies#tools) for our own consumption.

```console
$ go get -tool ./cmd/task
```

Task developers can now run tasks via `go tool task`, independently of the `task` installed in `$PATH`.

This is a first step before further experiments of integrations such as #2784.

## Meta

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -134,3 +134,5 @@ require (
 	google.golang.org/protobuf v1.36.11 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+tool github.com/go-task/task/v3/cmd/task


trulede (commented):
@dolmen are you sure this is correct? From what I have read, `go get -tool ...` would add task as a tool dependency to some project, and you might do it to pin a specific version of a tool.

It seems to not make much sense for a tool to declare itself as a tool.

trulede:
It was already decided not to do this in the related issue.

PR #2788, opened 2026-04-13
feat(templater): add absPath template function

### LINKED ISSUE #2681: Add a resolvePath or absPath function

mateenali66:
adds absPath as a thin wrapper over filepath.Abs so Taskfiles can resolve paths with .. or . components without shelling out. slots in next to joinPath and relPath in internal/templater/funcs.go, same one-liner style.

originally discussed in #2681. i first tried adding osAbs in slim-sprig (go-task/slim-sprig#23), but per @trulede's suggestion in https://github.com/go-task/task/issues/2681#issuecomment-4238152667 that repo is inactive, so moving the change here instead.

the integration test under testdata/abs_path verifies `{{absPath "foo/../bar" | base}}` resolves to "bar". also added a one-line example to the Path Functions section of the templating reference docs, and a CHANGELOG entry.

symlink resolution (filepath.EvalSymlinks) was also mentioned in the issue but keeping scope to just absPath for now, happy to follow up with a second PR if wanted.

Fixes #2681

### DIFF
--- a/testdata/abs_path/Taskfile.yml
+++ b/testdata/abs_path/Taskfile.yml
@@ -3,4 +3,4 @@ version: '3'
 tasks:
   default:
     cmds:
-      - cmd: echo '{{absPath "foo/../bar" | base}}'
+      - cmd: echo '{{absPath "foo/../bar" | osBase}}'
--- a/testdata/abs_path/Taskfile.yml
+++ b/testdata/abs_path/Taskfile.yml
@@ -3,4 +3,4 @@ version: '3'
 tasks:
   default:
     cmds:
-      - cmd: echo '{{absPath "foo/../bar" | osBase}}'
+      - cmd: echo '{{absPath "foo/../bar"}}'
--- a/testdata/abs_path/Taskfile.yml
+++ b/testdata/abs_path/Taskfile.yml
@@ -1,0 +1,6 @@
+version: '3'
+
+tasks:
+  default:
+    cmds:
+      - cmd: echo '{{absPath "foo/../bar" | base}}'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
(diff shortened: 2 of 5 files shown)


testdata/abs_path/Taskfile.yml
```diff
@@ -0,0 +1,6 @@
+version: '3'
+
+tasks:
+  default:
+    cmds:
+      - cmd: echo '{{absPath "foo/../bar" | base}}'
```
trulede: What happens if you remove base? 

I wonder if this test is really proving that the function is working. It is really checking the _last_ element of the path, and _not_ that the path is absolute.

You should be able to echo the TASKDIR to get the working dir, and then absPath should resolve to TASKDIR+"/bar", then you can programmatically read both lines, and construct the check condition from the first line, and compare to the second.

trulede (changes requested):
Looks good. I've suggested an improvement to the test.

testdata/abs_path/Taskfile.yml
mateenali66: switched to comparing against filepath.Join(cwd, "bar") computed at test time in 8c9d68f. the osBase pipe only checked the last element, you were right that it wasn't proving the result is absolute or that .. got resolved. the new assertion fails if either property breaks.

tried the TASKFILE_DIR approach first but absPath runs during template evaluation using the test process cwd (the repo root), not the taskfile dir, so TASKFILE_DIR + /bar wouldn't match the actual output. cwd-based construction was the closest equivalent that still validates both properties.

trulede (approved):
Looks good.

Low risk change.

vmaerten (approved):
Thanks!

Please next time, do not edit the changelog. We do it after merging

PR #2789, opened 2026-04-13
docs: add call internal task within a task example

42sol-eu:
Added example task to schema documentation with notes on usage.


trulede (changes requested):
I don't think the comment part is necessary. You might check the spelling/capitalisation for the rest of the example (i.e. remaining consistent with the existing content).

Otherwise it seems reasonable.

42sol-eu:
thank you

PR #2797, opened 2026-04-15
feat: use GH_PAT for goreleaser

pd93:
Need to recreate `GH_PAT` before merging.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -89,6 +89,7 @@ brews:
     repository:
       owner: go-task
       name: homebrew-tap
+      token: "{{secrets.GH_GORELEASER_TOKEN}}" # So that it runs as the task-bot user
     test: system "#{bin}/task", "--help"
     install: |-
       bin.install "task"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -89,6 +89,7 @@ brews:
     repository:
       owner: go-task
       name: homebrew-tap
+      token: "{{secrets.GH_PAT}}" # So that it runs as the task-bot user
```
andreynering:
```suggestion
      token: "{{secrets.GH_GORELEASER_TOKEN}}" # So that it runs as the task-bot user
```

.goreleaser.yml
```diff
@@ -130,6 +131,7 @@ winget:
       owner: go-task
       name: winget-pkgs
       branch: 'task-{{.Version}}'
+      token: "{{secrets.GH_PAT}}" # So that it runs as the task-bot user
```
andreynering:
```suggestion
      token: "{{secrets.GH_GORELEASER_TOKEN}}" # So that it runs as the task-bot user
```

andreynering (approved):
Just created and added the token and added as `GH_GORELEASER_TOKEN`.

PR #2799, opened 2026-04-15
feat: add security docs to website and update contributing

pd93:
Copies the security and incident response plan docs and updates the contributing file to be inline with the one added in https://github.com/go-task/.github/pull/4.

Also removes the `CONTRIBUTING.md` from this repo since that will now be handled by the one in `.github`.


andreynering (approved):
👏 👏 👏

PR #2802, opened 2026-04-19
fix: refresh for-loop commands after deps

### LINKED ISSUE #1494: Bug: `sources` are checked before running `deps`
Hello,
firstly thanks for making such a great tool!
I believe I found a bug where `sources` are checked before `deps` complete:
```
  compile_sources:
    deps: [preprocess_sources]
    sources:
      - "./build/**/*.cc"
    generates:
      - "./build/**/*.o"
    cmds:
      - for: sources
        cmd: '{{.CXX}} {{.CPPFLAGS}} -c -o $(basename {{.ITEM}} .cc).o {{.ITEM}}'

  preprocess_sources:
    sources:
      - "./src/**/*.cc"
    generates:
      - "./build/**/*.cc"
    cmds:
      # Preprocess src/**/*.cc into build/**/*.cc
      - cp src/main.cc build/main.cc     # dummy command
```

The `preprocess_sources` task generates sources that are then are used as sources for `compile_sources`. But this is the output that I get:
```
❯ ls build                                                 # empty dir
❯ task compile_sources
task: [preprocess_sources] cp src/main.cc build/main.cc    # compile_sources does not run at all
❯ ls build
main.cc                                                    # no main.o
❯ task -f compile_sources                                  # force run
task: [preprocess_sources] cp src/main.cc build/main.cc
task: [compile_sources] clang++ .....                      # compile_sources is run
❯ ls build
main.cc  main.o                                            # main.o exists
```

What I believe is happening is:
1. `compile_sources` is called
2. `compile_sources` checks it's `sources`, generates checksum
3. Call dependencies
  3.1 `preprocess_sources` generates `.cc` files in `build/`
4. `compile_sources` does not generate `.o` files in `build/`, because `build/` was empty when `sources` were checked

Expected behavior:
1. `compile_sources` is called
2. Call dependencies
  2.1 `preprocess_sources` generates `.cc` files in `build/`
3. `compile_sources` checks it's `sources`, generates checksum
4. `compile_sources` generates `.o` files from `sources` in `build/`

- Task version: 3.34.1
- Operating system: Ubuntu
- Experiments enabled: No

lawrence3699:
Fixes #1494

When a task uses `for: sources` and the matching files are created by a dependency, the first run can skip the task because the command list is compiled before the deps materialize those sources.

The issue in #1494 looks like a fingerprint ordering bug at first, but the stale piece is actually the compiled `cmds` list. `deps` already run before fingerprinting; the missing step is refreshing commands that loop over `sources` or `generates` after those deps finish.

This patch keeps the fix scoped to that command-refresh step. It rebuilds the compiled command list after deps complete, so `for: sources` / `for: generates` loops see newly created files without re-running dynamic `sh:` vars.

## Reproduction

```yaml
version: '3'

tasks:
  create_deps:
    cmds:
      - touch a.dep b.dep c.dep

  use_deps:
    deps:
      - create_deps
    sources:
      - '*.dep'
    cmds:
      - for: sources
        cmd: echo {{.ITEM}}
```

Before this change, a clean run only executed `create_deps`, and `use_deps` could then be treated as up to date because its compiled command list was still empty.

After this change, the same first run executes the expanded `use_deps` commands and prints all three generated files.

## Testing

- `go test ./... -run '^TestForCmdsRecompileAfterDeps$'`
- `go test ./...`
- `go run ./cmd/task -d /tmp/task1494-repro use_deps`

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -894,6 +894,47 @@ func TestForCmds(t *testing.T) {
 	}
 }
 
+func TestForCmdsRecompileAfterDeps(t *testing.T) {
+	t.Parallel()
+
+	dir := t.TempDir()
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "Taskfile.yml"), []byte(`
+version: '3'
+
+tasks:
+  create-deps:
+    cmds:
+      - mkdir -p generated
+      - printf 'a\n' > generated/a.dep
+      - printf 'b\n' > generated/b.dep
+      - printf 'c\n' > generated/c.dep
+
+  use-deps:
+    deps:
+      - create-deps
+    sources:
+      - generated/*.dep
+    cmds:
+      - for: sources
+        cmd: cat "{{.ITEM}}"
+`), 0o644))
+
+	var buffer SyncBuffer
+	e := task.NewExecutor(
+		task.WithDir(dir),
+		task.WithSilent(true),
+		task.WithStdout(&buffer),
+		task.WithStderr(&buffer),
+	)
+	require.NoError(t, e.Setup())
+	require.NoError(t, e.Run(t.Context(), &task.Call{
+		Task: "use-deps",
+		Vars: ast.NewVars(),
+	}))
+
+	require.Equal(t, "a\nb\nc\n", string(PPSortedLines(t, buffer.buf.Bytes())))
+}
+
 func TestForDeps(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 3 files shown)


trulede:
PR #2711 is a variant of the same general issue. If a dep changes something, should the Task be recompiled?

What are the somethings?
* dotenv files
* sources  (e.g. generated files)

So a recompile triggered by either of those _only_ would keep the performance hit limited to when you actually want it. Can we trigger only if sources are changed?

I think it might be better to just recompile the entire task? So that _any_ templated values get updated, with the idea that this PR gets combined with #2711.

PR #2808, opened 2026-04-23
chore(golangci-lint): improve and clean configuration

ldez:

A small improvement to your configuration file.

`gofmt` is redundant with `gofumpt`.
`rewrite-rules` can be replaced with [`modernize`](https://golangci-lint.run/docs/linters/configuration/#modernize).

There is no need to instruct all the formatters with the import group, `gci` already handles it.

I guess the options inside the exclusion section (`paths` and `generated`) have been created during the v1 to v2 migration, those elements was here only for some edge case compatibilities with the v1, but you don't need them.


andreynering (approved):
This is great, thank you!

PR #2811, opened 2026-04-26
feat: Improve task init template options

B13rg:
This PR adds an additional flag `--init-template` that allows specifying a template to create.
It adds an additional template (`full.yml`) that shows more complex task usage.

User call it with `task --init-template=default` or `task --init-template=full` to template the two types.

This is related to issue #2012 

Created with assistance from `Claude Haiku 4.5`.

### DIFF
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -89,7 +89,7 @@ func run() error {
 
 	// Handle --init-template flag (list or create with specified template)
 	if pflag.Lookup("init-template").Changed {
-		if flags.InitTemplate == "" {
+		if flags.InitTemplate == "list" {
 			// List available templates
 			log.Outf(logger.Default, "Available templates:\n")
 			for _, name := range task.ListTemplates() {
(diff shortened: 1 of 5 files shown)


B13rg:
The current `full.yml` file is a good start, but a little rough on the edges.  I want to simplify it a little and integrate examples from the docs more closely.  Passing `list` to view available templates isn't very ergonomic.

Documentation changes still need to be made.

As wished for in the issue, this could be expanded to allow the user to set a custom template file via environment variable.  It could be added as an additional option `env` alongside the new `default` and `full` types.
I think this would be useful to be able to set the in a `.env` project file or a `.bashrc`.
If specified, it could also override the template used by the plain `--init` flag.

It big enough to potentially be a separate PR though.

B13rg:
On further thought, I think this approach is too heavy handed. I worry about stuffing the full opinionated template in every binary and being very rarely used. 

I think configuring a template file via env var is cleaner and doesn't require an additional flag. I will think further and update the issue

PR #2813, opened 2026-04-26
Create the task.dir before evaluating ShVars.

### LINKED ISSUE #1001: `dir` does not create a directory in some cases

trulede:
In some limited cases a ShVar may run commands which require the configured (not default) Task Dir to be created. This is especially the case if the ShVar attempts to access the parent folder of a Task with a configured Dir.

An attempt is made to create the Task Dir before the Task is compiled via `CompiledTask()`. If this fails (i.e. Task Dir depends on a templated ShVar value) then the Task Dir will be created later, after the Task is compiled. The edge case that remains is not addressed by this PR.

The testcase includes an example.

Fixes #1001 
Alternative to #2800


task.go
```diff
@@ -156,6 +156,14 @@ func (e *Executor) RunTask(ctx context.Context, call *Call) error {
 		}
 	}
 
+	// Attempt to create the task dir before compiling the task so that ShVars can run/access
+	// the configured task dir. If the directory can't be created (e.g. contains unresolved
+	// template values), continue as if nothing happened (the directory is created again later).
+	if t.Dir != "" && !strings.Contains(t.Dir, "{") {
```
vmaerten, line 162: I think it does not work and this cause the extra dir created.

Maybe we can use something like : 

```
func dirDependsOnShVar(rawDir string, vars *ast.Vars) bool {
    if !strings.Contains(rawDir, "{{") {
        return false
    }
    for name, v := range vars.All() {
        if v.Sh != nil && strings.Contains(rawDir, name) {
            return true
        }
    }
    return false
}
```


```suggestion
	if t.Dir != "" && !dirDependsOnShVar(origTask.Dir, t.Vars) {
```

vmaerten:
With this taskfile : 

```yml
version: '3'

tasks:
  package:
    dir: 'releases/{{.VERSION}}/artifacts'
    vars:
      VERSION:
        sh: echo v1.2.3          # en vrai : git describe --tags
    cmds:
      - 'echo "cmds tournent dans : $(pwd)"'
    silent: true

```


an unexpected directory is created, `releases` under `artifacts` : 

CleanShot 2026-08-10 at 22 04 56

trulede:
That is the fundamental problem, in this edge condition they depend on each other. The guard, like you suggest, might be OK.

But I think the problem is, I think, that `t.Dir` is already "releases/artifacts" at this point (fast compile) .... and the `{{` has been lost.


Edit:
So, having a Task that should run `sh` commands in a custom Task Dir, and that custom Task Dir depends on an `sh` command ... then its a problem that is going nowhere. Perhaps its possible to detect if a `var` has unresolved `sh` vars?

trulede:
This is a very difficult problem. One possible approach might be:

1/ Setup a template function which will error if a template can't be resolved, option `missingkey=error`:
https://github.com/go-task/template/blob/a8e846539ce33dc7702e9f6015be44478006c7d5/option.go#L40

2/ Create a replace function, analogue of this one, but handles this particular condition as an error:
https://github.com/go-task/task/blob/e8404ea6e0289da8ec97ff78405c842092433655/internal/templater/templater.go#L104

3/ Use that function to template the `t.Dir` value in compiledTask(evalSh=false):
https://github.com/go-task/task/blob/e8404ea6e0289da8ec97ff78405c842092433655/variables.go#L137

and handle the error condition by restoring/retaining the original value which should contain the {{ markers.

PR #2820, opened 2026-04-30
perf(templater): skip template engine for strings without `{{`

romnn:
First commit from #2819 with comments removed. Speeds up `task --list` from 13.8 seconds to ~5 seconds on a large monorepo. See original PR for details.

### DIFF
--- a/internal/templater/templater.go
+++ b/internal/templater/templater.go
@@ -68,6 +68,13 @@ func ReplaceWithExtra[T any](v T, cache *Cache, extra map[string]any) T {
 		return v
 	}
 
+	// Optimization: skip if string is not a template
+	if s, ok := any(v).(string); ok {
+		if !strings.Contains(s, "{{") {
+			return v
+		}
+	}
+
 	// Initialize the cache map if it's not already initialized
 	if cache.cacheMap == nil {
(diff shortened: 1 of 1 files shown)


trulede (commented):
LGTM.

If you like to experiment further, AI suggests this at the start of the function.

```
func ReplaceWithExtra[T any](v T, cache *Cache, extra map[string]any) T {
	if cache.err != nil {
		return v
	}

	// Optimization: Direct check for common simple types
	// If it's a string and doesn't have "{{", return immediately.
	if s, ok := any(v).(string); ok {
		if !strings.Contains(s, "{{") {
			return v
		}
	}
```

romnn:
Great suggestion, this indeed saved another 1.5 to 2 seconds on our repo!

I've folded it into my single commit as per the contributing guide and decided to add a very concise comment in both places to highlight that those short circuits are purely performance optimizations and not behavioral. I think those are stylistically fine and improve clarity, but I can also adjust or remove them.

trulede:
> Great suggestion, this indeed saved another 1.5 to 2 seconds on our repo!
> 

That is a great outcome! The PR is fine.

andreynering (approved):
Very simple, yet we never thought about doing this optimization 🙂 

Thank you!

PR #2825, opened 2026-05-05
Add templating function 'joinPathList'

### LINKED ISSUE #2406: joinEnv - Just like joinPath but with ; for windows else :

trulede:
This PR implements a simple `joinPathList` templating function which uses the `filepath.ListSeparator` value (from same namespace as existing functions).

It does not add a special variables as that can be done with environment variables (.dotenv files). 
Also it does not add the `joinUrl` templating function because its too similar to existing `urlJoin` and `urlPath`. The existing `joinPath` should cover this anyway.

Fixes #2406 
Alternative to #2408


andreynering:
Closing as implemented by:

* #2408

Issue #2300, opened 2025-06-18
Support for excluding entire namespaces in addition to individual tasks


dgroh:
### Problem Statement

Currently, the excludes feature in Task only allows excluding individual tasks by their full names (e.g., "flight-recorder:start", "flight-recorder:dump"). This becomes cumbersome when you want to exclude all tasks within a specific namespace.

### Current Behavior

```yaml
version: "3"
includes:
  common:
    taskfile: ../common/Taskfile.yml
    excludes:
      - "flight-recorder:start"
      - "flight-recorder:dump"
      # Need to list each task individually
```

### Desired Behavior

Allow excluding entire namespaces to simplify the configuration:

```yaml
version: "3"
includes:
  common:
    taskfile: ../common/Taskfile.yml
    excludes:
      - "flight-recorder"  # Excludes all tasks in flight-recorder namespace
```

### Use Case

When including a large taskfile with multiple namespaces, it's common to want to exclude entire groups of related tasks rather than listing each one individually. 

This is especially useful for:
* Excluding debugging/monitoring tools (like flight-recorder) from production services
* Excluding development-only tasks from deployment pipelines
* Excluding test suites that aren't relevant for certain contexts

### Proposed Implementation

The excludes array could support both:

* Individual task names (current behavior): "namespace:task"
* Namespace patterns (new feature): "namespace" or "namespace:*"

### Related

This builds upon the existing excludes feature introduced in #1859.

trulede:
@dgroh it could be an idea to support regular expressions in the `excludes` list. Something like this:

```
func MatchTextAuto(input, pattern string) bool {
	// Characters that indicate a likely regex pattern
	regexMetaChars := `*+?^$()[]{}|`

	// If the pattern contains any meta-characters, treat it as regex
	if strings.ContainsAny(pattern, regexMetaChars) {
		re, err := regexp.Compile(pattern)
		if err == nil {
			return re.MatchString(input)
		}
	}
	// Fallback to string matching.
	return strings.Contains(input, pattern)
}
```

kjasn:
> [@dgroh](https://github.com/dgroh) it could be an idea to support regular expressions in the `excludes` list. Something like this:
> 
> ```
> func MatchTextAuto(input, pattern string) bool {
> 	// Characters that indicate a likely regex pattern
> 	regexMetaChars := `*+?^$()[]{}|`
> 
> 	// If the pattern contains any meta-characters, treat it as regex
[...]

PR #2838, opened 2026-05-14
feat(includes): add `only` field to limit which tasks are included

JakesMD:
### What changed
- Added `only` field to `includes` as the inverse of `excludes`:
  ``` yml
  includes:
    shared:
      taskfile: shared.yml
      only: [foo, baa]
  ```
- Added tests
- Updated guide.md, schema.md and schema.json

### Use cases
- include specific task from a "third party" or even a remote taskfile
- avoid having to keep updating `excludes` whenever tasks are added to included taskfile
- most useful when `internal: false` and the included tasks are part of the "public API"

### Disclosure
- [x] AI was used during implementation

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1467,6 +1467,36 @@ func TestIncludesWithExclude(t *testing.T) {
 	assert.Equal(t, "foo\n", buff.String())
 }
 
+func TestIncludesWithOnly(t *testing.T) {
+	t.Parallel()
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir("testdata/includes_with_only"),
+		task.WithSilent(true),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+	)
+	require.NoError(t, e.Setup())
+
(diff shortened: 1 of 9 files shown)


JakesMD:
Hi! I was wondering if there's a specific reason this hasn't been implemented before. When you only want to include a small subset of tasks from a larger taskfile, an `only` option makes more sense than a long `excludes` list.

You could split the included taskfile into multiple smaller ones, but for remote taskfiles that might not be an option. And when multiple taskfiles depend on an overlapping subset of tasks, that doesn't work either. The included taskfile also shouldn't need to know how it's being used, so even filtering by prefix or glob pattern isn't a good fit.

I'm not sure about the field name `only`. Other options could be `includes`, `select(s)` or `subset`.

trulede (commented):
@JakesMD perhaps you take a look at #2300 and think if a regex on the existing `excludes` might work better. In that case, tasks _not_ matching the exclude regex are included.

JakesMD:
@trulede Yes, that would work. I personally don't think it's quite as elegant but I can see how it solves 2 problems in one go.

PR #2860, opened 2026-05-27
fix: validate include 'taskfile' field, suggest 'vars:' if missing

RichardHopperProGrammar:
## Problem

When a user accidentally writes `includes:` with variable-like content instead of `vars:` — a common typo like this:

```yaml
# Taskfile.yml
version: '3'

includes:
  GOBIN:
    sh: echo $(go env GOPATH)/bin

tasks:
  test:
    cmds:
```

...the current behavior is to silently accept the malformed include entry (since any YAML mapping passes the `yaml.MappingNode` check in `Includes.UnmarshalYAML`), and later report a misleading **"include cycle detected between A <--> A"** error, because the empty `Taskfile` path resolves to the same file.

This is confusing and unhelpful — the real problem isn't a cycle, it's that the user put variables under `includes:` instead of `vars:`.

## Solution

Added validation in `taskfile/ast/include.go` to check that each include entry has a non-empty `Taskfile` field after decoding. If it doesn't:

1. Returns a clear error: `include "GOBIN" is missing the required "taskfile" field; did you mean to use "vars:" instead?`
2. Lists unexpected keys found (e.g., `sh`) so the user understands what went wrong
3. Provides actionable guidance (use `vars:` instead)

## Changes

- `taskfile/ast/include.go`: Added validation in `Includes.UnmarshalYAML` after decoding each include value, checking for empty `Taskfile` and producing a helpful error message
- `testdata/includes_missing_taskfile/`: Test fixtures reproducing the exact scenario from issue #1881
- `task_test.go`: New test `TestIncludesMissingTaskfile` verifying the new error message is shown instead of the misleading cycle error

## Testing

- All existing include-related tests pass (`TestIncludes`, `TestIncludeCycle`, `TestIncludesIncorrect`, etc.)
- New test specifically validates:
  - Error contains "missing the required \"taskfile\" field"
  - Error does NOT contain "include cycle detected"

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1064,6 +1064,25 @@ func TestIncludeCycle(t *testing.T) {
 	assert.Contains(t, err.Error(), "task: include cycle detected between")
 }
 
+func TestIncludesMissingTaskfile(t *testing.T) {
+	t.Parallel()
+
+	const dir = "testdata/includes_missing_taskfile"
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir(dir),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+		task.WithSilent(true),
+	)
+
+	err := e.Setup()
+	require.Error(t, err)
+	assert.NotContains(t, err.Error(), "include cycle detected", "should not report a cycle error for a missing taskfile field")
+	assert.Contains(t, err.Error(), "missing the required \"taskfile\" field")
+}
+
 func TestIncludesIncorrect(t *testing.T) {
 	t.Parallel()
 
--- a/taskfile/ast/include.go
+++ b/taskfile/ast/include.go
@@ -1,6 +1,7 @@
 package ast
 
 import (
+	"fmt"
 	"iter"
 	"sync"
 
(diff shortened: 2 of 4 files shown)


trulede:
See #1902

RichardHopperProGrammar:
Thanks @trulede — great call. #1902 already covers this and has been through review, so I'll close this PR to avoid duplication. Good to know about the existing work!

trulede:
@RichardHopperProGrammar I think the other PR is dormant .... so you might prefer to keep yours open, and incorporate the feedback. Small concise changes that don't change too much have a good chance of being accepted.

RichardHopperProGrammar:
Thanks @trulede — I'll keep it open then. Good call about #1902 being dormant and the project moving away from go generate. Let me incorporate the feedback (switch to `fmt.Errorf` with `%w`, add test for the include cycle error) and push an update soon. Appreciate the guidance!

RichardHopperProGrammar:
Thanks @trulede — I see #1902 is dormant so I will keep mine open and incorporate feedback. The test for missing taskfile vs. cycle error is already included (line 1069 in task_test.go — verifies the error says "taskfile not found" and not "include cycle detected"). Happy to make any other refinements you suggest.

vmaerten:
Thanks! I've merged this one https://github.com/go-task/task/pull/2892

PR #2861, opened 2026-05-29
fix: resolve loop-local refs in task expansion

### LINKED ISSUE #2505: cannot reference loop variable using .ref
### Description

in a for loop cmd, the loop variable cannot be passed using .ref method

this works:
```yaml
for:
  var: any_iterable
  as: this
task: process
vars:
  thing: '{{ .this }}'
```

This does not:
```yaml
for:
  var: any_iterable
  as: this
task: process
vars:
  thing:
    ref: .this
```

### Version

3.45.4

### Operating system

linux

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: 3

silent: true

vars:
  global_var: global
  any_iterable: some thing


tasks:

  process:
    vars:
      thing:
        ref: default "undefined" .thing
    cmd: echo '{{ .thing }}'

  default:
    cmds:
      - task: cannot-pass-loop-variable--by-ref
      - task: cannot-pass-loop-variable--by-ref-default-name
      - task: cannot-pass-loop-variable--default-by-caller-does-not-see-it
      - task: can-pass-variable-by-ref
      - task: can-pass-loop-variable--by-template

  can-pass-variable-by-ref:
    cmds:
      - echo '[!!!] no loop | using .ref with a global variable'
      - task: process
        vars:
          thing:
            ref: .global_var

  cannot-pass-loop-variable--by-ref-default-name:
    cmds:
      - echo '[!!!] loop | using .ITEM'
      - for:
          var: any_iterable
        task: process
        vars:
          thing:
            ref: .ITEM

  cannot-pass-loop-variable--by-ref:
    cmds:
      - echo '[!!!] loop | using varible alias'
      - for:
          var: any_iterable
          as: this
        task: process
        vars:
          thing:
            ref: .this

  can-pass-loop-variable--by-template:
    cmds:
      - echo '[!!!] loop | using template'
      - for:
          var: any_iterable
          as: this
        task: process
        vars:
          thing: '{{ .this }}'

  cannot-pass-loop-variable--default-by-caller-does-not-see-it:
    cmds:
      - echo '[!!!] loop | setting fallback on (invoker) task'
      - for:
          var: any_iterable
          as: this
        task: process
        vars:
          thing:
            ref: default "invoker_default" .this
```

Yongcheng123:
## Summary
- resolve `ref:` values against loop-local variables during task expansion
- add regression coverage for both default `.ITEM` and aliased loop vars in `cmds` and `deps`
- preserve existing non-loop `ref:` behavior by keeping the extra context scoped to loop expansion

## Testing
- `go test ./... -run 'TestForCmds|TestForDeps'`
- `go test ./...`
- `go vet ./...`

Closes #2505


trulede:
@Yongcheng123 why have you opened this PR for an issue that already has an open PR?

Yongcheng123:
Good catch — I missed the existing open PR for #2505. I am closing this one to avoid duplicating work. Thanks for pointing it out.

PR #2870, opened 2026-06-01
fix: report cancellation reason when parallel tasks fail in failfast mode

### LINKED ISSUE #1226: Report reason for cancellation of parallel task
When two tasks run in parallel, and one fails, the other seems to be cancelled, but this cancellation is not reported explicitly.

It would be significantly clearer to users what is happening if the reason for cancellation was explicitly reported.

So for example, instead of (GitHub Actions output):

```console
⯈lint:mypy
⯈CI:test
task: Failed to run task "CI:lint": task: Failed to run task "lint:mypy": exit status 1
Error: Process completed with exit code 201.
```

have

```console
⯈lint:mypy
⯈CI:test
task: Terminated `CI:test` because parallel running task `CI:lint` failed.
task: Failed to run task "CI:lint": task: Failed to run task "lint:mypy": exit status 1
Error: Process completed with exit code 201.
```

toller892:
## Problem

When failfast is enabled and multiple tasks run in parallel, if one task fails, the other running tasks are cancelled via context. However, the error message only shows `context cancelled` without explaining _why_ the task was cancelled. This makes it hard for users to understand what happened.

**Current behavior:**

    ⯈lint:mypy
    ⯈CI:test
    task: Failed to run task "CI:test": context canceled
    task: Failed to run task "CI:lint": exit status 1

The user sees `context canceled` for `CI:test` but has no idea it was cancelled because `CI:lint` failed.

## Solution

Use `context.Cause()` to extract the original error that triggered the cancellation and include it in the error message. The `errgroup.WithContext` function (used when failfast is enabled) stores the first error as the context cause via `context.WithCancelCause`.

**New behavior:**

    ⯈lint:mypy
    ⯈CI:test
    task: Failed to run task "CI:test": task: cancelled due to: exit status 1
    task: Failed to run task "CI:lint": exit status 1

## Changes

- `task.go`: Added `enrichCancellationError` helper that checks for `context.Canceled` errors and enriches them with the cause from `context.Cause(ctx)`
- Applied to both `Run()` (top-level parallel tasks) and `runDeps()` (dependency tasks)

## Testing

- All existing tests pass, including `TestFailfast`
- The change is minimal and only affects error messages, not behavior

Fixes #1226

### DIFF
--- a/task.go
+++ b/task.go
@@ -85,8 +85,10 @@ func (e *Executor) Run(ctx context.Context, calls ...*Call) error {
 	}
 
 	g := &errgroup.Group{}
+	var failfastCtx bool
 	if e.Failfast {
 		g, ctx = errgroup.WithContext(ctx)
+		failfastCtx = true
 	}
 	for _, c := range regularCalls {
 		if e.Parallel {
@@ -98,7 +100,7 @@ func (e *Executor) Run(ctx context.Context, calls ...*Call) error {
 		}
 	}
 	if err := g.Wait(); err != nil {
-		return err
+		return enrichCancellationError(err, ctx, failfastCtx)
 	}
 
 	if len(watchCalls) > 0 {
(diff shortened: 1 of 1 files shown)


trulede:
AI Bot.

PR #2873, opened 2026-06-04
feat: configure output flags via environment variables

### LINKED ISSUE #2872: Allow configuring `--output` / `--output-group-begin` / `--output-group-end` via environment variables
### Description

Task supports environment-variable equivalents for several CLI flags (for example `TASK_VERBOSE=1` enables verbose mode, the same as `--verbose`). However, there is no environment-variable equivalent for the output-related flags:

- `--output` (`-o`)
- `--output-group-begin`
- `--output-group-end`
- `--output-group-error-only`

These can currently only be set as CLI flags on the `task` invocation, or as the static `output:` key in a Taskfile.

### Motivation / use case

In a CI pipeline (GitLab CI in our case), we want grouped output with custom begin/end markers so each task's output renders as a collapsible job-log section. The two existing ways to enable this both have drawbacks:

1. **`output:` in the Taskfile** is global and unconditional. It also affects local runs and, critically, corrupts values captured from `sh: task ...` dynamic variables, because the grouped output wrapping is applied to the captured stdout. So we cannot safely set it in the Taskfile.

2. **CLI flags** (`task -o group --output-group-begin=... --output-group-end=...`) work and are scoped to the invocation, but they must be added to **every** `task` call across the CI configuration. In a large pipeline with dozens of task invocations across many job templates, this is repetitive and error-prone, and easy to forget on new jobs.

An environment variable (e.g. `TASK_OUTPUT`, `TASK_OUTPUT_GROUP_BEGIN`, `TASK_OUTPUT_GROUP_END`, `TASK_OUTPUT_GROUP_ERROR_ONLY`) would let us enable CI-friendly grouped output once at the pipeline/job level (a single `variables:` entry) without touching the Taskfile or every invocation. This keeps local runs clean (no env var set) while turning on grouping everywhere in CI.

### Current behavior (Task 3.50.0)

```console
$ TASK_VERBOSE=1 task        # works: verbose mode enabled via env var
task: "default" started
...

$ TASK_OUTPUT=group task     # ignored: no grouping applied
... plain output ...
```

### Proposed behavior

Recognize environment variables mirroring the flags, with the usual precedence (CLI flag > env var > Taskfile `output:` > default). Suggested names, consistent with the existing `TASK_*` convention:

| Flag | Proposed env var |
|------|------------------|
| `--output` | `TASK_OUTPUT` |
| `--output-group-begin` | `TASK_OUTPUT_GROUP_BEGIN` |
| `--output-group-end` | `TASK_OUTPUT_GROUP_END` |
| `--output-group-error-only` | `TASK_OUTPUT_GROUP_ERROR_ONLY` |

### Notes

- I searched existing issues and did not find a duplicate. The closest are #2806 (unique section identifiers for GitLab), #937 (grouping vs streaming), and #647 (the begin/end markers themselves) which address different concerns.
- Happy to contribute a PR if the maintainers are open to this.

liiight:
## Summary

Closes #2872.

Adds environment-variable equivalents for the output-related flags so grouped output can be enabled once at the CI/job level (a single `variables:` entry) instead of being repeated on every `task` invocation or set unconditionally via the Taskfile `output:` key:

| Flag | Env var |
|------|---------|
| `--output` | `TASK_OUTPUT` |
| `--output-group-begin` | `TASK_OUTPUT_GROUP_BEGIN` |
| `--output-group-end` | `TASK_OUTPUT_GROUP_END` |
| `--output-group-error-only` | `TASK_OUTPUT_GROUP_ERROR_ONLY` |

These are wired through the existing `getConfig` helper (the same mechanism used by `TASK_VERBOSE`, `TASK_COLOR`, etc.), so the usual precedence applies: **CLI flag > env var > default**. An explicit CLI flag still overrides the env var via pflag's `Changed` tracking.

## Changes

- `internal/flags/flags.go`: source the four output flag defaults from `TASK_*` env vars.
- `internal/flags/flags_test.go`: unit tests for each env var, fallback-to-default, invalid-bool handling, and env > taskrc-config precedence.
- `website/src/docs/reference/environment.md` + `cli.md`: document the new env vars and cross-link them from the flags.
- `CHANGELOG.md`: Unreleased entry.

## Testing

- `go build ./...` and `go vet ./...` pass.
- `go test ./internal/...` passes (new flags tests included).
- Verified end-to-end with a built binary:
  - `TASK_OUTPUT=group` enables grouping
  - `TASK_OUTPUT_GROUP_BEGIN`/`END` apply templates
  - CLI flag overrides the env var
  - validation still rejects begin/end without `--output=group`


trulede (approved):
The changes look good. Mods normally update the changelog themselves.

Comment: recently envars were added for (only) CLI options which are also present in the task config schema. For CI (non-interactive) use cases some other envars might be useful:

* status
* watch
* parallel
* exit-code
* dir
* taskfile
* interval
* global
* force
* force-all

liiight:
I'll remove the changelog changes.
Regarding the other option you suggested, would you like me to add them as part of this MR as well? I'd be happy to

trulede:
> I'll remove the changelog changes. Regarding the other option you suggested, would you like me to add them as part of this MR as well? I'd be happy to

Lets see what the mods say. I just noted it when reviewing the change.

liiight:
Gentle bump. I would greatly appreciate your time and eyes on this MR.
Thanks in advance 🙏

vmaerten:
I'll take a look tomorrow

vmaerten (approved):
Thanks!

liiight:
Thank you for merging!

PR #2875, opened 2026-06-07
fix: escape apostrophes in fingerprint globs

### LINKED ISSUE #2874: "reached EOF without closing quote" when folder name has a quote in it
### Description

I ran `task`, I expected it to work, but it threw the error:
```
task: Failed to run task "default": 1:56: reached EOF without closing quote `'`
```

Here is how I can reproduce it on Linux Mint Debian Edition.

1. Create a folder with a quote `'` in the name, for example `test'd`. (It looks like it is enough to have a quote anywhere in the absolute path.)
2. Put into the folder the following `Taskfile.yml`:
```yaml

version: '3'

tasks:
  default:
    generates:
      - test.out
    sources:
      - test.in
    cmds:
      - cp test.in test.out
```
```console
$ tree test\'d
test'd
└── Taskfile.yml
```
3. Run `task` from inside the folder.

I've found this example after a lot of debugging. The initial issue was much more subtle: a target was rebuilt every time regardless of `sources`. There were no quote in the folder name, but there was one somewhere in the absolute path.

### Version

3.51.1

### Operating system

LMDE 7

### Experiments Enabled

_No response_

### Example Taskfile

```YAML

version: '3'

tasks:
  default:
    generates:
      - test.out
    sources:
      - test.in
    cmds:
      - cp test.in test.out
```

cyphercodes:
## Summary

- Treat apostrophes in fingerprint glob paths as literal path characters before shell-style field expansion.
- Add a regression test that runs a Taskfile from a temporary directory whose name contains an apostrophe, with both `sources` and `generates` configured.

Fixes #2874.

## Tests

- `go test . -run TestStatusChecksumAllowsApostropheInWorkingDirectory -count=1`
- `go test ./internal/fingerprint ./internal/execext . -count=1`
- `go test ./...`
- `git diff --check`

AI assistance was used to help investigate and prepare this patch; I reviewed and tested the changes.

### DIFF
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -93,6 +93,7 @@ func RunCommand(ctx context.Context, opts *RunCommandOptions) error {
 func escape(s string) string {
 	s = filepath.ToSlash(s)
 	s = strings.ReplaceAll(s, " ", `\ `)
+	s = strings.ReplaceAll(s, "'", `\'`)
 	s = strings.ReplaceAll(s, "&", `\&`)
 	s = strings.ReplaceAll(s, "(", `\(`)
 	s = strings.ReplaceAll(s, ")", `\)`)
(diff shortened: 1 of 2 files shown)


trulede:
This is not really a good solution. Would you mind coming up with something better.

trulede:
Close in favour of #2879

PR #2879, opened 2026-06-10
Handle apostrophes in fingerprint glob paths

kv1sidisi:
- Escape apostrophes when expanding fingerprint globs so quoted paths are parsed correctly.
- Add a regression test for a directory name that contains an apostrophe.

### DIFF
--- a/internal/execext/exec.go
+++ b/internal/execext/exec.go
@@ -96,6 +96,7 @@ func escape(s string) string {
 	s = strings.ReplaceAll(s, "&", `\&`)
 	s = strings.ReplaceAll(s, "(", `\(`)
 	s = strings.ReplaceAll(s, ")", `\)`)
+	s = strings.ReplaceAll(s, "'", `\'`)
 	return s
 }
 
--- a/internal/fingerprint/glob_test.go
+++ b/internal/fingerprint/glob_test.go
@@ -1,0 +1,24 @@
+package fingerprint
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGlobHandlesApostropheInPath(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


internal/fingerprint/glob_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGlobHandlesApostropheInPath(t *testing.T) {
```
trulede, line 11: A table driven test, for each of the cases in the code, would be useful.

internal/execext/exec.go
```diff
@@ -96,6 +96,7 @@ func escape(s string) string {
 	s = strings.ReplaceAll(s, "&", `\&`)
 	s = strings.ReplaceAll(s, "(", `\(`)
 	s = strings.ReplaceAll(s, ")", `\)`)
+	s = strings.ReplaceAll(s, "'", `\'`)
```
trulede, line 99: Something like this would be better:
```
var pathReplacer = strings.NewReplacer(
	" ", `\ `,
	"&", `\&`,
	"(", `\(`,
	")", `\)`,
	"'", `\'`,
)

func escape(s string) string {
	s = filepath.ToSlash(s)
	return pathReplacer.Replace(s)
}
```

But I also wonder if there are other characters which need escaping?

trulede (commented):
Seems OK, I offered some feedback.

internal/fingerprint/glob_test.go
trulede: Actually, this test is not very useful. It would be better to make an Executor/Task test where you have a list of files including characters which should be escaped, then run the Task, then check the files are handled correctly. You might even find that such a test exists.

Such a test would cover the entire processing chain (in case it gets broken somewhere else too).

PR #2883, opened 2026-06-18
perf: avoid eager fingerprint variable evaluation

Napolitain:
This is a **stacked** optimization PR on top of the filesystem benchmark work in #2881. Please review or merge #2881 first; this branch intentionally depends on those benchmarks so the performance impact can be evaluated with the same many-small and few-large fixtures. In this way we can also easily revert a commit if needed, without affecting the test (which is naturally harmless).

That PR avoids duplicate fingerprint evaluation during task compilation

(I think, if I put the base as my branch, then I can't switch of origin)

```text
BenchmarkManySmallFiles/checksum-28                    3         139978116 ns/op           0.71 MB/s             0.09537 source_MiB/op     20000 source_files/op      679207408 B/op    248202 allocs/op
BenchmarkManySmallFiles/timestamp-28                   3          42650460 ns/op           2.34 MB/s             0.09537 source_MiB/op     20000 source_files/op      25159722 B/op     188222 allocs/op
BenchmarkManySmallFiles/native-mtime-28                3          18077746 ns/op           5.53 MB/s             0.09537 source_MiB/op     20000 source_files/op      11558392 B/op     123036 allocs/op
BenchmarkManySmallFiles/none-28                        3            701787 ns/op         2598624 B/op       3193 allocs/op

BenchmarkFewLargeFiles/checksum-28                     3          19877230 ns/op        27009.34 MB/s          512.0 source_MiB/op            4.000 source_files/op     517832 B/op       1717 allocs/op
BenchmarkFewLargeFiles/timestamp-28                    3            147128 ns/op        3648997.44 MB/s        512.0 source_MiB/op            4.000 source_files/op     261288 B/op       1730 allocs/op
BenchmarkFewLargeFiles/native-mtime-28                 3             17511 ns/op        30659066.42 MB/s               512.0 source_MiB/op             4.000 source_files/op      4701 B/op         44 allocs/op
BenchmarkFewLargeFiles/none-28                         3            967514 ns/op         2599981 B/op       3193 allocs/op
```

### DIFF
--- a/variables.go
+++ b/variables.go
@@ -5,6 +5,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/joho/godotenv"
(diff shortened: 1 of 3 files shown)


variables.go
```diff
@@ -326,6 +329,132 @@ func (e *Executor) compiledTask(call *Call, evaluateShVars bool) (*ast.Task, err
 	return &new, nil
 }
 
+func taskReferencesFingerprintVar(t *ast.Task, kind string) bool {
```
andreynering: We have a lot a new functions in this file. I propose to move all of them to its own file in the `taskfile/ast` package. What do you think?

Maybe `taskfile/ask/fingerprint.go`

andreynering (commented):
Hey @Napolitain

I added a comment below. Can you take a look + rebase with `main`?

variables.go
Napolitain: Yes, I moved them to `taskfile/ast/fingerprint.go`.

andreynering (approved):
Thank you!

PR #2886, opened 2026-06-20
fix: sanitize all invalid characters in checksum filenames

s3onghyun:
## Description

`normalizeFilename` (in `internal/fingerprint`) turns a task name into the on-disk checksum/timestamp filename, replacing invalid characters with `-`. It uses:

```go
var checksumFilenameRegexp = regexp.MustCompile("[^A-z0-9]")
```

`A-z` is the classic range bug: it spans ASCII 65 (`A`) to 122 (`z`), which **includes the six characters between `Z` and `a`** — `[ \\ ] ^ _ \``. Those are not sanitized. The intent (per the comment) was clearly `[A-Za-z0-9]`.

Concrete impact: a task name containing one of these characters (notably `\\`, a path separator on Windows) leaks into the checksum/timestamp file path, which can corrupt the file location and break `sources:`/`generates:` up-to-date detection for that task.

## Fix

`[^A-z0-9]` → `[^A-Za-z0-9]`.

## Test

Extended `TestNormalizeFilename` with cases for the leaked characters (`\\`, `_`, `[`, `]`, `^`, `\``). Fails before, passes after.

Note: for any existing task whose name contains one of these characters, the normalized filename changes once, so the first run after upgrade recomputes the checksum (treated as not-up-to-date) — expected and harmless.

### DIFF
--- a/internal/fingerprint/sources_checksum_test.go
+++ b/internal/fingerprint/sources_checksum_test.go
@@ -16,6 +16,10 @@ func TestNormalizeFilename(t *testing.T) {
 		{"foo/bar/baz", "foo-bar-baz"},
 		{"foo@bar/baz", "foo-bar-baz"},
 		{"foo1bar2baz3", "foo1bar2baz3"},
+		{"foo\\bar", "foo-bar"},
+		{"foo_bar", "foo-bar"},
+		{"foo[bar]baz", "foo-bar-baz"},
+		{"foo^bar`baz", "foo-bar-baz"},
 	}
 	for _, test := range tests {
 		assert.Equal(t, test.Out, normalizeFilename(test.In))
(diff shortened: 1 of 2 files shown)


internal/fingerprint/sources_checksum_test.go
```diff
@@ -16,6 +16,12 @@ func TestNormalizeFilename(t *testing.T) {
 		{"foo/bar/baz", "foo-bar-baz"},
 		{"foo@bar/baz", "foo-bar-baz"},
 		{"foo1bar2baz3", "foo1bar2baz3"},
+		// Characters in the ASCII range between 'Z' (90) and 'a' (97) must
```
trulede: I don't think this comment is necessary, but the checks are OK.

trulede (approved):
The change seems reasonable. 

I would actually do this, but that is a matter of preference.

`var timestampFilenameRegexp = regexp.MustCompile(`[^[:alnum:]]`)`

s3onghyun:
Thanks for the review, @trulede! Adopted both:

- Removed the explanatory comment in the test.
- Switched the regexp to `[^[:alnum:]]` as you suggested — cleaner, and in Go's RE2 it's exactly equivalent to `[^A-Za-z0-9]` (ASCII alnum), so behavior is unchanged.

`TestNormalizeFilename` still passes.

vmaerten (approved):
Thanks!

PR #2892, opened 2026-06-26
fix: reject include without taskfile or dir

### LINKED ISSUE #1881: Incorrect/confusing `include cycle detected` error message after a silly mistake

Lewin671:
Fixes #1881.

## What changed

- Reject advanced include mappings that specify neither `taskfile` nor `dir`.
- Add a regression test for the malformed include case from the issue.
- Assert the old misleading include-cycle error is no longer returned.

## Why

A mapping under `includes:` without a Taskfile path currently falls through to default Taskfile discovery. In the issue example, that makes a simple configuration mistake look like an include cycle. Failing during Taskfile decoding gives the user a direct configuration error instead.

## Validation

- `go test ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run`

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -1082,6 +1082,25 @@ func TestIncludesIncorrect(t *testing.T) {
 	assert.Contains(t, err.Error(), "Failed to parse testdata/includes_incorrect/incomplete.yml:", err.Error())
 }
 
+func TestIncludesMissingTaskfile(t *testing.T) {
+	t.Parallel()
+
+	const dir = "testdata/includes_missing_taskfile"
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir(dir),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+		task.WithSilent(true),
+	)
+
+	err := e.Setup()
+	require.Error(t, err)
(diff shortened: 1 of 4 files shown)


vmaerten (approved):
thanks ! I've added the schema validation as well

PR #2897, opened 2026-06-29
feat(completion): unify shell completions behind `task __complete`

vmaerten:
## Summary

Adds a shell completion engine: a single `task __complete` command written in Go, that drives Bash, Fish, Zsh and PowerShell.
The shell scripts become thin wrappers that forward the current words to `task __complete` and render the suggestions, so every shell gets the same, richer behaviour and there is no more per-shell drift.

The wire protocol mirrors cobra v2 (`value\tdescription` lines followed by a `:<directive>` bitfield) behavior.

## Why

It's painful to maintain 4 (or more) different completion scripts, and it's also difficult to test them. With this approach, all new flags will be included in every shell by default.

## What it bring more than old completion

It's not only a drop in replacement of the old completions.
Now we can have requires vars being autocompleted, and even with enum 🚀 

## Testing

Following the cobra approach, the protocol is tested in Go and the generated scripts are only smoke-tested for how they route each directive:

- `internal/complete/complete_test.go`: white-box tests of the engine.
- `completion/protocol_test.go`: black-box test of the real binary's `__complete` output (candidates + directive).
- `completion/tests/`: thin per-shell smokes (bash/zsh/fish/powershell) that assert directive routing (no-files, dir-only, extension filter, inline `--flag=`, …), runnable via `task test:completion`.
- A `completion` CI job runs the smokes on Ubuntu and macOS (with fish and PowerShell installed, strict mode).

> Disclaimer: I used AI to write the completion scripts & the testing protocol (in bash).


vmaerten:
@arturict, please don't comment on draft PRs. Also, your comments appear to be AI-generated. We're still open to meaningful, human-written comments. Finally, please avoid giving instructions to maintainers as if you were one of them.

arturict:
Understood, and sorry about that. I should not have commented on a draft or phrased the comment as an instruction. I'll step back.

PR #2899, opened 2026-06-30
maint: update dependencies to address CVE-2026-39832

bertold:
The current golang.org/x/crypto dependency has a [vulnerability](https://pkg.go.dev/vuln/GO-2026-5006). Simple updates to the dependencies using `go get -u` and `go mod tidy` would address this issues. Thank you for considering.


vmaerten:
Hello
Task is not affected by this vulnerability. We run [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) to ensure that our code does not contains any vulnerability.
The CVE is about the agent (golang.org/x/crypto/ssh/agent)  and we do not use it

bertold:
I understand, but I think y'all missing a point it gets flagged in security scans if the task tool is part of any container. It is nuisance to mute all these potential issues. And not to mention the interaction with other components/tools on the system cannot be guaranteed. I feel that this is a normal hygiene to keep dependencies up-to-date especially if they do not break the code and do not impact required minimum language levels.

vmaerten:
Thanks for the explanation!
I've updated our Renovate bot to take indirect deps into account so it's automatically done now : 
- https://github.com/go-task/task/pull/2903
- https://github.com/go-task/task/pull/2889

Thanks for the report

PR #2904, opened 2026-07-01
fix: remote taskfiles from ADO git path

### LINKED ISSUE #1317: Remote Taskfiles experiment

pd93:
Fixes https://github.com/go-task/task/issues/1317#issuecomment-4847494248.

Adds a condition to the `getScheme` function so that Azure Devops git paths are detected correctly.

Thanks for always being different Microsoft 🙄 

FYI @YoungOak

### DIFF
--- a/taskfile/node.go
+++ b/taskfile/node.go
@@ -2,6 +2,7 @@ package taskfile
 
 import (
 	"context"
+	"slices"
 	"strings"
 	"time"
 
@@ -89,7 +90,10 @@ func getScheme(uri string) (string, error) {
 		return "", err
 	}
 
-	if strings.HasSuffix(strings.Split(u.Path, "//")[0], ".git") && (u.Scheme == "git" || u.Scheme == "ssh" || u.Scheme == "https" || u.Scheme == "http") {
+	isDotGit := strings.HasSuffix(strings.Split(u.Path, "//")[0], ".git")
+	isUnderscoreGit := strings.Contains(strings.Split(u.Path, "//")[0], "/_git/")
+	schemeIsGitCompatible := slices.Contains([]string{"git", "ssh", "https", "http"}, u.Scheme)
+	if (isDotGit || isUnderscoreGit) && schemeIsGitCompatible {
 		return "git", nil
 	}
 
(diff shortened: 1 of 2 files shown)


vmaerten (approved):

YoungOak:
This is awesome, thanks for the quick turnaround!

PR #2906, opened 2026-07-02
feat: enable remote taskfiles

### LINKED ISSUE #1317: Remote Taskfiles experiment

pd93:
Related to #1317 

This PR enables the remote taskfile functionality globally without the need for an experiment flag/config. If users continue to attempt to enable the remote taskfile experiment, they will be presented with a warning telling them that they no longer need to do this.

Additionally, large changes have been made to the documentation on the website. Namely:

- There is a new dedicated page for Remote Taskfiles (consisting mostly of the same content as the experiment doc)
- The guide has been updated to show a couple of basic things (this then links to the full doc)
- The configuration that was previously in the experiment doc has been moved to the reference docs for the CLI/envinroment variables/config files respectively.


vmaerten (commented):
Nice !
You need to update completions files : https://github.com/go-task/task/blob/b5861b1e27163c1b2d6bd5307562aa488ea34b58/completion/zsh/_task#L131 & https://github.com/go-task/task/blob/b5861b1e27163c1b2d6bd5307562aa488ea34b58/completion/ps/task.ps1#L76 & https://github.com/go-task/task/blob/b5861b1e27163c1b2d6bd5307562aa488ea34b58/completion/fish/task.fish#L110

pd93:
> Nice ! You need to update completions files :

Thanks! Updated. Completion scripts aren't my forte, so please check them :laughing: 

Also fixed the broken links.

vmaerten (approved):
Looks good to me! 🔥

arturict:
Taskfile timeout docs and implementation update helpful. Could you include one sample command that demonstrates fallback/timeout semantics.

andreynering (approved):

Issue #2429, opened 2025-09-20
[Mac] [3.45.5] Task panics if user places 'pipefail' into `shopt` by mistake


posidoni:
### Description

Hello! First of all, I would like to say thank you for your work, I have 'sold' task to my teammates in multiple companies for go projects instead of `make`. This is my first issue for public OSS projet

Context: if I include pipefail in `shopt` - task panics without clear error message.

this code panics for me:

```
set:
  - pipefail
  - nounset
  - errexit
shopt:
  - pipefail
```

and this one 

```
# set: <- may be empty.
shopt:
  - pipefail
```

It is clear here the problem is invalid `pipefail` that is in `shopt`, however it has to be in `set`.

```sh
task: [default] echo 'I will not work'
shopt: invalid option name "abcd"
task: Failed to run task "default": exit status 1
```

This issue happens only with `pipefail` and not with other invalid non existing options.

Issue: task panics with weird  'panic: runtime error: index out of range [-1]'.

Expected: give user more clear and friendly information of what happened. Maybe add helpful message that tells move `pipefail` into `set` and explanation of possible options.

Or give user this message: `shopt: invalid option name "pipefail"`

Something bad happens here: https://github.com/mvdan/sh/blob/b717ad599e17a6975e15a67ed8937bd8d4909146/interp/builtin.go#L793

I suppose, the issue is with `mvdan/sh` library, but anyway I think it should be handled in this project.

Also, could you please explain, why don't you use recovery while running cmd that potentially panics? I have seen panics from sh interpreter in multiple versions, however now I have only this issue.

In recovery handler you could, for example, capture error message from CMD, but if you want stacktraces, you could manually add them to recovery logs.

Full panic stacktraces are below.

```txt
 > task                                                                                                                                                                               
task: [default] nginx -p "/Users/posidoni/powx/cicd/nginx" -c ./nginx.conf -T
panic: runtime error: index out of range [-1]

goroutine 1 gp=0x140000021c0 m=4 mp=0x14000100008 [running]:
panic({0x1030acd60?, 0x1400003a588?})
        /opt/homebrew/opt/go/libexec/src/runtime/panic.go:802 +0x150 fp=0x14000593a50 sp=0x140005939a0 pc=0x1026384b0
runtime.goPanicIndex(0xffffffffffffffff, 0x36)
        /opt/homebrew/opt/go/libexec/src/runtime/panic.go:115 +0x70 fp=0x14000593a90 sp=0x14000593a50 pc=0x102600910
[...]

PR #2911, opened 2026-07-10
fix: validate set and shopt options when reading a Taskfile

posidoni:

Follow-up to #2429. The panic from that issue was fixed upstream in `mvdan/sh`, but an invalid `set:`/`shopt:` option is still only reported when a command actually runs — once per command, with no hint about the actual mistake:

```text
task: [default] echo "hello"
shopt: invalid option name "pipefail"
task: Failed to run task "default": exit status 1
```

This PR validates `set` and `shopt` lists when the Taskfile is read (root, task and command level), following the approach of #2892:

```text
task: invalid shopt option "pipefail" (did you mean to put it in "set"?)
file: Taskfile.yml:1:1
```

- Valid values mirror the interpreter's unexported `posixOptsTable`/`bashOptsTable` (`mvdan.cc/sh/v3/interp`), i.e. exactly what works at runtime.
- `shopt: ["-o", "pipefail"]` keeps working: a `-o` entry switches validation of the remaining entries to `set` options, matching `shopt -s -o` semantics (covered by a regression test).
- When an option is placed in the wrong list, the error suggests the right one (`pipefail` → `set`, `globstar` → `shopt`).
- `schema.json`: the `shopt` enum was missing three supported options (`dotglob`, `extglob`, `nocaseglob`) — extended to all six; reference docs updated accordingly.

Note on behavior: an invalid option now fails at parse time even if the task declaring it never runs — consistent with other Taskfile validation such as #2892.

Tested with four new tests (an invalid option at each of the three levels, plus the `-o` regression) and a full `go test ./...` run.

Disclosure per the [AI usage policy](https://taskfile.dev/contributing/#ai-usage-policy): this PR (code and description) was prepared with AI assistance. I reviewed and tested all changes and can explain every line.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -2758,6 +2758,67 @@ func TestBashShellOptsCommandLevel(t *testing.T) {
 	assert.Equal(t, "globstar\ton\n", buff.String())
 }
 
+func TestShellOptsInvalidShoptGlobalLevel(t *testing.T) {
+	t.Parallel()
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir("testdata/shopts/invalid_global_shopt"),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+	)
+
+	err := e.Setup()
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), `invalid shopt option "pipefail" (did you mean to put it in "set"?)`)
+}
+
+func TestShellOptsInvalidSetTaskLevel(t *testing.T) {
+	t.Parallel()
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir("testdata/shopts/invalid_task_set"),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+	)
+
+	err := e.Setup()
+	require.Error(t, err)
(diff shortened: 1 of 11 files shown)


trulede (commented):
IMHO I think it would be better if errors with options were raised by mvdan (which I think does this already, perhaps?) than to have to keep such a list up to date.

End effect would be (is) the same, an error.

posidoni:
mvdan does raise an error today, but only at runtime, once per command, after Task has already started executing: every command in the task fails with `shopt: invalid option name "pipefail"` and there is no hint that the option belongs in `set:`. What I wanted here is the same thing #2892 did for includes: catch it when the Taskfile is read, and point at the right list, which mvdan cannot do since it does not know the `set:`/`shopt:` split exists.

On keeping the list up to date: fair point, though we already maintain exactly this list in schema.json for editor completion (it had drifted, this PR brings it back in sync). If a second copy in Go feels wrong, I can drop the list entirely and probe the interpreter at parse time instead: run `shopt -s <opt>` / `set -o <opt>` against a throwaway `interp.Runner` and translate the error. Same early failure, nothing to maintain. Happy to go either way.

trulede (changes requested):
I think its reasonable to update the schema.

However, adding programmatic checks would prevent passing new values to mvdan in the future. As it stands now, you get an error from mvdan ... fix the typo ... and move on.

PR #2915, opened 2026-07-13
fix: joinUrl no longer collapses the URL scheme

vsaraikin:
## What
`joinUrl` collapses the `//` in a URL scheme, so it produces an invalid URL:

```yaml
vars:
  SERVER: "http://localhost"
tasks:
  default:
    cmds:
      - echo "{{joinUrl .SERVER "path1" "path2"}}"
      # got:  http:/localhost/path1/path2
      # want: http://localhost/path1/path2
```

The documented behavior (docs/reference/templating.md, "URLs" section) shows `http://localhost/path1/path2`, so this is a clear bug — and it's the whole reason `joinUrl` exists separately from `joinPath`.

## Why
`joinUrl` was implemented as `path.Join(elem...)`. `path.Join` runs `path.Clean`, which collapses consecutive slashes and destroys the `//` after the scheme.

## Fix
Use `net/url.JoinPath`, which is purpose-built for joining URL path segments and preserves the scheme. The signature becomes `(string, error)`, consistent with the sibling `relPath`/`absPath` funcs (`filepath.Rel`/`filepath.Abs`), and the template engine already supports funcs that return an error.

## Tests
Added `TestJoinUrl` in `internal/templater/funcs_test.go` covering scheme preservation, a base with an existing path, and trailing-slash handling. `joinUrl` shipped without a test in #2408, which is how this slipped through.

### DIFF
--- a/internal/templater/funcs.go
+++ b/internal/templater/funcs.go
@@ -3,8 +3,8 @@ package templater
 import (
 	"maps"
 	"math/rand/v2"
+	"net/url"
 	"os"
-	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


trulede (approved):
Seems OK.

vmaerten:
Thanks!
Can you fix the lint ? https://github.com/go-task/task/actions/runs/29262496919/job/86931062202?pr=2915

vsaraikin:
Done — added `t.Parallel()` to `TestJoinUrl` and its subtests to satisfy the `paralleltest` linter. Thanks!

vmaerten (approved):

PR #2916, opened 2026-07-13
fix: escape regex metacharacters in WildcardMatch

vsaraikin:
## What

A task whose name contains a regex metacharacter breaks task matching for the whole Taskfile — either with a hard panic or a silent mis-match.

`(*Task).WildcardMatch` builds a regex from the task name, translating only `*`, and calls `regexp.MustCompile` on it:

```go
regexStr := fmt.Sprintf("^%s$", strings.ReplaceAll(taskName, "*", "(.*)"))
regex := regexp.MustCompile(regexStr)
```

The raw task name is injected into the pattern, so any other metacharacter is interpreted as regex syntax:

- **Panic / crash** — a task named `c++` yields `^c++$`, and `regexp.MustCompile` panics with `invalid nested repetition operator: ++`.
- **False positive** — a task named `a.b` matches the call `axb` (the `.` acts as a wildcard). A realistic footgun: `deploy.prod` gets run by a mistyped `task deploy-prod`.

`FindMatchingTasks` calls `WildcardMatch(call.Task)` on **every** task whenever the requested name isn't a direct/alias match, so a single task with such a name breaks matching for the entire Taskfile.

## Fix

Escape the task name with `regexp.QuoteMeta` before building the pattern, then turn the (now escaped) `\*` back into the wildcard group:

```go
regexStr := fmt.Sprintf("^%s$", strings.ReplaceAll(regexp.QuoteMeta(taskName), `\*`, "(.*)"))
```

`*` remains the only wildcard; everything else is matched literally.

## Testing

Added `TestTaskWildcardMatch` covering the existing `build-*` wildcard behavior plus the metacharacter cases (`c++`, `a.b`, `deploy.prod`). On the current code the test panics (`invalid nested repetition operator`); with the fix it passes. The full `taskfile/ast` package suite passes and the module builds clean.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -3417,6 +3417,26 @@ func TestWildcard(t *testing.T) {
 			call:           "wildcard-foo-bar",
 			expectedOutput: "Hello foo-bar\n",
 		},
+		{
+			name:           "regex metacharacters are matched literally",
+			call:           "c++",
+			expectedOutput: "Building c++\n",
+		},
+		{
+			name:    "regex metacharacters do not match as a pattern",
+			call:    "cxx",
+			wantErr: true,
+		},
+		{
+			name:           "a dot matches itself",
+			call:           "deploy.prod",
+			expectedOutput: "Deploying prod\n",
+		},
+		{
+			name:    "a dot is not a wildcard",
+			call:    "deploy-prod",
+			wantErr: true,
+		},
 	}
 
 	for _, test := range tests {
--- a/taskfile/ast/task.go
+++ b/taskfile/ast/task.go
@@ -87,7 +87,10 @@ func (t *Task) WildcardMatch(name string) (bool, []string) {
 	names := append([]string{t.Task}, t.Aliases...)
 
 	for _, taskName := range names {
-		regexStr := fmt.Sprintf("^%s$", strings.ReplaceAll(taskName, "*", "(.*)"))
+		// Escape regex metacharacters in the task name so a name like "c++" or
+		// "a.b" is matched literally (and doesn't panic in MustCompile); only
+		// "*" is a wildcard.
+		regexStr := fmt.Sprintf("^%s$", strings.ReplaceAll(regexp.QuoteMeta(taskName), `\*`, "(.*)"))
 		regex := regexp.MustCompile(regexStr)
 		wildcards := regex.FindStringSubmatch(name)
 
--- a/taskfile/ast/task.go
+++ b/taskfile/ast/task.go
@@ -87,7 +87,10 @@ func (t *Task) WildcardMatch(name string) (bool, []string) {
 	names := append([]string{t.Task}, t.Aliases...)
 
 	for _, taskName := range names {
-		regexStr := fmt.Sprintf("^%s$", strings.ReplaceAll(taskName, "*", "(.*)"))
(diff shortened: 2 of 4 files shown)


vmaerten:
Did you read our contribution guide? https://taskfile.dev/docs/contributing#ai-usage-policy
Any AI usage should be disclosed.
The PR description and the PR itself both look AI-generated.
Your [comment](https://github.com/go-task/task/pull/2915#issuecomment-4973517640) also looks AI-generated to me.
We want to communicate with humans, not bot.

vsaraikin:
> Did you read our contribution guide? https://taskfile.dev/docs/contributing#ai-usage-policy Any AI usage should be disclosed. The PR description and the PR itself both look AI-generated. Your [comment](https://github.com/go-task/task/pull/2915#issuecomment-4973517640) also looks AI-generated to me. We want to communicate with humans, not bot.

Fair point, sorry about that. Yeah I used an AI tool to clean up the writeup and I should've flagged the description – my bad. But the comments are mine.

trulede (changes requested):
I would request that you probably should write a normal Task test to capture the behaviour from end-2-end. Similar table driven approach, which is good. Then drop the unit test from the PR if you are happy with that.

This code (AI generated, but I suspected as much) shows a faster algorithm which I think should be considered in the PR.

```
func (t *Task) WildcardMatch(name string) (bool, []string) {
	names := append([]string{t.Task}, t.Aliases...)

	for _, taskName := range names {
		// First, quick check without regex if there are no wildcards
		if !strings.Contains(taskName, "*") {
			if taskName == name {
				return true, nil
			}
			continue
		}

		pattern := regexp.QuoteMeta(taskName)
		pattern = strings.ReplaceAll(pattern, `\*`, "(.*)")
		regex := regexp.MustCompile("^" + pattern + "$")
		wildcards := regex.FindStringSubmatch(name)
		if len(wildcards) > 1 {
			return true, wildcards[1:]
		}
	}

	return false, nil
}
```

vsaraikin:
Took the faster version, thanks

task_test.go
```diff
@@ -3417,6 +3417,16 @@ func TestWildcard(t *testing.T) {
 			call:           "wildcard-foo-bar",
 			expectedOutput: "Hello foo-bar\n",
 		},
+		{
+			name:           "regex metacharacters are matched literally",
+			call:           "c++",
+			expectedOutput: "Building c++\n",
+		},
+		{
+			name:    "a dot is not a wildcard",
```
trulede, line 3436: Should this have a "dot" somewhere in the call?

trulede:
Apparently we don't need the fast check:

> This check would only help if:
> 
> A task name has no wildcard (e.g., "build")
> And it's being checked via WildcardMatch
> Which means it already failed the direct match check in line 541
> So this is solving a non-problem — if call.Task == "build" and there's a task named "build", it would have already returned at line 541. The only time WildcardMatch is called on non-wildcard task names is when the call doesn't exactly match the task name (and isn't an alias either).
> 
> The real value of the PR is the regexp.QuoteMeta() fix, which solves the actual bug. The optimization is more of a defensive nicety — it avoids compiling a regex for non-wildcard names, but in practice those shouldn't reach WildcardMatch anyway due to the earlier checks.

 Please be sure to also check you cover the original conditions correctly in the test case. Otherwise it looks OK.

Issue #2917, opened 2026-07-14
Recursive source globs handle hidden files inconsistently


Napolitain:
### Description

After my PR introducing the optimized file-globbing path (#2884), Task now
matches hidden entries when a pattern takes that path and does not match them
when the pattern uses the fallback path.

I understand that not matching names beginning with `.` comes from traditional
Unix shell glob behavior, but it feels surprising that `**/*` does not mean
every file below that directory when it is used for Task sources.

This matters for caching. In my project, I had inputs beginning with `.` that I
expected to be matched. A familiar example would be `.env`: if it affects a
task and changes, I would expect the source fingerprint to change too. Whether
a file participates in the cache key should not depend on which internal
globbing path handles the pattern.

Exclusions can have the same inconsistency. An optimized inclusion can find a
hidden file that a fallback exclusion then fails to remove.

I see two possible directions:

1. Restore the previous behavior by making the optimized path skip hidden
   entries.
2. Go forward with recursive source globs including hidden entries consistently
   in both paths.

My preference is the second option. **"Hidden" is mostly a naming and display
convention; it does not mean that a file cannot affect a task**. However, this
would intentionally change the older behavior, so I would appreciate input
before implementing it.

I have a tests-only draft that exercises both paths, including hidden files,
hidden directories, exclusions, and their effect on source fingerprint caching.

### Version

`nightly` / `main` at 75b227ee. The inconsistency was introduced by 9200d422
(#2884).

### Operating system

Observed on Linux. The two globbing paths themselves are platform-independent.

### Experiments Enabled

None.

### AI assistance

I used Codex to help develop and validate the tests. I reviewed and understand
the changes and this report.

trulede:
@Napolitain your PR broke the existing behaviour? There was another recent PR that did something with .gitignore files, perhaps that is a factor.

trulede:
@andreynering FYI this might cause a regression. But not sure.

Napolitain:
> @Napolitain your PR broke the existing behaviour? There was another recent PR that did something with .gitignore files, perhaps that is a factor.

But note of the behavior, do we want a glob pattern to not ignore dotfiles ? If we don't want to match them, then it is a easy fix. We just need to ensure that is what we want.

trulede:
Task is using a standard glob library which replicates a linux shell, and its behaving in the expected way.

You can try `**/.*` or use shops `dotglob globstar` which would likely result in the desired behaviour.

PR #2918, opened 2026-07-14
test: cover hidden files in optimized and fallback source globs

Napolitain:
Related to #2917.

This draft adds tests for recursive source globbing and fingerprint caching
through both the optimized and fallback paths.

add tests for :
- dotfiles and files below dot-directories;
- exact file-set checks for optimized and fallback patterns;
- optimized inclusions combined with optimized and fallback exclusions;
- nonrecursive, wildcard-root, nested-suffix, brace, and multiple-`**`
  fallback patterns;
- source checksum invalidation through both globbing paths; and
- individually listed generated outputs, so every missing output is checked.

The tests propose that **source globs include hidden entries consistently**. They
currently demonstrate the mismatch on `main`: optimized cases pass, while
fallback inclusions omit hidden inputs and fallback exclusions fail to remove
hidden inputs found by an optimized inclusion.

This PR intentionally changes no production behavior and is not merge-ready.
It is a tests-only draft intended to make the inconsistency and its cache impact
concrete while we decide between preserving the historical shell behavior or
adopting consistent hidden-entry inclusion.

Validation:

- `golangci-lint run` passes with no issues.
- `go test ./internal/fsext -count=1` passes.
- The optimized glob and checksum controls pass.
- `go test ./... -count=1` fails intentionally only in the new fallback
  expectations under `internal/fingerprint`.
- Enabling `DotGlob` temporarily makes the new fingerprint and filesystem tests
  pass; that production change is deliberately not included here.

AI assistance: I used Codex to help develop and validate these tests. I reviewed
and understand the changes.

### DIFF
--- a/internal/fingerprint/glob_test.go
+++ b/internal/fingerprint/glob_test.go
@@ -1,0 +1,215 @@
+package fingerprint
+
+import (
+	"os"
+	"path/filepath"
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/go-task/task/v3/internal/filepathext"
+	"github.com/go-task/task/v3/internal/fsext"
+	"github.com/go-task/task/v3/taskfile/ast"
+)
+
+type routedGlob struct {
+	glob *ast.Glob
+	fast bool
+}
+
+func TestGlobsRecursiveIncludesAllFilesAndHonorsExclusions(t *testing.T) {
+	t.Parallel()
+
+	dir := t.TempDir()
+	files := []string{
+		"input tree/folder-a/.hidden-file-a.aaa",
+		"input tree/folder-a/file-b.bbb",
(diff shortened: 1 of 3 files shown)


Napolitain:
The behavior choice I would like input on is whether `**/*` for Task sources
should mean every file below the directory, including hidden entries.

My preference is to include them. A leading dot does not mean that a file is
irrelevant to a task or its cache; a hidden input can still change generated
output. I understand that the more backward-compatible alternative is to make
the optimized path skip hidden entries and preserve the older shell behavior.

The draft is intentionally red and keeps production code unchanged while that
choice is discussed.

Napolitain:
@andreynering hey, note that the tests fail on purpose. While I am in favor of matching dotfiles via `**/*`, that needs to be agreed upon and I can understand if you prefer not doing so.

trulede:
I do believe this problem is solved with usage of `shopt` or `**/.*`

Issue #524, opened 2021-07-18
Dynamic Variables only calculated once if "sh:" is identical


is-jonreeves:
- Task version: 3.6
- Operating System: Windows 10, Alpine Linux 3.13

It seems like Dynamic Variables that share the same verbatim `sh:` line will not get recomputed when run in their new context. For example if a global `var` is set to the current dir, but then a `task` with `dir: ../` re-sets it, this value does not update for that task. If you make a slight tweak to the `sh:` command, it does work.

### Example Taskfile showing the issue

```yaml
version: '3'
silent: true

vars:
  PROJECT_ROOT:
    sh: echo $PWD

tasks:

  # Use Global Dynamic Var
  example1:
    cmds:
      - echo "Example 1"
      - echo "{{.PROJECT_ROOT}}"

  # Redefine Global Dynamic Var as Local Dynamic Var (with own dir context)
  example2:
    dir: ../
    vars:
      PROJECT_ROOT:
        sh: echo $PWD
    cmds:
      - echo "Example 2"
      - echo "{{.PROJECT_ROOT}}"

  # Create a new Local Dynamic Var (with own dir context)
  example3:
    dir: ../../
    vars:
      REPO_ROOT:
        sh: echo $PWD
    cmds:
      - echo "Example 3"
      - echo "{{.REPO_ROOT}}"
```

## Expected:
```
Example 1
C:\Users\jon\Projects\taskfile-test
Example 2
C:\Users\jon\Projects
Example 3
C:\Users\jon
```

## Actual:
```
Example 1
C:\Users\jon\Projects\taskfile-test
Example 2
C:\Users\jon\Projects\taskfile-test
Example 2
C:\Users\jon\Projects\taskfile-test
```

## Workaround
If you tweak the `sh:` command to be unique each time its used, then it appears to force it to rerun correctly. So something like this will allow it to work:
```
sh: echo $PWD
sh: echo $PWD''
sh: echo $PWD''''  
```

## Side-Note/Feature on Directories
I'll probably open a feature request for this, but the reason that I am jumping through the above hoops is that there doesn't seem to be an official "special variable" to get the Current Working Directory, which would be really handy. For commands that requires a Abs paths like Docker, obtaining these in a cross-platform manner can be annoyingly repetitive to mange. If there was a "special variable" with the current working directory and it was correctly affected by the `dir:` within each `task`, this would help a lot especially with the "slim-sprig" functions.

SwitchedToGitlab:
[...]

PR #2931, opened 2026-07-22
fix: dynamic var uses wrong dir

### LINKED ISSUE #2928: Dynamic variables use the wrong directory when the same Taskfile is included multiple times
## Description

When the same Taskfile is included multiple times with different `dir` values, dynamic variables may be evaluated using the directory configured for another inclusion.

Each included task works correctly when executed separately. However, when both included tasks are executed in the same Task invocation, their dynamic-variable directory context appears to be incorrectly shared.

I reproduced this on the latest `main` commit:

81a49a40

Steps to reproduce

Create the following structure:

example/
├── Taskfile.yml
├── included/
│   └── Taskfile.yml
└── workspace/
    ├── backend/
    └── frontend/

Use the following root Taskfile.yml:

version: "3"

includes:
  backend:
    taskfile: ./included/Taskfile.yml
    dir: ./workspace/backend

  frontend:
    taskfile: ./included/Taskfile.yml
    dir: ./workspace/frontend

tasks:
  default:
    cmds:
      - task: backend:show
      - task: frontend:show

Use the following included Taskfile at included/Taskfile.yml:

version: "3"

tasks:
  show:
    vars:
      CURRENT_DIR:
        sh: basename "$(pwd)"

    cmds:
      - 'echo "{{.CURRENT_DIR}}" > result.txt'

Create the required directories:

mkdir -p workspace/backend workspace/frontend

Run both tasks in the same invocation:

task backend:show frontend:show

Inspect the generated files:

cat workspace/backend/result.txt
cat workspace/frontend/result.txt
Expected behavior

Each dynamic variable should be evaluated using its inclusion's configured directory:

workspace/backend/result.txt  -> backend
workspace/frontend/result.txt -> frontend
Actual behavior

Both files contain the backend directory name:

workspace/backend/result.txt  -> backend
workspace/frontend/result.txt -> backend

The frontend:show command itself executes inside workspace/frontend, but its dynamic variable contains the value evaluated for workspace/backend.

Additional observations

Both tasks work correctly when executed separately:

task backend:show
# Produces: backend

task frontend:show
# Produces: frontend

The problem occurs when both included tasks are executed within the same Task invocation:

task backend:show frontend:show

The following screenshot demonstrates the difference between running the tasks separately and together:

Dynamic variables produce correct values separately but share the wrong directory when tasks run together
Regression test

I also created a regression test using the existing fileContentTest helper.

It currently fails with:

expected: "frontend\n"
actual:   "backend\n"

I would be happy to investigate this further and work on a fix if the behavior is confirmed as unintended.

### Version

3.52.0

### Operating system

ProductName:            macOS ProductVersion:         26.3.1 ProductVersionExtra:    (a) BuildVersion:           25D771280a arm64

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: "3"

includes:
  backend:
    taskfile: ./included/Taskfile.yml
    dir: ./workspace/backend

  frontend:
    taskfile: ./included/Taskfile.yml
    dir: ./workspace/frontend

tasks:
  default:
    cmds:
      - task: backend:show
      - task: frontend:show
```

kjasn:
The shell command value which stored using the command as a cache key may have collisions with the same command 
but different dir. When the command value is related with dir, it will get a wrong command value.

Close #2928

I used AI assistant to inspect the repo structure and help with tests.

### DIFF
--- a/compiler.go
+++ b/compiler.go
@@ -29,8 +29,13 @@ type Compiler struct {
 
 	Logger *logger.Logger
 
-	dynamicCache   map[string]string
+	dynamicCache   map[dynamicCacheKey]string
 	muDynamicCache sync.Mutex
+}
+
+type dynamicCacheKey struct {
+	dir string
+	sh  string
 }
 
 func (c *Compiler) GetTaskfileVariables() (*ast.Vars, error) {
@@ -156,9 +161,12 @@ func (c *Compiler) HandleDynamicVar(v ast.Var, dir string, e []string) (string, 
 	}
 
 	if c.dynamicCache == nil {
-		c.dynamicCache = make(map[string]string, 30)
+		c.dynamicCache = make(map[dynamicCacheKey]string, 30)
 	}
-	if result, ok := c.dynamicCache[*v.Sh]; ok {
+
+	key := dynamicCacheKey{dir, *v.Sh}
+
+	if result, ok := c.dynamicCache[key]; ok {
 		return result, nil
 	}
 
@@ -184,7 +192,7 @@ func (c *Compiler) HandleDynamicVar(v ast.Var, dir string, e []string) (string, 
 	result := strings.TrimSuffix(stdout.String(), "\r\n")
 	result = strings.TrimSuffix(result, "\n")
 
-	c.dynamicCache[*v.Sh] = result
+	c.dynamicCache[key] = result
 	// Never print the resolved value of a secret variable, even in verbose mode
 	logResult := result
 	if v.Secret {
--- a/task_test.go
+++ b/task_test.go
@@ -1052,8 +1052,8 @@ func TestIncludes(t *testing.T) {
 			"./module2/included_directory_with_dir.txt": "included_directory_with_dir",
 			"./module2/included_taskfile_with_dir.txt":  "included_taskfile_with_dir",
 			"os_include.txt":                            "os",
-			"output1/result.txt":                        "output1",
-			"output2/result.txt":                        "output2",
+			"result.txt":                                "includes",
+			"common/result.txt":                         "common",
 		},
 	}
 	t.Run("", func(t *testing.T) {
--- a/task_test.go
+++ b/task_test.go
@@ -1052,6 +1052,8 @@ func TestIncludes(t *testing.T) {
 			"./module2/included_directory_with_dir.txt": "included_directory_with_dir",
 			"./module2/included_taskfile_with_dir.txt":  "included_taskfile_with_dir",
 			"os_include.txt":                            "os",
+			"output1/result.txt":                        "output1",
+			"output2/result.txt":                        "output2",
 		},
 	}
 	t.Run("", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


trulede:
This reminds me of #524 where a solution is presented `sh: 'task --offline env # {{now}}'`. There is a reason its this way, I think that in global vars you can get the same `sh` constantly evaluating (undesirable).

What I mean is, probably, the cause of the problem is not: "var uses the wrong dir".

Looking at this PR makes me wonder if the hash of a Task could be used as par of the index:
```
type dynamicCacheKey struct {
	hash string
	sh  string
}
```
for global vars the hash would be `""`, which is probably OK. Imports would need something too. I don't think `dir` is going to be robust in all situations, however it seems like a good idea regardless.

kjasn:
> This reminds me of #524 where a solution is presented `sh: 'task --offline env # {{now}}'`. There is a reason its this way, I think that in global vars you can get the same `sh` constantly evaluating (undesirable).
> 
> What I mean is, probably, the cause of the problem is not: "var uses the wrong dir".
> 
> Looking at this PR makes me wonder if the hash of a Task could be used as par of the index:
> 
> ```
> type dynamicCacheKey struct {
> 	hash string
> 	sh  string
> }
> ```
> 
> for global vars the hash would be `""`, which is probably OK. Imports would need something too. I don't think `dir` is going to be robust in all situations, however it seems like a good idea regardless.

I found it doesn't work for some situation like this: the same task run with different dirs when its `dir` is templated from `var`

```yml
tasks:
  default:
    cmds:
      - task: build
        vars: { PROJECT: dir1 }
      - task: build
        vars: { PROJECT: dir2 }

  build:
    dir: "{{.PROJECT}}"
    vars:
      HERE:
        sh: pwd
    cmds:
      - echo "{{.HERE}}"
      - echo "{{.PROJECT}}"
```
Expecte `dir1` and `dir2`, but actual both are`dir1` if `dynamicCacheKey` holds `hash`.

ikshantshukla123:
> > This reminds me of #524 where a solution is presented `sh: 'task --offline env # {{now}}'`. There is a reason its this way, I think that in global vars you can get the same `sh` constantly evaluating (undesirable).
> > What I mean is, probably, the cause of the problem is not: "var uses the wrong dir".
> > Looking at this PR makes me wonder if the hash of a Task could be used as par of the index:
> > ```
> > type dynamicCacheKey struct {
> > 	hash string
> > 	sh  string
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for global vars the hash would be `""`, which is probably OK. Imports would need something too. I don't think `dir` is going to be robust in all situations, however it seems like a good idea regardless.
> 
> I found it doesn't work for some situation like this: the same task run with different dirs when its `dir` is templated from `var`
> 
> ```yaml
> tasks:
>   default:
>     cmds:
>       - task: build
>         vars: { PROJECT: dir1 }
>       - task: build
>         vars: { PROJECT: dir2 }
> 
>   build:
>     dir: "{{.PROJECT}}"
>     vars:
>       HERE:
>         sh: pwd
>     cmds:
>       - echo "{{.HERE}}"
>       - echo "{{.PROJECT}}"
> ```
> 
> Expecte `dir1` and `dir2`, but actual both are`dir1` if `dynamicCacheKey` holds `hash`.

thats true

PR #2950, opened 2026-08-03
fix: preserve stdin taskfile bytes

cuishuang:

`StdinNode.Read` used `bufio.Scanner` to read a Taskfile line by line.

`Scanner` has a default maximum token size of 64 KiB, so a Taskfile containing a longer line failed with:

```text
bufio.Scanner: token too long
```

The previous implementation also removed the original line endings and added `\n` after every scanned line. This converted CRLF input to LF and added a trailing newline when the input did not contain one.

This differed from regular Taskfile reads, which use `io.ReadAll ` and preserve the input bytes.

### DIFF
--- a/taskfile/node_stdin_test.go
+++ b/taskfile/node_stdin_test.go
@@ -1,0 +1,65 @@
+package taskfile
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStdinNodeReadPreservesInput(t *testing.T) { //nolint:paralleltest // replaces process-wide stdin
+	node, err := NewStdinNode("")
+	require.NoError(t, err)
+
+	tests := []struct {
+		name  string
+		input []byte
+	}{
+		{
+			name:  "line longer than 64 KiB",
+			input: []byte(strings.Repeat("a", 64*1024+1)),
+		},
+		{
+			name:  "no trailing newline",
+			input: []byte("version: '3'"),
+		},
+		{
+			name:  "CRLF line endings",
+			input: []byte("version: '3'\r\ntasks:\r\n"),
+		},
+		{
+			name:  "empty input",
+			input: []byte{},
+		},
+	}
+
+	for _, tt := range tests { //nolint:paralleltest // subtests replace process-wide stdin
+		t.Run(tt.name, func(t *testing.T) {
+			replaceStdin(t, tt.input)
+
+			got, err := node.Read()
+			require.NoError(t, err)
+			assert.Equal(t, tt.input, got)
+		})
+	}
+}
+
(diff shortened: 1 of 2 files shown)


taskfile/node_stdin_test.go
```diff
@@ -0,0 +1,65 @@
...
+		{
+			name:  "empty input",
+			input: []byte{},
+		},
+	}
+
+	for _, tt := range tests { //nolint:paralleltest // subtests replace process-wide stdin
+		t.Run(tt.name, func(t *testing.T) {
+			stdin := replaceStdin(t, tt.input)
+			defer stdin()
```
vmaerten:
```suggestion
	replaceStdin(t, tt.input)

```

taskfile/node_stdin_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	stdin, err := os.Open(path)
+	require.NoError(t, err)
+
+	original := os.Stdin
+	os.Stdin = stdin
+	return func() {
+		os.Stdin = original
+		require.NoError(t, stdin.Close())
+	}
```
vmaerten:
```suggestion
	t.Cleanup(func() {
		os.Stdin = original
		require.NoError(t, stdin.Close())
    })
```

taskfile/node_stdin_test.go
```diff
@@ -0,0 +1,65 @@
...
+			require.NoError(t, err)
+			assert.Equal(t, tt.input, got)
+		})
+	}
+}
+
+func replaceStdin(t *testing.T, input []byte) func() {
+	t.Helper()
+
+	path := t.TempDir() + "/stdin"
```
vmaerten:
```suggestion
	path := filepath.Join(t.TempDir(), "stdin")
```
As Copilote suggested (he is right)

taskfile/node_stdin_test.go
vmaerten: `t.Cleanup` lets you drop the returned closure: no `defer stdin()` on the caller side (and stdin no longer names a restore function)

vmaerten (commented):
Looks good
Small comments

taskfile/node_stdin_test.go
cuishuang: Updated replaceStdin to register the cleanup with t.Cleanup, so callers no longer need to retain and defer a cleanup closure.

taskfile/node_stdin_test.go
cuishuang: Updated. replaceStdin now restores os.Stdin and closes the temporary file through t.Cleanup.

taskfile/node_stdin_test.go
cuishuang: Updated to use filepath.Join for cross-platform path construction.

cuishuang:
> Looks good Small comments

@vmaerten Modified. Please review it again.

vmaerten (approved):
Thanks!

PR #2955, opened 2026-08-04
docs: close two unterminated code fences and fix --trust flag name

no-hup:
Three small documentation fixes:

- **`reference/schema.md`**: the YAML example under `requires` is never closed, so the following `#### vars` heading and its description get rendered inside the code block (the next ````yaml```` is consumed as the closer).
- **`reference/config.md`**: the example under "Example Configuration" is left open at the end of the file.
- **`experiments/remote-taskfiles.md`**: refers to a `--trust` flag, but the registered flag is `--trusted-hosts` (`internal/flags/flags.go`). The same page already uses `--trusted-hosts` in its configuration section.

Docs only.


vmaerten (approved):
thanks	!

PR #2956, opened 2026-08-04
fix: accept []string and []int in matrix refs

### LINKED ISSUE #2544: Can't use computed lists in matrix loops
### Description

You can only use literal lists in matrix loops, either by reference, or directly. Using a list created from some other source, fails.

See the example taskfile, where `OS_KEYS` is a list produced by the `keys` template command.

Output for `task`:

```
[spew] OS_VAR: ([]interface {}) (len=3 cap=3) {
[spew]  (string) (len=7) "windows",
[spew]  (string) (len=5) "linux",
[spew]  (string) (len=6) "darwin"
[spew] }
[spew] OS_KEYS: ([]string) (len=3 cap=4) {
[spew]  (string) (len=6) "darwin",
[spew]  (string) (len=7) "windows",
[spew]  (string) (len=5) "linux"
[spew] }
[list-ref] windows/amd64
[list-ref] windows/arm64
[list-ref] linux/amd64
[list-ref] linux/arm64
[list-ref] darwin/amd64
[list-ref] darwin/arm64
task: Failed to run task "default": task: Failed to parse /tmp/repro/Taskfile.yml:
matrix reference ".OS_KEYS" must resolve to a list
```

I do note that `spew` shows that the `keys` list a `[]string` array, while `OS_VAR` is a `[]interface {}` array, but I wouldn't expect that to be the issue.

### Version

3.45.5

### Operating system

macos 15.6.1

### Experiments Enabled

_No response_

### Example Taskfile

```YAML
version: '3'
output: prefixed

vars:
  OS_MAP:
    map:
      windows: "Microsoft Windows"
      linux: "Linux kernel"
      darwin: "Apple macos"
  OS_VAR: ["windows", "linux", "darwin"]
  OS_KEYS: { ref: keys .OS_MAP }
  ARCH_VAR: ["amd64", "arm64"]
  
tasks:
  spew:
    silent: true
    cmds:
      - "echo -n 'OS_VAR: {{spew .OS_VAR}}'"
      - "echo -n 'OS_KEYS: {{spew .OS_KEYS}}'"
  list-ref:
    silent: true
    cmds:
      - for:
          matrix:
            OS: { ref: .OS_VAR }
            ARCH: { ref: .ARCH_VAR }
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
  keys-output:
    silent: true
    cmds:
      - for:
          matrix:
            OS: { ref: .OS_KEYS }
            ARCH: { ref: .ARCH_VAR }
        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
  default:
    cmds:
      - task: spew
      - task: list-ref
      - task: keys-output
```

no-hup:
Fixes #2544.

Matrix `ref:` only accepted a literal `[]any`, so a ref computed with a template function like `splitList` came back as `[]string` and failed with "must resolve to a list" even though its a perfectly good list. trulede already pinned this down on the issue, the type switch just never matched.

The fix adds a small helper (`resolvedAsAnySlice`) that accepts `[]any`, `[]string` and `[]int`. that set deliberately mirrors what `itemsFromFor` already takes for literal lists, so nothing that worked before behaves differently, no reflect involved. Enum refs shared the same `[]any` assertion so they get the same widening.

Tests: a table test for the conversion (any/string/int slices plus two error cases) and an end-to-end fixture task `loop-matrix-ref-computed` that loops over a `splitList` ref, with a golden file. `go test ./...` is green locally.

Disclosure: I used an AI assistant to help implement and test this change. I've reviewed and verified it and I'm happy to explain or revise anything here.

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -929,6 +929,7 @@ func TestForCmds(t *testing.T) {
 		{name: "loop-explicit"},
 		{name: "loop-matrix"},
 		{name: "loop-matrix-ref"},
+		{name: "loop-matrix-ref-computed"},
 		{
 			name:    "loop-matrix-ref-error",
 			wantErr: true,
--- a/testdata/for/cmds/Taskfile.yml
+++ b/testdata/for/cmds/Taskfile.yml
@@ -4,6 +4,7 @@ vars:
   OS_VAR: ["windows", "linux", "darwin"]
   ARCH_VAR: ["amd64", "arm64"]
   NOT_A_LIST: "not a list"
+  OS_CSV: "windows,linux,darwin"
 
 tasks:
   # Loop over a list of values
@@ -26,6 +27,16 @@ tasks:
           matrix:
             OS:
               ref: .OS_VAR
+            ARCH:
+              ref: .ARCH_VAR
+        cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
+
+  loop-matrix-ref-computed:
+    cmds:
+      - for:
+          matrix:
+            OS:
+              ref: 'splitList "," .OS_CSV'
             ARCH:
               ref: .ARCH_VAR
         cmd: echo "{{.ITEM.OS}}/{{.ITEM.ARCH}}"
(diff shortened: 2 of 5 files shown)


vmaerten (approved):
LGTM
Thanks!

PR #2957, opened 2026-08-04
docs: document use_gitignore in schema reference and guide

no-hup:
The `use_gitignore` setting shipped in v3.52.0 (#2773) and is in the JSON schema and changelog, but the schema reference and guide never picked it up, so the only way to find it right now is reading the schema file itself.

This adds the root-level and per-task entries to the schema reference (same format as the neighbouring settings, descriptions match the schema.json wording and the actual merge behaviour in `taskfile/ast/taskfile.go`) and a short tip in the guide's fingerprinting section, since excluding gitignored build artifacts from `sources`/`generates` globs is the main use case.

Docs only, no code changes.

Disclosure: I used an AI assistant to help draft this change. I've reviewed and verified it against the implementation and I'm happy to revise anything.

### DIFF
--- a/website/src/docs/guide.md
+++ b/website/src/docs/guide.md
@@ -898,6 +898,16 @@ of the running the task is considered as a generate.
 
 :::
 
+::: tip
+
+If your globs match files that are ignored by Git (build artifacts, caches,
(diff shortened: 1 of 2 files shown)


vmaerten (approved):
Thanks!

PR #2959, opened 2026-08-09
feat: support excluding task namespaces

### LINKED ISSUE #2300: Support for excluding entire namespaces in addition to individual tasks

xmxxc:
## Summary

- Allow `includes.excludes` entries ending in `:*` to exclude entire task namespaces.
- Preserve exact task-name matching for entries without `:*`.
- Scope namespace exclusions to descendants: `foo:*` excludes `foo:child`, but not the standalone `foo` task or tasks in similarly prefixed namespaces such as `foo-other`.
- Add coverage for exact task matching and namespace exclusions with regular and flattened includes.
- Update the guide, schema reference, and public JSON Schema description.

## Testing

- `go test ./... -run '^TestIncludesWithExclude$' -count=1`
- `go test ./taskfile/... -count=1`
- `go vet ./...`

## AI Assistance

AI assistance was used to review the changes. During the review, it identified that the `excludes` description in the public JSON Schema had not been updated to mention namespaces. I reviewed and fixed the inconsistency before submitting this pull request.

Fixes #2300

### DIFF
--- a/taskfile/ast/tasks.go
+++ b/taskfile/ast/tasks.go
@@ -130,9 +130,9 @@ func (t1 *Tasks) Merge(t2 *Tasks, include *Include, includedTaskfileVars *Vars) 
 		task.Internal = task.Internal || (include != nil && include.Internal)
 		taskName := name
 
-		// If the task or one of its parent namespaces is in the exclude list,
-		// don't add it to the merged taskfile.
-		if isTaskExcluded(name, include.Excludes) {
+		// If the task or its namespace is in the exclude list, don't add it to
+		// the merged taskfile.
+		if include.isTaskExcluded(name) {
 			continue
 		}
 
@@ -249,17 +249,4 @@ func taskNameWithNamespace(taskName string, namespace string) string {
 		return after
 	}
 	return fmt.Sprintf("%s%s%s", namespace, NamespaceSeparator, taskName)
-}
-
-func isTaskExcluded(name string, excludes []string) bool {
-	for _, exclude := range excludes {
-		exclude = strings.TrimSuffix(exclude, NamespaceSeparator+"*")
-		if exclude == "" {
-			continue
-		}
-		if name == exclude || strings.HasPrefix(name, exclude+NamespaceSeparator) {
-			return true
-		}
-	}
-	return false
 }
--- a/taskfile/ast/tasks.go
+++ b/taskfile/ast/tasks.go
@@ -130,8 +130,9 @@ func (t1 *Tasks) Merge(t2 *Tasks, include *Include, includedTaskfileVars *Vars) 
 		task.Internal = task.Internal || (include != nil && include.Internal)
 		taskName := name
 
-		// if the task is in the exclude list, don't add it to the merged taskfile
-		if slices.Contains(include.Excludes, name) {
+		// If the task or one of its parent namespaces is in the exclude list,
+		// don't add it to the merged taskfile.
+		if isTaskExcluded(name, include.Excludes) {
 			continue
 		}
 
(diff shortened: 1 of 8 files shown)


taskfile/ast/tasks.go
```diff
@@ -249,3 +250,16 @@ func taskNameWithNamespace(taskName string, namespace string) string {
...
+		exclude = strings.TrimSuffix(exclude, NamespaceSeparator+"*")
+		if exclude == "" {
+			continue
+		}
+		if name == exclude || strings.HasPrefix(name, exclude+NamespaceSeparator) {
+			return true
+		}
+	}
+	return false
+}
```
vmaerten: Can you put this logic in the `taskfile/ast/include.go`
Something like : 
```
func (include *Include) isTaskExcluded(name string) bool {
      if include == nil {
              return false
      }
      for _, exclude := range include.Excludes {
              exclude = strings.TrimSuffix(exclude, NamespaceSeparator+"*")
              if exclude == "" {
                      continue
              }
              if name == exclude || strings.HasPrefix(name, exclude+NamespaceSeparator) {
                      return true
              }
      }
      return false
}
```

vmaerten (commented):
Thanks for working on this

I know it's not like this in the issue but can you make `excludes` entries match task names exactly, and require a trailing `:*` for a namespace? So `excludes: [debug]` only excludes a task named `debug`, while `excludes: ['debug:*']` excludes everything under it.

Two reasons:
- Everywhere else in Task a task name means exactly that task. A bare name here would silently change behaviour for anyone who has both a `deploy` task and a `deploy:` namespace, which would be misleading for our users.
- It also keeps the door open for the glob/regex support if we want to add it later

The rest looks good
Happy to push these changes myself if you prefer.

xmxxc:
Thanks for the feedback and for offering to help! I've applied the requested changes in `15648b46`:

- Entries without `:*` now match task names exactly.
- Namespace exclusions require a trailing `:*` and only match tasks within that namespace, so `foo:*` does not exclude the standalone `foo` task or tasks under similarly prefixed namespaces.
- The matching logic has been moved to `(*Include).isTaskExcluded` in `taskfile/ast/include.go`.
- Tests and documentation have been updated accordingly.

The following checks all pass locally:

- `go test ./... -run '^TestIncludesWithExclude$' -count=1`
- `go test ./taskfile/... -count=1`
- `go vet ./...`

Please take another look when you have a chance.

taskfile/ast/tasks.go
xmxxc: Done in `15648b46` - I moved the exclusion logic to `taskfile/ast/include.go` as `(*Include).isTaskExcluded` and updated `Tasks.Merge` to call it.

vmaerten (approved):
Thanks !

PR #2969, opened 2026-08-11
fix: resolve enum refs that depend on dynamic variables in interactive prompts

### LINKED ISSUE #2958: --interactive shows a free-form text prompt (not a selection menu) when enum.ref points to a dynamic sh: variable
### Description

The guide documents that enum refs can reference dynamically-computed variables:

https://taskfile.dev/docs/guide#using-variable-references-for-enum-values

```yaml
version: '3'

vars:
  AVAILABLE_SERVICES:
    sh: printf 'api\nweb\ndb\n'

tasks:
  deploy:
    requires:
      vars:
        - name: SERVICE
          enum:
            ref: .AVAILABLE_SERVICES | splitLines | compact
    cmds:
      - echo "Deploying {{.SERVICE}}"
```

And the interactive-mode docs say "For variables with an `enum`, a selection menu is shown":

https://taskfile.dev/docs/guide#prompting-for-missing-variables-interactively

But when running `task --interactive deploy`, a **free-form text prompt** is shown instead of the api/web/db selection menu. Pressing enter without typing anything then fails with a confusing error:

```text
task: Task "deploy-dynamic" cancelled because it is missing required variables:
  - SERVICE has an invalid value : '' (allowed values : [api web db])
```

This is a remaining gap of the fix from #2927 (which closed #2817): that fix resolves `enum.ref` against the **fast-compiled** task vars, where `sh:` variables are not evaluated and resolve to `""`. The ref then produces an empty list, so `input.Prompter.Prompt` falls back to `Text`.

The values **are** resolvable though: running non-interactively works and validation lists the allowed values (as shown in the error above), so the menu could be built from them.

### Steps to reproduce

```yaml
# Taskfile.yml
version: '3'

vars:
  AVAILABLE_SERVICES:
    sh: printf 'api\nweb\ndb\n'

tasks:
  deploy:
    requires:
      vars:
        - name: SERVICE
          enum:
            ref: .AVAILABLE_SERVICES | splitLines | compact
    cmds:
      - echo "Deploying {{.SERVICE}}"
```

```bash
task --interactive deploy
# Expected: selection menu with api/web/db
# Actual:   free-form text input
```

A static ref (`ref: .ALLOWED_ENVS` where `ALLOWED_ENVS: [dev, staging, prod]`) does show the selection menu correctly.

### Impact

- Contradicts the docs and the changelog entry for #2927 ("enum.ref now show the selection list like static enums, instead of falling back to free-form input").
- Affects deps too, since `promptDepsVars` resolves their enum refs the same way.
- The guide's own examples for dynamic enum refs are exactly the affected case.

### Proposed fix

In `resolveEnumRefForPrompt` (`requires.go`), evaluate the referenced dynamic (`sh:`) variables before resolving the ref so the enum list is populated and a `Select` menu is shown.

ikshantshukla123:
Fixes #2958

## Summary

`task --interactive` showed a free-form text prompt (instead of a selection menu) when an `enum.ref` pointed to a dynamic `sh:` variable. The previous enum-ref support (#2927) resolved refs against the **fast-compiled** task vars, where `sh:` variables are not evaluated and resolve to `""`. The ref then produced an empty list and the prompter fell back to `Text` input.

This PR evaluates the relevant dynamic (`sh:`) variables first, so the enum ref resolves to the real values and a `Select` menu is shown.

## Changes

- `resolveEnumRefForPrompt` now evaluates dynamic (`sh:`) variables **before** resolving the enum ref, then resolves the ref a single time. It reuses the compiler's dynamic-var cache and mirrors `GetVariables` evaluation order, and a failing command is left unevaluated so the remaining variables can still resolve.
- Fixed a bug in the initial implementation: `FastGetVariables` stores un-evaluated dynamic vars as `{Value: "", Sh: ...}` (a non-nil empty value), so treating any non-nil value as static meant the `sh:` command never ran.
- All checks pass: `go build ./...`, `go vet ./...`, `golangci-lint run ./...`, and the full root-package test suite.

## Tests

- Added fixture-based integration test `TestPromptEnumRefResolution` driving the real prompt path (`FastCompiledTask` → `resolveEnumRefForPrompt`) against `testdata/enum_ref_prompt/`, covering a dynamic `sh:` ref, a static list ref, and an empty dynamic var.
- Reworked the `resolveEnumRefForPrompt` unit tests to assert the actual resolved enum values for realistic scenarios (fast-compiled var shape, empty dynamic output, failing unrelated `sh:` var).

Note: a golden end-to-end test like `testdata/prompt` isn't possible for this path because the `requires:` prompt uses a bubbletea-based `input.Prompter`, which can't be driven by the plain-stdin test harness (only the `prompt:` confirm field can).

## AI usage disclosure

I used DeepSeek (via the opencode CLI) to understand the fast-compiled var shape and the templater resolution flow, and to help write the regression tests included in this PR. The duplicate-logic review feedback was applied manually; the streamlined flow is written by hand.

### DIFF
--- a/requires.go
+++ b/requires.go
@@ -6,6 +6,7 @@ import (
 	"github.com/elliotchance/orderedmap/v3"
 
 	"github.com/go-task/task/v3/errors"
+	"github.com/go-task/task/v3/internal/env"
 	"github.com/go-task/task/v3/internal/input"
 	"github.com/go-task/task/v3/internal/templater"
 	"github.com/go-task/task/v3/internal/term"
@@ -46,7 +47,7 @@ func (e *Executor) promptDepsVars(calls []*Call) error {
 
 		for _, v := range getMissingRequiredVars(compiledTask) {
 			if !varsMap.Has(v.Name) {
-				varsMap.Set(v.Name, resolveEnumRefForPrompt(v, compiledTask.Vars))
+				varsMap.Set(v.Name, e.resolveEnumRefForPrompt(v, compiledTask.Vars, compiledTask.Dir))
 			}
 		}
 
@@ -220,13 +221,68 @@ func getEnumValues(e *ast.Enum) []string {
 
 // resolveEnumRefForPrompt returns a copy of v with its enum ref resolved into
 // concrete values, so the interactive prompter can show a Select. Refs that
-// depend on dynamic vars may not resolve here and fall back to free-form input.
-func resolveEnumRefForPrompt(v *ast.VarsWithValidation, vars *ast.Vars) *ast.VarsWithValidation {
+// point to dynamic (sh:) variables are resolved by evaluating those variables
+// first. Refs that still can't be resolved fall back to free-form input.
+func (e *Executor) resolveEnumRefForPrompt(v *ast.VarsWithValidation, vars *ast.Vars, dir string) *ast.VarsWithValidation {
 	if v.Enum == nil || v.Enum.Ref == "" || len(v.Enum.Value) > 0 {
 		return v
 	}
 	vCopy := v.DeepCopy()
 	cache := &templater.Cache{Vars: vars}
-	_ = resolveEnumRefs(&ast.Requires{Vars: []*ast.VarsWithValidation{vCopy}}, cache)
+	if err := resolveEnumRefs(&ast.Requires{Vars: []*ast.VarsWithValidation{vCopy}}, cache); err == nil && len(vCopy.Enum.Value) > 0 {
+		return vCopy
+	}
+	// The ref may depend on dynamic (sh:) variables that are not evaluated in
+	// the fast-compiled vars (they resolve to an empty value there). Evaluate
+	// the dynamic variables and retry, so the prompt can show a selection list
+	// instead of falling back to free-form input.
+	if !hasDynamicVars(vars) {
+		return vCopy
+	}
+	if evaluated := e.evaluateDynamicVarsForPrompt(vars, dir); evaluated != nil {
+		vCopy = v.DeepCopy()
+		_ = resolveEnumRefs(&ast.Requires{Vars: []*ast.VarsWithValidation{vCopy}}, &templater.Cache{Vars: evaluated})
+	}
 	return vCopy
+}
+
+// hasDynamicVars reports whether any variable in vars is backed by a shell
+// command (sh:) that has not been evaluated yet.
+func hasDynamicVars(vars *ast.Vars) bool {
+	if vars == nil {
+		return false
+	}
+	for v := range vars.Values() {
+		if v.Sh != nil && *v.Sh != "" {
+			return true
+		}
+	}
+	return false
+}
+
+// evaluateDynamicVarsForPrompt evaluates the dynamic (sh:) variables in vars,
+// returning a new Vars map with their resolved values. Individual commands that
+// fail are left unevaluated so the remaining variables can still be resolved.
+func (e *Executor) evaluateDynamicVarsForPrompt(vars *ast.Vars, dir string) *ast.Vars {
+	if e == nil || e.Compiler == nil || vars == nil {
+		return nil
+	}
+	result := ast.NewVars()
+	for k, v := range vars.All() {
+		cache := &templater.Cache{Vars: result}
+		newVar := templater.ReplaceVar(v, cache)
+		if newVar.Value != nil || newVar.Sh == nil {
+			result.Set(k, ast.Var{Value: newVar.Value, Secret: v.Secret})
+			continue
+		}
+		static, err := e.Compiler.HandleDynamicVar(newVar, dir, env.GetFromVars(result))
+		if err != nil {
+			// Leave the variable unevaluated if the command fails, so the rest
+			// of the variables can still be evaluated.
+			result.Set(k, ast.Var{Value: "", Secret: v.Secret})
+			continue
+		}
+		result.Set(k, ast.Var{Value: static, Secret: v.Secret})
+	}
+	return result
 }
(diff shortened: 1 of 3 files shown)


ikshantshukla123:
@trulede have a look.

requires.go
```diff
@@ -220,13 +221,68 @@ func getEnumValues(e *ast.Enum) []string {
...
+// point to dynamic (sh:) variables are resolved by evaluating those variables
+// first. Refs that still can't be resolved fall back to free-form input.
+func (e *Executor) resolveEnumRefForPrompt(v *ast.VarsWithValidation, vars *ast.Vars, dir string) *ast.VarsWithValidation {
 	if v.Enum == nil || v.Enum.Ref == "" || len(v.Enum.Value) > 0 {
 		return v
 	}
 	vCopy := v.DeepCopy()
 	cache := &templater.Cache{Vars: vars}
-	_ = resolveEnumRefs(&ast.Requires{Vars: []*ast.VarsWithValidation{vCopy}}, cache)
+	if err := resolveEnumRefs(&ast.Requires{Vars: []*ast.VarsWithValidation{vCopy}}, cache); err == nil && len(vCopy.Enum.Value) > 0 {
```
trulede: There is a lot of duplicated logic. I would try to streamline it. Evaluate vars, then resolve enum refs, done.

trulede: And see about writing tests similar to this if possible:
https://github.com/go-task/task/tree/main/testdata/prompt

Because AI generated tests are next-to-useless.

ikshantshukla123: Done. `resolveEnumRefForPrompt` is now: evaluate dynamic vars first (when any exist), then resolve the enum ref once — no more two-pass retry or duplicated `resolveEnumRefs`/`DeepCopy`.

For tests, since the interactive prompt is bubbletea-based (can't be driven by a plain stdin buffer in the harness, same reason the existing `testdata/prompt` tests use the `prompt:` field instead of `requires:`), I added a fixture-based integration test at `testdata/enum_ref_prompt` that drives the real prompt path (`FastCompiledTask` → `resolveEnumRefForPrompt`) and asserts the resolved enum values, plus targeted unit tests for the edge cases (empty dynamic output, failing unrelated `sh:` var).

ikshantshukla123: > And see about writing tests similar to this if possible: https://github.com/go-task/task/tree/main/testdata/prompt
> 
> Because AI generated tests are next-to-useless.

Added fixture-based tests at `testdata/enum_ref_prompt/` that run the real prompt path against a Taskfile (dynamic `sh:` var, static list var, and an empty dynamic var), added in `TestPromptEnumRefResolution`. The existing unit tests in `requires_internal_test.go` were reworked to assert the actual resolved enum values for realistic scenarios.

Note: a fully end-to-end golden test like `testdata/prompt` isn't possible here because the `requires:` prompt uses a bubbletea-based `input.Prompter`, which can't be driven by the plain-stdin test harness (only the `prompt:` confirm field can).

ikshantshukla123: @trulede Ready for rereview whenever you get a chance

requires.go
```diff
@@ -220,13 +221,63 @@ func getEnumValues(e *ast.Enum) []string {
...
 	return vCopy
 }
+
+// hasDynamicVars reports whether any variable in vars is backed by a shell
+// command (sh:) that has not been evaluated yet.
+func hasDynamicVars(vars *ast.Vars) bool {
+	if vars == nil {
+		return false
+	}
+	for v := range vars.Values() {
```
trulede, line 250: Should this be `for _, v ...`

trulede: and if wrong, perhaps why the unit test does not fail.

ikshantshukla123: > and if wrong, perhaps why the unit test does not fail.

Good question  vars.Values() returns iter.Seq[Var] (a single-value iterator, not (k, v) pairs), so for v := range vars.Values() yields each Var directly. for _, v := range wouldn't compile here since iter.Seq[Var] ranges over exactly one variable.
On the test question  it's guarded by TestPromptEnumRefResolution — if hasDynamicVars returned false for the dynamic sh: fixture, the enum would come back empty and that test would fail. (And hasDynamicVars is really just a cheap guard; evaluateDynamicVarsForPrompt handles the no-op case fine either way.)

vmaerten:
@trulede thanks for being so patient.

@ikshantshukla123 please refer to [our policy on AI](https://taskfile.dev/docs/contributing#ai-usage-policy).

It's clear that an LLM is posting with your account. It looks like you're pasting trulede's answers to an LLM and then pasting the response back into GitHub (with the associated commit, also generated)

Please don't do that.

ikshantshukla123:
> @trulede thanks for being so patient.
> 
> 
> 
> @ikshantshukla123 please refer to [our policy on AI](https://taskfile.dev/docs/contributing#ai-usage-policy).
> 
> 
> 
> It's clear that an LLM is posting with your account. It looks like you're pasting trulede's answers to an LLM and then pasting the response back into GitHub (with the associated commit, also generated)
> 
> 
> 
> Please don't do that.

Sorry sir but it was not blindly copy pasted , my English is weak so with my pointers I used ai to generate convo. Can you suggest another method?

ikshantshukla123:
> > @trulede thanks for being so patient.
> > 
> > 
> > 
> > @ikshantshukla123 please refer to [our policy on AI](https://taskfile.dev/docs/contributing#ai-usage-policy).
> > 
> > 
> > 
> > It's clear that an LLM is posting with your account. It looks like you're pasting trulede's answers to an LLM and then pasting the response back into GitHub (with the associated commit, also generated)
> > 
> > 
> > 
> > Please don't do that.
> 
> Sorry sir but it was not blindly copy pasted , my English is weak so with my pointers I used ai to generate convo. Can you suggest another method?

@trulede sir can you give some tips?

trulede:
Yes. Don't paste crap from AI, its always junk. Unit tests written by AI are junk too .... and the project has to maintain that ... which is actually quite some effort.

You simply write that: `var.Values() is a slice, the loop works.` Write it in you native language and use AI to translate. Simple things translate well.


Use AI to help you, but write the code yourself. And the PR too.

PR #2971, opened 2026-08-11
fix: return full path in getTaskPath() instead of bare filename

### LINKED ISSUE #2970: getTaskPath() in signals_test.go returns bare filename instead of full path
## Bug Description

In `signals_test.go`, the `getTaskPath()` helper function has a subtle but impactful bug: it calls `os.Stat("./bin/task")` and then returns `info.Name()`. The `FileInfo.Name()` method returns **only the base filename** (i.e. `"task"`), not the path that was passed to `os.Stat`.

### Affected Code

```go
// signals_test.go, line 190-193
func getTaskPath() (string, error) {
    if info, err := os.Stat("./bin/task"); err == nil {
        return info.Name(), nil  // ❌ returns "task", not "./bin/task"
    }
    ...
}
```

### Root Cause

From the Go standard library documentation for [`fs.FileInfo.Name()`](https://pkg.go.dev/io/fs#FileInfo):

> Name returns the name of the file **corresponding to this entry**, for a regular file, this is the **base name** of the file.

So `info.Name()` returns `"task"` — **not** `"./bin/task"`.

### Impact

When the signal tests call `getTaskPath()` and `./bin/task` exists, the function returns `"task"` (bare filename). This path is then passed to `exec.Command()`. On systems where `task` is also installed globally (in `$PATH`), the tests silently run against the **system-installed binary** instead of the locally-built one from `./bin/task`. This can mask failures in locally-built binaries and cause signal tests to pass even when the local build is broken.

### Expected Behavior

`getTaskPath()` should return `"./bin/task"` when the binary exists at that path.

### Fix

```go
func getTaskPath() (string, error) {
    const localTaskBin = "./bin/task"
    if _, err := os.Stat(localTaskBin); err == nil {
        return localTaskBin, nil  // ✅ returns "./bin/task"
    }
    ...
}
```

### Steps to Reproduce

1. Build task locally: `go build -o ./bin/task ./cmd/task`
2. Run signal tests: `go test -tags signals -run TestSignalSentToProcessGroup ./...`
3. Observe that `getTaskPath()` returns `"task"` instead of `"./bin/task"`

### Environment

- Go version: `go1.24`
- OS: Linux

Nithwin:
## Summary

Fixes #2970

## Problem

In `signals_test.go`, `getTaskPath()` uses `os.Stat("./bin/task")` and returns `info.Name()`. The `FileInfo.Name()` method only returns the **base filename** (`"task"`), not the path passed to `os.Stat` (`"./bin/task"`).

This means `exec.Command` receives `"task"` (bare name) and resolves it via `$PATH`, silently running signal tests against a system-installed binary instead of the locally-built one.

## Fix

Store the path in a constant and return it directly:

```diff
 func getTaskPath() (string, error) {
-    if info, err := os.Stat("./bin/task"); err == nil {
-        return info.Name(), nil
+    const localTaskBin = "./bin/task"
+    if _, err := os.Stat(localTaskBin); err == nil {
+        return localTaskBin, nil
     }
 ```

## Testing

- The fix is in the test file itself (`signals_test.go`)
- Verified that `os.FileInfo.Name()` only returns the base filename per the Go standard library docs

### DIFF
--- a/signals_test.go
+++ b/signals_test.go
@@ -188,8 +188,8 @@ func TestSignalSentToProcessGroup(t *testing.T) {
 }
 
 func getTaskPath() (string, error) {
-	if info, err := os.Stat("./bin/task"); err == nil {
-		return info.Name(), nil
+	if _, err := os.Stat("./bin/task"); err == nil {
+		return filepath.Abs("./bin/task")
 	}
 
 	if path, err := exec.LookPath("task"); err == nil {

### CODE REVIEW

**Comment by Nithwin** [comment]
@trulede Could you give me your review please.

File: signals_test.go [hunk 1]
```diff
@@ -188,8 +188,9 @@ func TestSignalSentToProcessGroup(t *testing.T) {
 }
 
 func getTaskPath() (string, error) {
-	if info, err := os.Stat("./bin/task"); err == nil {
-		return info.Name(), nil
+	const localTaskBin = "./bin/task"
```
trulede: I suggest this, with associated changes:

```
var sleepit, _ = filepath.Abs("./bin/sleepit")
var localTaskBinary, _ = filepath.Abs("./bin/task")

...

func TestSignalSentToProcessGroup(t *testing.T) {
	task := localTaskBinary
	if task == "" {
		// Fallback to system task if local build doesn't exist
		var err error
		task, err = exec.LookPath("task")
		if err != nil {
			t.Fatal(err)
		}
	}
	// ... rest of test
}
```

trulede (commented):
I make a suggestion, but otherwise fine.

signals_test.go
Nithwin: Thanks for the suggestion! 👍 

I've implemented your approach and it's much better.

Nithwin:
FYI - Using "Squash and merge" when merging this PR would ensure the contribution shows up in the contributors list. Thanks!

signals_test.go
```diff
@@ -21,11 +20,17 @@ import (
 )
 
 var SLEEPIT, _ = filepath.Abs("./bin/sleepit")
+var localTaskBinary, _ = filepath.Abs("./bin/task")
```
trulede: Fix sleepit too, style matters.

signals_test.go
```diff
@@ -21,11 +20,17 @@ import (
...
 
 func TestSignalSentToProcessGroup(t *testing.T) {
-	task, err := getTaskPath()
-	if err != nil {
-		t.Fatal(err)
+	task := localTaskBinary
```
trulede: Now you don't check if the path exists? So the following code is not hit. Right?

This style is better when possible: 
if task, err = exec.LookPath("task"); err != nil {
   t.Fatal(err)
}

Nithwin:
@trulede  Thanks for catching that! You were totally right the path wasn't being checked for existence, so the fallback was never hit.

I just pushed an update:
1. I used `exec.LookPath` for both `task` and `sleepit` so the style is consistent. If the local binaries don't exist, it now properly falls back to the system ones.
2. I also fixed a bug I found while testing: `for range tc.sendSigs - 1` was causing the loop to run 0 times, so the tests weren't even sending the signals! I changed it to `for range tc.sendSigs` and the tests pass properly now.

Thanks again for the help!
Looking forward to learn more from you

trulede (changes requested):
I would suggest starting again. This is going nowhere.

Just fix the original problem.

Nithwin:
@trulede I've reset the branch and pushed a single minimal commit that directly fixes #2970 by returning `filepath.Abs("./bin/task")` in `getTaskPath()` instead of `info.Name()`. 

Thanks for the feedback!

trulede:
What agent do you use to write this?

I suggest to put the original code into google.ai and look at what it suggests, and then really think about which of those suggestions is the best. And more importantly, if they are better than what you currently have.

pd93:
I'm closing this PR as it doesn't follow our contribution guidelines. Please refer to our [AI usage policy](https://taskfile.dev/docs/contributing#ai-usage-policy).

Nithwin:
Thank you so much for your feedback! It helped me realize that I have been overly dependent on AI. I will ensure to solve future problems myself.

PR #2974, opened 2026-08-13
feat: add blog post for remote taskfiles

pd93:
Blog post for the release of the remote taskfiles feature.

### DIFF
--- a/website/src/next/blog/remote-taskfiles.md
+++ b/website/src/next/blog/remote-taskfiles.md
@@ -1,0 +1,139 @@
+---
+title: Remote Taskfiles
+description:
+  Remote Taskfiles have been made generally available after nearly 3 years of
+  experimentation.
+author: pd93
+date: 2026-08-18
+tags: ['experiments', 'remote', 'new-features']
+outline: deep
+editLink: false
+---
+
+# Remote Taskfiles
+
+<AuthorCard :author="$frontmatter.author" />
+
+It's finally time! After nearly 3 years as an experiment, we're excited to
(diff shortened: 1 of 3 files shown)


website/src/next/blog/remote-taskfiles.md
```diff
@@ -0,0 +1,133 @@
...
+
+includes:
+  my-remote-namespace: https://raw.githubusercontent.com/go-task/task/main/website/src/public/Taskfile.yml
+```
+
+```yaml [Git over HTTP]
+version: '3'
+
+includes:
+  my-remote-namespace: https://github.com/go-task/task.git//website/src/public/Taskfile.yml?ref=main
```
vmaerten, line 117: I don't understand how it works. Where is stored the Taskfile ?
Maybe something is missing in the `public/_headers` ?

vmaerten (approved):

website/src/next/blog/remote-taskfiles.md
pd93: The example Taskfile is in our version control: https://github.com/go-task/task/blob/main/website/src/public/Taskfile.yml. I've just double checked this example and it works fine for me.

That file is also available via https://taskfile.dev/Taskfile.yml. The new `_headers` file is so that when you click that link, Netlify would give the correct content type so you can view it in the browser rather than it trying to download it.

vmaerten: Oops I think I have a bad configuration in my shell

PR #2976, opened 2026-08-13
feat(remote): add remote.headers to use HTTP Header

vmaerten:


executor.go
```diff
@@ -36,6 +37,7 @@ type (
 		Download            bool
 		Offline             bool
 		TrustedHosts        []string
+		RemoteAuth          taskfile.HostHeaders
```
trulede: I don't think Hostheaders is the correct term. It should probably just be Headers

taskfile/http_auth.go
trulede, line 1: @vmaerten Hi, just looking at this and wondering if a more generic "headers" approach would be viable. Similar to `curl` with its -H option. Its the same code, just without "auth" (also drop from the schema).

Rational is that headers can be set for a number of reasons, from which authorisation is only a subset.

Just for example:
```
for _, headers := range config.Remote.headers {
	byHost[auth.Host] = headers
}
```

Also, it might be useful, or necessary, to have different headers for requests against the same host. If I understand correctly, you are consolidating (last wins). 

But OK, I see that you put this in the taskrc file, and not the `includes`, so there is no solution for that.

PR #3005, opened 2026-08-29
fix: honour the global run setting from an included Taskfile

### LINKED ISSUE #1738: Honour the global-level 'run: once' setting in an included Taskfile
Given two Taskfiles in a project: one at the root and the other in a subdirectory.
If the subdirectory Taskfile has `run: once` set at the global/top level (rather than set explicitly per-task) and the root Taskfile includes the subdirectory Taskfile, the `task` calls from the root directory can potentially run the subdirectory Taskfile tasks more than once.

`/Taskfile.yaml`:
```yaml
version: '3'

includes:
  subdir: ./subdir/Taskfile.yaml

tasks:
  reproduce:
    deps:
      - task: subdir:once-only
      - task: subdir:once-only
```

`/subdir/Taskfile.yaml`:
```yaml
version: '3'

run: once

tasks:
  once-only:
    cmd: echo "Once"

  several:
    deps:
      - task: once-only
      - task: once-only
      - task: once-only
```

From inside the subdirectory, `run: once` works fine:

```shell
$ cd subdir
$ task several
task: [once-only] echo "Once"
Once
```

From the root directory, the subdirectory Taskfile's `run: once` setting has no effect:
```shell
$ task reproduce
task: [subdir:once-only] echo "Once"
task: [subdir:once-only] echo "Once"
Once
Once
```

---

These other issues/PRs seem similar/relevant, but distinct from this particular situation:
- #852
- #1563
- #1655

no-hup:
Fixes #1738

### The problem

If an included Taskfile sets `run` at the top level rather than per-task, the setting is dropped when it's included. Its tasks run once per call instead of once per run — no warning, no error. Running that same Taskfile directly works fine, which makes it a confusing one to track down.

### Why it happens

`Taskfile.Merge` already pushes two of the included file's global settings down onto its tasks — `silent` and `use_gitignore` — but never looks at `Run`:

https://github.com/go-task/task/blob/main/taskfile/ast/taskfile.go#L63-L79

Downstream, `GetHash` falls back to the root Taskfile:

```go
r := cmp.Or(t.Run, e.Taskfile.Run)
```

After `graph.Merge()`, `e.Taskfile` is the root file, so an included file's `Run` is unreachable from there.

### The change

One block in `Merge`, mirroring the `silent` block directly above it: when the included Taskfile sets `run`, apply it to its own tasks that don't set `run` themselves. Tasks with an explicit `run:` are untouched.

@trulede proposed this same fix in #2548 and described the cause in the issue thread — that PR was closed without tests, so I've added the coverage it was missing.

### Tests

New `testdata/run_once_included/` fixture: a root Taskfile including a library Taskfile that sets `run: once` globally, with `service-a` and `service-b` both depending on `library:build`. The test sits next to `TestRunOnceSharedDeps` and follows the same style.

- With the change: `go test ./...` → 817 passed across 33 packages; `go vet ./...` clean.
- Without it, the new test fails as you'd expect — the library task runs twice:

```
[FAIL] TestRunOnceGlobalInIncludedTaskfile
    task_test.go:2397: Not equal:
    expected: 1
    actual  : 2
```

### Behaviour change worth calling out

This does change behaviour for anyone currently affected by the bug, and I'd rather flag it than have it surface in review:

1. An included Taskfile with a global `run: once` will now actually run its tasks once. If someone was calling such a task repeatedly with different `vars` and relying on it re-running, they'll now get one execution — `run: once` hashes on task name only, not vars.
2. Where the root sets a global `run` and an included file sets a different one, the included file's value now wins for its own tasks. That's the same precedence `silent` and `use_gitignore` already have, so this makes `run` consistent with its siblings rather than introducing a new rule.

Both follow from honouring the setting at all, but if you'd prefer the root's global to keep winning, that's a one-line change and I'm happy to make it.

---

Disclosure: I used an AI assistant while writing and testing this change. I've reviewed it and verified the behaviour myself, and I'm happy to explain or revise anything here.

### DIFF
--- a/task_test.go
+++ b/task_test.go
@@ -2377,6 +2377,28 @@ func TestRunOnceSharedDeps(t *testing.T) {
 	assert.Contains(t, buff.String(), `task: [service-b:build] echo "build b"`)
 }
 
+func TestRunOnceGlobalInIncludedTaskfile(t *testing.T) {
+	t.Parallel()
+
+	const dir = "testdata/run_once_included"
+
+	var buff bytes.Buffer
+	e := task.NewExecutor(
+		task.WithDir(dir),
+		task.WithStdout(&buff),
+		task.WithStderr(&buff),
+		task.WithForceAll(true),
+	)
+	require.NoError(t, e.Setup())
+	require.NoError(t, e.Run(t.Context(), &task.Call{Task: "build"}))
+
+	// The included Taskfile sets `run: once` globally, so its build task should
+	// execute once even though both service-a and service-b depend on it.
+	assert.Equal(t, 1, strings.Count(buff.String(), `echo "build library"`))
+	assert.Contains(t, buff.String(), `task: [service-a] echo "build a"`)
+	assert.Contains(t, buff.String(), `task: [service-b] echo "build b"`)
+}
+
 func TestRunOnceSharedFailurePropagates(t *testing.T) {
 	t.Parallel()
 
--- a/taskfile/ast/taskfile.go
+++ b/taskfile/ast/taskfile.go
@@ -76,6 +76,13 @@ func (t1 *Taskfile) Merge(t2 *Taskfile, include *Include) error {
 			}
 		}
 	}
+	if t2.Run != "" {
+		for _, t := range t2.Tasks.All(nil) {
+			if t.Run == "" {
+				t.Run = t2.Run
+			}
+		}
+	}
 	t1.Vars.Merge(t2.Vars, include)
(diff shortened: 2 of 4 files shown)


no-hup:
Gentle bump on this one. It's green, and the two intended behaviour changes — var-aware include hashing, and a child global honouring the root `run` setting — are both called out in the PR body, with an offer to invert the precedence if you'd rather have the root win. Happy to adjust either way; just let me know.

PR #3015, opened 2026-09-03
feat: add interactive TUI for task execution

### LINKED ISSUE #2077: Add a interactive TUI powered by Bubbletea
### Description

Use [Bubbletea](https://github.com/charmbracelet/bubbletea) to have a fancy TUI to see a list of tasks, and optionally run then interactively. Would probably be accessible via a `--tui` flag.

Use [Wish](https://github.com/charmbracelet/wish) to allow to boot a SSH server to access it. Users would be able to run this server on a remote machine, and access via SSH to run tasks remotely. Would be a `--server` flag or similar.

I plan to implement this myself.

janluke:
# `task --tui`

## Summary

This PR adds an interactive terminal UI to Task, but it is really two separable things:

1. A public API that lets any client watch a run and answer the questions Task asks.
2. A TUI built on that API, shipped inside Task.

Perhaps, they could be split into two PRs, and I am happy to do that if you would prefer to look at them separately. The API is the part that deserves the closest reading: it is a v3 commitment, so whatever lands is what external clients build against for the life of v3. The TUI, by contrast, can be changed or replaced later without breaking anyone.

~6,000 of the lines added in this PR are in `internal/tui`.
**Cost to the rest of Task is +543 −43 across 12 files**, two thirds of which are the two new API files.

### `Listener` — watching a run

[`listener.go`](listener.go). A struct of optional callbacks rather than an interface:

| field | when it fires |
| --- | --- |
| `Scheduled(Invocation)` | a call is known about, before it runs |
| `Started(Started)` | a call begins, with the executor's own timestamp |
| `Finished(Finished)` | a call ends, with a `Result`, an `error` and a duration |
| `Joined(Joined)` | a call joined an existing `run: once` execution instead of running |
| `OutputFor(Invocation) (io.Writer, io.Writer)` | where this call's stdout and stderr should go |
| `OwnsScreen bool` | the client is drawing the display; Task must not write to the terminal |

I opted for a **struct of callbacks, not an interface.** Why? Adding a method to an interface breaks every implementation; adding a field to a struct of callbacks does not. `httptrace.ClientTrace` is the precedent in the standard library.

`Invocation` identifies one call: `ID`, `ParentID`, `RootID`, the `Task` name as written in the Taskfile, and a display `Name` (the task's label when it has one). IDs are unique per executor, so repeated calls to the same task are distinguishable, which is what makes a tree of invocations possible rather than a list of task names.

### `Prompter` — answering Task's questions

[`prompter.go`](prompter.go). Two methods:

```go
type Prompter interface {
    Confirm(task, message string) (bool, error)
    Ask(VarRequest) (any, error)
}
```

`Confirm` covers a task's `prompt:`.

`Ask` covers a required variable that was not supplied, and carries a `VarType`: a sum type that is `StringVar` or `EnumVar` today. The type is what lets a client choose the right widget and validate the answer, and being a sum type means new kinds can be added later without breaking clients.

Setting a `Prompter` is itself the statement that someone is there to answer, so no terminal and no `--interactive` are required. A client that wants questions refused can refuse them in its own `Prompter`, which is a better place for that decision than a flag, since it can answer some and decline others.

## CLI usage

```bash
task --tui                 # open the launcher
task --tui build test      # run these tasks in the TUI
```

| flag | meaning |
| --- | --- |
| `--tui`, `-T` | run in the TUI |
| `--tui-status [icons\|labels]` | how a task's state is shown in the task pane (default `icons`) |
| `--tui-task-navigator [list\|tree]` | how tasks are arranged (default `tree`) |

The two display options can be set once in `.taskrc.yml` instead of being repeated:

```yaml
tui:
  status: labels
  task-navigator: list
```

`TASK_TUI_STATUS` and `TASK_TUI_TASK_NAVIGATOR` work as well. Whether to use the TUI at all is deliberately **not** a config option: it needs a terminal, so a setting that turned it on by default would break every piped or scripted run.

`--output` keeps its normal meaning for tasks launched outside the TUI (see Ctrl+R below) and has no effect on tasks running inside it, where each pane shows raw command output.

## The launcher

<details open>
<summary>Launcher screenshot</summary>
launcher
</details>

Opening `task --tui` with no task names lists every task with a description, and a filter box above it. Typing filters on both name and description; Escape clears the filter.

<details>
<summary>Filtering screenshot</summary>
launcher-filter
</details>

Two ways to launch, because they answer different needs:

- **Enter** runs the task inside the TUI.
- **Ctrl+R** hands the terminal back and runs it exactly as `task <name>` would, with Task's normal output and your `--output` style.

After a run finishes, Escape or `b` returns to the launcher.

## The run view

<details open>
<summary>Run view screenshot</summary>
run-view
</details>

### The task navigator (left pane)

The task pane (or "task navigator") shows each **invocation**, not each task name, so a task called twice appears twice.

Two navigator styles are available:

- the "tree" view (default) shows the call tree;
- the "list" view flattens it to one level per root task.

<details open>
<summary>Tree vs list view</summary>
<table>
<tr>
  <td>
tree-view-crop
</td>
  <td>
list-view-crop
</td>
</table>
</details>

You can switch between them with `n`. The default one can be set in `.taskrc.yml` with `tui.task-navigator`. The default setting can be overridden on the command line with `--tui-task-navigator`.

When multiple tasks depend on or call a shared task configured with `run: once`, the shared task appears multiple times in the tree, marked with `↳`. In the list view, it appears only once.

If the same task is called multiple times, in the list view, each invocation is numbered with a prefix `#<n>` to distinguish them.

<details>
<summary>Tree vs list view: run-once dependency</summary>
<table>
<tr>
  <td>
tree-view-run-once
</td>
  <td>
list-view-run-once
</td>
</tr>
</table>
</details>

Each row carries a status, and an elapsed time that counts up while the task runs and settles when it finishes.

### The output pane (right pane)

The output pane's header names the selected task and the current status (in the top right corner). For a failed task, the status includes the exit code between parentheses, e.g. `failed (1)`.

A scrollbar appears on the pane's right border when the output is taller than the pane.

### Help view

`?` lists every key available in the current view. At the moment of writing:

<details>
<summary>Help screenshot</summary>
help
</details>

```
↑/↓       scroll the output
←/→/tab   switch pane
click     focus a pane
wheel     scroll the output
pgup/pgdn scroll a page
g         jump to start
G         jump to end
f         output fullscreen
y         copy output without ANSI codes
Y         copy output with ANSI codes
s         save output to a file
S         save every output to a folder
n         switch task view: tree or list
esc/b     stop, open launcher
q         stop and quit
?         show this list
```

#### Copy the output to the clipboard
- `y` copies ("yanks") the selected task's output with ANSI escape sequences stripped.
- `Y` keeps them, for pasting somewhere that renders them.

Copying uses OSC 52 and, where one exists, a clipboard helper (`wl-copy`, `pbcopy`, `xclip`, `xsel`, `clip.exe`), because OSC 52 travels over SSH but is ignored by some terminals, VTE-based ones included.

#### Save as file(s)
- `s` saves the selected task's output;
- `S` saves every task's output into a folder, one file per task.

Both ask where in the footer, suggesting `logs/<task>.<timestamp>.log` and `logs/<task>.<timestamp>/` beside the project.

Implementation notes:
- missing directories are created automatically
- saved output keeps its escape sequences, so `less -R` shows the colour
- a `logs` directory that Task itself creates is given a `.gitignore` that ignores itself, so it never shows up in `git status`; a `logs` directory that already existed is left alone.

#### Use full-screen view

You can use the full-screen view described below to select and copy a portion of the output.

### Full-screen output view

By pressing `f`, the output pane takes over the whole screen. Besides displaying the output on a larger area for an easier read, this view is useful to select and copy a portion of the whole output, without using a mouse (*). To do that:
- press `v` (or `V`) to start a selection (the `V` key comes from Vim's visual mode)
- use up/down arrow keys to expand/reduce the selection
- use `y` or `Y` to copy the selection to the clipboard, with or without ANSI escape sequences, respectively
- press `v` again to stop selection mode.

(*) Selection using a mouse is only partially supported: the selection is cleared every time you scroll with the mouse wheel, and you cannot select more text than is visible on the screen.

<details>
<summary>Full-screen mode: no active selection</summary>
full-screen-view
</details>

<details>
<summary>Full-screen mode: active selection</summary>
full-screen-view-selection
</details>

### Prompts and confirmations

<details>
<summary>Confirmation prompt</summary>
confirm-dialog
</details>

Required variables and `prompt:` confirmations are asked in a dialog over the run view, with the dialog's own keys shown inside it and a confirmation's default answer marked. Questions are asked whenever they arise, including partway through a long run, not only before it starts.

Because the TUI can always ask, `--tui` behaves as though `--interactive` were set — it is the `Prompter` that enables prompting, not the flag. `--tui --interactive=false` is therefore rejected with an error rather than silently ignored, since the two ask for contradictory things.

## Open questions and limitations

1. **`-T` as a shorthand for `--tui`.** Should we keep the `-T` flag?

2. **Naming.** The docs call the two-pane view "the dashboard". I don't like that name too much, and I am open to suggestions. "Run view" is used in this PR instead; it needs to be one or the other in both places.

3. **`interactive: true` tasks are refused under `--tui`.** They need the raw terminal the TUI is drawing on. I do not think this is solvable; Ctrl+R from the launcher is the way to run them.

4. **`--watch` is refused too**, but only for want of work in the TUI, not for want of API. `watchTasks` already runs tasks through `RunTask`, so a `Listener` sees a watch run today. What is missing is a decision about what the task tree should do across iterations, and a way to interrupt a watch loop that never returns.

5. **`testdata/tui/Taskfile.yml`** is a demo Taskfile; each task demonstrates one behaviour. No automated test uses it though. I would keep it, since it is how anyone would try this, but it can go.

6. **The `?` page does not scroll**, so on a terminal that is both narrow and short the tail of the key list is unreachable.

## Unrelated fixes and findings

- `task test:all` and `task test:watch` passed `-tags` to `gotestsum` rather than to `go test`. `gotestsum` parses its flags with pflag, which reads `-tags` as the shorthand `-t` with the value `ags` and fails. `test:all` had never worked; the line dates from `d6d51a2f` (Dec 2022) and no CI job calls `gotestsum`, which is why it went unnoticed. Fixed here since it blocks running the suite.

- **Not fixed, reported for information:** `--watch` never resets run state between iterations, so a `run: once` dependency runs on the first file change and never again, and `taskCallCount` accumulates until `MaximumTaskCall` trips. Reproduced on `main`, so it is not from this PR. `Executor.ResetRunState()`, which this PR adds, is the one-line fix. Happy to open a separate issue, or to include it here if you would rather.

## Trying it

```bash
go run ./cmd/task --dir ./testdata/tui --tui
```

## AI usage

I used Codex with ChatGPT 5.6 Sol in the initial phases of development, up until I opened this PR. I used Claude Code with Opus 5 since I first opened this PR.

## Related issue

Closes #2077.

## To do

- [ ] Review the documentation once the design is approved/stable.

## Checklist

- [x] I have read and followed the [Contribution Guide](https://taskfile.dev/contributing/).
- [x] I have disclosed the use of any AI-generated content in this pull request per the [AI Usage Policy](https://taskfile.dev/docs/contributing#ai-usage-policy).
- [x] I fully understand the changes and have hand-written the description (No AI) of this pull request. (*)

(*) I haven't reviewed the code line by line yet, but I took part in every decision, often steering the code agent in the direction I wanted. I'll check the code more in depth after I get feedback from the maintainers on the overall API and UI design.


janluke:
@andreynering I have iterated on this and made many changes, so I've rewritten the PR description from scratch.

The short version of what changed:

- **The TUI no longer piggybacks on `internal/output`.** This TUI experiment originally started as a fourth `--output tui` mode (this was before there was a launcher). The TUI  used to implement `output.Output` and be discovered by type-asserting that field. When I asked the agent to change approach and add a launcher, it somehow recycled the previous code. 
Now, there's a proper public API instead: `Listener` for watching a run, `Prompter` for answering the questions Task asks. The `internal/output` is back to a zero diff against `main`. The new `Listener` and `Prompter` API is the part I would most like reviewed.

- **Prompts and confirmations work inside the TUI.** Required variables and `prompt:` are asked in a dialog, whenever they come up, including partway through a run.

- **There are now several ways to get text out**: clipboard, save to file, and keyboard line selection in the full-screen view, which is the one I would try first.

- Many bug fixes and UI/UX improvements, including a help/keys screen, scrollbar in output pane, task duration times, status and exit code of each task in the output pane...

The Windows CI failure should be gone.

I'll stop here and wait for feedback now.

